PR #28, opened 2019-07-25
Port for Android

JoeyJiao:
Orig from android-afl,
Example to build this for android using llvm:
export AFL_TRACE_PC=true
mm

Example to enable android-test with afl:
cd android-test
TEST_CLANG_FAST_AARCH64=true mm

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -1,0 +1,1 @@
+Makefile
--- a/afl-analyze.c
+++ b/afl-analyze.c
@@ -21,6 +21,9 @@
 
 #define AFL_MAIN
 
+#ifdef __ANDROID__
+  #include "android-ashmem.h"
+#endif
 #include "config.h"
 #include "types.h"
 #include "debug.h"
--- a/afl-fuzz.c
+++ b/afl-fuzz.c
@@ -28,6 +28,9 @@
 #endif
 #define _FILE_OFFSET_BITS 64
 
+#ifdef __ANDROID__
+  #include "android-ashmem.h"
+#endif
 #include "config.h"
 #include "types.h"
 #include "debug.h"
@@ -11318,6 +11321,7 @@ static void check_term_size(void) {
 
   if (ioctl(1, TIOCGWINSZ, &ws)) return;
 
+  if (ws.ws_row == 0 || ws.ws_col == 0) return;
   if (ws.ws_row < 24 || ws.ws_col < 79) term_too_small = 1;
 
 }
@@ -12370,8 +12374,8 @@ int main(int argc, char** argv) {
     if (unicorn_mode) FATAL("-U and -n are mutually exclusive");
 
   }
-
-  if (index(argv[optind], '/') == NULL) WARNF(cLRD "Target binary called without a prefixed path, make sure you are fuzzing the right binary: " cRST "%s", argv[optind]);
+  
+  if (strchr(argv[optind], '/') == NULL) WARNF(cLRD "Target binary called without a prefixed path, make sure you are fuzzing the right binary: " cRST "%s", argv[optind]);
 
   OKF("afl++ is maintained by Marc \"van Hauser\" Heuse, Heiko \"hexcoder\" Eissfeldt and Andrea Fioraldi");
   OKF("afl++ is open source, get it at https://github.com/vanhauser-thc/AFLplusplus");
--- a/afl-gotcpu.c
+++ b/afl-gotcpu.c
@@ -28,6 +28,9 @@
 #define AFL_MAIN
 #define _GNU_SOURCE
 
+#ifdef __ANDROID__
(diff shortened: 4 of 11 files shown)


hexcoder-:
Thanks for your contribution!

There seem to be some changes that are not strictly related to Android (like [llvm7 and AFLTRACE_PC=1](https://github.com/vanhauser-thc/AFLplusplus/pull/28/commits/dbf59fdccc3f670edd8bb9b9b45f2d8fbf41122e) or the test file crash.c).
I would like to add some tests, but once for all supported platforms.

The integration of ashmem is very clean (and does only minimally touch the afl++-sources) ++.
Is it correct, that only llvm_mode (aka afl-clang-fast) is currently supported , but not afl-clang or afl-gcc?

I am not so happy with a separate Makefile for android, because it raises the maintenance costs for all future changes. 
Could it be possibly integrated with our main Makefile?

Thank again!

vanhauser-thc:
Hey some feedback from me:

I agree that there should only be one Makefile. @JoeyJiao if you have difficulties to combine them I can help you (although I dont know what would work on Android)

also - why did you rename llvm_mode/afl-llvm-pass.cc  to llvm_mode/afl-llvm-pass.cpp? 
Dont. really dont. all c++ files in the project have a .cc extensions. and original afl introduced that, and thats why its not changed (although I dont like .cc either)

JoeyJiao:
llvm_mode/afl-llvm-pass.cc to llvm_mode/afl-llvm-pass.cpp? => if using cc, Android complains that the alf-llvm-pass.o.cc not used.
And android-test is put there to verify if the built afl-clang-fast works, but it can be removed if having other test cases.

JoeyJiao:
For the Makefile, I don't know there is way Android accepts. As far as I know, Android only accepts Android.mk and Android.bp to build.

vanhauser-thc:
@JoeyJiao ah ok, so seems there really is not much you could change/fix.

for the .cc -> .cpp - how about if the llvm_mode Makefile creates a symlink from .cc -> .cpp if Android is detected? that would be better IMHO

hexcoder-:
For the Makefile we could also try with a symlink.

vanhauser-thc:
Hey, what is happening with the PR @JoeyJiao ?

I am more than happy to support Android here, but it must be in a way that is compatible as otherwise the Android port part will become outdated and not compile etc. and that is not a good thing for afl++ or for people who want to use it on Android.

So ... using symlinks would be the best way if thats possible. have you tested it?

JoeyJiao:
> Hey, what is happening with the PR @JoeyJiao ?
> 
> I am more than happy to support Android here, but it must be in a way that is compatible as otherwise the Android port part will become outdated and not compile etc. and that is not a good thing for afl++ or for people who want to use it on Android.
> 
> So ... using symlinks would be the best way if thats possible. have you tested it?

I tried to delete the Android.mk and ln -s Makefile Android.mk, but the build failed: ninja: error: unknown target 'MODULES-IN-AFLplusplus'.
Per my understanding to Android, it's another build system other than make, so I don't think the symbol link way will work.

vanhauser-thc:
of course just symlinking will not work as some variables are not set.

can you please try the following:

mv Android.mk Makefile.test
ln -s Makefile.test Android.mk

and then try to build ... does that work?
If so the symlink solution works and we would "just" need to modify our Makefile

JoeyJiao:
> of course just symlinking will not work as some variables are not set.
> 
> can you please try the following:
> 
> mv Android.mk Makefile.test
> ln -s Makefile.test Android.mk
> 
> and then try to build ... does that work?
> If so the symlink solution works and we would "just" need to modify our Makefile

I tried to mv Andorid.mk Makefile.test && ln -s Makefile.test Android.mk, build succeed.

so I just upload the changes excluding the Android.mk at the moment.

vanhauser-thc:
please also change back the .cpp to .so.cc - then I can accept the PR.
I will see how to get the Android.mk file + llvm_mode stuff to work. but as I dont have Android you will need to test that :)

JoeyJiao:
> please also change back the .cpp to .so.cc - then I can accept the PR.
> I will see how to get the Android.mk file + llvm_mode stuff to work. but as I dont have Android you will need to test that :)

Updated, pls check again.
And for sure, will test once new Makefile ready

vanhauser-thc:
I am currently in on a time extensive customer project, I will check it out on Friday+ or so

vanhauser-thc:
can you please attack your original Android.mk file here?

vanhauser-thc:
@JoeyJiao can you please attach here your original Android.mk to a comment ?

hexcoder-:
[Google developer guide for Android.mk](https://developer.android.com/ndk/guides/android_mk)

(serves as a bookmark for me)

Issue #29, opened 2019-07-25
Pull "changes" from upstream afl (?)


domenukk:
Google just released AFL on github officially, apparently.
The code lives here:
https://github.com/google/AFL
The version has been bumped to 2.5**3**b. 
Should afl++ merge the one line of readme changes and adapt to the upstream versioning?

vanhauser-thc:
it was about time :)
but I doubt anything else will happen in that repository. but lets see.

funny that they do a 2.53b. our 2.53c is likely coming tomorrow ;)

the changes are just putting link markups into the README and renaming it to README.md. 
thats something we could pull as its on our TODO list as well.

hexcoder-:
I would copy the new travis CI file, looks useful.

Heiko

vanhauser-thc:
they were thorough with the copyright stuff but very sloppy with the documentation, also the readme.md is so incomplete with the links ... 

its done, but mostly by hand.

PR #31, opened 2019-07-26
Upstream Changes of 2.53b

domenukk:
This addresses #29 :
I have tried to merge the upstream changes as well as I could. 
The main changes were a bit of docu and the License header in each file (it is now seperated from the file description).
Hope nothing else broke :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.o
 *.so
 .gitignore
+# Binaries produced by "make".
 afl-analyze
 afl-as
 afl-clang
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: c
+
+env:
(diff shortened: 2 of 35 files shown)


vanhauser-thc:
thanks for the effort!

there is a lot the google guys missed in the documentation which I added by hand and your pull request reverses stuff that we put in :) (e.g. changes to  docs/QuickStartGuide.txt)

As I only want some of the changes anyway I did it by hand already, but forgot to commit ... sorry!

domenukk:
Welp sorry for the mess then ;)
It could be nice to have the upstream commits in afl++ in case they start adding things

vanhauser-thc:
yeah that would be. but dunno how that could be done without starting a new repository.

but lets see if they really add something. I kinda doubt that. the next step would be to basically rework how afl works

PR #32, opened 2019-07-26
Rebased onto Upstream AFL

domenukk:
I played around with git a little.
You'll have to decide how nice this looks and/or wait for the future to see if it's really worth it, but here's a rebase onto afl master. There may be nicer ways around, also in case somebody knows git well. 
The steps I took were
```bash
git checkout -b upstream
git remote add afl https://github.com/google/AFL.git    
# theirs = afl++ / logic reversed here due to rebase...   
git rebase afl/master -X theirs 
git merge origin master --allow-unrelated-histories           
git checkout --theirs .  
git add .
git commit -am "Rebased on upstream"
```
Pro: diffs against upstream work
Con: all commit timestamps (not author timestamps) are set to today

`git diff master --` only shows one change, an added `CONTRIBUTING.md` file. 
Do with this PR whatever you want. ;)


vanhauser-thc:
wow that is really cool! didnt think its actual posible :)

I am split on if I want this now, or when there is actual coding and/or pulls on afl.

@andreafioraldi @hexcoder-  ... ?

domenukk:
Looks like it might not be super relevant https://twitter.com/Dor3s/status/1154737061787660288

vanhauser-thc:
sad. but then again it would make more sense to do something new that is better and more flexible.
the future afl++ releases that is something we want to steer to.

vanhauser-thc:
@andreafioraldi @hexcoder- what is your opinion? maybe its not a bad idea the longer I think about it. people will push to that repository for general fixes and minor enhancements which would make it easier for us just to pull the changes

andreafioraldi:
When we will split afl-fuzz.c in mutiple files I think will be more simple just to look at the code added in the google AFL repo and port it int afl++ than resolving millions of git merge conflicts

vanhauser-thc:
that is a valid point ...

vanhauser-thc:
We are moving further and further away, plus we will indent all code with a slightly different style ... so I think this is sadly not helpful.

thanks for the effort Dominik! (nearly typed Demonik - would be a cool name too ;) )

PR #35, opened 2019-08-01
Custom mutator

vanhauser-thc:
@kyakdan I like that addition. are you still working on it? Or is it done?

for a good pull request the custom changes to the Makefile and a simple example library + a README.custom_mutator would be great :)


kyakdan:
@vanhauser-thc thanks, I am preparing a pull request that I'll submit shortly.

vanhauser-thc:
sounds great! :)

vanhauser-thc:
Thanks!
If you now could also add a README plus an example library that would be great :)

kyakdan:
I'll get to this today or over the weekend.

jonathanmetzman:
@vanhauser-thc 
What do you think of supporting [libFuzzer's interface for custom mutators](https://cs.chromium.org/chromium/src/testing/libfuzzer/libfuzzer_exports.h?l=28) instead of a different one?

Pros:
* Compatibility will probably make life easier for people.

Cons:
* The name isn't great for anyone but libFuzzer: `LLVMFuzzerCustomMutator`, unfortunately, I don't think we can change it at this point.
* I don't think libFuzzer will support `afl_pre_save_handler` (at least under its current name). What is its purpose, I'm not sure I understand it.

Alternatively, we can just ask users to implement one of the custom mutators as a wrapper for the other.

This isn't something that will be immediately useful to me since I'm not using AFLPlusPlus now. I'm just hoping to avoid fragmentation if possible.

vanhauser-thc:
@kyakdan please dont forget a short README and a small example, thanks!

vanhauser-thc:
@jonathanmetzman  I like the new flexible addition. And I dont like the libfuzzer one, its complicated ... LLVMFuzzerMutate, LLVMFuzzerCustomMutator, LLVMFuzzerTestOneInput, ... 
It would make it more difficult for people to use this extension.

but: having a libmutator for this PR that calls LLVMFuzzerMutate* etc. with which you link the libfuzzer code - that would be the best solution IMHO.

or what do you think?

jonathanmetzman:
I think there might be a little confusion about what the different libFuzzer functions do, since not everything I linked to has to do with custom mutators (my bad, but I'll try to clear things up).

`LLVMFuzzerTestOneInput` doesn't have anything to do with custom mutators. It's the name of the harness/target function a user defines to fuzz something with libFuzzer. See [here](https://cs.chromium.org/chromium/src/third_party/zlib/contrib/tests/fuzzers/uncompress_fuzzer.cc?l=14) for example.

`LLVMFuzzerMutate` is actually defined by libFuzzer, not clients. It allows custom mutators to use libFuzzer's default mutators on a specified chunk of bytes. See [here](https://cs.chromium.org/chromium/src/content/test/fuzzer/renderer_tree_fuzzer.cc?l=57) for example.

`LLVMFuzzerCustomMutator` seems equivalent to `afl_custom_mutator`, a function defined by clients to mutate testcases.

`LLVMFuzzerCustomCrossOver` seems like the main thing that could "confuse" users. This is a function clients define so that they can combine two testcases together in a custom manner. But this is optional and actually doesn't need to be defined to get benefits out of custom mutation with libFuzzer.

>but: having a libmutator for this PR that calls LLVMFuzzerMutate* etc. with which you link the libfuzzer code - that would be the best solution IMHO.

This could definitely work. I slightly prefer the solution I proposed but I don't see anything terribly wrong with this approach.

In fact, we use [a similar approach](https://cs.chromium.org/chromium/src/third_party/libFuzzer/src/afl/afl_driver.cpp?l=260) to fuzz all of these libFuzzer targets with AFL.

I guess the downside of not using the same interface is that compatibility work might need to be done in AFL (to use libFuzzer custom mutators) and libFuzzer (to use AFL custom mutators). But obviously I understand if making libFuzzer devs' lives easier isn't your top priority :-)

But having one of these solutions will probably be useful since libFuzzer already has some custom mutators written for it. The most important of these were written using [libprotobuf-mutator](https://github.com/google/libprotobuf-mutator) supporting this interface in AFLplusplus will add support for these fuzzers as well.

jonathanmetzman:
I noticed another potential problem (maybe it's an advantage?) with this implementation (this doesn't have anything to do with compatibility). In libFuzzer, when a custom mutator is defined, only the custom mutator is used. As I understand it, here the custom mutator won't be the only mutator used by AFL.

If you have a custom mutator that builds an AST from a "JavaScript" testcase, I think it may not play nicely with AFL's default mutators which do things like bitflipping.

The possible upside of doing this is I think combining the "smart" custom mutations with the "dumb" default mutations will lead to max coverage the fastest.

vanhauser-thc:
actually that is a good point. There should be an env variable which specifies if the other mutators should run or if the custom ones should run exclusive. I put that on my TODO list

vanhauser-thc:
@kyakdan we want to release tomorrow and still need a simple small custom mutator example and a README for it. thanks!

kyakdan:
Sorry. I was extremely busy in the last weeks. I will add a a README and an example in today. At what time is the release planned?

vanhauser-thc:
@kyakdan today or tomorrow

kyakdan:
As promised: https://github.com/vanhauser-thc/AFLplusplus/pull/54

Let me know if more information is needed

jonathanmetzman:
@vanhauser-thc I suppose I haven't convinced you to support the libFuzzer APIs natively right?

jonathanmetzman:
>I don't think libFuzzer will support afl_pre_save_handler (at least under its current name). What is its purpose, I'm not sure I understand it.

I understand it now thanks to
[this](https://github.com/vanhauser-thc/AFLplusplus/pull/54/files#diff-c3ff00fd91b2ce4654a9155142c2a7fdR27)

I think this feature is probably necessary for AFL but not for libFuzzer (since libFuzzer targets always have an LLVMFuzzerTestOneInput harness function, whereas with AFL one might not even have the source code), so maybe the APIs can't be compatible.

PR #38, opened 2019-08-06
Unset AFL_CC

floyd-fuh:
If we install QEMU mode later after having already done some fuzzing runs and setting AFL_CC, QEMU mode installation will fail if AFL_CC is something else than TEST_CC.


vanhauser-thc:
Thanks!

PR #41, opened 2019-08-08
Use arc4random(3) on *BSD/macOS

t6:
Solution not involving file descriptors, seeded upon fork and on a regular basis.

This was done by @devnexen in https://github.com/google/AFL/pull/8. I only rebased it. I am hoping it can find a home here.

### DIFF
--- a/afl-fuzz.c
+++ b/afl-fuzz.c
@@ -62,6 +62,7 @@
 
 #if defined(__APPLE__) || defined(__FreeBSD__) || defined (__OpenBSD__)
 #  include <sys/sysctl.h>
+#  define HAVE_ARC4RANDOM 1
 #endif /* __APPLE__ || __FreeBSD__ || __OpenBSD__ */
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
I saw the PR and read the Google employees rational why they did not accept it - code complexity.

Its little complexity, so I am fine with that, but the question I have is - how does that make afl-fuzz better? Is it faster or is needed to run on a target platform?

We need good randomness, and /dev/urandom does that very well, and it is available on a lot of platforms. and we are not really depleting the random pool as there is only an occasional reseed from /dev/urandom. so to add something that is specific to a subset of platforms must add something useful.

paging @hexcoder- in who is taking care of *BSD & portability :)

vanhauser-thc:
I read the google PR and after thinking about it ... i dont think its worth it, but it doesnt hurt either. so I dont care and accept it ;)

vanhauser-thc:
@t6 @devnexen  - thanks

PR #44, opened 2019-08-10
llvm-mode: Get the clang version correctly

GoodDayGeorge:
When using clang-8.0, The previous command in the
Makefile will get two 8.0.0, thus a warning message print.


hexcoder-:
Thanks! I forgot that there might be two or more instances of the version triple.

PR #45, opened 2019-08-12
Revert 1 master

michaelpdu:


hexcoder-:
What problem does this solve?

vanhauser-thc:
@michaelpdu 

1 - please give a reasoning why this is important. really, take a but of effort. it is also costing our time. and please react to someone who has questions.
2 - why revert the whole thing? if there is a problem that you encounter, submit a fix for that. the commit you would like to reverse also fixes typos etc. so reverting everything is a bad move.

PR #46, opened 2019-08-18
system-config: making it more compatible with BSD systems.

devnexen:
The following knobs are Linux specifics but have few counterparts
in those systems.


vanhauser-thc:
oh very good, thank you!

PR #49, opened 2019-08-27
Qemu neverzero

andreafioraldi:
I added neverzero counters for qemu_mode and unicorn_mode when compiled for x86 and x86_64.
Doing some tests results that showmap logs more coverage in this way (it is correct?).
Execution speed is almost the same.

However IMO this needs some other tests before merge.

Neverzero for both QEMU and Unicorn can be disabled commenting the `#define AFL_QEMU_NOT_ZERO` line in config.h.

### DIFF
--- a/config.h
+++ b/config.h
@@ -339,6 +339,10 @@
 #define  CTEST_CORE_TRG_MS  1000
 #define  CTEST_BUSY_CYCLES  (10 * 1000 * 1000)
 
+/* Enable NeverZero counters in QEMU mode */
+
+#define AFL_QEMU_NOT_ZERO
+
 /* Uncomment this to use inferior block-coverage-based instrumentation. Note
    that you need to recompile the target binary for this to have any effect: */
 
--- a/qemu_mode/patches/afl-qemu-translate-inl.h
+++ b/qemu_mode/patches/afl-qemu-translate-inl.h
@@ -42,11 +42,25 @@ extern abi_ulong afl_start_code, afl_end_code;
 
 void tcg_gen_afl_maybe_log_call(target_ulong cur_loc);
 
-void afl_maybe_log(target_ulong cur_loc) { 
+void afl_maybe_log(target_ulong cur_loc) {
 
   static __thread abi_ulong prev_loc;
 
-  afl_area_ptr[cur_loc ^ prev_loc]++;
+  register target_ulong afl_idx = cur_loc ^ prev_loc;
+
+#if (defined(__x86_64__) || defined(__i386__)) && defined(AFL_QEMU_NOT_ZERO)
+  asm volatile (
+    "incb (%0, %1, 1)\n"
+    "seto %%al\n"
+    "addb %%al, (%0, %1, 1)\n"
+    : /* no out */
+    : "r" (afl_area_ptr), "r" (afl_idx)
+    : "memory", "eax"
+  );
+#else
+  afl_area_ptr[afl_idx]++;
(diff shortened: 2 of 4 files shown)


domenukk:
This, in general, seems like a good idea. 
However as a discussion point:
Wouldn't it make sense, on overflow, to increment some other buffer like:
```c
afl_area_ptr[afl_idx +1] += ++afl_area_ptr[afl_idx] ? 0 : 1;
```
That way, on overflow, the instrumentation map actually increases (unless that happens exactly 255 times :D )
Haven't tested it could still be fairly quick, godbolt doesn't output too many instructions?

Other idea would be leaving it at MAX_CHAR instead of wrapping around...
```c
afl_area_ptr[afl_idx] = afl_area_ptr[afl_idx] == 0xFF ? 0xFF : afl_area_ptr[afl_idx] + 1;
```

andreafioraldi:
This are fair ideas but IMO is expensive to insert an additional branch in afl_maybe_log.

andreafioraldi:
Neverzero, ATM, can be implemented adding an adc instruction after the inc.

```
+    "incb (%0, %1, 1)\n"
+    "adc 0, (%0, %1, 1)\n"
+    : "r" (afl_area_ptr), "r" (afl_idx)
```

domenukk:
No additional branch but potentially a few more moves. Your last solution is beautiful :1st_place_medal:

andreafioraldi:
Ty but is from Marc ahaha
I'm pushing this
```c
#if (defined(__x86_64__) || defined(__i386__)) && defined(AFL_QEMU_NOT_ZERO)
#  define INC_AFL_AREA(loc) \
    asm volatile ( \
      "incb (%0, %1, 1)\n" \
      "adc $0, (%0, %1, 1)\n" \
      : /* no out */ \
      : "r" (afl_area_ptr), "r" (loc) \
      : "memory", "eax" \
    )
#else
#  define INC_AFL_AREA(loc) \
  afl_area_ptr[loc]++
#endif
```
To the uc_compcov branch (I accidentally merged Unicorn CompCov with Neverzero and now the two branched will live in uc_compcov and merged in master at the same time)

andreafioraldi:
Mmm
```
        movzx   eax, BYTE PTR [rsp+rsi]
        cmp     al, -1
        setne   dl
        add     eax, edx
        mov     BYTE PTR [rsp+rsi], al
```
Is interesting, maybe something like
```
incb (afl_area_ptr, afl_idx, 1)
cmovo $0xff, (afl_area_ptr, afl_idx, 1)
```
Is it the same right?

vanhauser-thc:
@domenukk 
```
afl_area_ptr[afl_idx +1] += ++afl_area_ptr[afl_idx] ? 0 : 1;
```
that is an option however the issue is that if the value there was zero then afl thinks a new edge has been discovered and put this into the queue with high priority.
so overall I think that would make path discovery bad by putting more focus on queue inputs that should not have a higher priority.

@andreafioraldi 
```
incb (afl_area_ptr, afl_idx, 1)
cmovob $0xff, (afl_area_ptr, afl_idx, 1)
```
this was already tried by heiko and is implemented for historical purposes in this branch: https://github.com/vanhauser-thc/AFLplusplus/tree/saturated_counters
it is slower than plain afl and worse path discovery than plain afl.

never_zero is a bit slower than plain afl, but also has a bit better path discovery than plain afl.
and because path discovery is more important than speed, never_zero is a good thing to have.

PR #55, opened 2019-09-05
Unbreak build on FreeBSD

t6:


src/afl-forkserver.c
```diff
@@ -40,12 +40,18 @@
 #include <sys/wait.h>
 #include <sys/resource.h>
 
+#if defined(__APPLE__) || defined(__FreeBSD__) || defined(__OpenBSD__)
+#define HAVE_ARC4RANDOM 1
+#endif                           /* __APPLE__ || __FreeBSD__ || __OpenBSD__ */
+
```
t6: Should HAVE_ARC4RANDOM maybe be moved to `include/config.h` instead?

vanhauser-thc: in my repository state it is in ./include/afl-fuzz.h which is fine

vanhauser-thc: ah sorry, now I get it. yes I will move it.

vanhauser-thc: will move it to config.h

vanhauser-thc:
OK made the changes, and fixed a compile break you introduced for non-BSD :)

please check that everything still works

t6:
> OK made the changes, and fixed a compile break you introduced for non-BSD :)

Oops!

> please check that everything still works

Yes, everything still works. Thank you!

PR #61, opened 2019-09-18
Checking CPU scaling on MacOS

devnexen:
Checking optimal cpu performance or ignore if the AFL_SKIP_CPUFREQ
env is set.


vanhauser-thc:
thanks!
could you please also send another PR for the afl-system-config script that configures good settings (like cpu scaling) for MacOS?

devnexen:
Ah yes but these are read only values was just to warn the end user more than anything else.

PR #67, opened 2019-09-25
use the original data as input to custom fuzzer modules

aoh:
Custom fuzzer module seems to be repeatedly getting the output buffer as sample data instead of the original sample data. This leads to crashes and generally poor results.


vanhauser-thc:
thanks for the fix!

PR #68, opened 2019-09-25
cpu binding for FreeBSD support.

devnexen:


hexcoder-:
Thanks. I included your patch with minor editing.

PR #73, opened 2019-10-01
Build statically-linked binaries

mattz0rt:
These are changes that should help with fuzzing closed-source embedded devices.

* Adds support for statically compiling both AFL and QEMU binaries
* Improved error message when unable to reach the fork server
* Related documentation updates


vanhauser-thc:
that is useful, thanks!

PR #74, opened 2019-10-03
Multiarch CompCov

andreafioraldi:
Implements CompCov for at least the ARM QEMU target


andreafioraldi:
Seems to work for QEMU ARM 32 bit (I tested readelf).
MAP_SIZE needs to be enlarged to at least 2^18.
I also swapped the bytes compared in compcov to allow the bypassing of < and > comparisons.

andreafioraldi:
ASAP I will port it to unicorn. I would like to cover also MIPS but I have no idea how the comparisons in MIPS works.

PR #75, opened 2019-10-03
Binding to the first free cpu, porting to FreeBSD

devnexen:


vanhauser-thc:
Hi!

thanks for the PR.

I am not happy with this:

#if defined(__linux__)
  ...
#else
  ...

as it assumes everything is BSD style.  Have you tested OSX, OpenBSD and NetBSD as well that you changes work there? If not reversing it to  "if FreeBSD" but be the better first step IMHO

devnexen:
No you re right it slipped my mind.

vanhauser-thc:
thanks!

PR #79, opened 2019-10-11
NetBSD binding to cpu, ignore sleeping processes.

devnexen:


hexcoder-:
Thanks, works ok, although now we need root to set the cpu binding.

hexcoder-:
Just checked also for FreeBSD. I set 2 cpu cores on VirtualBox and now have the same problem there as well (No more free CPU cores). Could you adapt the same thing for FreeBSD, please?
Thanks!

devnexen:
Ah yes with FreeBSD I have more powerful hardwares so that problems "slipped" :-) ok looking into it.

PR #84, opened 2019-10-23
Porting libtokencap to Darwin.

devnexen:
Reading only main addresses and read only's.

### DIFF
--- a/libtokencap/libtokencap.so.c
+++ b/libtokencap/libtokencap.so.c
@@ -22,17 +22,23 @@
 #include <stdio.h>
 #include <string.h>
 #include <ctype.h>
+#include <unistd.h>
 
 #include "../types.h"
 #include "../config.h"
 
-#if !defined(__linux__) && !defined(__APPLE__)
+#if !defined(__linux__) && !defined(__APPLE__) && !defined(__FreeBSD__)
 #error "Sorry, this library is unsupported in this platform for now!"
 #endif                                                        /* !__linux__ */
 
 #if defined(__APPLE__)
 #include <mach/vm_map.h>
 #include <mach/mach_init.h>
+#elif defined(__FreeBSD__)
+#include <sys/types.h>
+#include <sys/sysctl.h>
+#include <sys/user.h>
+#include <sys/mman.h>
 #endif
 
 /* Mapping data and such */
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
@devnexen : thanks!
@hexcoder- : how about porting this to OpenBSD at the same time?
(or even FreeBSD/NetBSD if one of you two is able to (system availabilty and time))

devnexen:
I can help of any BSD if necessary.

hexcoder-:
Great, thanks!
Yes, let us compare *BSDs APIs and integrate all of them (if possible).
I am able to run NetBSD, FreeBSD, OpenBSD in VirtualBox and Darwin on a Mac mini, but need to investigate the APIs before.

devnexen:
@hexcoder since you seem to like OpenBSD maybe you would like to have a take on ? Did the FreeBSD implementation maybe we can merge it it is ok or still work from this branch ?

libtokencap/libtokencap.so.c
```diff
@@ -102,6 +108,48 @@ static void __tokencap_load_mappings(void) {
...
+     struct kinfo_vmentry *region = (struct kinfo_vmentry *)low;
+     size_t size = region->kve_structsize;
+
+     if (size == 0) break;
+
+     /* We go through the whole mapping of the process and track read-only addresses */
+     if ((region->kve_protection & KVME_PROT_READ) &&
+	 !(region->kve_protection & KVME_PROT_WRITE)) {
+          __tokencap_ro[__tokencap_ro_cnt].st = (void *)low;
+          __tokencap_ro[__tokencap_ro_cnt].en = (void *)high;
```
hexcoder-: should be
`__tokencap_ro[__tokencap_ro_cnt].en = (void *)(low +size);`

hexcoder-:
It checked the Darwin part and it is ok. Thanks very much. I enabled the corresponding test in the script for `make tests`. FreeBSD has problems yet, since the test does not succeed.
It is ok to work here for me.

hexcoder-:
Test in subdirectory `test` with
```
cc -o test-compcov test-compcov.c
AFL_TOKEN_FILE=token.out LD_PRELOAD=../libtokencap.so ./test-compcov foobar
grep BUGMENOT token.out
```

libtokencap/libtokencap.so.c
devnexen: ah yes in fact even the line just above is wrong too ... let me fix this

hexcoder-:
Bingo!

hexcoder-:
ok, can investigate NetBSD and OpenBSD tomorrow.

vanhauser-thc:
good job guys!

hexcoder-:
@devnexen: Could you help out with NetBSD? It partly works, but seems to grab invalid addresses,  which later lead to seg faults :-(

devnexen:
Try to refactor for NetBSD case instead as :
`
for (I = 0; I<len; i++) {
... getting region as region[I]..
}
`
or maybe mmap flags to be changed or making a simple malloc for NetBSD.
Hope it helps.

hexcoder-:
ok, I ask at the NetBSD mailing list then.

devnexen:
might help in case but I took little time to look at it and in fact the segfault occurs when dumping the tokens. When I redirect to say stderr, it does not segfault.

hexcoder-:
Thanks, that is interesting. So the iteration might be ok, then.

PR #87, opened 2019-10-25
Little compiler plugins rework regarding block location picked up.

devnexen:


hexcoder-:
Thanks, what is the motivation/gain for making the compile time random values seeded with time?
Are there collisions?

devnexen:
To make it consistent with the gcc plugin above all

hexcoder-:
I see. @vanhauser-thc: should we sync llvm_mode to gcc_plugin or vice versa?

vanhauser-thc:
it just improves the random number generation a bit, nothing wrong with that, so its nice to have

PR #100, opened 2019-11-04
Basic docker build proposal.

devnexen:


andreafioraldi:
Seems fine, great, thank you!

vanhauser-thc:
I disagree, sorry.
For docker we want llvm 9 and gcc 9. anything older is bad as the performance improvement is quite noticable.

devnexen:
> I disagree, sorry.
> For docker we want llvm 9 and gcc 9. anything older is bad as the performance improvement is quite noticable.

Ah yes the never zero part ... forgot.

vanhauser-thc:
its not only the notzero for llvm, for radamsa (upcoming in a few days I hope) gcc9 is required as it is faster with it. and faster is ... still very slow ;)

Thanks!

PR #102, opened 2019-11-05
qemu mode: enabling other platforms if possible.

devnexen:


vanhauser-thc:
@andreafioraldi ?

andreafioraldi:
I'm 100% sure that this will not work for macOS. QEMU user is supported for BSD but I never used it and I don't know the state and the percentage of compatibility. Remove the references to Darwin and test it on *BSD please.

andreafioraldi:
@devnexen why `--no-check-certificate`? We download QEMU directly from the official website and if the certificate of qemu.org expires is better to not download at all.

andreafioraldi:
also, libcompcov is a linux-only library, insert an if statement before calling make -C libcompcov

devnexen:
Because of NetBSD ... does not handle SSL in their packages (as git and so on)

krytarowski:
NetBSD does handle SSL, there is need to install mozilla-rootcerts.

krytarowski:
On the other hand... qemu bsd-user is broken since ever. There are downstream FreeBSD patches for it but they never hit the upstream tree.

andreafioraldi:
I think that, in addition, the patches to the files into linux-user/ should be ported to bsd-user in order to make it work.

devnexen:
regardin

PR #103, opened 2019-11-06
libdislocator, optional huge pages support.

devnexen:


hexcoder-:
Nice addition, thanks.
Where is the definition of `USEHUGEPAGE`?
mmap with huge pages can fail, so we should do a regular mmap as fall back.

```
buf = mmap(NULL, LENGTH, PROTECTION,
#ifdef MAP_HUGETLB
           MAP_HUGETLB |
#endif
           MAP_PRIVATE | MAP_ANONYMOUS,
           0, 0);
#ifdef MAP_HUGETLB
  if (buf == MAP_FAILED) {
    // try again without huge pages:
    buf = mmap(NULL, LENGTH, PROTECTION, 
               MAP_PRIVATE | MAP_ANONYMOUS, 0, 0);
  };
#endif
if (buf == MAP_FAILED) {
  ...
}
```

PR #104, opened 2019-11-08
test, /sbin not necessarily in the path os casual users.

devnexen:


hexcoder-:
Thanks, this is probably always valid on Linux...

vanhauser-thc:
I had to fix this. not all sysctl are in /sbin. e.g. on systems with busybox ...

devnexen:
The second call passes ?

vanhauser-thc:
ah overlooked the second one, thanks

PR #106, opened 2019-11-11
Radamsa

andreafioraldi:
Now we have libradamsa.so in the project root. `make radamsa` to build it.
TODO: add a test of -RR to test.sh and update the doc.


vanhauser-thc:
did QA, added to the Makefile and integrated in test.sh.

so only DOC is missing?

andreafioraldi:
yep seems that only the doc is missing

andreafioraldi:
I added the info about radamsa in the readme. If you want to add additional info in the docs/ folder just do it in master, I will merge the branch now.

PR #107, opened 2019-11-12
gcc plugin little update proposal.

devnexen:
like its llvm counterpart, passing the compiler chain used
to avoid using necessarily AFL_CC/AFL_CXX.


vanhauser-thc:
That is a good idea however it is pointless if its not documented :)

vanhauser-thc:
thanks!

PR #108, opened 2019-11-12
libradamsa forgotten possible undefined symbol.

devnexen:
Note: do not know if it should be upstreamed first or not though...


vanhauser-thc:
radamsa is on gitlab which makes upstream patches more work. also libradamsa.c is a file that is dynamically generated in the radamsa project.

I created an issue in their gitlab and reported it.

thanks!

PR #109, opened 2019-11-13
Qbdi mode

hac425xxx:


vanhauser-thc:
Hi,

thanks for the pull request.
Please only send pull requests that are done. This one is working only for you and no-one else:
```
compiler_prefix="/home/hac425/workspace/android-standalone-toolchain-21-x86/bin/i686-linux-android-"
```
the build script should try to detect the compiler and also let the user define one

also there is zero documentation no how to use it - or why to use it.

please add that.

andreafioraldi:
Hi, what is the purpose to use a DBI for binary-only fuzzing? QEMU mode is a magnitude faster. It has sense when it can run binaries that QEMU can't, like DynamoRIO in WinAFL, but I don't really see any reason to ship this instrumentation in AFL++. Try to convince me if you see particular needs.

hac425xxx:
> Hi, what is the purpose to use a DBI for binary-only fuzzing? QEMU mode is a magnitude faster. It has sense when it can run binaries that QEMU can't, like DynamoRIO in WinAFL, but I don't really see any reason to ship this instrumentation in AFL++. Try to convince me if you have particular needs.

qemu mode don't support android now, qbdi support andorid x86 && x64, and It could support android arm and aarch64 soon.

hac425xxx:
> Hi,
> 
> thanks for the pull request.
> Please only send pull requests that are done. This one is working only for you and no-one else:
> 
> ```
> compiler_prefix="/home/hac425/workspace/android-standalone-toolchain-21-x86/bin/i686-linux-android-"
> ```
> 
> the build script should try to detect the compiler and also let the user define one
> 
> also there is zero documentation no how to use it - or why to use it.
> 
> please add that.

ok, I could add the document soon

andreafioraldi:
Ok if your target is Android native libraries is ok for me to add an addition instrumentation. But one question, why not frida-gum? With it you can instrument both native and generated code from ART.

hac425xxx:
frida is not stable for collect code  coverage, I  have try frida before,  but It can't correctly get the code coverage of my test program, some code path could missing....

hac425xxx:
Using qbdi, we just need load the library by dlopen, don't need  inject QBDI in an Android app.
An usage demo
```
https://blog.quarkslab.com/android-native-library-analysis-with-qbdi.html
```

andreafioraldi:
Got your point, but Frida improved a lot in recent times. I did some stuffs with Frida more than 1 year ago and there was some problems (mainy on x86, not in their main target that is ARM) but now should works very well. I was thinking about somehting like this https://gist.github.com/andreafioraldi/5f352c17869e8dd91477fcfab0a0c14c that is a generic and performant instrumentation for both native and generated code. Maybe @oleavr can say something about the speed and the correctness of Frida and if is ready for fuzzing.
Regards your patch, document it well and run the code formatter (see the `code-format` target in the root makefile), we will problebly merge it after some tuning and tests.

hac425xxx:
add document and modify build.sh, also format the code

vanhauser-thc:
this looks much better now :)
the build script is not for ARM or AARCH64 - is there a reason?
Also you specify exact versions to use for the library and the build tools - is that necessary or would it also work with older/newer versions? If it can be more flexible (e.g. switching 0.6.3 with VER and saying "(verfified to work with v0.6.3)". Otherwise the documentation has to be updated every time

oleavr:
@andreafioraldi 

> Maybe @oleavr can say something about the speed and the correctness of Frida and if it is ready for fuzzing.

It's definitely ready for that, especially with the new `CModule` integration. Not aware of any correctness-issues, but if any are found they will be easy to fix – `Stalker` is quite small and self-contained. Only depends on a few internal components used by Frida's `Interceptor` (`{X86,Arm64}Relocator` and `{X86,Arm64}Writer`).

hac425xxx:
> this looks much better now :)
> the build script is not for ARM or AARCH64 - is there a reason?
> Also you specify exact versions to use for the library and the build tools - is that necessary or would it also work with older/newer versions? If it can be more flexible (e.g. switching 0.6.3 with VER and saying "(verfified to work with v0.6.3)". Otherwise the documentation has to be updated every time

QBDI don't release arm  and aarch64 android version now, but I could release this after a few months, I just test 0.7.0 now, and I could update document  every time qbdi update until  arm  and aarch64 android version  relase

vanhauser-thc:
@andreafioraldi  - this is rather your expertise, I leave this PR in your hands, OK?

andreafioraldi:
yes fine, I'll setup a proper env to test it with also other binaries and then i will merge it.

andreafioraldi:
I pushed my changes and for me we can merge w/o problems. The only issue here is that we can't insert a test for qbdi_mode in test.sh.
I will test it by hand every time that there will be a change, I don't see any other options (or we can install the Android NDK and Emulator in Travis lol).

PR #110, opened 2019-11-13
Fix docker build

k0ss:
The docker build has never worked for me.  I fixed all the build errors, mostly due to missing dependencies.  I also fixed silent errors that would allow for the build to finish but wouldn't contain features such as python mutators or QEMU's unsigaction module.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,3 +29,4 @@ ARG LLVM_CONFIG=llvm-config-9
 COPY . /app
 RUN cd /app && make clean && make distrib && \
     make install && cd .. && rm -rf /app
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
Thanks!
isnt a 
```
RUN mkdir /app
```
missing?

k0ss:
The `/app` dir is created by the `COPY . /app` command. It saves us some space by not creating an unnecessary layer for another RUN command.

vanhauser-thc:
ah I am not a docker guy, so I assumed the COPY command would not create the target directory :)

PR #111, opened 2019-11-16
gcc plugin fix proposal, line of code misplacement mainly.

devnexen:


vanhauser-thc:
oh, I accidently commited something where I was checking out where the problem in the gcc_plugin is, that was not meant for committing ... damn

PR #113, opened 2019-11-20
the plot data does not generate proper gnuplot data

devnexen:
when killed or for a defined amount of time.


vanhauser-thc:
thanks!

PR #114, opened 2019-11-22
compiler plugins, just create const constants once.

devnexen:


vanhauser-thc:
@hexcoder- is on vacation so I accept this on his behalf

PR #115, opened 2019-11-23
gcc plugin add stack_protect attribute on demand proposal.

devnexen:


vanhauser-thc:
@hexcoder-  - this is your turf :)

hexcoder-:
Hi David, thanks.
What is the intention here? AFL_HARDEN normally switches on stack protection globally for the target compilation, but your patch optionally modifies just the afl_trace function from the gcc runtime. Sorry, I did not understand it. Did you mean to modify the gcc wrapper?

devnexen:
Yes sorry that s surely not so useful after 2nd thoughts, thought apply stack to avoid possible wrong map limit ...

PR #120, opened 2019-11-26
Python module build fix in exotic oses

devnexen:


vanhauser-thc:
ah good idea! thanks!

PR #121, opened 2019-11-28
libdislocator updates improvements proposal.

devnexen:
- optional tls alloc_canary value.
- slight change of memory limit value checking.


vanhauser-thc:
what is the goal of this? I am not sure I understand this

devnexen:
It is mainly to do somewhat like libc allocators using randomised canaries.

PR #122, opened 2019-11-28
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override build date with `SOURCE_DATE_EPOCH`
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

Note: This date call is designed to work with different flavors
of date (GNU, (Free)BSD and others).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,13 @@ AFL_FUZZ_FILES = $(wildcard src/afl-fuzz*.c)
 
 PYTHON_INCLUDE	?= /usr/include/python2.7
 
+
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
the build date in the patch will only be in the man pages ... shouldnt that information also be in e.g afl++ -h output?

I will accept this, however this looks incomplete to me.

bmwiedemann:
It was enough to make our openSUSE afl package build reproducibly, so I dont need more.

PR #127, opened 2019-12-02
LLVM 10 build fix proposal

devnexen:
c++14 is the minimum for this version, thus the hardcoded gnu++11
standard flag makes the build fails.
Missing header only for this version seemingly for the option.

### DIFF
--- a/llvm_mode/LLVMInsTrim.so.cc
+++ b/llvm_mode/LLVMInsTrim.so.cc
@@ -13,6 +13,7 @@
 #include "llvm/IR/Module.h"
 #include "llvm/Pass.h"
 #include "llvm/Support/raw_ostream.h"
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
you are saying it needs c++14 to compile but you do not set this in the Makefile ... why?

devnexen:
I allow to override it via LLVM_STDCXX variable in case anyone uses llvm 6 for ex I did not dare to hardcode it.

vanhauser-thc:
yes you allow to override, but if someone does not do it it will still not compile with llvm-10 right? then this is pretty pointless :)
the better move is to detect the llvm version and set the c++14 version flag if its 10 or higher

vanhauser-thc:
thanks!

PR #128, opened 2019-12-02
Enable libtokencap on DragonFlyBSD

devnexen:


libtokencap/libtokencap.so.c
```diff
@@ -33,7 +33,7 @@
 #include "../config.h"
 
 #if !defined __linux__ && !defined __APPLE__ && !defined __FreeBSD__ && \
-    !defined __OpenBSD__ && !defined __NetBSD__
+    !defined __OpenBSD__ && !defined __NetBSD__ && !defined __DragonFly__
```
hexcoder-, line 36: @devnexen : Hi, just looked at DragonFly and libtokencap. The test fails, so I had a look in the sources. 
There is currently no real support for getting the required read-only mappings from the kernel, right?
I looked at sysctl() and it seems not to be supported by the API.

If so, IMO we should consequently drop support for DragonFly.

devnexen: There is but for Dragonfly it is different than FreeBSD, you would need more to rely on kvm api.

PR #129, opened 2019-12-03
libdislocator, typo fix.

devnexen:


vanhauser-thc:
I saw that change to uppercase but it compiled for me so I was fine with it ... whats up with this?

devnexen:
the upper case is one is "just" the default value whereas the lower's is the tls value which can be overriden (this is how I found out the issue).

vanhauser-thc:
thanks!

PR #131, opened 2019-12-06
IOS build fix.

devnexen:
system call is disabled on this platform.


vanhauser-thc:
does it still compile on MacOS with this?

devnexen:
Yes just tried.

PR #133, opened 2019-12-09
Updating regexp.dict

RootUp:
Cleaning up few regex which weren't escaping properly while fuzzing and adding new more :)
Reference: https://twitter.com/RandomDhiraj/status/1204089908131979264


vanhauser-thc:
thanks I had seen the PR @ google afl but hadnt had the time to grab it :-)

PR #137, opened 2019-12-13
afl-unicorn define those same macros, leading to spurious warnings.

devnexen:


andreafioraldi:
afl unicorn will be completely replaced

devnexen:
ah sorry :)

andreafioraldi:
np

PR #140, opened 2019-12-15
Concealing the no build test to apple devices.

devnexen:


devnexen:
to fix [139](https://github.com/vanhauser-thc/AFLplusplus/issues/139)

vanhauser-thc:
so when compiling for iOS then it would try to build? because iOS != Darwin?

the correct fix is to not set LLVM_UNSUPPORTED if its LLVM10.
and LLVM_UNSUPPORTED should not disable the build, just warn and still try - unless we know it wont build (like with apple's Xcode)

devnexen:
Ok tested with Ubuntu with the LLVM PPA repo and Mac when it fails as expected.

vanhauser-thc:
thanks!. that UNSUPPORTED thing was a bug I introduced.

PR #146, opened 2019-12-18
Tidied up python examles and fixed bug in uc_afl_fuzz

domenukk:
This PR removes unneeded code from the python examples and fixes a tiny bug in uc_afl_fuzz where files would not be unmapped.


unicorn_mode/samples/simple/simple_test_harness.py
```diff
@@ -120,6 +102,8 @@ def main():
 
     uc.mem_map(STACK_ADDRESS, STACK_SIZE)
     uc.reg_write(UC_MIPS_REG_SP, STACK_ADDRESS + STACK_SIZE)
+    
+    print(STACK_ADDRESS + STACK_SIZE)
```
vanhauser-thc: is that a debug print that is still in there?

domenukk: Removed :)

vanhauser-thc:
thanks!

PR #148, opened 2019-12-20
On non Linux systems, (g)make distrib stops halfway

devnexen:
because of QEMU not supported.
Symplifying cores counting data gathering.


vanhauser-thc:
thanks!

PR #150, opened 2019-12-21
minor fix for llvm_mode build with non-standard paths

afflux:
plus cleanup of the unicorn submodule.

I'm not sure if the version checking logic in llvm_mode/Makefile still makes sense after 123d97bfb896aefca24ff8bdc235b900fe714fde - but at least it should use the same llvm-config.

### DIFF
--- a/llvm_mode/Makefile
+++ b/llvm_mode/Makefile
@@ -136,8 +136,8 @@ else
 endif
 
 ifneq "$(CLANGVER)" "$(LLVMVER)"
-  CC = $(shell llvm-config --bindir)/clang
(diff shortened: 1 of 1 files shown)


andreafioraldi:
ty for the patch.
@vanhauser-thc look at this travis build, xenial-arm64 with GCC 5:
```c
afl-llvm-pass.so.cc:39:10: fatal error: 'llvm/IR/DebugInfo.h' file not found

#include "llvm/IR/DebugInfo.h"
```
The build doesn't fail cause the LLVM test is skipped.
The problems seems that with old LLVM the file is `llvm/DebugInfo.h` and not `llvm/IR/DebugInfo.h`

PR #152, opened 2019-12-21
argvfuzz preload for fuzzing binaries' argv

afflux:
LD_PRELOAD library, hooks `__libc_start_main` to replace `argv` with data from stdin (uses logic from `argv-fuzz-impl.h`).

Instead of hooking __libc_start_main, a more generic approach might be to provide a `__attribute__((constructor))` function which also gets argc, argv and envp as arguments. However - unless I've missed something - such a function would not be able to change them without doing stack unwinding (see e.g., https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/x86_64/dl-machine.h;h=58260c7876fd361a5a6f30e11a341ceb605a199c;hb=HEAD#l141)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -316,6 +316,7 @@ clean:
 	-$(MAKE) -C gcc_plugin clean
 	$(MAKE) -C libdislocator clean
 	$(MAKE) -C libtokencap clean
+	$(MAKE) -C experimental/argv_fuzzing clean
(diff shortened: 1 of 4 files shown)


andreafioraldi:
Hi ty for the contribution. I'll test it when I'll have free time.

vanhauser-thc:
thanks kjell!

PR #156, opened 2019-12-27
Fixed memory leak in afl-fuzz-python.c - trim_case_python.

n0pFlux:
retbuf gets malloced in 'trim_py', however, it's never freed afterwards in 'trim_case_python'.

### DIFF
--- a/src/afl-fuzz-python.c
+++ b/src/afl-fuzz-python.c
@@ -345,7 +345,10 @@ u8 trim_case_python(char** argv, struct queue_entry* q, u8* in_buf) {
     fault = run_target(argv, exec_tmout);
     ++trim_execs;
 
-    if (stop_soon || fault == FAULT_ERROR) goto abort_trimming;
+    if (stop_soon || fault == FAULT_ERROR) {
+        free(retbuf);
+        goto abort_trimming;
+    }
 
     cksum = hash32(trace_bits, MAP_SIZE, HASH_CONST);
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
oh good catch!

Looking at the code - would it not be enought to have one free() in the abort_trimming part? it looks like the function has only one return.

and did you find it by actually using the python module feature? I am curious :)

n0pFlux:
retbuf is in the scope of the surrounding while loop, https://github.com/vanhauser-thc/AFLplusplus/blob/cc3bf762ece81fb1007bc368f4bd604804afd095/src/afl-fuzz-python.c#L327 , and in every iteration is a malloc in 'trim_py'.
Call to 'trim_py' within the loop: https://github.com/vanhauser-thc/AFLplusplus/blob/cc3bf762ece81fb1007bc368f4bd604804afd095/src/afl-fuzz-python.c#L336
The malloc in 'trim_py': https://github.com/vanhauser-thc/AFLplusplus/blob/cc3bf762ece81fb1007bc368f4bd604804afd095/src/afl-fuzz-python.c#L294
Therefore, retbuf would be out of scope with a single free() in the abort_trimming part and it'd just free the allocation of the last iteration.

I actually did find it by using the Python module feature. It also got me to read a lot about Python's memory management, switching from Python2 to Python3 (because Python2 might have some memory issues, which are fixed with Python3!), and rewriting my module a few times, because I suspected the mistake in Python/my Python usage. So, I've learned a few thinks in the process ;).

vanhauser-thc:
thanks!

also if you have nice python mutator scripts you can share - that would be awesome for other people. just put a PR if so

PR #162, opened 2020-01-03
libtokencap: *chr string functions

devnexen:


vanhauser-thc:
Does this make sense?
usually the haystack comes from external data and the needle is what the developer puts in, so the other way around then what we usually do. so we would log what we already send?

andreafioraldi:
I agree with @vanhauser-thc , tokencap dumps the haystack alongside the needle for other routines cause "why not, we are already here" but for this single char routines are useless.

PR #164, opened 2020-01-05
main install task fix for osx mainly as -D flag is not supported

devnexen:


devnexen:
spotted when trying to make home-brew package.

vanhauser-thc:
ah ... we have 8+ .8 man pages ... and your patch only installs two afterwards?
the correct fix is to remove the -D flag as it is unnecessary - the directories are already created in the previous command, so I will remove it.

PR #166, opened 2020-01-07
llvm_mode little update proposal.

devnexen:
avoiding intrinsics generated by LLVM for the instrumentation.

### DIFF
--- a/llvm_mode/afl-llvm-pass.so.cc
+++ b/llvm_mode/afl-llvm-pass.so.cc
@@ -156,9 +156,14 @@ bool AFLCoverage::runOnModule(Module &M) {
 
   /* Instrument all the things! */
 
+  const char *IntrinsicPrefix = "llvm.";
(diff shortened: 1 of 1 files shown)


devnexen:
@hexcoder ;-)

vanhauser-thc:
out of curiosity - what functions are these? do they normally exist in the created binary?

andreafioraldi:
This should be redundant with -fno-builtin I guess

devnexen:
> out of curiosity - what functions are these? do they normally exist in the created binary?

As an example debug metadata as llvm.dbg.declare callinst are generated.

vanhauser-thc:
@andreafioraldi well it cant hurt ... in afl-clang-fast are specfic -fno-builtin-foos ... and if llvm11 has new ones they would not be covered

hexcoder-:
Thanks, that is useful. I had not thought about that optimization before.

PR #171, opened 2020-01-11
experimental OpenBSD build fix

devnexen:


vanhauser-thc:
have you also checked NetBSD and FreeBSD? Do they need -ldl? or do they fail with it too?

devnexen:
yes it does work :-)

vanhauser-thc:
in the future please describe your pull request. if you say  "openbsd" and the patch just makes a difference for linux ...

PR #174, opened 2020-01-18
libdislocator: reallocarray API introduction

devnexen:


domenukk:
Shouldn't the realloc fail when a possible overflow was detected?

devnexen:
ah yes I used wrong diff my bad.

PR #183, opened 2020-02-03
Broader Python Support

### LINKED ISSUE #182: Support More Python 3 Versions
Arch now ships with Python 3.8 (released end of last year), the latest Ubuntu LTS (18.04) still uses Python 3.6. 
Right now, however, the Makefile fixes the required python version to 3.7, see https://github.com/vanhauser-thc/AFLplusplus/blob/c7167190a6abe9884787b0aae9e0728848971163/Makefile#L66
I suggest to, instead, fix it to python3 (any) and potentially parse `python3 --version` if we want to enforce a min version.

domenukk:
This fixes #182 and should(tm) work on major operating systems (as long as Python is in the PATH).
Maybe someone can check on BSDs if it works there as expected, too, before merging into master

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,42 +63,20 @@ CFLAGS     += -Wall -g -Wno-pointer-sign -I include/ \
 
 AFL_FUZZ_FILES = $(wildcard src/afl-fuzz*.c)
 
-ifneq "$(filter %3.7m, $(shell python3.7m-config --includes 2>/dev/null))" ""
-  PYTHON_INCLUDE  ?= $(shell python3.7m-config --includes)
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
first you do python3 stuff and then you overwrite the with python stuff if that is there too?

domenukk:
It's a `?=`, so it only overwrites if the variable was not set before. Do you think I should do it more explicitly (elseif)?

vanhauser-thc:
nah overlooked it, i'm in stress :)

vanhauser-thc:
works fine for me

PR #186, opened 2020-02-03
add AFL_USE_UBSAN

andreafioraldi:
Support to ubsan in afl-clang-fast. I choosed the `-fsanitize=undefined -fsanitize-undefined-trap-on-error -fno-sanitize-recover=all` flags, should be correct.


vanhauser-thc:
needs changelog, env_vars and llvm/readme entries though

PR #192, opened 2020-02-08
libdislocator FreeBSD build fix. max_align_t had been define from the…

devnexen:
… 12th release.


andreafioraldi:
Lol so FreeBSD was not compliant with the C11 standard for so much time (v12 is 2018 IIRC)?

hexcoder-:
Ah, yes. Thanks for the quick fix! I had only FreeBSD version 11 lying around.

PR #203, opened 2020-02-20
Regular expression dictonary

RootUp:
Updating the `dict` with few new seeds, has I found this during my continuous fuzzing in VIM.


domenukk:
Sounds like fun, thanks :)

PR #213, opened 2020-02-26
Fix hanging fork and child with -V -E

Kusoku:
If we let multiple fuzzers end with -V or -E option, it will cause it to think we are still occupying the cores, even if they are free, once we try to run another job it would return an error that no free nodes are available.

This change fixes that problem.


vanhauser-thc:
thanks!

PR #216, opened 2020-02-28
Added missing dependency to the documnetation. Tested on Debian 10

anon8675309:
This dependency appears to be required for UnicornAFL.  Running the previous apt command ad then `make distrib` on a clean Debian 10 install should reproduce the issue.  The sanity checks did work and very helpfully pointed me to the package, but I figured it should be listed here too. :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,7 @@ These build options exist:
 * STATIC - compile AFL++ static
 * ASAN_BUILD - compiles with memory sanitizer for debug purposes
 * AFL_NO_X86 - if compiling on non-intel/amd platforms
+* LLVM_CONFIG - if your distro doesn't use the standard name for llvm-config (e.g. Debian)
 
 e.g.: make ASAN_BUILD=1
 
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ afl++ has many build options.
 The easiest is to build and install everything:
 
 ```shell
-$ sudo apt install build-essential libtool-bin python3 automake bison libglib2.0-dev libpixman-1-dev clang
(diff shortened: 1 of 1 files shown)


domenukk:
Shouldn't that be `python3-setuptools` instead of `python-setuptools` if we list python3 as dependency in the first place?
At least that package exists in debain repos too.

anon8675309:
The component requires python-setuptools, not python3-setuptools.

Whether that's what it "should" require is debatable, but ./Dockerfile and unicorn_mode/build_unicorn_support.sh both mention the legacy Python package.

I also double checked the Debian 10 machine I just set up and it does not have python3-setuptools installed and I was able to build everything without any issue.

domenukk:
I'll double check if it installs unicornafl into python3 or python then. Maybe it's time to update to python3-setuptools in 2020 :)

domenukk:
Thanks for the pr either way :) extending the docs makes sense.

anon8675309:
I added one more to make the LLVM_CONFIG easier to find and indicate that Debian uses a non-standard name for llvm-config (llvm-config-7).

Part of me things it's reasonable to auto-detect that alternative name (as is done for BSD), but documenting it seemed easier.  It took me a while to notice that the afl-clang-fast was missing because it was buried in an error message in the output of `make distrib` (quiet failure).

There were some others that failed too due to missing dependencies, but there was a message that explicitly mentioned that it was okay that it failed (just some test programs by the looks of it).

vanhauser-thc:
looks good to me. Dominik: maybe accept this and take care of python3 afterwards? I fear python3 will be a harder task to get right for all the Linux, BSD and MacOS platforms.

domenukk:
All right.

PR #219, opened 2020-03-02
Updates for easy installations

ankit2001:
i have fixed some errors and documentation while installing afl in all modes, with this a user will be able to install it in all modes without any difficulty
Screenshot from 2020-03-02 17-44-54
Screenshot from 2020-03-02 17-43-19
Also added 32 to 64 bit binary on readme.


andreafioraldi:
Hi @ankit2001 , thanks for your contribution.
However, we don't ship script with distro-specific commands.
Like for the QEMU mode script, we just suggest to the user what he has to install.

domenukk:
It does seem like we have at least `llvm` and `flex` as dependencies missing, then, though?
I'm not against extending the docs if this is the case.

vanhauser-thc:
We should at least mention it IMHO
llvm is a kinda must-have for source base fuzzing, but for people who do binary fuzzing only its of zero interested and no requirement at all.

andreafioraldi:
Yep I was talking about `gcc_plugin/install.sh`, not the readme stuffs.

vanhauser-thc:
yes :)

NO to an install.sh script that uses a distribution specific tool (apt)
YES to mentions clang/flex in the main README in the install section

Issue #191, opened 2020-02-08
Uniform python and custom mutator API


andreafioraldi:
Python and custom mutators do exactly the same thing in two languages.
We should uniform the API (maintaining backward compatibility but should be easy).

I propose to define the following functions in the C mutator:
+ afl_custom_fuzz (equivalent to the old afl_custom_mutator that is used if exported and afl_custom_fuzz is not defined)
+ afl_custom_init
+ afl_custom_init_trim
+ afl_custom_trim
+ afl_custom_post_trim
+ afl_custom_pre_save (equivalent to the old afl_pre_save_handler thai is used if exported and afl_pre_save_handler is not defined)
+ afl_custom_queue_get (called when a queue entry is pulled from the queue and given to fuzz_one())

I propose to add the following method to the python mutator:
+ pre_save (does the same job of afl_custom_pre_save in the custom).
+ queue_get

In this case, a python mutator is a case of a custom mutator. We can use a single entry in the UI cause custom and python will be never used combined (and we have to check this cause they are mutually exclusive).

domenukk:
Wrt mutually exclusive, could there be use cases where adding multiple mutators might be beneficial?

vanhauser-thc:
they are mutually exclusive? I am not so sure about this. I thought they can be used both. and would be nice to have both.

unifying the function names etc - all for it.

domenukk:
How about 
- supporting multiple custom mutators (multiple mutator .so files can be added with simicolons)
- refactoring the python mutators to be one custom mutator lib (or a dummy lib)
- if `AFL_PYTHON_MODULE` is set, add the python mutator lib to the path
or do you think that would be overkill?

vanhauser-thc:
I like the idea. Although I doubt it will be a common thing people would use so not sure its something we want to put our time in? dunno.

It goes into the same direction I want to do for qemu_mode, instead (or additionally) to afl_code_start/end have afl_code_regions=0x123-0x456,0x789-0xabc,.. etc.
(but first I will do the lto instrim stuff. thats for me the absolute priority. tomorrow I will start with that)

andreafioraldi:
I produced some examples about how will be the unified API:

C mutator: https://github.com/vanhauser-thc/AFLplusplus/blob/custom_mutator_unified_api/examples/custom_mutators/new_custom_mutator.c
[...]

PR #220, opened 2020-03-03
[Discussion] AFL without* globals

domenukk:
This is a first draft how a non-global AFL could look like.
It now passes all test cases but still has many rough edges.

The new structs are still stored globally for some destructuring magic, sigalarm is not handled for more than one instance of afl/the forksever and I may have missed some global vars.

Please try it out, test the speed, see how you like the naming, etc. :)


unicorn_mode/samples/persistent/harness.c
```diff
@@ -8,7 +8,7 @@
 
    Run under AFL as follows:
 
-   $ cd <afl_path>/unicorn_mode/samples/persistent/
+   $ cd <afl_path>/afl->unicorn_mode/samples/persistent/
```
vanhauser-thc: really  "afl->unicorn_mode"?

domenukk: Hehe that one was a renaming bug. 🙃
It's in both harness files, too.

vanhauser-thc:
in general looks good to me.

Some comments:
 - could also already define MAP_SIZE as a variable where it is used? for future usage where its variable. looks like it fits well within this project item
 - We should not merge it the next two weeks :) and first do proper testing that everything works. basically when test.sh runs through fine on all our target platforms incl. *BSD
 - I am also curious if there is a performance penalty. is afl-fuz compiling and working so far?

domenukk:
A variable `MAP_SIZE` is a good idea and is already partly in, namely the following are in `sharemem_t`:
` size_t         size_alloc; /* actual allocated size */
  size_t         size_used;  /* in use by shmem app */
`
But are not yet used widely.

afl-fuzz works. I'd be cool to leverage googles https://google.github.io/fuzzbench/ to benchmark it.

domenukk:
Turns out getting rid of MAP_SIZE is a bigger undertaking: 

- There is `MAP_SIZE_POW2` also
- Some arrays are allocated based on `MAP_SIZE`  
- Probably other things

I think the better way is to keep MAP_SIZE almost everywhere and instead use `shm->size_used` (and have `size_used_pow2`) where applicable.

domenukk:
Welp... Something seems to be wrong, still :/

image

domenukk:
Nevermind, that was only in my local changes. not pushing :)

domenukk:
The latest commit added linked lists and tidied the apis a bit, if anybody now wants to build a proper timer system, feel free (@gsoc students?)

vanhauser-thc:
I did some speed tests, it is about a 0.3-0.7% speed decrease which is not much in my opinion. But of course the new alarm handling can also impact it further when it is there.

domenukk:
> I did some speed tests, it is about a 0.3-0.7% speed decrease which is not much in my opinion. But of course the new alarm handling can also impact it further when it is there.

A new alarm handling won't make it into this PR, I think we're better off starting to use threads instead of signals after all.
Most timeouts waiting for can be replaced with `select`.

vanhauser-thc:
Is this currently waiting for the signal handling portion?
As this is a big patch cluster and it is already not merging because of the code changes I would find it preferable if this is merged sooner than later :)

And as our trunk is in "testing" anyway because of LTO this would be a perfect time to but afl-fuzz & co in testing as well.

domenukk:
I've done my best to merge the changes.
As a quick&dirty fix for the custom mutator API drawbacks I posted in #191, I added an afl to each custom mutator call.
I'm pro merging this fast, keeping this branch up to date takes a long time

vanhauser-thc:
go for it once you caught up. I will keep you busy today :p

domenukk:
I've merged it. Fingers crossed :see_no_evil:

vanhauser-thc:
it is colliding with a different issue - leak sanatizer that was not a problem is now reporting leaks everywhere and terminating travis early ... sigh

domenukk:
In afl-fuzz three is a leak when argv @@ is replaced by the filename. I'm pretty sure this leak always existed... I've added a todo note there 😅
I didn't find the time to leak check afl-showmap, got to go to work now

vanhauser-thc:
@domenukk afl-cmin fails (very likely because of afl-showmap)  and afl-tmin too - can you investigate?
you can see the failures in test.sh output

PR #228, opened 2020-03-05
Few memory leak fixes proposal.

devnexen:


domenukk:
Looks correct to me, thanks!

PR #235, opened 2020-03-06
Fixing 2 little mistakes

antonio-morales:
This example doesn't compile due to two little errors:
- There is a missing semicolon
- "data" array doesn't exist. I think "buf" should be used instead.


andreafioraldi:
ty for the patch

PR #238, opened 2020-03-07
Two new hooks for the custom mutator

h1994st:
- `afl_custom_queue_get` is added before checking `pending_favored`
- `afl_custom_queue_new_entry` is added at the end of `add_to_queue`
- Update the corresponding document and examples
- Update `.gitignore` to exclude `afl-clang-lto`, `afl-clang-lto++`, `afl-clang-lto.8`, and `afl-clang-lto++.8`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@ afl-clang
 afl-clang++
 afl-clang-fast
 afl-clang-fast++
+afl-clang-lto
+afl-clang-lto++
 afl-fuzz
 afl-g++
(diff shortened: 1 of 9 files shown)


andreafioraldi:
Great! merged

domenukk:
It might be f good to add some test cases for this to prevent regressions

h1994st:
@domenukk Yes, that would be great. For now, I manually run some tests for the custom mutator. I can think about integrating them into the test scripts.

PR #239, opened 2020-03-08
fix qemu persistent mode for 32-bit target on 64-bit host

0x1eadbead:
When running persistent qemu-mode with CPU_TARGET=i386 on x86_64 host like this:
```sh
AFL_DEBUG_CHILD_OUTPUT=1 AFL_QEMU_PERSISTENT_ADDR=$function_start_addr afl-fuzz -Q -i in -o out -- ./test @@
```
afl aborts with message:
  /AFLplusplus/qemu_mode/qemu-3.1.1/tcg/tcg-op.c:2646: tcg fatal error
  qemu: uncaught target signal 11 (Segmentation fault) - core dumped
This patch fixes the error

### DIFF
--- a/qemu_mode/patches/afl-qemu-cpu-translate-inl.h
+++ b/qemu_mode/patches/afl-qemu-cpu-translate-inl.h
@@ -35,7 +35,7 @@
 #include "tcg.h"
 #include "tcg-op.h"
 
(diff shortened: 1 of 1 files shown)


andreafioraldi:
thank you!

PR #241, opened 2020-03-08
Use select to monitor forkserver for timeouts instead of alarm

rish9101:


src/afl-forkserver.c
```diff
@@ -292,19 +293,31 @@ void afl_frk_srv_start(afl_forkserver_t *frk_srv, char **argv) {
...
-  setitimer(ITIMER_REAL, &it, NULL);
+      }
+
+    } else {
+
+      rlen = read(frk_srv->fsrv_st_fd, &status, 4);
+
+    }
+
+  }
```
domenukk: The case `exec_tmout == 0` no longer reads 4 bytes and will fail

src/afl-fuzz-cmplog.c
```diff
@@ -175,19 +178,32 @@ void init_cmplog_forkserver(afl_state_t *afl) {
 
   if (afl->frk_srv.exec_tmout) {
```
domenukk: Also here the `else` case is missing/broken (or can it never be 0?)

src/afl-fuzz-cmplog.c
```diff
@@ -175,19 +178,32 @@ void init_cmplog_forkserver(afl_state_t *afl) {
...
-  setitimer(ITIMER_REAL, &it, NULL);
+    int sret =
+        select(afl->cmplog_fsrv_st_fd + 1, &readfds, NULL, NULL, &timeout);
 
-  rlen = read(afl->cmplog_fsrv_st_fd, &status, 4);
+    if (sret == 0) {
 
-  it.it_value.tv_sec = 0;
-  it.it_value.tv_usec = 0;
+      if (afl->cmplog_forksrv_pid > 0) {
```
domenukk: This also always will be true.
Either remove the check or add an assert /FATAL to spot future bugs

src/afl-forkserver.c
rish9101: I think exec tmout cannot be zero. Because even if we dont pass it the `-t` option, it still takes a certain default timeout.

domenukk: What about `-t 0`?

src/afl-fuzz-run.c
```diff
@@ -44,6 +58,13 @@ u8 run_target(afl_state_t *afl, u32 timeout) {
      territory. */
 
   memset(afl->fsrv.trace_bits, 0, MAP_SIZE);
+  memset(&timer_signal_event, 0, sizeof(struct sigevent));
+
+  timer_signal_event.sigev_notify = SIGEV_THREAD;
+  timer_signal_event.sigev_notify_function = timeout_handle;
+  timer_signal_event.sigev_value.sival_ptr = &afl;
+  timer_create(CLOCK_REALTIME, &timer_signal_event, &timer);
```
domenukk: I think we should use `CLOCK_MONOTONIC` for this use case, see https://stackoverflow.com/a/3527632/1345238
Then we may even be able to do without `-lrt` but I'm not sure.

src/afl-forkserver.c
```diff
@@ -290,21 +291,25 @@ void afl_fsrv_start(afl_forkserver_t *fsrv, char **argv) {
 
   /* Wait for the fork server to come up, but don't wait too long. */
 
-  if (fsrv->exec_tmout) {
```
domenukk: Please add this again, see my other comment 🙃
Might be a handy feature for some.

vanhauser-thc:
I can do speed tests if the merge takes place latest tomorrow at lunch time

domenukk:
@vanhauser-thc You can do a speed test on this branch already if you find the time, only minor nitpicks are left.

src/afl-forkserver.c
rish9101: Doesn't work. It gives a timeout too low warning and exits  :sweat_smile:

PR #243, opened 2020-03-09
Remove alarm and add POSIX timers

rish9101:


domenukk:
I just tried and it doesn't work if the child immediately blocks. I think you set the timer up too early. At least in my tests the `child_pid` in the timer handler was 0, so the child never gets killed.

rish9101:
Hi, I don't understand. The timer is set at the same time as last time. Also, `child_pid=0` means the function was called in the child, right? How do I fix this? Also, what do you mean by child blocks immediately? The child exits immediately or waits for some time?

domenukk:
My target is a program that just calls sleep() for 30 seconds.
My expectation would be that afl recognizes the timeout, but instead it does nothing for 30 seconds.
I'm not sure why child id is 0 but for sure the timer should only be in the parent, after the fork.
Strace or gdb can give you further information

rish9101:
I think this might be due to the dry run that is done by the fuzzer. It runs the binary without any timeout initially I guess. I checked it with the previous timer(without POSIX) and it also hung. We'd have to change the behaviour of the dry run for this.

domenukk:
Ah thx didn't know. Still the timeout fired for some reason (?)

rish9101:
I finally figured out the problem. It's in the argument passing of the timeout_handler. It's basically running it in another thread, right? So, I cannot pass the `afl` struct directly to it. That is why it is showing the wrong pid in the handler. I'll have to see how to pass arguments to such handlers. Right now, everywhere what I'm seeing is to pass it as a global variable, but I don't think we should do that, right?

rish9101:
It's working as expected now. For binaries that block, the fuzzer quits due to a timeout now.

domenukk:
Great! 
Now the only thing missing are checks for return values of the timers.
You can use FATAL if something fails (for exapmle `FATAL("Could not set timer");`).

rish9101:
I've added the checks :)

domenukk:
Almost there, however the timers are still leaking.
I think we'll need to call one `timer_delete` for each create, see valgrind output below:
```
[...]
==168974==    by 0x487D5F9: timer_create (in /usr/lib/librt-2.31.so)
==168974==    by 0x116A39: run_target (afl-fuzz-run.c:182)
==168974==    by 0x117296: common_fuzz_stuff (afl-fuzz-run.c:900)
==168974==    by 0x12E4EB: fuzz_one_original (afl-fuzz-one.c:935)
==168974==    by 0x10F442: UnknownInlinedFun (afl-fuzz-one.c:4390)
==168974==    by 0x10F442: main (afl-fuzz.c:1073)
==168974== 
==168974== 2,904 bytes in 33 blocks are still reachable in loss record 13 of 15
==168974==    at 0x483977F: malloc (vg_replace_malloc.c:309)
==168974==    by 0x487D5F9: timer_create (in /usr/lib/librt-2.31.so)
==168974==    by 0x116A39: run_target (afl-fuzz-run.c:182)
==168974==    by 0x117296: common_fuzz_stuff (afl-fuzz-run.c:900)
==168974==    by 0x12E46F: fuzz_one_original (afl-fuzz-one.c:935)
==168974==    by 0x10F442: UnknownInlinedFun (afl-fuzz-one.c:4390)
==168974==    by 0x10F442: main (afl-fuzz.c:1073)
==168974== 
==168974== 3,520 bytes in 40 blocks are still reachable in loss record 14 of 15
==168974==    at 0x483977F: malloc (vg_replace_malloc.c:309)
==168974==    by 0x487D5F9: timer_create (in /usr/lib/librt-2.31.so)
==168974==    by 0x116A39: run_target (afl-fuzz-run.c:182)
==168974==    by 0x118DDE: calibrate_case (afl-fuzz-run.c:493)
==168974==    by 0x134668: perform_dry_run (afl-fuzz-init.c:473)
==168974==    by 0x10ECE5: main (afl-fuzz.c:994)
==168974== 
==168974== 129,712 bytes in 1,474 blocks are still reachable in loss record 15 of 15
==168974==    at 0x483977F: malloc (vg_replace_malloc.c:309)
==168974==    by 0x487D5F9: timer_create (in /usr/lib/librt-2.31.so)
==168974==    by 0x116A39: run_target (afl-fuzz-run.c:182)
==168974==    by 0x117296: common_fuzz_stuff (afl-fuzz-run.c:900)
==168974==    by 0x12AA0A: fuzz_one_original (afl-fuzz-one.c:2108)
==168974==    by 0x10F442: UnknownInlinedFun (afl-fuzz-one.c:4390)
==168974==    by 0x10F442: main (afl-fuzz.c:1073)
==168974== 
==168974== LEAK SUMMARY:
==168974==    definitely lost: 0 bytes in 0 blocks
==168974==    indirectly lost: 0 bytes in 0 blocks
==168974==      possibly lost: 272 bytes in 1 blocks
==168974==    still reachable: 147,312 bytes in 1,674 blocks
==168974==         suppressed: 0 bytes in 0 blocks
==168974== 
==168974== For lists of detected and suppressed errors, rerun with: -s
==168974== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
```

rish9101:
I've added the timer delete function :)

domenukk:
One more to go ;)
```
==199840== 
==199840== HEAP SUMMARY:
==199840==     in use at exit: 272 bytes in 1 blocks
==199840==   total heap usage: 10,612 allocs, 10,611 frees, 1,835,972 bytes allocated
==199840== 
==199840== 272 bytes in 1 blocks are possibly lost in loss record 1 of 1
==199840==    at 0x483BB65: calloc (vg_replace_malloc.c:762)
==199840==    by 0x40133E1: allocate_dtv (in /usr/lib/ld-2.31.so)
==199840==    by 0x4013D51: _dl_allocate_tls (in /usr/lib/ld-2.31.so)
==199840==    by 0x4C8B15E: pthread_create@@GLIBC_2.2.5 (in /usr/lib/libpthread-2.31.so)
==199840==    by 0x487E75C: ??? (in /usr/lib/librt-2.31.so)
==199840==    by 0x4C9292E: __pthread_once_slow (in /usr/lib/libpthread-2.31.so)
==199840==    by 0x487D5C2: timer_create (in /usr/lib/librt-2.31.so)
==199840==    by 0x116A49: run_target (afl-fuzz-run.c:182)
==199840==    by 0x118DFE: calibrate_case (afl-fuzz-run.c:495)
==199840==    by 0x134A58: perform_dry_run (afl-fuzz-init.c:473)
==199840==    by 0x10ECF5: main (afl-fuzz.c:994)
==199840== 
==199840== LEAK SUMMARY:
==199840==    definitely lost: 0 bytes in 0 blocks
==199840==    indirectly lost: 0 bytes in 0 blocks
==199840==      possibly lost: 272 bytes in 1 blocks
==199840==    still reachable: 0 bytes in 0 blocks
==199840==         suppressed: 0 bytes in 0 blocks
==199840== 
==199840== For lists of detected and suppressed errors, rerun with: -s
```

domenukk:
Also it might be better/quicker to keep one timer in the `afl_state_t` for each use-case instead of creating and destroying them all the time, what do you think?
Less syscalls. No idea about performance in this case, though.

rish9101:
I think we can put the `timer_id` and the `sigevent` struct in `afl_state_t`, that might reduce the memory leaks. Yes, I can see the performance changes it causes( By putting the timer_create code somewhere else), I'll try it.

vanhauser-thc:
travis is working again so restarting the travis check for this PR to check if it is safe :)

vanhauser-thc:
It seems this change makes afl-fuzz slightly faster (1%)

PR #252, opened 2020-03-11
Move afl-fuzz related env variables into afl_state_t

rish9101:


src/afl-fuzz-init.c
```diff
@@ -1486,7 +1486,7 @@ void check_crash_handling(void) {
       "    sudo launchctl unload -w ${SL}/LaunchDaemons/${PL}.Root.plist\n");
 
 #endif
-  if (!get_afl_env("AFL_I_DONT_CARE_ABOUT_MISSING_CRASHES"))
+  if (!)get_afl_env("AFL_I_DONT_CARE_ABOUT_MISSING_CRASHES")
```
domenukk: This looks ... Odd.

src/afl-fuzz-globals.c
```diff
@@ -75,7 +76,7 @@ list_t afl_states = {0};
 
 /* Initializes an afl_state_t. */
 
-void afl_state_init(afl_state_t *afl) {
+void afl_state_init(afl_state_t *afl, char **envp) {
```
domenukk: We can think about moving this out of the init but it's fine if we do it later, also

src/afl-fuzz-init.c
rish9101: Ah. Didn't notice it. I'm sorry. I'll correct it. :sweat_smile:

src/afl-fuzz-globals.c
rish9101: I thought of making it a new function. I can do that if you say. :)

rish9101:
Btw, if these changes are okay. I'll probably move the other environment variables (for afl-analyze, afl-tmin etc) out into appropriate structs too.

src/afl-fuzz-globals.c
```diff
@@ -119,6 +120,221 @@ void afl_state_init(afl_state_t *afl) {
...
+
+void set_afl_environment(afl_state_t *afl, char **envp) {
+
+  int   index = 0, found = 0;
+  char *env;
+  while ((env = envp[index++]) != NULL) {
+
+    if (strncmp(env, "ALF_", 4) == 0) {
+
+      WARNF("Potentially mistyped AFL enqqqvironment variable: %s", env);
```
domenukk: Can you change this typo? :)

src/afl-fuzz.c
```diff
@@ -762,22 +762,22 @@ int main(int argc, char **argv_orig, char **envp) {
 
   }
 
-  if (get_afl_env("AFL_NO_FORKSRV")) afl->no_forkserver = 1;
-  if (get_afl_env("AFL_NO_CPU_RED")) afl->no_cpu_meter_red = 1;
-  if (get_afl_env("AFL_NO_ARITH")) afl->no_arith = 1;
-  if (get_afl_env("AFL_SHUFFLE_QUEUE")) afl->shuffle_queue = 1;
-  if (get_afl_env("AFL_FAST_CAL")) afl->fast_cal = 1;
+  if (afl->afl_env.afl_no_forksrv) afl->no_forkserver = 1;
```
domenukk: I think all of these are redundant.
Of they are only read here onc, we might not need a second .env variable 🤔 what's everybody's opinion on this?

src/afl-fuzz-globals.c
rish9101: Done. Sorry for this :sweat_smile:

src/afl-fuzz.c
rish9101: That is true. Then, we'd have to add flags for it, right? Or, can we remove the variables from afl_state and keep them in env_struct?

domenukk: Well in these cases (`no_arith`, `shuffle_queue`, `fast_cal`, `no_forkserver`) there are already some flags.
I'd suggest we leave those out of the env struct for now, then decide what to do for the others later.

domenukk:
@andreafioraldi Any idea, where `[!] afl-fuzz is not working correctly with qemu_mode compcov` comes from in Travis? I don't see any changes that could have affected this.

src/afl-fuzz.c
```diff
@@ -681,8 +683,6 @@ int main(int argc, char **argv_orig, char **envp) {
         "Using -M master with the AFL_CUSTOM_MUTATOR_ONLY mutator options will "
         "result in no deterministic mutations being done!");
 
-  check_environment_vars(envp);
```
domenukk: What did this do and where did it go?

domenukk:
@hexcoder- this would be an important step to get everything running in OpenBSD again, however I have no clue why Travis fails. Maybe you have an idea :)

domenukk:
Ah no nevermind I mixed different things up 🙃 but still I have no clue

vanhauser-thc:
> @andreafioraldi Any idea, where `[!] afl-fuzz is not working correctly with qemu_mode compcov` comes from in Travis? I don't see any changes that could have affected this.

My guess it was just unlucky that cmplog could not solve the test in time. I restarted that build check

vanhauser-thc:
yes that fixed travis

vanhauser-thc:
yes that fixed travis

PR #254, opened 2020-03-12
Update README.md add install of llvm package

ChakshuGupta:
The build and install steps are missing the installation of "llvm" package which is a pre-requisite for building llvm_mode.

Added the "llvm" package in the installation line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ afl++ has many build options.
 The easiest is to build and install everything:
 
 ```shell
-$ sudo apt install build-essential libtool-bin python3 automake bison libglib2.0-dev libpixman-1-dev clang python-setuptools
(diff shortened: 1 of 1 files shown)


domenukk:
Thanks! :)
It's technically not necessary for AFL itself, though, so I'm not sure about listing it here.
I'll wait for the other's optinions.

vanhauser-thc:
for source code as a good part of what makes afl++ amazing is the llvm_mode stuff so it should be added. The issue is however - how? llvm9 is what should be installed however non-current distros (e.g. Ubuntu 16.04, not sure about 18.04) does not have it.

ChakshuGupta:
Okay. noted.

vanhauser-thc:
I will add a text line to the README, not apt install command, I dont see another way.

PR #259, opened 2020-03-15
Just compute the current len once.

devnexen:
Fix env var types assignments.


vanhauser-thc:
makes sense! thanks! can you please adapt to the fixed I made in between? a pull from the master should suffice

PR #270, opened 2020-03-24
afl-whatsup "time without finds" stat

### LINKED ISSUE #197: Feature request: stop condition
Thank you for your awesome work guys!

I use AFL++ a lot and i think it would be great if it had optional condition to stop fuzzing jobs after certain amount of minutes (hours) without new paths discovered.

For example: stop all fuzzing jobs if **each** status screen shows at least 30 minutes without new paths.

fuzzah:
Fixes #197 
Now afl-whatsup also prints time since the most recent path has been found.
One can then do some greps to check if required time has passed across all the afl-fuzz instances to perform some actions.
Example:
```
afl-whatsup out | grep "Time without finds: 3 hours"
```

### DIFF
--- a/afl-whatsup
+++ b/afl-whatsup
@@ -122,6 +122,7 @@ fmt_duration()
 
 FIRST=true
 TOTAL_WCOP=
+TOTAL_LAST_PATH=0
 
 for i in `find . -maxdepth 2 -iname fuzzer_stats | sort`; do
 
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
but isnt that the opposite of what you want?
if you have 4 fuzzers, three finding new paths all the time, but one not finding anything - then you would want to abort? I thought you would want to know when the last path ever was from a longer time onwards. (so basically a `-lt` rather than `-gt` in your code)

vanhauser-thc:
damn wrote to the wrong PR :) but same issue

vanhauser-thc:
And no. -gt means newer. I am exhausted from performing an online training :) you are correct

PR #271, opened 2020-03-24
add "time without finds" to afl-whatsup

### LINKED ISSUE #197: Feature request: stop condition

fuzzah:
Fixes #197 
Now afl-whatsup also prints time since the most recent path has been found across all synchronized afl-fuzz instances.
One can then do some greps to check if required time has passed to perform some actions.
Example:
```
afl-whatsup out | grep "Time without finds: 3 hours"
```

### DIFF
--- a/afl-whatsup
+++ b/afl-whatsup
@@ -122,6 +122,7 @@ fmt_duration()
 
 FIRST=true
 TOTAL_WCOP=
+TOTAL_LAST_PATH=0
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thanks!

PR #282, opened 2020-03-28
Multiple custom mutators

rish9101:
This is a WIP pull request(not ready yet), in order to integrate multiple custom mutators in the fuzzer. I'll continue working on this and feedback related to it is valuable. :smile_cat:


include/afl-fuzz.h
```diff
@@ -615,6 +615,7 @@ typedef struct afl_state {
 
   u8 *   ex_buf;
   size_t ex_size;
+  list_t * list_mutators;
```
domenukk, line 618: Why allocate this dynamically? Use `list_t list_mutators` instead.

src/afl-fuzz-mutators.c
```diff
@@ -26,27 +26,43 @@
...
+    while (fn_token)
+    {
+      mut = load_custom_mutator(afl, fn_token);
+      list_append(afl->list_mutators, mut);
+      fn_token = strsep(&fn, ";");
 
-    return;
+    }
+
+    mut = get_head(afl->list_mutators)->prev->data;
```
domenukk, line 64: Make use of the functions and defines in list.h (`list_append` and later `LIST_FOREACH`)

include/afl-fuzz.h
rish9101: OK. Will make the changes.

rish9101:
I've made a few more changes but I have a few questions regarding the mutators. In `afl-fuzz-onee` during havoc stage, a `stacked custom probability` is calculated using the custom mutator, how would this be calculated when there are multiple mutators? Also, in `trim_test_case` should we trim the test case with each of the custom mutator's trimming function? Similarly, how  should `custom_pre_save` be done now?

include/afl-fuzz.h
domenukk: You can see usage of list.h in the test case https://github.com/AFLplusplus/AFLplusplus/blob/fix/list/test/unittests/unit_list.c
(it still fails for long lists, see issue #291 - but will be fine for this use case)

domenukk:
I think each probablility should be calculated individually.
All mutator actions can just be done one after the other, in order of custom mutators listed for now (foreach...) until we think of something smarter. cc @andreafioraldi

andreafioraldi:
Yes each custom mutator has its own probability. use an array not a list for all the things, we can support multiple mutators but not infinite, set a bound to e.g. 8 mutators (maybe use a define).

subho-13:
We can use the same strategy as MOpt-AFL. The implementation is rather easy and it can automatically adjust the probabilities.

domenukk:
@rish9101 any progress on this?

rish9101:
I'm a bit busy right now with some online college work, (I've informed Marc about the same) but I'll start working on it(and the framework too) from 20 or 22nd. Sorry for the delay guys.

rish9101:
For `afl-python` mutators, should I make similar changes and simply add them to the list of mutators?

domenukk:
> For `afl-python` mutators, should I make similar changes and simply add them to the list of mutators?

Python mutators... I guess the python mutator could be one entry in the normal mutators list? That would need a bit of refactoring though, so I'm not sure.

rish9101:
> I mean, it's a bit hard to understand and use, but If you do use the `list_t`, you don't have an upper max on the custom mutator count.
> Maybe this testcase makes the usage more clear:
> https://github.com/AFLplusplus/AFLplusplus/blob/dev/test/unittests/unit_list.c

I thought we wanted to limit the number of custom mutators to some limit. As mentioned here

> Yes each custom mutator has its own probability. use an array not a list for all the things, we can support multiple mutators but not infinite, set a bound to e.g. 8 mutators (maybe use a define).

That is why I changed the list type to an array and defined an upper limit (`MAX_MUTATORS_COUNT`) :sweat_smile:

rish9101:
> > For `afl-python` mutators, should I make similar changes and simply add them to the list of mutators?
> 
> Python mutators... I guess the python mutator could be one entry in the normal mutators list? That would need a bit of refactoring though, so I'm not sure.

Yes, I wasn't sure about how to approach this. Because python mutators are quite different from the regular ones.

domenukk:
> I thought we wanted to limit the number of custom mutators to some limit. As mentioned here
> 

Ah I see. Ask 5 people, get 5 opinions :D
I feel like limiting the number of mutators is a premature optimization, the list makes sure everything stays on the same page so it _shouldn't_ be much slower. Anyway no need to do additional coding for now, if it works.


> Yes, I wasn't sure about how to approach this. Because python mutators are quite different from the regular ones.
Since they got refactored to have the same api, they are not too different anymore - each python mutator function is wrapped by a thin layer only.
Either way, I guess it's fine if we just support one python mutator for now.

How's the overall status of the PR?

rish9101:
> > I thought we wanted to limit the number of custom mutators to some limit. As mentioned here
> 
> Ah I see. Ask 5 people, get 5 opinions :D
> I feel like limiting the number of mutators is a premature optimization, the list makes sure everything stays on the same page so it _shouldn't_ be much slower. Anyway no need to do additional coding for now, if it works.
> 
> > Yes, I wasn't sure about how to approach this. Because python mutators are quite different from the regular ones.
> 
> Since they got refactored to have the same api, they are not too different anymore - each python mutator function is wrapped by a thin layer only.
> Either way, I guess it's fine if we just support one python mutator for now.
> 
> How's the overall status of the PR?

I think I only need to add the python mutator support now. I don't think any thing else is left in this.

vanhauser-thc:
does this include the post module support? post module should be integrated with custom module (it is just one function that fits well in there)

domenukk:
> does this include the post module support? post module should be integrated with custom module (it is just one function that fits well in there)

No it does not.
The custom mutators already have some kind of pre-write hook but that is oddly at a different position as the post library hook..

src/afl-fuzz-mutators.c
```diff
@@ -82,166 +103,125 @@ void setup_custom_mutator(afl_state_t *afl) {
...
   /* Initialize the custom mutator */
-  if (afl->mutator->afl_custom_init) {
+  if (mutator->afl_custom_init)
+    mutator->data = 
+      mutator->afl_custom_init(afl, rand_below(afl, 0xFFFFFFFF));
 
-    afl->mutator->data =
-        afl->mutator->afl_custom_init(afl, rand_below(afl, 0xFFFFFFFF));
+  mutator->stacked_custom = (mutator && mutator->afl_custom_havoc_mutation);
+  mutator->stacked_custom_prob = 6; // like one of the default mutations in havoc
```
domenukk, line 218: Should the custom mutator maybe have the possibility to decide on its own probability? Open question

src/afl-fuzz-one.c
```diff
@@ -384,17 +384,17 @@ u8 fuzz_one_original(afl_state_t *afl) {
...
     /* The custom mutator will decide to skip this test case or not. */
 
-    if (!afl->mutator->afl_custom_queue_get(afl->mutator->data,
-                                            afl->queue_cur->fname)) {
+    LIST_FOREACH(&afl->custom_mutator_list, struct custom_mutator, {
 
-      return 1;
-
-    }
+      if (el->afl_custom_queue_get && !el->afl_custom_queue_get(el->data, afl->queue_cur->fname)) {
```
domenukk, line 393: Does it ever make sense to specify multiple custom queues? I'm not 100% sure how they are supposed to work.

src/afl-fuzz-run.c
```diff
@@ -65,21 +65,40 @@ void write_to_testcase(afl_state_t *afl, void *mem, u32 len) {
...
+  if (unlikely(afl->number_of_custom_mutators)) {
 
     u8 *new_buf = NULL;
+    ssize_t new_size;
+
+    LIST_FOREACH(&afl->custom_mutator_list, struct custom_mutator, {
+
+      if (el->afl_custom_pre_save) {
+        new_size = el->afl_custom_pre_save(
+          el->data, mem, len, &new_buf
```
domenukk, line 79: next round should get `new_buf` and `new_size` of last round.
Or do we only want to allow a single valid `pre_save` hook?

src/afl-fuzz-run.c
```diff
@@ -488,10 +507,23 @@ void sync_fuzzers(afl_state_t *afl) {
...
+  if (afl->number_of_custom_mutators) {
+
+    u8 trimmed_case;
+    bool custom_trimmed = false;
+
+    LIST_FOREACH(&afl->custom_mutator_list, struct custom_mutator, {
+
+      if (el->afl_custom_trim) {
+        
+        trimmed_case = trim_case_custom(afl, q, in_buf, el);
```
domenukk, line 519: Not sure how often trim should be called/if they should be called all the time?

src/afl-fuzz-mutators.c
```diff
@@ -26,27 +26,46 @@
...
+      afl->number_of_custom_mutators++;
 
-    return;
+    } else {
+
+      while (fn_token) {
+
+        mutator = load_custom_mutator(afl, fn_token);
+        list_append(&afl->custom_mutator_list, mutator);
+        afl->number_of_custom_mutators++;
```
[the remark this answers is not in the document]
rish9101: Hi, Dominik, what do you mean by "this never wraps around". couldn't get that :sweat_smile:

src/afl-fuzz-mutators.c
rish9101: That actually happens I think, the mutators `stacked_custom_prob` is actually calculated before the `havoc stage` starts.

src/afl-fuzz-run.c
rish9101: Isn't trim called only once for each test case we have?

src/afl-fuzz-mutators.c
```diff
@@ -82,166 +103,125 @@ void setup_custom_mutator(afl_state_t *afl) {
...
+      if (el->pre_save_buf) {
+        ck_free(el->pre_save_buf);
+        el->pre_save_buf = NULL;
+        el->pre_save_size = 0;
+      }
 
-    }
+    } );
 
-    ck_free(afl->mutator);
```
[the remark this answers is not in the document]
rish9101: I think `LIST_FOREACH_CLEAR` takes care of both of these things? I mean the `afl->mutator` member wouldn't be needed after this custom_mutator implementation. (Unless we want a "default" custom mutator out of the given ones?)

src/afl-fuzz-mutators.c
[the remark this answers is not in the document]
domenukk: If you pass 256 mutators or more, `number_of_custom_mutators` will wrap around (as it's larger than u8).
Would break a lot of things in this case.

src/afl-fuzz-mutators.c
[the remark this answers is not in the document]
domenukk: `LIST_FOREACH_CLEAR` doesn't free anything you allocced earlier, it only clears this part of the list. It merely removes the containig datastructure (`el_box`)
As you do a:
`struct custom_mutator *mutator = ck_alloc(sizeof(struct custom_mutator));`
you'll have to `ck_free(el);` inside the `LIST_FOREACH_CLEAR` loop yourself.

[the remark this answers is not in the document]
domenukk: See the foreach implementation here btw 
https://github.com/AFLplusplus/AFLplusplus/blob/ce9c6df45639db922feafb55259e20c7e82af0bf/include/list.h#L101

src/afl-fuzz-run.c
domenukk: I guess. Still, you call one trim for each mutator, and then only return the result of the last in case it worked?
I don't think this works for n>1 mutators, what do you think?

src/afl-fuzz-mutators.c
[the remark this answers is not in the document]
vanhauser-thc: then make it a u32? :)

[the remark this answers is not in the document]
rish9101: Yes. Then we should make `number_of_custom_mutators` as u32.

src/afl-fuzz-one.c
vanhauser-thc: the function type decides if a fuzzing of the queue entry should be performed or not. if any of them would say no it should be skipped.

src/afl-fuzz-run.c
rish9101: Yes. I mean this and the `custom_pre_save` thing could be solved by making the first custom_mutator as "defualt" mutator? Then these heavy operations (like `trimming` and `pre_save`) could only be defined in that one mutator. What do you think?

src/afl-fuzz-mutators.c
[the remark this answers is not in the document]
domenukk: Then you should still check if the `u32` wraps (practically impossible but it's always better to do these things explicitly in C, IMHO)...
Just check if the value is smaller than before -> `FATAL`.
See also:
https://wiki.sei.cmu.edu/confluence/display/c/INT30-C.+Ensure+that+unsigned+integer+operations+do+not+wrap

src/afl-fuzz-run.c
vanhauser-thc: it returns 1 if there was trimming, right? if so it should return one if any of the custom trimmer trimmed the entry

src/afl-fuzz-one.c
domenukk: Ah I see, thx. By the name of the function I thought the mutator would implement its own queue...
It's more like a selector/filter function then. In that case, stacking them make sense, also.

src/afl-fuzz-mutators.c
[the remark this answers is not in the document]
vanhauser-thc: pointless, cant happen. the environment variable memory is not big enough to hold so many :)

src/afl-fuzz-one.c
vanhauser-thc: docs/custom_mutators.md ;)

src/afl-fuzz-run.c
vanhauser-thc: IMHO if a user wants different trimmers he should have them

rish9101: > it returns 1 if there was trimming, right? if so it should return one if any of the custom trimmer trimmed the entry

Yes, that is what is happening right now. :slightly_smiling_face:

domenukk: Just looked it up in the code - The return values are... odd.
However I think the current code is correct, then, the only thing is it could return earlier on errors.

src/afl-fuzz-mutators.c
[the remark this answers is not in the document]
domenukk: Yes, you are right - but also that's the C coder's mindset that makes fuzzers necessary in the first place :P

src/afl-fuzz-run.c
rish9101: Doesn't the `trim_case_custom` function handle this? I mean it exits on any error in trimming, e,g if the new corpus size increases than before

src/afl-fuzz-mutators.c
```diff
@@ -119,8 +117,6 @@ void destroy_custom_mutator(afl_state_t *afl) {
```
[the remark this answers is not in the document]
rish9101: > Still needs
> 
> ```c
> ck_free(el);
> ```

Yes I've added it just now. :slightly_smiling_face:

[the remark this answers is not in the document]
domenukk: Nice thx :)

src/afl-fuzz-python.c
```diff
@@ -295,88 +295,94 @@ void deinit_py(void *py_mutator) {
...
+struct custom_mutator * load_custom_mutator_py(afl_state_t *afl, char *module_name) {
 
-  afl->mutator = ck_alloc(sizeof(struct custom_mutator));
-  afl->mutator->pre_save_buf = NULL;
-  afl->mutator->pre_save_size = 0;
+  struct custom_mutator * mutator;
 
-  afl->mutator->name = module_name;
+  mutator = ck_alloc(sizeof(struct custom_mutator));
+  mutator->pre_save_buf = NULL;
```
domenukk, line 303: Shouldn't this be part of the python mutator struct? (And also be cleare there?) It's not used anywhere else.

domenukk:
What's there to be done? Should we merge?

rish9101:
> What's there to be done? Should we merge?

Yes, I don't think anything is left in this. :sweat_smile:

PR #288, opened 2020-03-31
llvm_mode: using MaybeAlign wrapper over the deprecated setter.

devnexen:
seems to be available even on LLVM 3.7

### DIFF
--- a/llvm_mode/afl-llvm-lto-instrumentation.so.cc
+++ b/llvm_mode/afl-llvm-lto-instrumentation.so.cc
@@ -378,7 +378,7 @@ bool AFLLTOPass::runOnModule(Module &M) {
         M, Int32Ty, true, GlobalValue::ExternalLinkage, 0, "__afl_final_loc", 0,
         GlobalVariable::GeneralDynamicTLSModel, 0, false);
     ConstantInt *const_loc = ConstantInt::get(Int32Ty, afl_global_id);
-    AFLFinalLoc->setAlignment(4);
+    AFLFinalLoc->setAlignment(MaybeAlign(4));
     AFLFinalLoc->setInitializer(const_loc);
 
   }


vanhauser-thc:
@hexcoder- what do you think? your speciality field :)

hexcoder-:
Thanks David!

hexcoder-:
Oops, now with clang 9.0.1 I get
```
/usr/bin/clang++ `llvm-config --cxxflags` -Wl,-znodelete -fno-rtti -fpic -O3 -funroll-loops -D_FORTIFY_SOURCE=2 -Wall -g -I ../include/ -DVERSION=\"++2.62d\" -Wno-variadic-macros -Wno-writable-strings -fno-rtti -fPIC -std=gnu++11 -shared afl-llvm-lto-instrumentation.so.cc MarkNodes.cc -o ../afl-llvm-lto-instrumentation.so `llvm-config --ldflags` 
afl-llvm-lto-instrumentation.so.cc:381:31: error: use of undeclared identifier 'MaybeAlign'
    AFLFinalLoc->setAlignment(MaybeAlign(4));
                              ^
1 error generated.
make[1]: *** [GNUmakefile:285: ../afl-llvm-lto-instrumentation.so] Error 1
```
Is there some include missing?

vanhauser-thc:
there is an include missing, but one that does not exist in llvm9 as this is a llvm10 feature .. removing it ...

hexcoder-:
For an april fool a bit early...

hexcoder-:
But how could travis succeed then?????

vanhauser-thc:
its available on older and new llvm, but not some in between. weird but true. 

Also LTO is only for llvm >= 9, and from llvm 10 it is back. the highest we have in travis is llvm 8 I think

hexcoder-:
Thanks, yes.

domenukk:
Can we add a travis target with current llvm to be on the safe side?

PR #293, opened 2020-04-07
Snapshot for LLVM mode

andreafioraldi:
snapshot is automatically used when the module is loaded. This patch does nothing if the module is not loaded, IMO we can merge it.


vanhauser-thc:
can we first merge to dev and then merge everything to master for the release?
We have a little bit of time as the new cmin bug needs to be investigated anyway.

also, I would prefer something in the UI to show that snapshot is being used (e.g. the word SNAP in green in the top line or beside the CPU usage). otherwise one does not know if the lkm feature is used. (e.g. someone might accidently use an old compiled target or the lkm is not loaded, etc.)

PR #295, opened 2020-04-07
Android build fix proposal.

devnexen:
LTO flag is recognised but however broken on Android (tested with armv7 arch).
Thus giving the choice not to enable it.
In fortify mode, open required O_CREAT or O_TMPFILE when mode is set.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -36,6 +36,7 @@ SH_PROGS    = afl-plot afl-cmin afl-cmin.bash afl-whatsup afl-system-config
 MANPAGES=$(foreach p, $(PROGS) $(SH_PROGS), $(p).8) afl-as.8
 ASAN_OPTIONS=detect_leaks=0
 
+ifeq "$(findstring android, $(shell $(CC) --version 2>/dev/null))" ""
 ifeq "$(shell echo 'int main() {return 0; }' | $(CC) $(CFLAGS) -Werror -x c - -flto=full -o .test 2>/dev/null && echo 1 || echo 0 ; rm -f .test )" "1"
 	CFLAGS_FLTO ?= -flto=full
 else
@@ -46,6 +47,7 @@ else
 	CFLAGS_FLTO ?= -flto
   endif
  endif
+endif
 endif
 
 ifneq "$(shell uname)" "Darwin"
--- a/src/afl-fuzz-extras.c
+++ b/src/afl-fuzz-extras.c
@@ -451,7 +451,7 @@ void load_auto(afl_state_t *afl) {
     u8 *fn = alloc_printf("%s/.state/auto_extras/auto_%06u", afl->in_dir, i);
     s32 fd, len;
 
-    fd = open(fn, O_RDONLY, 0600);
+    fd = open(fn, O_RDONLY);
 
     if (fd < 0) {
 
--- a/src/afl-showmap.c
+++ b/src/afl-showmap.c
@@ -154,7 +154,8 @@ static u32 write_results_to_file(afl_forkserver_t *fsrv, u8 *outfile) {
 
   if (!strncmp(outfile, "/dev/", 5)) {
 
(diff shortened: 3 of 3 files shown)


src/afl-showmap.c
```diff
@@ -154,7 +154,11 @@ static u32 write_results_to_file(afl_forkserver_t *fsrv, u8 *outfile) {
 
   if (!strncmp(outfile, "/dev/", 5)) {
 
+#if !defined(__ANDROID__)
     fd = open(outfile, O_WRONLY, 0600);
+#else
+    fd = open(outfile, O_WRONLY | O_CREAT, 0600);
```
domenukk, line 160: Can we always use `O_CREAT` here, then?

devnexen: Hmmm no ... mistake of me.

GNUmakefile
```diff
@@ -36,6 +36,7 @@ SH_PROGS    = afl-plot afl-cmin afl-cmin.bash afl-whatsup afl-system-config
 MANPAGES=$(foreach p, $(PROGS) $(SH_PROGS), $(p).8) afl-as.8
 ASAN_OPTIONS=detect_leaks=0
 
+ifndef AFL_NO_LTO
```
vanhauser-thc, line 39: instead of this - is it possible to do ifneq "$(shell uname)" "Android" or something similar instead? Or check on a file that exists only on Android? Otherwise we need to document this somewhere and this makes things just more messy.
the more auto magic the better.

src/afl-showmap.c
```diff
@@ -154,7 +154,11 @@ static u32 write_results_to_file(afl_forkserver_t *fsrv, u8 *outfile) {
 
   if (!strncmp(outfile, "/dev/", 5)) {
 
+#if !defined(__ANDROID__)
     fd = open(outfile, O_WRONLY, 0600);
+#else
```
vanhauser-thc, line 159: Anyone knows why this has the permissions/mode is in there?
On normal Linux it is only for O_CREAT operations and otherwise ignored. and this has no O_CREAT.
@hexcoder-  - does *BSD need that?

GNUmakefile
devnexen: We re cross building here but I can try a little "assumption" here instead.

src/afl-showmap.c
devnexen: seemingly not

GNUmakefile
```diff
@@ -36,6 +36,7 @@ SH_PROGS    = afl-plot afl-cmin afl-cmin.bash afl-whatsup afl-system-config
 MANPAGES=$(foreach p, $(PROGS) $(SH_PROGS), $(p).8) afl-as.8
 ASAN_OPTIONS=detect_leaks=0
 
+ifeq "$(findstring android, $(shell $(CC) --version 2>/dev/null))" ""
```
vanhauser-thc, line 39: this looks like a good solution to me! :)

vanhauser-thc:
looks good for me :)

src/afl-showmap.c
hexcoder-: BSD needs permissions only for O_CREAT, too. Does not make sense otherwise IMO.

src/afl-fuzz-extras.c
```diff
@@ -451,7 +451,11 @@ void load_auto(afl_state_t *afl) {
     u8 *fn = alloc_printf("%s/.state/auto_extras/auto_%06u", afl->in_dir, i);
     s32 fd, len;
 
+#if !defined(__ANDROID__)
     fd = open(fn, O_RDONLY, 0600);
+#else
+    fd = open(fn, O_RDONLY);
```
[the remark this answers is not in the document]
hexcoder-: no, makes no sense.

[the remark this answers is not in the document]
hexcoder-: I mean remove it.

PR #299, opened 2020-04-08
fix limits.h include

vanhauser-thc:


domenukk:
Why does it make a difference?
Also, I was under the impression it's clean C code to put angle bracket includes before (local) quoted ones (?)

vanhauser-thc:
I forgot the 4th.
but the change was only necessary in one. I think the reason is that afl-fuzz.h includes the _GNU_SOURCE define, and if limits.h is included before afl-fuzz it is not there and then function definitions are not available.

PR #300, opened 2020-04-08
Integer swapping mutation in havoc

subho-13:
Added a mutation rule in surgical_havoc_mutate

Function :
Swap 2 remote chunks in the same data buffer. This is inspired by crossover mutation within the same data buffer. The mutation is unique.

Performance:
There has been No Degradation of performance due to the addition of the custom rule. The following graph proves the same .

For creating the graph, we take an output buffer of size varying from 3 to 1000. For each out_buf, we run the surgical_havoc_mutate for 200 times, one time without mutation, one time with mutation.
filename

### DIFF
--- a/examples/custom_mutators/custom_mutator_helpers.h
+++ b/examples/custom_mutators/custom_mutator_helpers.h
@@ -27,7 +27,7 @@ static void surgical_havoc_mutate(u8 *out_buf, s32 begin, s32 end) {
   static s16 interesting_16[] = {INTERESTING_8, INTERESTING_16};
   static s32 interesting_32[] = {INTERESTING_8, INTERESTING_16, INTERESTING_32};
 
-  switch (RAND_BELOW(12)) {
+  switch (RAND_BELOW(13)) {
 
     case 0: {
 
@@ -277,7 +277,8 @@ static void surgical_havoc_mutate(u8 *out_buf, s32 begin, s32 end) {
       if(end - begin <= 8) break;
       u32 offset_1 = RAND_BELOW((end - begin -8)) / 2;
       u32 offset_2 = (offset_1 ^ 0xffadeb)%(end - begin -8);
-
+      
+      /* If the offsets are different swap */
       if(offset_1 != offset_2) {
           SWAPQ((*(u64 *)(out_buf + offset_1)), (*(u64 *)(out_buf + offset_2)));
       }      
--- a/examples/custom_mutators/custom_mutator_helpers.h
+++ b/examples/custom_mutators/custom_mutator_helpers.h
@@ -272,6 +272,17 @@ static void surgical_havoc_mutate(u8 *out_buf, s32 begin, s32 end) {
       break;
 
     }
+    case 12: {
+      /* Swap 2 random qwords in the out_buf */
+      if(end - begin <= 8) break;
+      u32 offset_1 = RAND_BELOW((end - begin -8)) / 2;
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
can you explain more? if this is a general good mutation then maybe this should rather be in src/afl-fuzz-one.c? how does this improve edge discovery?

also never commit to master, only commit to dev. I will fix this.

waiting for @andreafioraldi or @domenukk to say something about this.

domenukk:
Sorry, but I don't really understand the `size of out_buffer` metric.
To see if this is actually beneficial, some actual use case, or at a plot of the coverage over time would be appreciated. :)
That being said, we could merge this and add it to one of the fuzzbench use-cases?

subho-13:
The size of out_buffer was used for 2 reasons :
1. Checking the stability of the new addition for the new mutation strategy.
2. The time taken for the swapping operation, if we increase the size of the output buffer.

I'm trying to get a plot of coverage with/without the new rule. I'll attach them here, if as soon as I get the graphs.

andreafioraldi:
Basically the added code is:

```c
    case 12: {
      /* Swap 2 random qwords in the out_buf */
      if(end - begin <= 8) break;
      u32 offset_1 = RAND_BELOW((end - begin -8)) / 2;
      u32 offset_2 = (offset_1 ^ 0xffadeb)%(end - begin -8);

      /* If the offsets are different swap */
      if(offset_1 != offset_2) {
          SWAPQ((*(u64 *)(out_buf + offset_1)), (*(u64 *)(out_buf + offset_2)));
      }      
      break;  
  }
```

You don't have to put this here, but int he fuzzer code (afl-fuzz-one.c), this is just an helper header that custom mutator developers can use.

Evaluating a change in the default mutator is hard. You need to benchmark a lot of targets.
So do it for 2/3 targets, and if the results are good we will benchmark it more exaustively in fuzzbench.

For coverage collection of the queue you can use afl-cov or my https://github.com/andreafioraldi/afl-qemu-cov is you want to profile uninstrumented binaries.

andreafioraldi:
In general, I think this can be ok for structured formats, maybe randomly choosing the size of the integer to swap and not swap always a qword

subho-13:
I'll try to gather the testcase coverage as much as I can. But, I may need some help with the tools(afl-cov or afl-qemu-cov). Secondly, we can obviously may the size variable, but it's effective on performance needs to be checked.

subho-13:
Can you suggest some libraries I can use for fuzzing?

andreafioraldi:
libpng, libjpeg-turbo, take libraries from OSS-Fuzz

subho-13:
Okay, I'm on it.

domenukk:
Maybe we can add a likelyness to have functions?

subho-13:
Actually, I was thinking about that. We can bias the selection of havoc functions using afl feedback. So, we can add some priority points if the tool was able to find new paths and redact some if no paths are found. So, the process of choosing the havoc mutations are still random, but with added bias.

subho-13:
Also, I'm trying to benchmark the Integer swapping, using  tools from binutils. 
I'm using the afl-fuzz in dumb mode for a period of 12 hours, using the same inputs and using "fast" schedule. 
After that, I'm taking the output and using afl-qemu-cov to check the number of paths found. The one which is able to find more paths is more effective.
Am I doing it right here?

subho-13:
I've opened a new pull request for this, because the files changed were different.

PR #304, opened 2020-04-09
fixes crash triage example

Toizi:
The script uses an otherwise unused variable to collect the arguments for the target and thus always executes it without arguments


vanhauser-thc:
@domenukk you merged into master, it should have been dev

domenukk:
Ah :/
Sorry my bad.

Issue #305, opened 2020-04-09
ki_oncpu=-1 in FreeBSD


neoni:
Hi, I perfromed parallel fuzzing in my FreeBSD vitual machine, but it binds the same CPU always.

So I dived into the code and found that 

> procs[i].ki_oncpu = -1

In my FreeBSD virtual machine, every proc's ki_oncpu is -1.  
Maybe we can use ki_lastcpu when ki_onpu = -1. ( It works in my machines. ) Here's [reference](https://lists.freebsd.org/pipermail/svn-src-stable-9/2013-June/004608.html) .

By the way, I think we can ignore [idle] process when deciding which CPU is free.

hexcoder-:
Thanks, since I merged your corresponding PR#306, I am closing this issue.

PR #306, opened 2020-04-09
fix when ki_oncpu=-1 in FreeBSD

neoni:
Using ki_lastcpu when ki_onpu = -1;

Ignoring [idle] process when deciding which CPU is free;

fix issue #305


hexcoder-:
Thanks, nice to see FreeBSD users here also.
@devnexen could you please have a look at the patch? Thanks.

hexcoder-:
Looks good to me too. Merged.

PR #307, opened 2020-04-10
LTO optimization, variable map size, autodictionary

vanhauser-thc:
This is quite a big change and due the volatility of the dev branch I want to merge it fast.

However I am sure several things can be improved.
So please take some time to look at the changes, comment what should be changed (not nitpicks, or nice to have, but real issues that need to be changed).

I will then merge and fix the comments in dev.


src/afl-fuzz-extras.c
```diff
@@ -305,10 +305,14 @@ static inline u8 memcmp_nocase(u8 *m1, u8 *m2, u32 len) {
 }
 
 /* Maybe add automatic extra. */
+/* Ugly hack: afl state is transfered as u8* because we import data via
```
domenukk, line 308: I suggest making this a `void *data` ptr, not `u8 *`. I feel that, semantically, u8 is wrong here. It's not super ugly, we actually may be able to reuse this for the coming custom lib refactorings.
For the future, do we want to add a meta afl struct for all shared env vars between tools? Right now, they are still global in afl-showmap et al. but that doesn't make sense imho.

src/afl-fuzz-queue.c
```diff
@@ -268,14 +270,17 @@ void update_bitmap_score(afl_state_t *afl, struct queue_entry *q) {
...
   u32                 i;
 
+  if (len == 0) len = 1;
+
   if (afl->dumb_mode || !afl->score_changed) return;
 
   afl->score_changed = 0;
 
-  memset(temp_v, 255, MAP_SIZE >> 3);
+  u8 temp_v[len];
```
domenukk, line 282: AFAIK most code styles prevent the use of variable length arrays.
They have a tendency to break software in random ways (as the stack may explode, etc.) and lead to a lot of slower added code. Also, they are C99+. In the case of the Linux kernel, seee https://lwn.net/Articles/749064/ 
Instead, I suggest (re)using a shared buffer in `afl` that you `ck_maybe_grow` to the `len` before use - or leave it at a size with fixed maximum.

domenukk:
Can the map size ever be larger than `MAP_SIZE >> 3`?

llvm_mode/README.lto.md
```diff
@@ -96,6 +87,13 @@ CC=afl-clang-lto CXX=afl-clang-lto++ ./configure
 make
 ```
 
+## AUTODICTIONARY feature
+
+Setting `AFL_LLVM_LTO_AUTODICTIONARY?  will generate a dictionary in the
```
domenukk, line 92: The `?` is out of place here

vanhauser-thc:
> Can the map size ever be larger than `MAP_SIZE >> 3`?
it can be up to MAP_SIZE (0 - (MAP_SIZE - 1)).

depends on the context. in 64 bit compares not.

vanhauser-thc:
@andreafioraldi if snapshot is used it is now displayed in the bottom right of the UI

Issue #316, opened 2020-04-12
llvm-bindir does not contain clang on NixOS


Mindavi:
On NixOS, the LLVM directory does not contain the clang binary. Rather, it's in another directory. In the Makefile there is a comment about this:

```
# We were using llvm-config --bindir to get the location of clang, but
# this seems to be busted on some distros, so using the one in $PATH is
# probably better.
```

The Makefile works around this by having some conditional logic that ultimately delegates to the PATH (which is a good default IMO, but maybe there's good reasons for having the full path).

However, in the code itself the LLVM_BINDIR is still assumed to contain clang/clang++.

https://github.com/AFLplusplus/AFLplusplus/blob/504529c3aa5c80937f9f722f90d0ec55e09c6dd2/llvm_mode/afl-clang-fast.c#L185

This results in a crash when trying to run the application:

```
[nix-shell:~/tmp]$ ./result/bin/afl-clang-fast++ test.cpp 
afl-clang-fast++2.63c by <lszekeres@google.com> in DEFAULT mode

[-] PROGRAM ABORT : Oops, failed to execute '/nix/store/c7wxcbjvpw3c7dq9271dd2jhzn373bp4-llvm-9.0.1/bin/clang++' - check your PATH
         Location : main(), afl-clang-fast.c:809
```

Clang on NixOS is in another path (`clang --version`): 
```
[nix-shell:~/tmp]$ clang --version
clang version 9.0.1 
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /nix/store/k03pk8rckgmbg9kfgdbg3jwdms5dwn1d-clang-9.0.1/bin
```

To fix this, it might be good to find the actual path of clang, or maybe consider not hardcoding the path in the binary. I think hardcoding the path makes the application less portable between machines or distros.

Mindavi:
Oh, I see this can be worked around by setting the AFL_CC and AFL_CXX path. However, I think it's not a very good assumption that the bindir contains clang, even if it might be so 'by accident' in most distros.

Should I keep this open for discussion?

andreafioraldi:
Hi, PATH as default is not good. Many ppl have multiple LLVM builds and want that afl-clang-fast uses only one of them. Said that, the solution to this problem should be more Makefile magic, I'm working on it.

andreafioraldi:
[...]

PR #318, opened 2020-04-14
afl-clang-fast: fail when binary name can't be used to determine build mode

Mindavi:
I'm not too sure about the specific error message, but this may be helpful when someone is trying to do things with the filename, such as wrapping it in an `exec` script. Another solution might be to generate a warning here, but doing nothing might make people shoot themselves in the foot.

In my case, the program was wrapped in a script that setup some environment stuff. Because the script called `exec`, the program name was the name of the script. Trying to compile c++ binaries failed because the c compiler was invoked (the `else` branch was used), resulting in errors like `fatal error: 'new' file not found`. Generating a failure or warning because the wrong compiler might be invoked seems like a valid choice.

Ultimately, the aforementioned script is unnecessary after applying the solution to #316, so it doesn't affect me anymore. However, it would be a good idea to warn or fail when users are trying to do such things. In the future, it might be a good improvement to do partial matching or allowing users to specify which binary to run using environment variables. However, that's not relevant right now, I think.

### DIFF
--- a/llvm_mode/afl-clang-fast.c
+++ b/llvm_mode/afl-clang-fast.c
@@ -184,7 +184,7 @@ static void edit_params(u32 argc, char **argv, char **envp) {
       sprintf(llvm_fullpath, CLANGPP_BIN);
     cc_params[0] = alt_cxx && *alt_cxx ? alt_cxx : (u8 *)llvm_fullpath;
 
-  } else {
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
unlikely but doesnt hurt, thanks :)

PR #321, opened 2020-04-17
Push for next release

vanhauser-thc:


gcc_plugin/afl-gcc-rt.o.c
```diff
@@ -138,8 +138,9 @@ static void __afl_map_shm(void) {
 
 static void __afl_start_forkserver(void) {
 
-  static u8 tmp[4];
-  s32       child_pid;
+  u8  tmp[4] = {0, 0, 0, 0};
```
domenukk, line 141: this could be initialized with `{0}` but for 4 chars it doesn't matter either way

vanhauser-thc:
pfff only 64 files changed. lazy 2 weeks. we have to pick this up again! ;)

qemu_mode/patches/afl-qemu-cpu-inl.h
```diff
@@ -293,7 +293,8 @@ static void print_mappings(void) {
 
 void afl_forkserver(CPUState *cpu) {
 
-  static unsigned char tmp[4];
+  u32                  map_size = 0;
+  static unsigned char tmp[4] = {0, 0, 0, 0};
```
domenukk, line 297: This is still static?

src/afl-analyze.c
```diff
@@ -795,8 +796,10 @@ static void usage(u8 *argv0) {
       "              (must contain abort_on_error=1 and symbolize=0)\n"
       "MSAN_OPTIONS: custom settings for MSAN\n"
       "              (must contain exitcode="STRINGIFY(MSAN_ERROR)" and symbolize=0)\n"
-      "AFL_PRELOAD: LD_PRELOAD / DYLD_INSERT_LIBRARIES settings for target\n"
       "AFL_ANALYZE_HEX: print file offsets in hexadecimal instead of decimal\n"
+      "AFL_MAP_SIZE: the shared memory size for that target. must be >= the size\n"
```
domenukk, line 800: `AFL_MAP_SIZE` or `AFL_MAPSIZE`? I've seen both supported in this pr, maybe decide on exactly one?

src/afl-analyze.c
```diff
@@ -982,18 +934,27 @@ int main(int argc, char **argv, char **envp) {
...
+
+  }
+
   use_hex_offsets = !!get_afl_env("AFL_ANALYZE_HEX");
 
   check_environment_vars(envp);
 
   sharedmem_t shm = {0};
-  trace_bits = afl_shm_init(&shm, MAP_SIZE, 0);
+  trace_bits = afl_shm_init(&shm, map_size, 0);
```
domenukk, line 951: The initial idea was to keep track of the current and the actual map size in `afl_shm` - there are variables for both, namely:
https://github.com/AFLplusplus/AFLplusplus/blob/504529c3aa5c80937f9f722f90d0ec55e09c6dd2/include/sharedmem.h#L47
https://github.com/AFLplusplus/AFLplusplus/blob/504529c3aa5c80937f9f722f90d0ec55e09c6dd2/include/sharedmem.h#L48
Now parts of it are in the afl state/global, parts in the forkserver state (`fsrv.map_size`), yet the vars in the shared map still exist. Not blocking for this pr but we may want to clean up 2 of the 4 vars afterwards. 
I still feel like they belong close to the shared mem struct, but it's fine either way

src/afl-forkserver.c
```diff
@@ -339,12 +391,113 @@ void afl_fsrv_start(afl_forkserver_t *fsrv, char **argv) {
...
+      if ((status & FS_OPT_MAPSIZE) == FS_OPT_MAPSIZE) {
+
+        u32 tmp_map_size = FS_OPT_GET_MAPSIZE(status);
+
+        if (!fsrv->map_size) fsrv->map_size = MAP_SIZE;
+
+        if (unlikely(tmp_map_size % 8)) {
+
+          // should not happen
+          WARNF("Target reported non-aligned map size of %ud", tmp_map_size);
```
domenukk, line 417: Should this be a `FATAL`?

src/afl-fuzz-queue.c
```diff
@@ -268,14 +269,17 @@ void update_bitmap_score(afl_state_t *afl, struct queue_entry *q) {
...
 
   struct queue_entry *q;
-  u8                  temp_v[MAP_SIZE >> 3];
+  u32                 len = (afl->fsrv.map_size >> 3);
   u32                 i;
+  u8 *                temp_v;
 
   if (afl->dumb_mode || !afl->score_changed) return;
 
+  temp_v = ck_alloc(afl->fsrv.map_size >> 3);
```
domenukk, line 278: I suggest (re)using `ck_maybe_grow` on a buf in `afl_state_t`  here, instead of an allocation each time.

src/afl-showmap.c
```diff
@@ -681,14 +540,24 @@ int main(int argc, char **argv_orig, char **envp) {
...
 
   char **argv = argv_cpy_dup(argc, argv_orig);
 
   afl_forkserver_t  fsrv_var = {0};
   afl_forkserver_t *fsrv = &fsrv_var;
   afl_fsrv_init(fsrv);
 
+  if ((ptr = getenv("AFL_MAP_SIZE")) || (ptr = getenv("AFL_MAPSIZE"))) {
+
+    map_size = atoi(ptr);
```
domenukk, line 553: This code is duplicated a few times

src/afl-showmap.c
```diff
@@ -1011,14 +884,15 @@ int main(int argc, char **argv_orig, char **envp) {
...
-  u32 ret = child_crashed * 2 + fsrv->child_timed_out;
+  u32 ret = child_crashed * 2 + fsrv->last_run_timed_out;
 
   if (fsrv->target_path) ck_free(fsrv->target_path);
 
   afl_fsrv_deinit(fsrv);
   if (stdin_file) ck_free(stdin_file);
 
   argv_cpy_free(argv);
+  if (fsrv->qemu_mode) free(use_argv[2]);
```
domenukk, line 895: How about `wine_mode`? Probably has the same leak

src/afl-common.c
```diff
@@ -348,12 +357,78 @@ char **get_wine_argv(u8 *own_loc, u8 **target_path_p, int argc, char **argv) {
 
 }
 
+/* Find binary, used by analyze, showmap, tmin
+   @returns the path, allocating the string */
+
+u8 *find_binary(u8 *fname) {
+
+  // TODO: Merge this function with check_binary of afl-fuzz-init.c
```
domenukk, line 365: Just FYI, here's a new todo, in case anybody gets bored during lockdown :)

src/afl-analyze.c
vanhauser-thc: documented is AFL_MAP_SIZE however I code to expect typos and hence AFL_MAPSIZE is secretly also supported (eg. we also support AFL_LLVM_LAF_TRANSFORM_COMPARES and LAF_TRANSFORM_COMPARES, etc)

src/afl-showmap.c
andreafioraldi: if wine_mode is set also qemu_mode is set

src/afl-analyze.c
vanhauser-thc: you are right. the map_size should be in sharedmem (and I renamed it to that to be obvious what it is and the name it is used everywhere else). the used size should be in the forkserver though as this is something the forkserver learns IMHO. removed the others

src/afl-forkserver.c
vanhauser-thc: we dont need to fail. will not result in failures (as afl-fuzz can take care of this). it should just not happen and would mean that there is a bug in afl-clang-fast/lto

src/afl-fuzz-queue.c
vanhauser-thc: I have no clue what that is :) too much happening in our code :) can you fix it?

src/afl-showmap.c
vanhauser-thc: do you want to have that in afl-common.c?

src/afl-showmap.c
vanhauser-thc: in the code if wine mode is used it also sets qemu_mode. hence this is all that is needed

src/afl-showmap.c
vanhauser-thc: moved it to afl-common

src/afl-analyze.c
vanhauser-thc: new change ok?

qemu_mode/patches/afl-qemu-cpu-inl.h
vanhauser-thc: I didnt want to touch it as this is Andreas domain but yeah ... just removed it

src/afl-analyze.c
domenukk: Yes thx :)

domenukk:
Should we also bind this release to the current `unicornafl` version?

vanhauser-thc:
as you like, its your baby and you know best

PR #325, opened 2020-04-20
push new formatting

vanhauser-thc:
As we have new formatting I think it is a good idea to push to master now before we have larger changes and we can't distinguish them anymore.


README.md
```diff
@@ -160,7 +160,8 @@ These build targets exist:
 * distrib: everything (for both binary-only and source code fuzzing)
 * man: creates simple man pages from the help option of the programs
 * install: installs everything you have compiled with the build options above
-* clean: cleans everything. for qemu_mode and unicorn_mode it means it deletes all downloads as well
+* clean: cleans everything compiled, not downloads (unless not on a checkout)
+* deelclean: cleans everything including downloads
```
domenukk, line 164: typo in deepclean

GNUmakefile
```diff
@@ -415,7 +424,17 @@ clean:
 	$(MAKE) -C qemu_mode/unsigaction clean
 	$(MAKE) -C qemu_mode/libcompcov clean
 	$(MAKE) -C src/third_party/libradamsa/ clean
-	-rm -rf unicorn_mode/unicornafl
+	rm -rf qemu_mode/qemu-3.1.1
+ifeq "$(IN_REPO)" "1"
+	$(MAKE) -C unicorn_mode/unicornafl clean || true
```
domenukk, line 429: Should probably also `$(MAKE) -C qemu_mode/qemu clean || true`? cc @andreafioraldi

test/test.sh
```diff
@@ -759,7 +768,14 @@ test -e ../afl-qemu-trace && {
       test "$SYS" = "i686" -o "$SYS" = "x86_64" -o "$SYS" = "amd64" -o "$SYS" = "i86pc" -o "$SYS" = "aarch64" -o ! "${SYS%%arm*}" && {
         $ECHO "$GREY[*] running afl-fuzz for persistent qemu_mode, this will take approx 10 seconds"
         {
-          export AFL_QEMU_PERSISTENT_ADDR=`expr 0x4$(nm test-instr | grep "T main" | awk '{print $1}' | sed 's/^.......//' )`
+          if file test-instr | grep -q "32-bit"; then
```
domenukk, line 771: For usability, a tool/shellscript to map a relocatable addr to the qemu-internal load addr might be beneficial. Using `AFL_QEMU_PERSISTENT_ADDR` needs a lot of internal knowledge about qemu at the moment.

GNUmakefile
```diff
@@ -415,7 +424,17 @@ clean:
...
+else
+	rm -rf qemu_mode/qemu-3.1.1.tar.xz
+	rm -rf unicorn_mode/unicornafl
+endif
+
+deepclean:	clean
+	rm -rf qemu_mode/qemu-3.1.1.tar.xz
+	rm -rf unicorn_mode/unicornafl
```
domenukk, line 437: technically wrong if in git, because then unicornafl is a submodule, removing it changes the state of the repo. Resetting the git repo might be the better/more correct solution?

GNUmakefile
vanhauser-thc: no as the patches are reapplied when the build script is called. if qemu has to be rebuild then because we changed the patches, so a make clean does not help

GNUmakefile
vanhauser-thc: deepclean can also be called by people not having git. then maybe `rm -rf unicorn_mode/unicornafl/*`?

vanhauser-thc: no, there is no unicornafl subdirectory in the repo so deepclean is correct

test/test.sh
vanhauser-thc: I thought about this too. but how? it only applies if PIE, AFL_QEMU_PERSISTENT_ADDR is set but not set to a 0x4.... address.  IMHO for now the documenation in the README and the FATAL message in afl-fuzz are ok. but because I dont know of a better way :)

domenukk: We could finds out using readelf (or some other way) if it's PIE, then
no_pie => return input
x86 => add 0x4...
x64 => ...

GNUmakefile
domenukk: For the git repo, unicornafl is a submodule - so it's there if you recursively clone or manually pull the submodules using `git submodule init && git submodule update`.
In this case, a way to "deep clean" is to do a `git submodule foreach --recursive git clean -xfd` (reset all changed files for each submodule).
If you simply remove the folder, git will think you changed the aflplusplus repo to no longer contain the submodule (leaving the parent repo in a changed state)
Another way is probably to clean the parent repo after the removal, that way the submodule should also be back (git `git reset --hard` or sthm)
Sorry for the confusion. Should I try to come up with a solution?

README.md
```diff
@@ -160,7 +160,8 @@ These build targets exist:
 * distrib: everything (for both binary-only and source code fuzzing)
 * man: creates simple man pages from the help option of the programs
 * install: installs everything you have compiled with the build options above
-* clean: cleans everything. for qemu_mode and unicorn_mode it means it deletes all downloads as well
+* clean: cleans everything compiled, not downloads (unless not on a checkout)
```
hexcoder-, line 163: avoid double negation (unless not) -> (when)

include/android-ashmem.h
```diff
@@ -105,7 +105,9 @@ static inline void *shmat(int __shmid, const void *__shmaddr, int __shmflg) {
 
 }
 
-#endif /* __ANDROID__ */
+#endif                                                       /* __ANDROID__ */
+
+#endif
```
hexcoder-, line 110: In total three `#if` and four `#endif`. That does not match! Kill the last `#endif`

include/forkserver.h
```diff
@@ -30,6 +30,7 @@
 
 #include <stdio.h>
 #include <stdbool.h>
+#include "types.h"
```
hexcoder-, line 33: types.h was already included in the next line :-)

unicorn_mode/build_unicorn_support.sh
```diff
@@ -146,21 +146,30 @@ fi
...
-test -d unicornafl || {
-   CNT=1
-   while [ '!' -d unicornafl -a "$CNT" -lt 4 ]; do
-     echo "Trying to clone unicornafl (attempt $CNT/3)"
-     git clone https://github.com/AFLplusplus/unicornafl
-     CNT=`expr "$CNT" + 1`
-   done
-}
+
+git status 1>/dev/null 2>/dev/null
```
hexcoder-, line 150: Please add `git` as a new requirement in the checks up front, since people could have used a zip archive from github to install it.

GNUmakefile
vanhauser-thc: ah understood. I dont use git :) yeah if you could find a solution ...

test/test.sh
vanhauser-thc: lets put that on the todo list

src/afl-fuzz.c
```diff
@@ -266,11 +273,14 @@ int main(int argc, char **argv_orig, char **envp) {
   afl->init_seed = tv.tv_sec ^ tv.tv_usec ^ getpid();
 
   while ((opt = getopt(argc, argv,
-                       "+c:i:I:o:f:m:t:T:dnCB:S:M:x:QNUWe:p:s:V:E:L:hRP:")) > 0)
+                       "+c:i:I:o:f:m:t:T:dnCB:S:M:x:QNUWe:p:s:V:E:L:hRP:")) >
+         0) {
```
hexcoder-, line 277: really ugly formatting IMO

GNUmakefile
domenukk: should(tm) be solved, please feel free to try without git :)

include/android-ashmem.h
domenukk: done

include/forkserver.h
domenukk: great find :)

.custom-format.py
```diff
@@ -29,14 +29,14 @@
 if CLANG_FORMAT_BIN is None:
     o = 0
     try:
-        p = subprocess.Popen(["clang-format-8", "--version"], stdout=subprocess.PIPE)
+        p = subprocess.Popen(["clang-format-10", "--version"], stdout=subprocess.PIPE)
```
domenukk, line 32: On arch, I have clang-format version 10.0.0 installed but `clang-format-10` is not in my path. 
I need to use `CLANG_FORMAT_BIN=clang-format make code-format` but that feels off.

unicorn_mode/build_unicorn_support.sh
vanhauser-thc: it is in there already

.custom-format.py
vanhauser-thc: why dont you put a symlink in /usr/local/bin?

src/afl-fuzz.c
vanhauser-thc: I agree. but its the formatter so what do?

hexcoder-: Never look at it again...

.custom-format.py
domenukk: I'm not a fan of messing with my distro's PATH, and there is no actual benefit (if an upgrade upgrades clang-format, the symlink won't change)

src/afl-fuzz.c
domenukk: We could 
- raise the number of allowed chars per line
- declare the ops as variable above the loop
- never look at it again

vanhauser-thc:
@domenukk unicorn build still fails:
```
make[1]: Entering directory '/home/travis/build/AFLplusplus/AFLplusplus/unicorn_mode/unicornafl'
cd qemu && \
./configure --cc="gcc" --extra-cflags="-DUNICORN_HAS_X86 -DUNICORN_HAS_ARM -DUNICORN_HAS_ARMEB -DUNICORN_HAS_M68K -DUNICORN_HAS_ARM64 -DUNICORN_HAS_ARM64EB -DUNICORN_HAS_MIPS -DUNICORN_HAS_MIPSEL -DUNICORN_HAS_MIPS64 -DUNICORN_HAS_MIPS64EL -DUNICORN_HAS_SPARC -fPIC -DUNICORN_AFL -fvisibility=hidden" --target-list="x86_64-softmmu, arm-softmmu, armeb-softmmu, m68k-softmmu, aarch64-softmmu, aarch64eb-softmmu, mips-softmmu, mipsel-softmmu, mips64-softmmu, mips64el-softmmu, sparc-softmmu,sparc64-softmmu," --python=python --disable-debug-info --python=/usr/bin/python
gcc -O3 afl.c -o afl.o
/usr/lib/gcc/x86_64-linux-gnu/7/../../../x86_64-linux-gnu/Scrt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
collect2: error: ld returned 1 exit status
Makefile:311: recipe for target 'afl.o' failed
make[1]: *** [afl.o] Error 1
make[1]: *** Waiting for unfinished jobs....
Source path       /home/travis/build/AFLplusplus/AFLplusplus/unicorn_mode/unicornafl/qemu
```

PR #326, opened 2020-04-20
Add malloc_usable_size to libdislocator.so

dpmdpm2:
I was trying to fuzz a program with AFL_PRELOAD=/path/to/libdislocator.so

The program called malloc_usable_size and, because libdislocator.so.c did not wrap this function, it crashed. I added the function and cleaned up two comments.


vanhauser-thc:
looks good, thanks!

PR #331, opened 2020-04-23
afl-gcc and afl-clang: fail when binary name can't be used to determine build mode

Mindavi:
This is a continuation of PR #318.
The goal is to prevent issues where binaries with the wrong name will
silently pass control to the C compiler instead of failing.
This makes it more explicit that aflplusplus relies on the name of the
binary for correct compiler execution.


vanhauser-thc:
thanks :)

PR #335, opened 2020-04-25
Integer-Swapping Havoc Mutator

subho-13:
Some changes were made in src/afl-fuzz-one.c and types.h . A primary report on fuzzer  performance, before and after patching is given below.

| Details   |                                                    |
|-------------|-------------------------------------------|
| OS          | Ubuntu Bionic |
| System   | Google Compute N1 custom (10 vCPUs, 9 GB memory) [ Free Tier ] |
| Test Lib   |   Binutils-2.30 |

The binaries were compiled using afl-clang-fast.

The command line argument used is ::
`` $ afl-fuzz -i afl_in -o afl_o_# -d -V 43200 # $ @@ ``

Each binutil was fuzzed only once using patched and unpatched afl-fuzz-one.c .
* where # stands for the following binutils used
* where $ stands for the arguments used along with them

|    #    |    $    |
|---------|---------|
| readelf |   -a    |
| strings |         |
| nm      |   -C    |
| objdump |   -d    |
| size    |         |

The coverage data was collected using afl-cov. 

|    binutil    |  patched?  |  lines  |  functions  |
|---------------|------------|---------|-------------|
|    readelf    |     No     |   6.7%  |    4.7%     |
|    strings    |     No     |   2.1%  |    2.6%     |
|    nm         |     No     |   3.4%  |    3.8%     |
|    objdump    |     No     |   5.4%  |    5.0%     |
|    size       |     No     |   0.1%  |    0.1%     |
|    readelf    |    Yes     |   6.7%  |    4.7%     |
|    strings    |    Yes     |   7.5%  |   11.6%     |
|    nm         |    Yes     |  12.3%  |   15.1%     |
|    objdump    |    Yes     |   7.3%  |    7.7%     |
|    size       |    Yes     |   4.9%  |    4.8%     |

In case of readelf, the patched version has slightly more code coverage 
   *  Unpatched -> 9466/142239 lines, 253/5350 functions
   *  Patched   -> 9501/142239 lines, 253/5350 functions

In the above cases, patched afl-fuzz-one.c resulted in improved code coverage. 

I'll run the fuzzers again to confirm the observation. For the testcases and the output, I can mail them to any member. There are some crash reports in them.

### DIFF
--- a/include/types.h
+++ b/include/types.h
@@ -107,6 +107,15 @@ typedef int64_t s64;
                                                                                \
   })
 
+  #define SWAPQ(_x, _y) \
+  (                     \
+      {                 \
+          u64 _t = _x;  \
+          _x     = _y;  \
+          _y     = _t;  \
+      }                 \
+  ) 
+
 #ifdef AFL_LLVM_PASS
 #if defined(__linux__)
 #define AFL_SR(s) (srandom(s))
--- a/src/afl-fuzz-one.c
+++ b/src/afl-fuzz-one.c
@@ -1750,7 +1750,7 @@ havoc_stage:
       }
 
       switch (rand_below(
-          afl, 15 + ((afl->extras_cnt + afl->a_extras_cnt) ? 2 : 0))) {
+          afl, 16 + ((afl->extras_cnt + afl->a_extras_cnt) ? 2 : 0))) {
 
         case 0:
 
@@ -2034,11 +2034,22 @@ havoc_stage:
           break;
 
         }
+            
+        case 15 : {
+          /* Swap QWORDS */
+          if(temp_len <= 8) break;
+
+          u32 offset_1 = rand_below(afl, temp_len-8);
+          u32 offset_2 = rand_below(afl, temp_len-8);
+
+          SWAPQ((*(u64 *)(out_buf + offset_1)), (*(u64 *)(out_buf + offset_2)));     
+          break;
+        }
 
           /* Values 15 and 16 can be selected only if there are any extras
              present in the dictionaries. */
 
-        case 15: {
+        case 16: {
 
           /* Overwrite bytes with an extra. */
 
@@ -2076,7 +2087,7 @@ havoc_stage:
 
         }
 
-        case 16: {
+        case 17: {
 
           u32 use_extra, extra_len, insert_at = rand_below(afl, temp_len + 1);
           u8 *new_buf;


vanhauser-thc:
I will merge it into a new experimental branch for testing havoc changes so we can easily make tests or add more test cases

vanhauser-thc:
I played a bit with it and here are some thoughts.

1. IMHO it is not a havoc routine but if added should rather be a deterministic fuzzing function, so similar to bit flips and byte flips this could switch 1, 2 or 4 bytes.
2. this needs a check if the x and y values are the same and if so skip the test. e.g if the quad value is 0x0000000000000000 it is pointless.
3. I am still not sure this is useful fuzzing. in my (very limited) test it degraded the path discovery and exec/s.

I propose implementing 1. and 2. instead of having it in havoc and then putting it to test on Google's fuzzbench.

@andreafioraldi what is your opinion?

vanhauser-thc:
P.S. @subho-13 whenever you do such tests add "-s 123" to your afl-fuzz command line options, as only then the random numbers generated are the same and you have comparable results

subho-13:
Okay, I'll start implementing 1 and 2. As of path degradation, I've run the tests again and I'll report if it degrades the execs/sec.

subho-13:
Hi Marc,
I gave a thought about this Integer Swapping as a deterministic mutation. And I don't think it's going to improve the fuzzer performance. There are several issues that crops up :
1. The swapping operation is going to be O(n^2) operation. We have to traverse the entire binary, check the effector map if a zone is prone to bit-changes and swap it with another zone. Also there can redundant swaps, in which the value with which a zone was swapped with is repeated. This is going to degrade the fuzzer performance. 
2. The swapping operation can be thought of as changing a few bits, essentially what is done in arithmetic steps and bit flips. So repeating the same through Swapping is redundant.

I took the idea from AFLSmart, where chunk based mutation is present, which is inapplicable here, since the internal structure of the input is unknown.

We can have this mutation in havoc stage: If the input is highly sensitive to mutations, then we can perform this mutation, else we can exclude this.

@vanhauser-thc  what is your opinion?

vanhauser-thc:
if you add any kind of new mutations to a fuzzer it will find new paths.
if the the new mutations are "good" can be assessed for two ways:

1- for the otherwise same configuration and same run time more new paths (or crashes) are found
2- for the otherwise same configuration and same run time edges are discovered that otherwise would not have been discovered.

both tests need to run for a longer time (say 24h+) and several targets (8+) each.

When thinking about the swapping mutation I think too it does not help much. but I have learned that this is a bad prediction how the results really are. it is better to test every idea and if it looks promising do more testing then come to a conclusion.

subho-13:
Okay, I'll implement this ASAP and submit a patch.

PR #336, opened 2020-04-25
pull again ...

vanhauser-thc:
Although before we take this one in we should have the unicorn build fixed (it - seemingly - randomly fails)


vanhauser-thc:
with "unicornafl updaetd" the problems started

hexcoder-:
To me it looks like parallel GNUmake is screwing things up. Can we try without -j and see, if problems vanish?

vanhauser-thc:
> To me it looks like parallel GNUmake is screwing things up. Can we try without -j and see, if problems vanish?
> If so, we have a hint to the cause.

lets try that. I put that commit ...

unicorn_mode/build_unicorn_support.sh
```diff
@@ -182,7 +182,8 @@ echo "[+] Configuration complete."
 echo "[*] Attempting to build unicornafl (fingers crossed!)..."
 
 $MAKECMD clean  # make doesn't seem to work for unicorn
-UNICORN_QEMU_FLAGS="--python=$PYTHONBIN" $MAKECMD -j$CORES || exit 1
+#UNICORN_QEMU_FLAGS="--python=$PYTHONBIN" $MAKECMD -j$CORES || exit 1
+UNICORN_QEMU_FLAGS="--python=$PYTHONBIN" $MAKECMD || exit 1
```
domenukk, line 186: The unicorn-engine Makefile will automatically set `-j4` if not specified otherwise. Maybe we should set `-j1` here, at least until the race condition is fixed?
Probably open a new issue for it, too.

src/afl-fuzz-state.c
```diff
@@ -81,7 +81,15 @@ void afl_state_init(afl_state_t *afl, uint32_t map_size) {
   and out_size are NULL/0 by default. */
   memset(afl, 0, sizeof(afl_state_t));
 
-  if (!map_size) { afl->shm.map_size = MAP_SIZE; }
+  if (!map_size) {
```
domenukk, line 84: If you like one-liners, this one could be
```c
afl->shm.map_size = map_size ? map_size : MAP_SIZE;
```

src/afl-forkserver.c
```diff
@@ -55,6 +55,8 @@
 
 list_t fsrv_list = {.element_prealloc_count = 0};
 
+void report_error_and_exit(int error);
```
domenukk, line 58: this function can instead be `static`

src/afl-forkserver.c
```diff
@@ -55,6 +55,8 @@
 
 list_t fsrv_list = {.element_prealloc_count = 0};
```
domenukk, line 56: Now that I see it, this list can be `static`

include/config.h
```diff
@@ -402,12 +402,26 @@
...
+#define FS_ERROR_MMAP 16
+
+/* Reporting options */
 #define FS_OPT_ENABLED 0x8f000001
 #define FS_OPT_MAPSIZE 0x40000000
 #define FS_OPT_SNAPSHOT 0x20000000
 #define FS_OPT_AUTODICT 0x10000000
+// FS_OPT_MAX_MAPSIZE is 8388608 = 0x800000 = 2^23 = 1 << 22
+#define FS_OPT_MAX_MAPSIZE ((0x00fffffe >> 1) + 1)
 #define FS_OPT_GET_MAPSIZE(x) (((x & 0x00fffffe) >> 1) + 1)
```
domenukk, line 422: Should this really be part of the config header?

hexcoder-:
Regarding parallel make: I think the problem might be within the unicornafl subproject:
https://github.com/AFLplusplus/unicornafl/blob/25ae270c1b949a5d1c2c5460b778f0a35bfe67d7/Makefile#L18 (`SMP_MFLAGS := -j4`)
referenced here
https://github.com/AFLplusplus/unicornafl/blob/25ae270c1b949a5d1c2c5460b778f0a35bfe67d7/Makefile#L311
and
https://github.com/AFLplusplus/unicornafl/blob/25ae270c1b949a5d1c2c5460b778f0a35bfe67d7/Makefile#L314

include/config.h
vanhauser-thc: it is needed in afl-forkserver and all afl-*-rt.o variants. so either in config.h or types.h.

vanhauser-thc:
since it is -j 1 this seems to have fixed the unicorn build.
merge now or do you want to do more research @domenukk ?

domenukk:
Let's merge.
I'm all for keeping config for configurables, though. Let's move it to
types or a proper common?

vanhauser-thc:
I moved the forkserver defines to types and ensured everything is compiling.
if travis is happy then I will merge

PR #349, opened 2020-05-03
Fix Unicorn support build script

assafcarlsbad:
Hi, I recently tried to build Unicorn mode support for AFL++ from my WSL distribution (Ubuntu-18.04). Unfortunately, the Unicorn build script repeatedly failed because it couldn't generate the test harness, but after applying the suggested changes it managed to compile and run successfully. Hope that helps 😅

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,7 +42,7 @@ qemu_mode/libcompcov/compcovtest
 as
 ld
 qemu_mode/qemu-*
-unicorn_mode/samples/*/\.test-*
+unicorn_mode/samples/*/test-*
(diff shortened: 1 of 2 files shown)


domenukk:
I have changed the target to the `dev` branch.

Can you please test if `./.test-instr0` instead of `./test-instr0` also works in WSL (with leading dot)?|
Pretty sure it should - and then, the gitignore also doesn't have to be touched. Other parts of afl might make use of `.test-*` files during build.

Either way, thank you for the PR!

domenukk:
Merging :)

PR #356, opened 2020-05-09
NetBSD build fix

devnexen:


hexcoder-:
Thanks! I assume you are on a recent NetBSD (newer than 8.0) where this file has been removed?

devnexen:
yes 9.0 here :-)

PR #359, opened 2020-05-11
push to master

vanhauser-thc:
before we reformat the code (indent directives) we should push to master, then reformat and push again. otherwise we dont see the changes because of the reformatting.


unicorn_mode/samples/persistent/Makefile
```diff
@@ -48,3 +48,6 @@ debug: harness-debug.o
 
 fuzz: harness
 	../../../afl-fuzz -m none -i sample_inputs -o out -- ./harness @@
+
+debugmake:
```
domenukk, line 52: why not `debug` but `debugmake`?
Also what's the content of _LRT?

src/afl-fuzz-mutators.c
```diff
@@ -26,27 +26,55 @@
...
-
+    if (afl->limit_time_sig)
       FATAL(
           "MOpt and custom mutator are mutually exclusive. We accept pull "
           "requests that integrates MOpt with the optional mutators "
-          "(custom/radamsa/redquenn/...).");
+          "(custom/radamsa/redqueen/...).");
 
-    }
+    u8 *fn_token = (u8 *)strsep((char **)&fn, ";:,");
```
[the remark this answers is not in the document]
vanhauser-thc: I really, really like it if we say "it has to be X" but we also parse Y and Z if it doesnt hurt us. sometimes people remember the wrong thing. Same as expecting environment variables mistyped etc.

src/afl-fuzz-mutators.c
```diff
@@ -26,27 +26,55 @@
...
+      mutator = load_custom_mutator(afl, fn);
+      list_append(&afl->custom_mutator_list, mutator);
+      afl->custom_mutators_count++;
 
-    return;
+    } else {
+
+      while (fn_token) {
+
+        if (*fn_token) {  // strsep can be empty if ";;"
```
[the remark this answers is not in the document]
vanhauser-thc: the code would break if ";;" and that is unnecessary. again - user-friendliness which costs nothing .... me big fan :)

unicorn_mode/samples/persistent/Makefile
vanhauser-thc: who added this? @hexcoder- ? not me :)

src/afl-fuzz-mutators.c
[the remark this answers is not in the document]
domenukk: So what's your suggestion? Add whitepace handling or something else?

unicorn_mode/samples/persistent/Makefile
hexcoder-: yes, i did. The make target 'debug' was used already in other Makefiles, so i used something unique for debugging the Makefile.
The content of _LRT is some unique string ('_QINU_'), following by the output of 'uname -s'.
The following is the only portable way (known to me) to do conditional logic (works for POSIX, gnu and bsdmake).
`$(_LRT:$(_UNIQ)Linux=-lrt)` means if _LRT ends with '_QINU_Linux', return '-lrt', otherwise return $_LRT.
The next expression $(__LRT:$(_UNIQ)$(UNAME_S)=) means, if __LRT ends in the same as _LRT, return '' otherwise return $__LRT.
The effect is, that $LRT is '-lrt' on Linux, and empty otherwise.

src/afl-fuzz-mutators.c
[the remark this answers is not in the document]
rish9101: I can add a whitespace handling or something. What would you guys suggest? 🙂

unicorn_mode/samples/persistent/Makefile
vanhauser-thc: makefile magician :)

src/afl-fuzz-mutators.c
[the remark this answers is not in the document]
vanhauser-thc: the empty string is already handled. whitespace - could be easily done, on the other hand I dont think people put whitespace. If you want add it

test/test-multiple-mutators.c
```diff
@@ -0,0 +1,24 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+int main(int argc, char **argv) {
+
+  int  a = 0;
+  char s[16];
+  memset(s, 0, 16);
+  read(0, s, 0xa0);
```
hexcoder-, line 17: @rish9101 : this seems just to be a simple buffer overflow. Sorry, I can't see, how that tests _multiple_ mutators.

test/test.sh
```diff
@@ -986,6 +996,25 @@ test "1" = "`../afl-fuzz | grep -i 'without python' >/dev/null; echo $?`" && {
       # Clean
       rm -rf out errors
 
+      #Run afl-fuzz w/ multiple C mutators
+      $ECHO "$GREY[*] running afl-fuzz with multiple custom C mutators, this will take approx 20 seconds"
+      {
+        AFL_CUSTOM_MUTATOR_LIBRARY="${CUSTOM_MUTATOR_PATH}/libexamplemutator.so;${CUSTOM_MUTATOR_PATH}/libexamplemutator.so" ../afl-fuzz -V20 -m ${MEM_LIMIT} -i in -o out -- ./test-multiple-mutators >>errors 2>&1
```
hexcoder-, line 1002: @rish9101 Shouldn't we test here with different custom mutators?

src/afl-fuzz-mutators.c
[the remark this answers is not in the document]
domenukk: yea, I think we don't need it. Closing :)

test/test-multiple-mutators.c
vanhauser-thc: the s buffer too small for the amount that is read, should be increased. and there should be two different mutators, and the 2nd sets [1] = 0 or something like that so we have a meaningful test.

PR #360, opened 2020-05-12
new code formatting + applied

vanhauser-thc:
Heiko made changes to the formatting python script and now we can indent directives.
I like it better than before, what do you guys thing?


vanhauser-thc:
by the way - I propose to make a new release (after this merge or if we drop the new indention), because of a crash that can happen in the laf-intel string splitting function

rish9101:
> I am reading the description and it seems they are the same.
> 
> but "pre_save" sounds like it is called before saving it in a queue or something,, not before it is send. then "pre_send" or "postprocess" or something like that maybe?
> 
> pre_save and post library are in the same place. pre_save is just before calling common_fuzz, and post lib is at the start of common_fuzz.

Yes, the `pre save` function __could__ be named better actually. Should I make a PR on dev regarding this?

vanhauser-thc:
> > I am reading the description and it seems they are the same.
> > but "pre_save" sounds like it is called before saving it in a queue or something,, not before it is send. then "pre_send" or "postprocess" or something like that maybe?
> > pre_save and post library are in the same place. pre_save is just before calling common_fuzz, and post lib is at the start of common_fuzz.
> 
> Yes, the `pre save` function **could** be named better actually. Should I make a PR on dev regarding this?

yeah, select a good name :)
and if you can then merge the code for afl POST lib feature to use the custom mutator (so the env var and function name are just put into an extra custom mutator entry) that would be good

vanhauser-thc:
I'd say we wait for rishi's PR for pre_save and postprocessing and then we merge this PR and then we release the next afl++ version.

domenukk:
Let's just call the post lib deprecated and remove it with the next version?

vanhauser-thc:
@hexcoder-  @domenukk 
```
[*] Using python binary /bin/python3
[-] we cannot test unicorn_mode because it is not present
```
the patch skips unicorn_mode tests  :(

domenukk:
Wouldn't a name like `pre_execute` or `pre_send` fit better in the future? or maybe `post_mutate`

vanhauser-thc:
yup. I did a "make distrib" and:
```
marc /prg/AFLplusplus/dev/test # python2 -c 'import unicornafl' && echo ok
ok
marc /prg/AFLplusplus/dev/test # python3 -c 'import unicornafl' && echo ok
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'unicornafl'
```
looks like the build script installs for python2, the test.sh script selects python3.
please fix

src/afl-fuzz-init.c
```diff
@@ -297,29 +297,27 @@ void setup_post(afl_state_t *afl) {
...
-  if (!afl->post_init) { FATAL("Symbol 'afl_postprocess_init' not found."); }
-  afl->post_deinit = dlsym(dh, "afl_postprocess_deinit");
-  if (!afl->post_deinit) {
+  struct custom_mutator * mutator;
+  mutator = ck_alloc(sizeof(struct custom_mutator));
+  memset(mutator, 0, sizeof(struct custom_mutator));
+
+  mutator->afl_custom_post_process = dlsym(dh, "afl_postprocess");
+  if (!mutator->afl_custom_post_process) { FATAL("Symbol 'afl_postprocess' not found."); }
+  mutator->afl_custom_init = dlsym(dh, "afl_postprocess_init");
```
domenukk, line 311: This should no longer be needed, the init is the same as the custom mutator initialization, right?

vanhauser-thc: it is for format of the post lib feature and that is what is name was/is.
it should not break what people are already using.
or what do you mean?

vanhauser-thc:
@hexcoder- bug is still there

vanhauser-thc:
There is also a bug in the main GNUmakefile because (from travis):
```
[*] Testing: qemu_mode
[-] qemu_mode is not compiled, cannot test
[*] Testing: unicorn_mode
[*] Using python binary /opt/pyenv/shims/python3
[-] we cannot test unicorn_mode because it is not present
```
in travis (but not on my system!) a "make distrib ; make tests ; " does not build qemu or unicorn:
```
make[1]: Leaving directory '/home/travis/build/AFLplusplus/AFLplusplus/libtokencap'
#make -C examples/afl_network_proxy
#make -C examples/socket_fuzzing
#make -C examples/argv_fuzzing
[*] starting afl++ test framework ...
```

vanhauser-thc:
ok BUILD_ASAN was broken because of wrong indents. fixed it.

src/afl-fuzz-init.c
domenukk: `afl_postprocess_init` and `afl_custom_mutator_init` have exactly the same use. It makes no sense to support both/is error prone.

domenukk: Same for deinit.
IMHO after merge, people will have to rename the function and that's that.

vanhauser-thc: so you want to break the stuff people are using?
the afl_postprocess_init is what is in the post library, and we removed that extra code and handle as if it would be a custom mutator.

vanhauser-thc: so far we have the policy not to break what vanilla afl does. removing the post lib feature would change that. I am all for that, but for a future release where we also change defaults and remove unnecessary features (as we collect in the two issues). so breaking changes in one big release rather than every other release something might break. or do you think that is not a good approach?

domenukk: `afl_postprocess_init` is already AFL++ stuff. I've added that for globals. Nobody complained so far.
Also we drop a function in the custom mutator, so why not do it right?

vanhauser-thc: that is not correct.
```
google-afl # grep afl_post  experimental/post_library/post_library.so.c 
const unsigned char* afl_postprocess(const unsigned char* in_buf,
```
it is in vanilla afl

vanhauser-thc: oh sorry, you meant the _init/_deinit. you are right, that has to be fixed. doesnt exist in vanille, so it must be optional now

vanhauser-thc:
> Wait, that's not what I meant :D
> Instead of properly merging the post lib and the custom mutator lib, we now have postlib and custom mutator lib, somewhat squished into the same code.
> My take is to either keep post lib and mutators seperated or merge them, and if we merge them, get rid of `afl_postprocess_init` and `afl_postprocess_deinit` completely (since they are the same as the custom mutator versions)
> 
> If we want to keep it 100% compatible to the old post libaray, simply removing the functions doesn't help since function signature also changed!
> That happened weeks ago, so far, nobody complained.
> 
> So I am for dropping "old" post support and merging it completely into mutators.

you are right, the API changed and that makes it obsolete. actually we should not have touched post lib :( damn
so either revert or remove.

domenukk:
Great. :)
Sorry for breaking the original post library in the first place, I should probably have asked the group before these kind of changes. I was in the flow. 
But I personally like the outcome now. If people complain it should be easy to build a shim for the old lib on top of custom mutators

vanhauser-thc:
removed post lib support.

now we still have to fix this bug before we can merge:

>  "make distrib" and:
> ```
> marc /prg/AFLplusplus/dev/test # python2 -c 'import unicornafl' && echo ok
> ok
> marc /prg/AFLplusplus/dev/test # python3 -c 'import unicornafl' && echo ok
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
> ModuleNotFoundError: No module named 'unicornafl'
> ```
> 
> looks like the build script installs for python2, the test.sh script selects python3.
> please fix

@domenukk @hexcoder-

domenukk:
Should be done

vanhauser-thc:
as you know I am clueless when it comes to python :)
is it possible to add the unicorn module for python2 and python3 if both are installed?

domenukk:
Sure, but now that python2 is EOL I guess it's okay to keep it as python3 default? We don't want to be too obtrusive. :)
I think I will try to publish unicornafl as pypi module soon, so it will simply be installable via pip, taking away some of the headache.

vanhauser-thc:
@domenukk fix didnt work.
unicorn compiled fine and test is still:
```
*] Testing: unicorn_mode
[*] Using python binary /bin/python3
[-] we cannot test unicorn_mode because it is not present
```

vanhauser-thc:
Now I get when trying to build unicorn:
```
[-] Error: Python setup-tools not found. Run 'sudo apt-get install python-setuptools'.
```

vanhauser-thc:
wtf? 
```
# apt-get install python-setuptools
Reading package lists... Done
Building dependency tree       
Reading state information... Done
python-setuptools is already the newest version (44.0.0-2).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
marc /prg/AFLplusplus/dev/unicorn_mode # ./build_unicorn_support.sh 
UnicornAFL build script

[*] Performing basic sanity checks...
[-] Error: Python setup-tools not found. Run 'sudo apt-get install python-setuptools'.
```

domenukk:
Hmm well for python3 you'll now need to `apt install python3-setuptools`... :/

vanhauser-thc:
@domenukk @hexcoder- 
I am not happy with that "fix"
1st if it requires setuptools 3 - why is that not the message being displayed?
2nd the real issue is that test.sh selects the python differently than the build_unicorn script.

and why do we break things unnecessarily? yes python3 is current, python2 is deprecated, but this does not help people who work (have to) work on an Ubuntu 16.04 and maybe not having the modules the need for python3 etc.
if my environment for python2 is fine but for python3 is not then python2 should be used, by test.sh and build_unicorn. if python3 is fine then python3. I would even go as so far as to say if python2 and python3 are installed, why not install it for both? I really dont get that.

vanhauser-thc:
> Interesting. What error does it fix?

you put tabs in there. you might have put spaces but your editor then put tabs. and that broke it.

hexcoder-:
Regarding unicornafl/python/setuptools: I will add logic to select the python version we have setuptools for (and prefer version 3.x if available) to the build script. Probably this evening after work...

rhtenhove:
> Regarding unicornafl/python/setuptools: I will add logic to select the python version we have setuptools for (and prefer version 3.x if available) to the build script. Probably this evening after work...

Just created a pull request before finding out you were on to it as well... See [here](https://github.com/AFLplusplus/AFLplusplus/pull/362).

vanhauser-thc:
@hexcoder-  got it to work first, but maybe @rhtenhove can check if this can be improved?

vanhauser-thc:
we did it! man that was a lot of effort this time ...

PR #362, opened 2020-05-14
Change to python3 only

rhtenhove:
The new default ubuntu (20.04) container has python3.8 by default, which does not cause any issues on building AFL++, qemu-mode and unicorn-mode. I've made some changes to ensure python3 is the only available python, and is used on all build paths.

Building QEMU still expects `python` to be on PATH, so for that reason the handy but hopefully soon redundant `python-is-python3` is added.

`easy_install` is also deprecated, and doesn't seem to be used anyway. So unicorn now only tests for the `setuptools` package.


vanhauser-thc:
Hi!

we do not accept PRs against master, only against dev (and push from dev to master every few weeks to keep the master always stable).
Sadly our patch does conflict against dev.
Either fix this by further commits, or revoke this PR and create a new one from dev.
But note that we currently work on the same thing (and its buggy currently). see our long discussion in #360 - so maybe you just help fixing this bug :)

vanhauser-thc:
ah and please note: we care about supporting old systems. We still need to support python2 (if python3 is not present). some fuzzing has to be performed on old systems (e.g. Ubuntu 14.04).

rhtenhove:
Closed this in favor of #363

PR #363, opened 2020-05-14
Recommend python3

rhtenhove:
Improved PR after a faulty #362.

Still removing the _check_ for `easy_install`, since it is never used. `setuptools` is used though, and its existence is now tested in a much simpler fashion. Since `setuptools` provides `easy_install` it will still be available if it is used anywhere.

Docker builds and manual builds are pointed to use python3 in the Dockerfile and the README.

## Older and/or python2 systems
I've tested such a python3 build in a `ubuntu:trusty` Docker container, and both the build and `test.sh` ran succesfully in there. Regardless of that, I've tried the same base image in a fresh container with just python2, and python3 completely removed. Again both the build and `test.sh` ran fine. This is to be expected since the build scripts don't care anymore which python is used.

Let's see what travis thinks.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,3 +20,8 @@ No camel case at all and use the AFL's macros wherever possible
 Remember that AFLplusplus has to build and run on many platforms, so
 generalize your Makefiles (or your patches to our pre-existing Makefiles)
 to be as much generic as possible.
+
+
+# Local testing
+
+To try
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,8 +20,3 @@ No camel case at all and use the AFL's macros wherever possible
 Remember that AFLplusplus has to build and run on many platforms, so
 generalize your Makefiles (or your patches to our pre-existing Makefiles)
 to be as much generic as possible.
-
-
-# Local testing
-
-To try
--- a/qemu_mode/build_qemu_support.sh
+++ b/qemu_mode/build_qemu_support.sh
@@ -76,17 +76,13 @@ for i in libtool wget automake autoconf sha384sum bison flex iconv patch pkg-con
 
 done
 
-PYTHON_TEMP_SYMLINK=0
-if ! command -v python 1>/dev/null; then
-  if command -v python3 1>/dev/null; then
-    echo "[*] 'python' not found. 'python3' was found. Creating temporary symlink."
-    update-alternatives --install /usr/bin/python python "$(command -v python3)" 0 1>/dev/null
-    PYTHON_TEMP_SYMLINK=1
-  else
-    echo "[-] Error: 'python' not found, please install using 'sudo apt install python3'."
-    PREREQ_NOTFOUND=1
-  fi
+PYTHONBIN=`command -v python3 || command -v python || command -v python2`
+
+if [ "$PYTHONBIN" = "" ]; then
+  echo "[-] Error: 'python' not found, please install using 'sudo apt install python3'."
+  PREREQ_NOTFOUND=1
 fi
+
 
 if [ ! -d "/usr/include/glib-2.0/" -a ! -d "/usr/local/include/glib-2.0/" ]; then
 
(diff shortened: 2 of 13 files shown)


vanhauser-thc:
hmm it is failing with:
```
cd unicorn_mode && sh ./build_unicorn_support.sh

UnicornAFL build script

*] Performing basic sanity checks...

[-] Error: Python setuptools not found. Run 'sudo apt-get install python3-setuptools'.
```

vanhauser-thc:
if you made changes to the Dockerfile you likely also have to update `.travis.yml`

domenukk:
As stated in #362 we'll also have to make sure qemu-mode builds without python2 installed

rhtenhove:
I have no experience with Travis, so to avoid trying out all sorts of confusing commits, can you point to where I should add the command to have `python3-setuptools` installed?

My guess is to add `python3-setuptools` to `sudo apt install -y` of both `- if [ "$MODERN" = "yes" ]` and `- if [ "$MODERN" = "no" ]`.

vanhauser-thc:
yes.

also note that there are Dockerfiles that are what is in travis in test/travis/[ubuntu-name] - you can experiment there until it works instead of comitting and wait for ages for travis

rhtenhove:
Alright, those two changes should work. Thanks for the pointers!

It seems `ubuntu:trusty` has python3 installed by default, so I've added the python3 version of setuptools to that travis test image.

The temporary symlink solves the issue of the user not having a `python` but having a `python3`. It also recommends to install `python3` if neither is found, for which the symlink will be created during the next try.

CONTRIBUTING.md
```diff
@@ -20,3 +20,8 @@ No camel case at all and use the AFL's macros wherever possible
 Remember that AFLplusplus has to build and run on many platforms, so
 generalize your Makefiles (or your patches to our pre-existing Makefiles)
 to be as much generic as possible.
+
+
+# Local testing
```
vanhauser-thc, line 25: please remove this ;)

qemu_mode/build_qemu_support.sh
```diff
@@ -76,6 +76,18 @@ for i in libtool wget python automake autoconf sha384sum bison flex iconv patch
 
 done
 
+PYTHON_TEMP_SYMLINK=0
+if ! command -v python 1>/dev/null; then
+  if command -v python3 1>/dev/null; then
+    echo "[*] 'python' not found. 'python3' was found. Creating temporary symlink."
+    update-alternatives --install /usr/bin/python python "$(command -v python3)" 0 1>/dev/null
```
vanhauser-thc, line 83: this does not work. the user might not be root or /usr might be mounted read-only.
does alias work? e.g. `alias python=python3` ?
if that does not work create a symlink in the qemu directory and set the search path to look there first. on errors and exit that symlink would need to be removed though

vanhauser-thc:
travis fails in test.sh (make tests) - the python binary detected is the wrong one (not a binary):
```
[*] Testing: unicorn_mode
[*] Using python binary /opt/pyenv/shims/python3
[*] running afl-fuzz for unicorn_mode, this will take approx 25 seconds
...
[-] PROGRAM ABORT : Program '/opt/pyenv/shims/python3' is a shell script
         Location : check_binary(), src/afl-fuzz-init.c:2044
```

rhtenhove:
Went for the --python flag in ./configure for QEMU, seemed a much cleaner approach.

I see @hexcoder- decided to test if _any_ python has setuptools, although that doesn't guarantee that the python used eventually will have setuptools.
Furthermore, I can't see any use of `easy_install` specifically, so I'm not sure why that is being tested instead of the existence of `setuptools`.



> travis fails in test.sh (make tests) - the python binary detected is the wrong one (not a binary):
> 
> ```
> [*] Testing: unicorn_mode
> [*] Using python binary /opt/pyenv/shims/python3
> [*] running afl-fuzz for unicorn_mode, this will take approx 25 seconds
> ...
> [-] PROGRAM ABORT : Program '/opt/pyenv/shims/python3' is a shell script
>          Location : check_binary(), src/afl-fuzz-init.c:2044
> ```

Not sure yet what is happening here.

hexcoder-:
Hi, thanks for your efforts. It took me a while to get this mess sorted out. Now I have a working solution with travis and I am open to suggestions to improve it.

domenukk:
This has been merged-ish with 4c253aedae6aec48d527668d699d124962d82821 - let's see what happens.

domenukk:
Seems to work now, thanks a lot! :)

PR #364, opened 2020-05-16
llvm_mode: perform right-shift of prev_loc if a single-block-loop is detected

risicle:
One problem with total removal of the right-shift before xor-ing with the current loc is that if both locs are the same, it will always xor to zero, and we'll therefore be unable to identify these edges. fortunately it's easy to identify BBs where this is possible because they will have themselves as one of their predecessors.

On my target this seems to result in a ~3% performance penalty, but on 6h runs in nondeterministic mode, fixed seed, `fast` power schedule, afl++ 2.64 with this patch was able to discover ~2900 paths vs ~2700 unpatched.

Tested with LLVM 6, 8 & 9.


vanhauser-thc:
but there is the right shift in the code:
```
Store = IRB.CreateStore(ConstantInt::get(Int32Ty, cur_loc >> 1), AFLPrevLoc);
```
The ShiftRight by llvm is unnecessary as we know the value at compile time.

or do you mean something different?

vanhauser-thc:
also we only accept PRs against dev

risicle:
Oh - you do the shift at _storage_ time. Has it always been this way or am I hallucinating...?

Looking at the original, it _has_ always been this way so my memory must have been failing me and I was looking in the wrong place for the shift.

It's curious then why I saw better performance - I guess any change to the hash function will alter which tuples collide with others and if I were to perform _enough_ runs any "advantage" would evaporate.

risicle:
Incidentally, what is your "gold standard" for testing how a change affects coverage and is it easy to reproduce? Clearly "a 6h run" is not great.

vanhauser-thc:
testing is difficult. in "normal" instrumentation two compilations with the same options will result in different binaries because the edge IDs are random. hence one might randomly generate more or different collisions than others for edges.

So I would say the "gold"  standard is testing against 3 different targets for a minimum of 2 runs each, and it gives a better result in 4 of the 6 runs.

For "how long" that depends on the target. sometimes 10 minutes is enough if it just a difference in how something is done to see if it increases or decreases speed. For instrumentation this is longer, but again depends on the target. a good indicator is running afl-plot and watching until the difference is stabilizing.

As an example: [https://www.fuzzbench.com/reports/2020-05-13-aflplusplus/lcms-2017-03-21_coverage_growth.svg](https://www.fuzzbench.com/reports/2020-05-13-aflplusplus/lcms-2017-03-21_coverage_growth.svg)
if what you are comparing would be aflplusplus_instrim_cmplog_mopt then you would know after 2h it is good. whereas if you would compare aflplusplus_instrim_laf_mopt vs aflplusplus_instrim_laf_mopt_17 it would need around 7 hours to declare the winner.

If something looks good (4 out of 6 runs), then it might be good and I would recommend adding the variant to fuzzbench (https://github.com/google/fuzzbench) - or doing longer tests against more targets.

but dont get discouraged. your one test case shows you might be on to something - or it might be a fluke.
I would recommend: change the code to do (cur_loc >> 2) - maybe shifting by 2 is better than shifting by 1. and then compare against the 3 targets x 2 runs each.

and about comparing coverage, not paths found. paths found is an indicator what is better, but a lower one can have a better real coverage. I recommend [afl-cov](https://github.com/vanhauser-thc/afl-cov). maybe you did that already, just want to be sure.

using the same options and setting the same -s seed option is exactly the right thing to do.

risicle:
Cool - thanks for this

PR #365, opened 2020-05-16
Better sync?

vanhauser-thc:
This is a new method to sync.

it is slightly faster on intial imports and very slightly slower afterwards.

the advantage: much fewer entries are synced to the queue, because I sort the entries and add them reverse, hence the deeper running queue entries get added first, and shallow ones from the beginning will not.

I think this is good, but I want your opinion on this

Timings - (import queue with 2234 files)

new - 978 imports, 23.58s initial import, 0.00118 per reassessment without adding anything

old - 1113 imports, 24.66s initial import, 0.00096 per reassessment without adding anything

Note that I already made the following changes to dev:
 * only masters sync from everyone, and slaves only sync from a master.
 * when a master starts and detects another master running already a warning is printed
 * when a slave starts and detects no master a warning is printed and a 5 second wait

What do you think a) about the new algo, and b) about the changes of slaves only syncing from master?

### DIFF
--- a/src/afl-fuzz-run.c
+++ b/src/afl-fuzz-run.c
@@ -367,13 +367,45 @@ abort_calibration:
 
 }
 
+/* this is a memstr() function where the needle has a fixed length */
+
+static u8 *next_entry(u8 *entry, u8 *file_list, u32 file_list_len) {
+
+  register int i;
+
+  for (i = 0; i < file_list_len - 6; i++)
+    if (memcmp(file_list + i, entry, 6) == 0) return (file_list + i);
+
+  return NULL;
+
+}
+
+/* Generate next entry to search for in the saved directory listing.
+   Dirty hack: only use sprintf if the new number ends in a 0, otherwise
+   we can be cheap and just increase the least significant byte */
+static void update_entry(u8 *entry, u32 next_accept) {
+
+  if (next_accept % 10 == 0)
+    sprintf(entry, "%06u", next_accept);
+  else
+    entry[5] += 1;
+
+}
+
 /* Grab interesting test cases from other fuzzers. */
 
 void sync_fuzzers(afl_state_t *afl) {
 
+  u64             profile_start, profile_end;
+  struct timespec spec;
+  clock_gettime(CLOCK_REALTIME, &spec);
+  profile_start = (spec.tv_sec * 1000000000) + spec.tv_nsec;
+
   DIR *          sd;
   struct dirent *sd_ent;
   u32            sync_cnt = 0;
+  u8 *           file_list, path[PATH_MAX], entry[8] = {0};
+  size_t         file_list_size = 65536;
 
   sd = opendir(afl->sync_dir);
   if (!sd) { PFATAL("Unable to open '%s'", afl->sync_dir); }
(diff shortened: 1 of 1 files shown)


andreafioraldi:
`only masters sync from everyone, and slaves only sync from a master.`
What about when there is no master?

vanhauser-thc:
> `only masters sync from everyone, and slaves only sync from a master.`
> What about when there is no master?

"when a slave starts and detects no master a warning is printed and a 5 second wait"

I also updated all documentation to state that one master is needed.

vanhauser-thc:
yes this can go wrong for people who are used to it without a master. but it hurts performance if everyone syncs from everyone

andreafioraldi:
Master is useless with structured formats. We should suggest to use -M -d instead of -S for that case.

andreafioraldi:
Btw there are tools, like QSYM, that run only one instance of AFL and it can not be a master. What about, if not master is found, do a leader election between the slaves? We can just sort the slaves by name and pick the first.

vanhauser-thc:
> What about, if not master is found, do a leader election between the slaves? We can just sort the slaves by name and pick the first.

That is an option, it then increases complexity. say, what if the master is started later. the newly elected slave would need to see that another master is there now and become a slave again. or if the slave that became a master is terminated, that a different slave picks it up then.

hmmm
but actually it is not that hard to code, and the overhead is limited.

vanhauser-thc:
@andreafioraldi have a look: 25fbec6

src/afl-fuzz-run.c
```diff
@@ -528,11 +535,21 @@ void sync_fuzzers(afl_state_t *afl) {
     closedir(qd);
     ck_free(qd_path);
     ck_free(qd_synced_path);
+    if (n > 0)
```
domenukk, line 538: some brakcets wouldn't hurt :D

src/afl-fuzz-run.c
```diff
@@ -528,11 +535,21 @@ void sync_fuzzers(afl_state_t *afl) {
...
 
   }
 
   closedir(sd);
 
+  clock_gettime(CLOCK_REALTIME, &spec);
+  profile_end = (spec.tv_sec * 1000000000) + spec.tv_nsec;
+  sprintf(path, "echo %016llu %u >> /tmp/profile.out",
+          profile_end - profile_start, afl->queued_imported);
+  if (system(path) != 0) WARNF("system");
```
domenukk, line 551: Why not open, write, close? This seems like it would break eventually on random systems.

src/afl-fuzz-run.c
```diff
@@ -528,11 +535,21 @@ void sync_fuzzers(afl_state_t *afl) {
     closedir(qd);
```
domenukk, line 535: `qd` seeems to be no longer used. This breaks travis?

src/afl-fuzz-run.c
vanhauser-thc: this is just for profiling and testing how long it takes. nothing for production code.

src/afl-fuzz-run.c
vanhauser-thc: right. removed.

src/afl-fuzz-run.c
vanhauser-thc: :p

vanhauser-thc:
> What about if the master dies or crashes half way through fuzzing?

if afl-fuzz crashes then people should create an issue :)

andreafioraldi:
> > What about if the master dies or crashes half way through fuzzing?
> 
> if afl-fuzz crashes then people should create an issue :)

Probably @domenukk is talking about OOM kill bu out of memory of out of disk space (?)

andreafioraldi:
Or maybe the user stops one fuzzer at time to use cmin and for some time the master is down

vanhauser-thc:
if kill -STOP on the master, then it stops. yes.
OOM or disk fill would affect all fuzzers.

I mean of course there are unlikely circumstances or bad usage that can result in problems. but then these are clear problems. IMHO its fine.

vanhauser-thc:
because of conflicts in the sources that github is unable to correctly mark I am closing this. inserted the code by hand. will run a long running test case soon to see the differences.

PR #369, opened 2020-05-21
Update canvas_harness.html

geeknik:
When running this locally as file:///path/to/canvas_harness.html and optionally when calling itself inside the iframe at the end, there is some spam about insecure this or that in the Firefox console. This addition appears to silence that noise.


geeknik:
Eh don't commit this. I silenced some spam and created more. Haha!

vanhauser-thc:
just push again when you have a fix :)

PR #373, opened 2020-05-24
Initialize work for framework executors

rish9101:
Hi, this  is a WIP PR for the executors for the proposed fuzzing framework extension of AFL. I'll be updating this branch as I continue workin on it. Comments and suggestions are very welcome :slightly_smiling_face:

### DIFF
--- a/include/afl-framework.h
+++ b/include/afl-framework.h
@@ -1,0 +1,19 @@
+#include "afl-fuzz.h"
+#include "list.h"
+
+typedef struct afl_executor
+{
+    bool has_forkserver;
+    bool has_fauxserver;
+    afl_state_t * afl_state;
+    u32 map_size;
+} afl_executor_t;
+
+typedef struct executors
+{
+    list_t executors_list;
+    u8 number_of_executors;
+    /* data */
+} afl_executors;
+
+
--- a/src/afl-framework/libexecutors.so.c
+++ b/src/afl-framework/libexecutors.so.c
@@ -1,0 +1,16 @@
+#include "afl-fuzz.h"
+#include "afl-framework.h"
+
+void * executor_run () {
+    //This function runs the exeutor (probably using the afl backend)
+}
+
+
+void * executor_init (afl_executor_t * afl_executor) {
+
+    afl_executor->afl_state = ck_alloc(sizeof(afl_state_t));
+
+    afl_state_init(afl_executor->afl_state, afl_executor->map_size);
+
+}
+


include/afl-framework.h
```diff
@@ -0,0 +1,19 @@
...
+    bool has_forkserver;
+    bool has_fauxserver;
+    afl_state_t * afl_state;
+    u32 map_size;
+} afl_executor_t;
+
+typedef struct executors
+{
+    list_t executors_list;
+    u8 number_of_executors;
```
vanhauser-thc, line 15: u8? I would propose u32. 256+ cores is a thing :)
plus maybe we will have remote executors

include/afl-framework.h
```diff
@@ -0,0 +1,19 @@
+#include "afl-fuzz.h"
+#include "list.h"
+
+typedef struct afl_executor
+{
+    bool has_forkserver;
+    bool has_fauxserver;
+    afl_state_t * afl_state;
+    u32 map_size;
+} afl_executor_t;
```
vanhauser-thc, line 10: would there not be a forkserver state? or is that in the afl_state?
and why is the map_size extra? would that not be in the afl_state or the forkserver_state? just asking so I understand this

src/afl-framework/libexecutors.so.c
```diff
@@ -0,0 +1,16 @@
+#include "afl-fuzz.h"
```
vanhauser-thc, line 1: will there be multiple libraries or just one? if there is just one then this should be called libafl++.so.c IMHO

include/afl-framework.h
rish9101: The forkserver state is in the afl_state itself. Regarding the map_size, we set the map size through an environment variable or it is the default value, Since we're not exactly exposing the real forkserver struct (which finally contains the map size) I thought including it here would be a good way.

include/afl-framework.h
rish9101: Ohh. Ok. I'll change this. :slightly_smiling_face:

src/afl-framework/libexecutors.so.c
rish9101: This I'm not sure about yet. We're supposed to have executors, mutators, seed schedulers etc, so having different libraries for them, is it a good idea?

include/afl-framework.h
vanhauser-thc: then maybe "initial_map_size" or "map_size_default" or something like that?
if it is needed?

src/afl-framework/libexecutors.so.c
vanhauser-thc: hmm If they can be independent than we could have different libraries. if however a feature requires features from other libraries then splitting them up provides no value IMHO.
Maybe we first have only one library and if it makes sense to split it up this is future work? @domenukk ? @hexcoder- ?

vanhauser-thc:
I think @andreafioraldi said he wanted to help on the design/architecture of this starting this week. so perfect time for this :)

rish9101:
That's great!! I'll contact him regarding this. :slightly_smiling_face:

vanhauser-thc:
dominik and andrea are busy until thursday, cut them 2 days slack :)

vanhauser-thc:
Also I think this is the repo where the work should be done: https://github.com/AFLplusplus/LibAFL

rish9101:
Yes, I'll be closing this PR tomorrow and we'll be shofting our work over to that repo. Thanks 🙂

src/afl-framework/libexecutors.so.c
hexcoder-: I would use extra libraries for optional things (which need not to be loaded every time). But that would be an optimization, so having one library at first seems reasonable.

rish9101:
I am closing this PR as the work on the library is shifted to a separate repository

PR #375, opened 2020-05-25
llvm_mode compare-transform-pass: add handling of sized comparisons with non-const size

risicle:
Don't know if this is of any use to you, but I've been using this modification with some success.

Some targets use sized comparisons with non-const `n` values (in cases I've seen this is usually the length of the variable-string's buffer). These can be handled by insertion of a "length check" bb for each unrolled character, branching to the phi node with a zero return value if the n argument equals each character's position index.

To make this easier to tackle I did a bit of refactoring of the length determination and renaming of variables.

And to give myself more confidence that I'm not breaking any of this existing logic I added some extra cases to `test-compcov.c` and extra steps to `test.sh` that actually test the correctness of the resulting binary's behaviour, also adding calls to these checks in other places the file is compiled to what should be a functioning binary. Excuse my horrible bash.

Oh, and I also added handling of a corner case where the const string has embedded null characters. In such a case, the `str*` functions should end comparison after the first-encountered null, not continue comparing to the end of the const.

### DIFF
--- a/llvm_mode/compare-transform-pass.so.cc
+++ b/llvm_mode/compare-transform-pass.so.cc
@@ -304,17 +304,24 @@ bool CompareTransform::transformCmps(Module &M, const bool processStrcmp,
           if (!(HasStr1 || HasStr2)) continue;
 
           if (isMemcmp || isStrncmp || isStrncasecmp) {
-
             /* check if third operand is a constant integer
              * strlen("constStr") and sizeof() are treated as constant */
             Value *      op2 = callInst->getArgOperand(2);
             ConstantInt *ilen = dyn_cast<ConstantInt>(op2);
-            if (!ilen) continue;
-            /* final precaution: if size of compare is larger than constant
-             * string skip it*/
-            uint64_t literalLength = HasStr1 ? Str1.size() : Str2.size();
-            if (literalLength + 1 < ilen->getZExtValue()) continue;
+            if (ilen) {
+              uint64_t len = ilen->getZExtValue();
+              // if len is zero this is a pointless call but allow real
+              // implementation to worry about that
+              if (!len) continue;
 
+              if (isMemcmp) {
+                // if size of compare is larger than constant string this is
+                // likely a bug but allow real implementation to worry about
+                // that
+                uint64_t literalLength = HasStr1 ? Str1.size() : Str2.size();
+                if (literalLength + 1 < ilen->getZExtValue()) continue;
+              }
+            } else continue;
           }
 
           calls.push_back(callInst);
@@ -341,7 +348,7 @@ bool CompareTransform::transformCmps(Module &M, const bool processStrcmp,
     Value *     VarStr;
     bool        HasStr1 = getConstantStringInfo(Str1P, Str1);
     bool        HasStr2 = getConstantStringInfo(Str2P, Str2);
-    uint64_t    constLen, sizedLen;
+    uint64_t    constStrLen, constSizedLen, unrollLen;
     bool        isMemcmp =
         !callInst->getCalledFunction()->getName().compare(StringRef("memcmp"));
     bool isSizedcmp = isMemcmp ||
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
Quite some change, will have a look on it on Wednesday

vanhauser-thc:
also - travis failed

risicle:
Looks like a `bash`ism on my part.

llvm_mode/compare-transform-pass.so.cc
```diff
@@ -304,17 +304,27 @@ bool CompareTransform::transformCmps(Module &M, const bool processStrcmp,
...
-            uint64_t literalLength = HasStr1 ? Str1.size() : Str2.size();
-            if (literalLength + 1 < ilen->getZExtValue()) continue;
-
+            if (ilen) {
+              uint64_t len = ilen->getZExtValue();
+              // if len is zero this is a pointless call but allow real
+              // implementation to worry about that
+              if (!len) continue;
+
+              if (isMemcmp) {
```
vanhauser-thc, line 317: why do you limit this check to memcmp? just a curious question

llvm_mode/compare-transform-pass.so.cc
```diff
@@ -389,99 +389,133 @@ bool CompareTransform::transformCmps(Module &M, const bool processStrcmp,
...
       VarStr = Str1P;
-      constLen = isMemcmp ? sizedLen : TmpConstStr.length();
 
     }
 
-    /* properly handle zero terminated C strings by adding the terminating 0 to
-     * the StringRef (in comparison to std::string a StringRef has built-in
-     * runtime bounds checking, which makes debugging easier) */
+    // add null termination character implicit in c strings
     TmpConstStr.append("\0", 1);
```
vanhauser-thc, line 411: this is not your change ... though thinking about it - for memcmp we should not add a \0 maybe?

llvm_mode/compare-transform-pass.so.cc
risicle: Because in string comparison it's perfectly valid - if you supply an `n` greater than the const string's length the comparison will still terminate at the end of the const string because either

 - the variable string won't have a null char in that position or
 - the variable string _does_ have a null char in that position, but of course that denotes the end of the variable string and so the comparison should terminate.

This might be a perfectly reasonable thing to do if e.g. the code just uses the var string's allocation length for `n` to guard against missing terminators.

In the `memcmp` case, the presence of a null char won't terminate the comparison and the expected behaviour would be to go on comparing past the end of the const string. I'd hope the compiler or `FORTIFY` flags would catch this kind of thing but I didn't look too far into it.

llvm_mode/compare-transform-pass.so.cc
risicle: A string literal in c will have a null termination character added to it whether it then goes on to be used in a memcmp comparison or not. And it doesn't make the unrolling any longer, because the `unrollLen` of a memcmp is governed by `n` anyway.

This is a perfectly valid thing to do:

```c
memcmp(bar, "foo", 4);
```

because the 4th char of `foo` can be expected to be a null.

vanhauser-thc: yes. but memcmp can also compare non-strings, e.g. memcmp(bar "\x01\x02\x03\x04\x05", 5)
so here a null should only be added if strlen() of the string is one less than the constlen.

llvm_mode/compare-transform-pass.so.cc
vanhauser-thc: true

llvm_mode/compare-transform-pass.so.cc
risicle: You should find that in that case the `unrollLen` does indeed get calculated to be 5 unless I've got my logic wrong.

```c
unrollLen = constSizedLen < constStrLen ? constSizedLen : constStrLen;
```

Adding the null here doesn't actually make any difference to it in the end, I just think it's clearer to add it anyway

llvm_mode/compare-transform-pass.so.cc
```diff
@@ -341,31 +348,20 @@ bool CompareTransform::transformCmps(Module &M, const bool processStrcmp,
...
-    uint64_t    constLen, sizedLen;
+    uint64_t    constStrLen, constSizedLen, unrollLen;
     bool        isMemcmp =
         !callInst->getCalledFunction()->getName().compare(StringRef("memcmp"));
     bool isSizedcmp = isMemcmp ||
                       !callInst->getCalledFunction()->getName().compare(
                           StringRef("strncmp")) ||
                       !callInst->getCalledFunction()->getName().compare(
                           StringRef("strncasecmp"));
+    bool isConstSized = isSizedcmp && isa<ConstantInt>(callInst->getArgOperand(2));
```
hexcoder-: What about getArgOperand(1)? *cmp() function's first argument could be constSized, right?

vanhauser-thc: constSize means: memcmp(foo, bar, 5). if there is a '2' it is a constant len. in contrast to memcmp(foo, bar, len) where it is a variable that can not be resolved. operand(0) is foo, operand(1) is bar, operand(2) is the length parameter, either a const value or a variable. its fine.

hexcoder-: Oops, i mixed up the arg numbering, sorry. Thanks for clarification.

vanhauser-thc:
thank you!

PR #376, opened 2020-05-27
Fix build on powerpc64

pkubaj:
powerpc64 needs the same treatment that aarch54 does.


vanhauser-thc:
#377 proposes a better solution :)

PR #383, opened 2020-05-29
Fix read_timed when accumulating short reads

dgmelski:
The existing code appears to use 'len_read' in several places where
'total_read' was intended.  The function may work if the first 1 or 2
iterations of the loop read the requested 'len' bytes.

If the first two reads are "short" and a third read is done, the bytes
will be placed over previously read bytes in buf and more than 'len'
bytes may be read in total, though buf is never overrun.

This commit changes read_timed to

  (1) correctly append short reads in buf
  (2) correctly terminate when the sum of the short reads
      equals the requested 'len' bytes
  (3) return an error when read() returns -1 or 0

The function also depends on select() decrementing the timeout
structure, as it does on Linux.  On other platforms, the exec_ms
returned is likely incorrect.  This patch does not attempt to address
this issue.


vanhauser-thc:
@hexcoder- : we can change the destination of the PRs  via the Edit button to the right of the PR's title. just did it for this one

vanhauser-thc:
> The function also depends on select() decrementing the timeout
structure, as it does on Linux. On other platforms, the exec_ms
returned is likely incorrect. This patch does not attempt to address
this issue.

@devnexen can you please have a look and give feedback if we have to handle select differently for BSD platforms?

dgmelski:
My apologies for putting the PR against the wrong branch. Thank you for adjusting it.

GitHub is now showing this PR as involving 4 commits and 4 files, while I only touched one. Did switching the target branch somehow change what is included in the PR? I cannot speak to the other commits and whether or not they belong on dev. FYI, in cases someone wants to review them.

vanhauser-thc:
> My apologies for putting the PR against the wrong branch. Thank you for adjusting it.
> 
> GitHub is now showing this PR as involving 4 commits and 4 files, while I only touched one. Did switching the target branch somehow change what is included in the PR? I cannot speak to the other commits and whether or not they belong on dev. FYI, in cases someone wants to review them.

yeah that can happen in cases were we have to rebase.
But I am surprise @hexcoder- did you commit the  "LDFLAGS += -Wl,-z,notext -L /usr/local/lib/" and test.sh stuff to master?

hexcoder-:
Oops, Yes, i did that, forgot to switch the newly installed repo to dev before fixing it (for me default master branch is pretty evil!). Sigh, yet another commit to the wrong branch...

dgmelski:
If it helps, I can put in a new PR for the one commit to dev. Though probably you can sort out what you want with what you have?

vanhauser-thc:
I work with subversion and dont have this issue :)
It looks like this on the filesystem:
```
AFLplusplus/trunk
AFLplusplus/branches/dev
AFLplusplus/branches/foo
```
and I symlinked the dev branch to make it simpler for me :)

domenukk:
@dgmelski looks good, thanks for spotting that! :)
@vanhaser-thc @hexcoder-  we can block pushes to master alltogether, should we do that? Then, only PRs can be accepted to master. There should never be a reason to push directly (and if there is, you can disable it for a short time)

PR #384, opened 2020-05-29
 Grammar for fuzzing Perl

RootUp:
As part of my continuous fuzzing with `AFL`. I utilized some common functions for fuzzing basic I/O in `perl5`. Hope this helps others while fuzzing `perl5`. I'll keep updating this dictionary in future.


hexcoder-:
Sigh, I forgot to check the branch for the PR (which is the wrong one). We would accept PRs only against the dev branch (and not master as you did).

PR #385, opened 2020-05-29
Support UNIX-style wildcards in AFL_LLVM_WHITELIST file

sirmc:
This patch allows use of wildcards (e.g., `*/path/to/*.cpp`) in the LLVM whitelist file. To keep backwards-compatibility, this feature is only enabled when the env variable `AFL_LLVM_WHITELIST_FNMATCH` is set.

Alternatively, we could set this to the default behavior, and automatically prepend `*/` to every whitelist entry, however, this might break backwards-compatibility for absolute paths in the whitelist.


vanhauser-thc:
Thanks that is useful :)
Why an env var to enable this? as I guess the '*' is not required for a line there is nothing that would break for existing users. Or am I missing something?

sirmc:
The issue is that the current behavior does a "suffix" match on the file path.

For example, the current "suffix"-based behavior would match `a.cpp` as well as `sub/dirs/a.cpp` for the whitelist entry `a.cpp`. 

If enabling the fnmatch behavior by default, something like `a.cpp` would only match the exact file `a.cpp`, and not `sub/dirs/a.cpp`.

vanhauser-thc:
but this could be combined so that
test.c and foo/bar/test.c and test.* and foo/bar/test.* are all basically the same thing.

requiring paths or matching on them leads to failure, as many environments do "cd src ; make ; cd contrib ; make" stuff in Makefiles and sub-Makefiles

sirmc:
You're right that matching on whole paths might not make sense here:
I think we can simple match on `*<entry>`. This should yield the same behavior as before, while also allowing patterns in the entry name itself (e.g., `Debug*.cpp`).

vanhauser-thc:
looks good, thanks!

vanhauser-thc:
what happens if the whitelist is "*/foo.c" ? is the double * a problem?

vanhauser-thc:
I also added it to llvm_mode/afl-llvm-lto-whitelist.so.cc

sirmc:
The double `*` should not be an issue. A whitelist with the following`*/b.c` only matches files `b.c` in subdirectories AFAIK (so no match in the root level).

Is there any infrastructure to add tests for AFL++? It might make sense to have some test cases to see that the behavior is as expected.

vanhauser-thc:
that is in test/   mostly in test/test.sh but also a few unittests with cmocka

PR #390, opened 2020-05-30
OpenBSD build update and llvm mode fix

devnexen:
Unlike upstream version, LLVM in OpenBSD enable by default anti ROP gadget
 leading to bigger binaries and lower performances.
On OpenBSD, it needs to link to c++ abi for th unwind symbols.


hexcoder-:
Thanks! Now testing with 6.6-64-bit and 6.7-32-bit...

PR #392, opened 2020-06-02
Push to master

vanhauser-thc:
Basically this has:
  wildcard support in whitelist (llvm mode so far only)
  fix for timed read to also working on non-linux
  shmem fuzzing for persistent mode (and unicorn at time we merge)
  fuzzbench driver for shmem (not of interesting to anyone else ;)
  master/slave renaming
  powerpc and other 64 bit support


vanhauser-thc:
I would say we could add wildcard support for whitelisting with gcc_plugin but as we get something better in a week or two I dont think that effort is useful.

for shmem fuzzing:
 - unicorn:  @domenukk is doing the part at the moment
 - qemu: we dont need that, we have something better to write directly into registers

I am thinking about writing a preload module that hooks read and open and depending on the @@/-f/none parameter then injects there. not sure it makes sense for the effort.
what do you guys think?

examples/persistent_demo/test-instr.c
```diff
@@ -0,0 +1,67 @@
...
+#include <fcntl.h>
+
+__AFL_FUZZ_INIT();
+
+int main(int argc, char **argv) {
+
+  __AFL_INIT();
+  unsigned char *buf = __AFL_FUZZ_TESTCASE_BUF;
+
+  while (__AFL_LOOP(2147483647)) {
```
domenukk, line 28: ... MAX_INT Constant? Or -1? That's an odd number to use in an example. At least comment what it means :)

src/afl-fuzz-run.c
```diff
@@ -412,17 +412,17 @@ void sync_fuzzers(afl_state_t *afl) {
 
     entries++;
 
-    // a slave only syncs from a master, a master syncs from everyone
-    if (likely(afl->is_slave)) {
+    // secondary nodes only syncs from main, the main node syncs from everyone
+    if (likely(afl->is_secondary_node)) {
 
-      sprintf(qd_path, "%s/%s/is_master", afl->sync_dir, sd_ent->d_name);
+      sprintf(qd_path, "%s/%s/is_main_node", afl->sync_dir, sd_ent->d_name);
```
domenukk, line 419: If `afl->sync_dir "/" sd_ent->d_name` is already at `PATH_MAX`, this will break...
Nothing we need to fix necessarily, but we could increase the `qd_path` buffer size by a bit and check the length here.

examples/persistent_demo/test-instr.c
vanhauser-thc: max int needs another include that might be different on BSD etc. so I avoided it :) I will add a comment

src/afl-fuzz-run.c
vanhauser-thc: It is impossible to exceed MAX_PATH, as this is the restriction of the OS unless you stuff to the string that has nothing to do with the path/filename

domenukk: Yes but if the path length is already close to max, we still have a buffer overflow here that we could catch

docs/parallel_fuzzing.md
```diff
@@ -57,26 +57,26 @@ Each fuzzer will keep its state in a separate subdirectory, like so:
...
 experimental support for parallelizing the deterministic checks. To leverage
 that, you need to create -M instances like so:
 
 ```
-./afl-fuzz -i testcase_dir -o sync_dir -M masterA:1/3 [...]
-./afl-fuzz -i testcase_dir -o sync_dir -M masterB:2/3 [...]
-./afl-fuzz -i testcase_dir -o sync_dir -M masterC:3/3 [...]
+./afl-fuzz -i testcase_dir -o sync_dir -M mainA:1/3 [...]
+./afl-fuzz -i testcase_dir -o sync_dir -M mainB:2/3 [...]
+./afl-fuzz -i testcase_dir -o sync_dir -M mainC:3/3 [...]
```
hexcoder-, line 76: Did we test that experimental mode lately? Just curious.

src/afl-forkserver.c
```diff
@@ -839,8 +894,6 @@ fsrv_run_result_t afl_fsrv_run_target(afl_forkserver_t *fsrv, u32 timeout,
 
   MEM_BARRIER();
 
-  if (fsrv->shdmem_fuzz_len) write_value += (fsrv->shdmem_fuzz_len << 8);
```
hexcoder-, line 842: Don't really understand this change. Is it not needed anymore?

src/afl-fuzz-run.c
vanhauser-thc: I think now I understand what you mean. dunno. we would need to at 256 bytes to the vars, and for a case that will never happen and I would account for "not very intelligent usage" ;) but a stack cost for everyone.

docs/parallel_fuzzing.md
vanhauser-thc: no one. but I made a change that writes a warning if another master is already running when starting up. someone could/should test it though (someone != me  ;)

src/afl-forkserver.c
vanhauser-thc: no, the length is now transfered via the shm fuzz map. more flexible

vanhauser-thc:
btw - travis is failing because of unicorn not succeeding in test.sh

PR #396, opened 2020-06-07
fix a fundamental bug: favor random branches

Kyle-Kyle:
In `never zero counter` patch, it uses `inc` and then `adc` to make sure  the coverage count will never be set to 0.(if an overflow happens, `adcb $0, (%edx, %edi, 1)` will make the value become 1).

It looks fine until I found out that `inc` actually does not update carry flag and `adc` actually uses the carry flag to decide whether an addition should be performed.

The last time the carry flag gets updated is in `shrl $1, %ecx\n`, where the `ecx` is the branch index(randomly assigned during compilation). 
So the logic becomes: if the branch index is odd, add 2 to the counter, else add 1.
This makes `AFLplusplus` favor some random branches for no reason.


hexcoder-:
Thanks very much for your analysis. This error is kind of embarassing, since I am sure, at sometime it worked like intended...

hexcoder-:
BTW: I merged this to the dev branch, since we don't accept PRs to master...

vanhauser-thc:
I hope no one uses afl-gcc anymore. its so outdated and not supported anymore ... llvm_mode does not have this bug :)
well spotted!

PR #404, opened 2020-06-14
Porting to Haiku.

devnexen:
getrusage does not implement resident memory gathering, no shm api neither.


vanhauser-thc:
haiku looks like a moving target, over time it will receive more features which will then need to be activated here.
thanks!

PR #418, opened 2020-06-20
Fix when env. PATH contains spaces

jonasmollerlarsen:
I don't advocate having spaces in a directory name, but this change should fix test/test.sh in that case.

By the way, is the travis CI status visible somewhere? afl-fuzz -R (radamsa) somehow stopped working on my machine, and it would be useful to know whether it's a local issue or if something might've broken in the code tree.


jonasmollerlarsen:
Answering my own question, https://travis-ci.com/github/AFLplusplus/AFLplusplus
it must be a local issue with radamsa then.

vanhauser-thc:
thanks!

PR #429, opened 2020-06-30
Big renaming

vanhauser-thc:
Please help me fix the renaming of blacklist/whitelist if anything went wrong or is not optimal.


vanhauser-thc:
please fix directly and push instead of nit picking :)

andreafioraldi:
I see a huge risk in this PR, maybe should land after the release

andreafioraldi:
I don't see bugs, good that you didn't break backward compatibility

vanhauser-thc:
fine with me. what is the risk though? travis is all green.

andreafioraldi:
mmmm we did a hotfix release for last minutes features before the previous release more than one time, just saying, maybe the changes are so little that is ok

vanhauser-thc:
we only had one bug release.
There was another time we did a short time release but that was for fixing bugs that were introduced in previous versions but important.
but yes. if it is a risk we should move it. I dont think it is one, but when we all check over the changes we get a better feeling if it is a complex change or not and then decide. we have 13 hours for that :)

vanhauser-thc:
the change here does still process AFL_*_WHITELIST env vars, however afl-common will report it is unknown.
I am now not sure this is good. This might make people think it is not working when however it does.
I want to discourage people using the old env var so they switch, but also not break backward compatability.
opinions?

andreafioraldi:
just show a deprecatio warning, not a not exists warning

vanhauser-thc:
we dont have something like this yet.
should I add it?

PR #433, opened 2020-07-01
final push to stable before release?

vanhauser-thc:


andreafioraldi:
+1 from me, tested some mix of valid/invalid/deprected and worked

PR #435, opened 2020-07-01
libtokencap illumos/solaris support proposal.

devnexen:


hexcoder-:
Thanks! Cannot verify just now. Tests are being passed on illumos?

devnexen:
Yes it does now :-)

PR #441, opened 2020-07-05
Fix generation of afl-system-config.8

bmwiedemann:
Without this patch, afl-system-config.8 varied between build hosts
because it contained lines such as
```
 ./afl-system-config: line 30: sysctl: command not found
 ./afl-system-config: line 31: /sys/kernel/mm/transparent_hugepage/enabled: Permission denied
 It is recommended to boot the kernel with lots of security off
```

This PR was done while working on reproducible builds for openSUSE.


vanhauser-thc:
I this patch incomplete?

it only contains
```
- test "$1" = "-h" && {
+ test "$1" = "-h" -o "$1" = "-hh" && {
```

bmwiedemann:
That is correct. See the `%.8` target in the top `GNUmakefile` that calls commands with `-hh` to fill the OPTIONS section

It seems `./afl-whatsup -hh` could also need some improvement, but at least it prints a usage.

vanhauser-thc:
ok.
please push to dev in the future, we do not accept PRs to stable/master.

PR #442, opened 2020-07-05
Dont always use march=native

bmwiedemann:
When a reproducible distribution build is wanted,
it is not good to make build results depend on the build machine's CPU

There might be other good ways to fix this?

See https://reproducible-builds.org/ for why this matters.

This PR was done while working on reproducible builds for openSUSE.


vanhauser-thc:
I agree for reproducible builds. hurts performance though but thats the disadvantage ppl have to live with then

Issue #424, opened 2020-06-24
low_freq.png graph displays integers and should therefore scale its y-values without decimal places 


toralf:
well, it is just cosmetic, but ....

[edit]: and the x-axis should indicates that times are in UTC[/edit]

vanhauser-thc:
@toralf I have no experience with gnuplot. I also do not see that as a big issue.
please send a pull request :)

PR #444, opened 2020-07-05
Fix y axis of low freq.png

toralf:
pls see https://github.com/AFLplusplus/AFLplusplus/issues/424#issuecomment-653878828


toralf:
ick, made against wrong branch (stable instead dev) - will redo a PR

PR #449, opened 2020-07-09
illumos littlefixes: little typo for cpu binding and

devnexen:
even tough gcc plugin less good than LLVM, clang
is more buggy on this os.


vanhauser-thc:
thanks!

PR #453, opened 2020-07-16
Tree mutation

adrian-rt:
Superion ported as a library to do grammar based mutation for (javascript/php/jerryscript etc) see
https://github.com/zhunki/Superion/


devnexen:
cmake produces a lot of content for its build doesn't it ? ;-)

vanhauser-thc:
@am06 thank you for your PR, here are some pointers:

1. always check first with a project how the PR process works. for us you can see in the README.md that we only accept PRs to the dev branch and "make code-format" has to be done (plus added for newly created subdirectories). I just changed the push target for you.
2. always check the diff generated by the PR - often debug stuff to other files, temporary files etc are part of a large PR, and these need then to be cleaned. in your case you would have noticed that the PR would have changed a lot you didnt do - because you pushed to the wrong branch :)
3. custom mutators have their own subdirectory: custom_mutators/ please put it there (there is only one other at the moment, radamsa)
4. I would not accept the PR because the solution uses cmake. I dont know the opinions of the others but I think they agree - no other build systems allowed other then build shell scripts and Makefile/GNUmakefile
5. shouldnt there be a README.md? the build files suggest that, but it is not in the PR

for 4) that means that there are two options. Either you remove the cmake stuff and put what is required in a Makefile (that usually works easily. cmake is overblown for most projects anyway.) - or you host it externally and we link to it from READMEs

vanhauser-thc:
as it was pointed out to me the code quality is also not as it should be (I have not looked at your code yet).
I dont mind shepherding it until it is where it should be, it would be quite some time and work for you though.

adrian-rt:
Hi all,

thanks for the advice, I'm not a professional developer as you probably noticed. It's my first attempt back at C/coding in 15 years.

Unfortunately I can't dedicate more time to this for refactoring.  I just wanted to get it working and share it with you and the community.

No idea where the README.md came from. probably generated by cmake..... 

The original project used cmake so......I had to work with that.

Regarding instruction to use I simply used your env variables:

export AFL_CUSTOM_MUTATOR_ONLY=1
export AFL_CUSTOM_MUTATOR_LIBRARY=~/Downloads/afl++/tree_mutation/js_parser/libTreeMutation.so

you have the build.sh script folder inside to compile the library (which uses cmake, yep) .

Please let me know if you need anything else from me. It should be easy to get it up & running.

vanhauser-thc:
If you do not have the time then create new repository, e.g. `superion_aflplusplus_custom_mutator` and put the files from tree_mutation in there with a short README.md
I will reference to it from our main README and our custom_mutator/README.
This way people will find it and can use it - and maybe even improve on it

domenukk:
No need to be a professional C coder at all @am06  ;) Thanks for the contribution.
It's definitely cool to have superion available as plugin in any case!

tree_mutation/js_parser/TreeMutation.cpp
```diff
@@ -0,0 +1,315 @@
...
+                       uint8_t **out_buf, // output buffer
+                       uint8_t *add_buf, size_t add_buf_size,  // add_buf can be NULL
+                       size_t max_size) {
+
+	int fd, i=0, n=0, new_hit_cnt=0, orig_hit_cnt;
+	size_t  retlen;					 
+	char* retbuf;  
+    static s32 splicing_with = -1;        /* Splicing with which test case?   */
+	static s32 stage_cur, stage_max;      /* Stage progression     */
+	afl_state_t *afl = (afl_state_t * )data;
```
domenukk, line 108: `afl_state_t` is the internal representation and may change when recompiling afl++. It's probably not safe to use here. 
Do you think we should add an official way to set a custom name for the stage?
cc @andreafioraldi

adrian-rt:
@vanhauser-thc  ok, I can definetely do that and will try to submit in the next few days. 
@domenukk you're welcome, it's been a pleasure to get back in the game and thanks for all your support.

Please let me know, if you need any other changes like that afl_state_t thing there....Yes, initially I wanted to add a custom name to the stage :) .....

tree_mutation/js_parser/TreeMutation.cpp
vanhauser-thc: we could do that but that would hurt performance a bit. IMHO it is not so important? "custom" is enough I think, and if someone has multiple custom mutators there is no way to see which one is more successful than others hence I dont see much value there either - besides "the UI shows a specific name".

domenukk: We do support stacked/multiple custom mutators, setting it once per stage shouldn't be a big issue I think? With the same reasoning we could hide the current stage completely (because who cares what stage found the bug?) - but it's still nice to know what is going on right now.

vanhauser-thc:
this one is obsolete

PR #454, opened 2020-07-16
Update post_library_gif.so.c

wizche:
Wrong function prototype


domenukk:
Thank you :)
Hm I feel like this should have been in some tests

PR #458, opened 2020-07-18
Allow the fuzzer to write empty buffer to the file

h1994st:
Sometimes, the custom mutator will generate an empty buffer for the test case (e.g., trimming in the grammar mutator). The only change in this pull request is to allow the fuzzer to write empty buffer (i.e., `len == 0`) to the file while using custom mutators.

In my case, the custom mutator does not define `custom_post_process`, so `new_buf` will be `NULL`. If `len` is 0, `new_size` will remain 0 as well. When the fuzzer checks `!new_buf && (new_size < 0)` in `afl-fuzz-run.c` Line 112, it will generates an error.

I have run simple tests. No error is observed. Hope it works fine.

### DIFF
--- a/src/afl-fuzz-mutators.c
+++ b/src/afl-fuzz-mutators.c
@@ -286,6 +286,15 @@ u8 trim_case_custom(afl_state_t *afl, struct queue_entry *q, u8 *in_buf,
           "Trimmed data returned by custom mutator is larger than original "
           "data");
 
+    } else if (unlikely(retlen == 0)) {
+
+      /* Do not run the empty test case on the target. To keep the custom
+         trimming function running, we simply treat the empty test case as an
+         unsuccessful trimming and skip it, instead of aborting the trimming. */
+
+      ++afl->trim_execs;
+      goto unsuccessful_trimming;
+
     }
 
     write_to_testcase(afl, retbuf, retlen);
@@ -324,6 +333,8 @@ u8 trim_case_custom(afl_state_t *afl, struct queue_entry *q, u8 *in_buf,
       }
 
     } else {
+
+unsuccessful_trimming:
 
       /* Tell the custom mutator that the trimming was unsuccessful */
       afl->stage_cur = mutator->afl_custom_post_trim(mutator->data, 0);
--- a/src/afl-fuzz-run.c
+++ b/src/afl-fuzz-run.c
@@ -109,7 +109,7 @@ void write_to_testcase(afl_state_t *afl, void *mem, u32 len) {
 
     });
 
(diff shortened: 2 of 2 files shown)


domenukk:
What is the reason to run the target with an empty file, though?
It's rather likely the mutator is broken in that case, isn't it?
Maybe we should allow an early return here?

h1994st:
I am working on the grammar mutator, which applies mutations and trimmings to a tree representation. Sometimes, the mutated tree just represents an empty string, which is still a valid test case though. That's why I want to run the target with an empty file.

domenukk:
How frequently does this occur? At the very least we should add special handling to never execute an empty test, then? 
Also, I just assume most of the times this will be a bug in a mutator

vanhauser-thc:
I agree that the FATAL is a bad choice. sending 0 byte to the target is not good either though.
I think the right move is to goto abadon_entry on 0 byte results and FATAL on < 0.

h1994st:
> How frequently does this occur? At the very least we should add special handling to never execute an empty test, then?

I haven't tested the frequency, but it is definitely not a common case.

> Also, I just assume most of the times this will be a bug in a mutator

I guess this should not be a bug for the mutator, because some rules in the given grammar can generate the empty string.

h1994st:
> I agree that the FATAL is a bad choice. sending 0 byte to the target is not good either though.
> I think the right move is to goto abadon_entry on 0 byte results and FATAL on < 0.

Yes, I agree that skipping the 0-byte test case is better than sending 0-byte test case to the target.

Will update the pull request later.

vanhauser-thc:
your first commit was in write_to_testcase, your last commit removed that and only has a change in trimming. I think you want both?

h1994st:
Hi @vanhauser-thc ,

I assume that AFL++ without the custom mutator handles 0-byte test cases well, so I only handle the empty test cases generated by the custom mutator. In my case, only `afl_custom_fuzz` and `afl_custom_trim` may generate empty test cases.

I discarded the changes in `write_to_testcase`, in which `afl_custom_post_process` in invoked, because I think `afl_custom_post_process` should not return 0-byte buffer.

For `afl_custom_fuzz`, [afl_fuzz_one.c:1742](https://github.com/AFLplusplus/AFLplusplus/blob/0da0b5cba05d1b49b7ca10987d11a74ab54892e0/src/afl-fuzz-one.c#L1742) indicates that the fuzzer will skip the 0-byte test case.

For `afl_custom_trim`, my second commit handles the 0-byte test case generated by the custom trimming.

vanhauser-thc:
ah ok, thanks for clarifying! :)

PR #459, opened 2020-07-18
Fix the bug of reading environment variables

xiaoxiaoafeifei:
Hello，when I read the code, I found a bug when the **read_afl_environment** function reads environment variables

### DIFF
--- a/src/afl-fuzz-state.c
+++ b/src/afl-fuzz-state.c
@@ -164,229 +164,227 @@ void afl_state_init(afl_state_t *afl, uint32_t map_size) {
 
 void read_afl_environment(afl_state_t *afl, char **envp) {
 
-  int   index = 0, issue_detected = 0;
+  int index = 0, issue_detected = 0;
   char *env;
   while ((env = envp[index++]) != NULL) {
 
-    if (strncmp(env, "ALF_", 4) == 0) {
+    if (strncmp(env, "ALF_", 4) != 0) {
 
       WARNF("Potentially mistyped AFL environment variable: %s", env);
       issue_detected = 1;
+      continue;
+    }
 
-    } else if (strncmp(env, "AFL_", 4) == 0) {
+    int i = 0, match = 0;
+    while (match == 0 && afl_environment_variables[i] != NULL) {
 
-      int i = 0, match = 0;
-      while (match == 0 && afl_environment_variables[i] != NULL) {
+      size_t afl_environment_variable_len =
+          strlen(afl_environment_variables[i]);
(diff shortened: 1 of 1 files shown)


src/afl-fuzz-state.c
```diff
@@ -164,229 +164,227 @@ void afl_state_init(afl_state_t *afl, uint32_t map_size) {
 
 void read_afl_environment(afl_state_t *afl, char **envp) {
 
-  int   index = 0, issue_detected = 0;
+  int index = 0, issue_detected = 0;
   char *env;
   while ((env = envp[index++]) != NULL) {
 
-    if (strncmp(env, "ALF_", 4) == 0) {
+    if (strncmp(env, "ALF_", 4) != 0) {
```
vanhauser-thc, line 171: you inverse this. why? this makes it all wrong

vanhauser-thc:
I think you misinterpreted the code.
some people write AFL_DEBUG when they mean AFL_DEBUG and this is what the function warns about. your change removes that feature.

src/afl-fuzz-state.c
xiaoxiaoafeifei: void read_afl_environment(afl_state_t *afl, char **envp) {

  int   index = 0, issue_detected = 0;
  char *env;
  while ((env = envp[index++]) != NULL) {

    if (strncmp(env, "ALF_", 4) == 0) {

      WARNF("Potentially mistyped AFL environment variable: %s", env);
      issue_detected = 1;

    } else if (strncmp(env, "AFL_", 4) == 0) {


I have a doubt， what is the meaning of the same condition as if and else branch？

xiaoxiaoafeifei:
> I think you misinterpreted the code.
> some people write AFL_DEBUG when they mean AFL_DEBUG and this is what the function warns about. your change removes that feature.

void read_afl_environment(afl_state_t *afl, char **envp) {

int index = 0, issue_detected = 0;
char *env;
while ((env = envp[index++]) != NULL) {

**if (strncmp(env, "ALF_", 4) == 0) {**

  WARNF("Potentially mistyped AFL environment variable: %s", env);
  issue_detected = 1;

**} else if (strncmp(env, "AFL_", 4) == 0) {**
**I have a doubt， what is the meaning of the same condition as if and else branch？**

src/afl-fuzz-state.c
vanhauser-thc: look closely, it is not the same. one is aLf, the other aFl :)

xiaoxiaoafeifei: > look closely, it is not the same
Ok, thanks for your reply

Issue #457, opened 2020-07-17
afl_custom_post_process hook called during initialization only


wizche:
Howdy,

Just noticed that whe using `afl_custom_post_process`, this hook seems to be called only during calibration phase (aka before the status screen).

Meaning that the the .cur_input doesnt go through the hook when in fuzzing mode. Should I see the same content in `*out` and in `.cur_input` right? 

Actually it seems that after a while `afl_custom_post_process` is being called again but still the `.cur_input` doesnt reflect the returned content from the hook. Am I missing something?

I'm testing with 2.66c.

vanhauser-thc:
it is called in write_to_testcase() which is just before any testcase is given to the target. so this is as it should be. 

however I see it is not in trim_case() as there is another write of the testcase (there via write_with_gap()). @rish9101 you know the custom API better - I think post_process would need to be in trim_case() as well, right?

rish9101:
The `trim_case()` also does the normal `write_to_testcase` if there is any `custom_trimming` available, do we want to call the post-process function even if it has been trimmed by the "default" method? It might lead to some changed testcases, meaning we might trim the wrong part of the data due to post-processing.

vanhauser-thc:
Depends on what the postprocessor does. I expect that to be checksum fixes, etc and these would need to be updated to be valid.
For other cases it might not be useful. However a postprocessor’s job is to ensure the data is fine, not to mutate it.

rish9101:
Yes, that is true. But what I was thinking that if the `post-processing` somehow reduces the size of the mutated/trimmed data (there could be possible use cases for this?) then calling it could lead to an `OOB` read.

vanhauser-thc:
that could be - should not but could be. the length needs to be updated after the module is called.
so we need it in the trim function.

@wizche  this discussion does not help you though ;) were you able to find out what the real issue is? because postprocess is called for every single fuzz.

rish9101:
Yes, we do need it. I think we can do it like this, we copy of the testcase with the gaps trimmed, and then call `post-process` on this, and then copy this `trimmed` and `post-process` returned mem into the file/afl_fuzz_memory. That I think removes any possibility of   OOB

vanhauser-thc:
sounds good. can/do you want to do this? Or are you too busy with libafl and life?

rish9101:
[...]

PR #460, opened 2020-07-19
Add post-process functionality in write_with_gap

rish9101:
We had ported post-process to custom_mutators, but write_with_gap function was still missing the important post-process usage. This commit adds that as raised in #457 .

### DIFF
--- a/src/afl-fuzz-run.c
+++ b/src/afl-fuzz-run.c
@@ -144,12 +144,12 @@ static void write_with_gap(afl_state_t *afl, void *mem, u32 len, u32 skip_at,
 
   /*
   This memory is used to carry out the post_processing(if present) after copying
-  the testcase by removing the gaps
+  the testcase by removing the gaps. This can break though
   */
-  u8 mem_trimmed[skip_at + tail_len +
+  u8 mem_trimmed[len - skip_len +
                  1];  // 1 extra size to remove chance of overflow
 
-  ssize_t new_size = skip_at + tail_len;
+  ssize_t new_size = len - skip_len;
   void *  new_mem = mem;
   u8 *    new_buf = NULL;
 
--- a/src/afl-fuzz-run.c
+++ b/src/afl-fuzz-run.c
@@ -142,34 +142,54 @@ static void write_with_gap(afl_state_t *afl, void *mem, u32 len, u32 skip_at,
   s32 fd = afl->fsrv.out_fd;
   u32 tail_len = len - skip_at - skip_len;
 
-  /* We first copy the mem into a new memory region removing the gaps
-    and then carry out any post-processing work on them. Then copy them out to
-    shared-mem or write to file */
-
-  void *mem_trimmed =
-      ck_alloc(skip_at + tail_len +
-               1);  // 1 extra size allocated to remove chance of overflow
-
-  if (skip_at) { memcpy(mem_trimmed, mem, skip_at); }
-
-  if (tail_len) {
-
-    memcpy(mem_trimmed + skip_at, (u8 *)mem + skip_at + skip_len, tail_len);
-
-  }
+  /*
+  This memory is used to carry out the post_processing(if present) after copying
+  the testcase by removing the gaps
+  */
+  u8 mem_trimmed[skip_at + tail_len +
+                 1];  // 1 extra size to remove chance of overflow
 
   ssize_t new_size = skip_at + tail_len;
-  void *  new_mem = mem_trimmed;
+  void *  new_mem = mem;
   u8 *    new_buf = NULL;
 
+  bool post_process_skipped = true;
+
   if (unlikely(afl->custom_mutators_count)) {
+
+    new_mem = mem_trimmed;
 
     LIST_FOREACH(&afl->custom_mutator_list, struct custom_mutator, {
 
       if (el->afl_custom_post_process) {
 
+        // We copy into the mem_trimmed only if we actually have custom mutators
+        // *with* post_processing installed
+
+        if (post_process_skipped) {
+
+          if (skip_at) { memcpy(mem_trimmed, (u8 *)mem, skip_at); }
+
+          if (tail_len) {
+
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
Hey rishi, thanks for the PR ... it looks wrong on so many levels :)
first you seem to always copy the buffer even if there is no custom trimmer. second the copy buffer is not using ck_maybe_alloc (or what that function was called) so it is costly every single trim attempt. lastly - why not update the length after the custom postprocess? and fail if it is a bad value?

rish9101:
Ohh, sorry about it. I'll correct the redundant copying part. Since ck_alloc (or any allocation) may be costly, I think we can use a local buffer for this? And the `new_size` is updated after every `post_process` function in the list, I do need to throw a FATAL though.

vanhauser-thc:
hey @rish9101 no hurry, just want to let you know: once this PR is done we will push again from dev to stable :) this PR should be in there

rish9101:
Hey @vanhauser-thc, sorry for the delay. I've removed the redundant copying and added error handling for `new_size` and `new_buf` now :) . Also, I've used a local buffer for the temporary copy of the trimmed data (if it requires making a copy for post-proessing).

src/afl-fuzz-run.c
```diff
@@ -142,18 +142,82 @@ static void write_with_gap(afl_state_t *afl, void *mem, u32 len, u32 skip_at,
...
   u32 tail_len = len - skip_at - skip_len;
 
+  /*
+  This memory is used to carry out the post_processing(if present) after copying
+  the testcase by removing the gaps
+  */
+  u8 mem_trimmed[skip_at + tail_len +
+                 1];  // 1 extra size to remove chance of overflow
+
+  ssize_t new_size = skip_at + tail_len;
```
vanhauser-thc: can you please switch this to `new_size = len - skip_len`? that makes it easier to follow :) also for `mem_trimmed` above

src/afl-fuzz-run.c
```diff
@@ -142,18 +142,82 @@ static void write_with_gap(afl_state_t *afl, void *mem, u32 len, u32 skip_at,
...
+    LIST_FOREACH(&afl->custom_mutator_list, struct custom_mutator, {
+
+      if (el->afl_custom_post_process) {
+
+        // We copy into the mem_trimmed only if we actually have custom mutators
+        // *with* post_processing installed
+
+        if (post_process_skipped) {
+
+          if (skip_at) { memcpy(mem_trimmed, (u8 *)mem, skip_at); }
```
vanhauser-thc, line 171: why not move this just before the FOREACH loop? if there is at least 1 mutator to run, this has to be done, but it does not need to be checked for every entry in the loop.

src/afl-fuzz-run.c
```diff
@@ -142,18 +142,82 @@ static void write_with_gap(afl_state_t *afl, void *mem, u32 len, u32 skip_at,
   s32 fd = afl->fsrv.out_fd;
   u32 tail_len = len - skip_at - skip_len;
 
+  /*
+  This memory is used to carry out the post_processing(if present) after copying
+  the testcase by removing the gaps
+  */
+  u8 mem_trimmed[skip_at + tail_len +
+                 1];  // 1 extra size to remove chance of overflow
+
```
vanhauser-thc, line 151: this is something you cannot know - trimming (and post processing) of a custom module can increase the size (e.g. for grammar, a simpler grammar can make the file size larger). I dont know what a good safe value is, but maybe add 4096 instead of 1. and please make a comment that this can break.

src/afl-fuzz-run.c
```diff
@@ -142,18 +142,82 @@ static void write_with_gap(afl_state_t *afl, void *mem, u32 len, u32 skip_at,
   s32 fd = afl->fsrv.out_fd;
   u32 tail_len = len - skip_at - skip_len;
 
+  /*
+  This memory is used to carry out the post_processing(if present) after copying
+  the testcase by removing the gaps
+  */
+  u8 mem_trimmed[skip_at + tail_len +
```
domenukk: So far, we tried to avoid the use of variable length arrays, as they are error-prone (mainly, the can be too large for the stack). I think we may want to reuse one of the alrady allocated buffers in `alf_state_t` or allocate a new one (?)

domenukk: Alternatively, the array could be fixed size to the maximum testcase size, that feels even cleaner.

src/afl-fuzz-run.c
rish9101: This has an advantage that we only copy the trimmed data into a new buffer if we actually have some post processing to do though. And that I think should be better than the cost of a bool check maybe?

src/afl-fuzz-run.c
rish9101: We can have extra buffer. But, `write_with_gap` actually is called only when AFL's own trimming function is called, so I don't think they can ever exceed the length given here. Also, `post_process` is supposed to return the processed data in its own buffer. So, the length of this buffer I think should be fine :)

src/afl-fuzz-run.c
rish9101: I didn't know about the allocated buffers in `afl_state_t` actually. 😅

src/afl-fuzz-run.c
vanhauser-thc: I think you are right, leave it like this!

src/afl-fuzz-run.c
vanhauser-thc: again, I think you are correct. nothing to see here :)

src/afl-fuzz-run.c
vanhauser-thc: You can safely do  `u8 *mem_trimmed = ck_maybe_grow(BUF_PARAMS(trim), len - skip_len + 1);`  here with no need to free or do anything else. After that the PR is read to land :)

vanhauser-thc:
sorry this takes too long, doing it myself now ... ;)

PR #463, opened 2020-07-21
Allow the custom mutator to generate larger trimmed data

h1994st:
For some use cases, like grammar mutator, the definition of "size" may have different meanings.

For example, the trimming function in a grammar mutator aims at reducing the objects in a grammar structure, but does not guarantee to generate a smaller binary buffer.

Thus, we allow the custom mutator to generate the trimmed data that is larger than the original data.


domenukk:
Sounds good

PR #465, opened 2020-07-22
Fixes AFLplusplus/AFLplusplus#464 winepath path translation for .cur_…

### LINKED ISSUE #464: afl-wine-trace breaks file input for .cur_input translation
When afl-wine-trace is invoked to translate from local path (e.g. /home/user/out/.cur_input) to winepath (Z:\\\\home\\\\user\\\\out\\\\.cur_input) it does so using the winepath utility. However, it does not chomp/rstrip the newline part of winepath's output, so the target binary for fuzzing eventually ends up with an input file argument that doesn't exist (i.e. "Z:\\\\home\\\\user\\\\out\\\\.cur_input\n").

I have tested using rstrip() on the winepath output and it resolves the issue (and also gives me more paths than 1).

Shall I propose a pull request for this issue?

electricworry:
Fixes #464.

### DIFF
--- a/afl-wine-trace
+++ b/afl-wine-trace
@@ -68,7 +68,12 @@ else:
 argv = sys.argv[1:]
 for i in range(len(argv)):
     if ".cur_input" in argv[i]:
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
@andreafioraldi me dont know or understand python. your code. you decide ;)

andreafioraldi:
The fix is ok, merged, ty @electricworry

PR #468, opened 2020-07-24
Fix bug in qemu mode

Pwnrin:
Fix bug in function afl_wait_tsl at qemu_mode/patches/afl-qemu-cpu-inl.h.
In order to improve the speed of fuzz,  the forkserver uses tb_add_jump to make TB jmp to another directly. But sometimes the TranslationBlock may generated without jmp target.
```
    /* The following data are used to directly call another TB from
     * the code of this one. This can be done either by emitting direct or
     * indirect native jump instructions. These jumps are reset so that the TB
     * just continues its execution. The TB can be linked to another one by
     * setting one of the jump targets (or patching the jump instruction). Only
     * two of such jumps are supported.
     */
    uint16_t jmp_reset_offset[2]; /* offset of original jump target */
#define TB_JMP_RESET_OFFSET_INVALID 0xffff /* indicates no jump generated */
    uintptr_t jmp_target_arg[2];  /* target address or offset */
```
In this case:
jmp_reset_offset[0]=jmp_reset_offset[1]=TB_JMP_RESET_OFFSET_INVALID;
jmp_target_arg[0]= jmp_target_arg[1]=0;
tb_add_jump(last_tb, c.tb_exit, tb)   ->  tb_target_set_jmp_target(tc_ptr, tc_ptr + offset, addr);
( offset = tb->jmp_target_arg[n]=0;    tc_ptr = tb->tc.ptr;)
As a result, the start codes of TranslationBlock (tb->tc.ptr) will change to illegal instructions, which will lead to a terrible crash.

### DIFF
--- a/qemu_mode/patches/afl-qemu-cpu-inl.h
+++ b/qemu_mode/patches/afl-qemu-cpu-inl.h
@@ -620,7 +620,8 @@ static void afl_wait_tsl(CPUState *cpu, int fd) {
 
         last_tb = tb_htable_lookup(cpu, c.last_tb.pc, c.last_tb.cs_base,
                                    c.last_tb.flags, c.cf_mask);
-        if (last_tb) { tb_add_jump(last_tb, c.tb_exit, tb); }
+#define TB_JMP_RESET_OFFSET_INVALID 0xffff
+        if (last_tb && (last_tb->jmp_reset_offset[c.tb_exit] != TB_JMP_RESET_OFFSET_INVALID)) { tb_add_jump(last_tb, c.tb_exit, tb); }
 
       }
 


andreafioraldi:
Hi @Kirin-say thank you for the patch!
We accept PRs only towards the dev branch, so you should change from stable to dev.
Can you also share some info about how to reproduce the bug also? I'm aware of offset invalid but never tought that it could be processed in wait tsl.

vanhauser-thc:
@andreafioraldi  we can change that merge target too, to the far right of the title is an "edit" button. just did that.

Pwnrin:
@andreafioraldi I have a binary to reproduce the bug. But I'm not sure under what circumstances the TranslationBlock will be generated without jmp target.(It is related to QEMU's translation mechanism.)
Crash:
```
gdb-peda$  p *(struct TranslationBlock *) 0x7f0fed23be40
$1 = {
  pc = 0x4000a2a383,
  cs_base = 0x0,
  flags = 0x40c0b3,
  size = 0x2,
  icount = 0x2,
  cflags = 0x0,
  trace_vcpu_dstate = 0x0,
  tc = {
    ptr = 0x7f0fed23bf00 <static_code_gen_buffer+2604496>,
    size = 0x49
  },
  orig_tb = 0x0,
  page_next = {0x7f0fed23ba40, 0x0},
  page_addr = {0x4000a2a000, 0xffffffffffffffff},
  jmp_lock = {
    value = 0x0
  },
  jmp_reset_offset = {0xffff, 0xffff},
  jmp_target_arg = {0x0, 0x0},
  jmp_list_head = 0x0,
  jmp_list_next = {0x7f0fed248600, 0x0},
  jmp_dest = {0x7f0fed248740, 0x0}
}
gdb-peda$ x/10i 0x7f0fed23bf00
   0x7f0fed23bf00 <static_code_gen_buffer+2604496>:     cld
   0x7f0fed23bf01 <static_code_gen_buffer+2604497>:     enter  0x0,0x85
=> 0x7f0fed23bf05 <static_code_gen_buffer+2604501>:     in     eax,dx
   0x7f0fed23bf06 <static_code_gen_buffer+2604502>:     jl     0x7f0fed23bf3d <static_code_gen_buffer+2604557>
   0x7f0fed23bf0c <static_code_gen_buffer+2604508>:     mov    rbp,QWORD PTR [r14+0x28]
   0x7f0fed23bf10 <static_code_gen_buffer+2604512>:     mov    rbx,QWORD PTR [rbp+0x0]
   0x7f0fed23bf14 <static_code_gen_buffer+2604516>:     add    rbp,0x8
   0x7f0fed23bf18 <static_code_gen_buffer+2604520>:     mov    QWORD PTR [r14+0x28],rbx
   0x7f0fed23bf1c <static_code_gen_buffer+2604524>:     mov    QWORD PTR [r14+0x20],rbp
   0x7f0fed23bf20 <static_code_gen_buffer+2604528>:     mov    rbx,QWORD PTR [rbp+0x0]
gdb-peda$ x/10xg  0x7f0fed23bf00
0x7f0fed23bf00 <static_code_gen_buffer+2604496>:        0x8c0fed850000c8fc      0x286e8b4900000031
0x7f0fed23bf10 <static_code_gen_buffer+2604512>:        0x08c58348005d8b48      0x206e8949285e8949
0x7f0fed23bf20 <static_code_gen_buffer+2604528>:        0x08c58348005d8b48      0x809e8949206e8949
0x7f0fed23bf30 <static_code_gen_buffer+2604544>:        0xa5e8fe8b49000000      0x058d48e0ffff7f86
0x7f0fed23bf40 <static_code_gen_buffer+2604560>:        0xd850cfe9fffffeff      0x00290001200000ff
```
Before tb_add_jump:
```
0x7f0fed23bf00: 0x8c0fed85dc6e8b41 => mov    ebp,DWORD PTR [r14-0x24];  test   ebp,ebp; ......
```

andreafioraldi:
The strange fact is that this should crash also the child process but ofc doesn't happen (otherwise would be a qemu bug). I think this is related to code that is different from parent to child, like self-mod code or dlopened code.

andreafioraldi:
Btw I'll merge this workaround, remind to myself to investigate deeper.

Kirin-say:
@andreafioraldi The child process will crash after the  ' fake tb_add_jump' happened. Because that will happen in forkserver, and change the TranslationBlock in parent process. Then the TranslationBlock with illegal instructions will exist in all the child process forked.

PR #476, opened 2020-07-29
New snapshot lkm support

vanhauser-thc:


include/snapshot-inl.h
```diff
@@ -35,25 +34,82 @@
...
+struct afl_snapshot_vmrange_args {
+
+  unsigned long start, end;
+
+};
 
 static int afl_snapshot_dev_fd;
 
-static int afl_snapshot_init(void) {
+static int afl_snapshot_init() {
```
domenukk, line 69: Does this need variable args?

vanhauser-thc: no args to that function

domenukk: Then we shoudn't drop the `void` here :)

vanhauser-thc: why? AFAIK () means no parameter, (void) means ... I dont know :) same would be my guess

PR #477, opened 2020-07-30
Push to stable

vanhauser-thc:
once again ... :)
4 weeks is pretty long this time.


Dockerfile
```diff
@@ -46,17 +46,15 @@ RUN update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-10 0
...
-ARG CXX=g++-10
-ARG LLVM_CONFIG=llvm-config-11
+ENV LLVM_CONFIG=llvm-config-11
+ENV AFL_SKIP_CPUFREQ=1
 
-RUN git clone https://github.com/AFLplusplus/AFLplusplus
-RUN cd AFLplusplus && export REAL_CXX=g++-10 && make distrib && \
-    make install && make clean
+RUN git clone https://github.com/AFLplusplus/AFLplusplus /AFLplusplus
+RUN cd /AFLplusplus && export REAL_CXX=g++-10 && export CC=gcc-10 && \
```
domenukk: Instead of cd-ing around and adding the path to every command, we could use the `WORKDIR` command.
```BASH
WORKDIR /AFLplusplus
RUN export REAL_CX...
```

GNUmakefile
```diff
@@ -96,7 +98,7 @@ ifneq "$(shell uname -m)" "x86_64"
 endif
 
 CFLAGS     ?= -O3 -funroll-loops $(CFLAGS_OPT)
-override CFLAGS += -Wall -g -Wno-pointer-sign -Wmissing-declarations\
+override CFLAGS += -Wall -g -Wno-pointer-sign -Wmissing-declarations -Wno-unused-result \
```
domenukk: How come we need this flag now? Don't we miss potential bugs by ignoring unused results?

custom_mutators/honggfuzz/honggfuzz.c
```diff
@@ -0,0 +1,143 @@
...
+
+*/
+
+/* here we run the honggfuzz mutator, which is really good */
+
+size_t afl_custom_fuzz(my_mutator_t *data, uint8_t *buf, size_t buf_size,
+                       u8 **out_buf, uint8_t *add_buf, size_t add_buf_size,
+                       size_t max_size) {
+
+  /* set everything up, costly ... :( */
```
domenukk, line 116: why can't you work on `buf` directly? Because the size can grow?

examples/afl_untracer/afl-untracer.c
```diff
@@ -738,7 +744,11 @@ int main(int argc, char *argv[]) {
 
 }
 
-static void fuzz() {
+#ifndef _DEBUG
+inline
+#endif
+    static void
+    fuzz() {
```
domenukk: This is probably  missing `void` in its decl

include/snapshot-inl.h
```diff
@@ -35,25 +34,82 @@
...
+struct afl_snapshot_vmrange_args {
+
+  unsigned long start, end;
+
+};
 
 static int afl_snapshot_dev_fd;
 
-static int afl_snapshot_init(void) {
+static int afl_snapshot_init() {
```
domenukk: This is probably  missing `void` in its decl

src/afl-fuzz-init.c
```diff
@@ -365,6 +438,159 @@ static void shuffle_ptrs(afl_state_t *afl, void **ptrs, u32 cnt) {
...
+        }
+
+        continue;
+
+      }
+
+      if (nl_cnt == 0) {
+
+        if (first)
+          WARNF("directory %s is currently empty",
```
domenukk, line 494: should we have brackets for this? Looks super odd right now.

src/afl-fuzz-mutators.c
```diff
@@ -325,6 +349,8 @@ u8 trim_case_custom(afl_state_t *afl, struct queue_entry *q, u8 *in_buf,
 
     } else {
 
+    unsuccessful_trimming:
```
domenukk: a goto into the else branch is.. unexpected? :)

src/afl-fuzz-queue.c
```diff
@@ -100,6 +103,108 @@ void mark_as_redundant(afl_state_t *afl, struct queue_entry *q, u8 state) {
...
+    if (isascii((int)buf[offset]) || isprint((int)buf[offset])) {
+
+      ascii++;
+      // we continue though as it can also be a valid utf8
+
+    }
+
+    // non-overlong 2-byte
+    if (((0xC2 <= buf[offset + 0] && buf[offset + 0] <= 0xDF) &&
+         (0x80 <= buf[offset + 1] && buf[offset + 1] <= 0xBF))) {
```
domenukk: This is a potential oob read if `offset == len -1`

src/afl-fuzz-queue.c
```diff
@@ -100,6 +103,108 @@ void mark_as_redundant(afl_state_t *afl, struct queue_entry *q, u8 state) {
...
+      utf8++;
+      comp--;
+      continue;
+
+    }
+
+    // excluding overlongs
+    if ((buf[offset + 0] == 0xE0 &&
+         (0xA0 <= buf[offset + 1] && buf[offset + 1] <= 0xBF) &&
+         (0x80 <= buf[offset + 2] &&
```
domenukk: More out of bounds here :)

src/afl-fuzz-redqueen.c
```diff
@@ -262,6 +263,58 @@ static u8 its_fuzz(afl_state_t *afl, u8 *buf, u32 len, u8 *status) {
...
 }
 
+static long long strntoll(const char *str, size_t sz, char **end, int base) {
+
+  char        buf[64];
+  long long   ret;
+  const char *beg = str;
+
+  for (; beg && sz && *beg == ' '; beg++, sz--)
+    ;
```
domenukk: empty brackets for nicer formatting ?

src/afl-fuzz-queue.c
```diff
@@ -100,6 +103,108 @@ void mark_as_redundant(afl_state_t *afl, struct queue_entry *q, u8 state) {
...
+      offset += 4;
+      utf8++;
+      comp -= 3;
+      continue;
+
+    }
+
+    offset++;
+
+  }
```
domenukk, line 199: Should we add unit tests for this? And potentially even fuzz-testcases ? :D

src/afl-fuzz-redqueen.c
```diff
@@ -262,6 +263,58 @@ static u8 its_fuzz(afl_state_t *afl, u8 *buf, u32 len, u8 *status) {
 
 }
 
+static long long strntoll(const char *str, size_t sz, char **end, int base) {
```
domenukk, line 266: Another candidate for unittests?

domenukk (approved):

GNUmakefile
domenukk: Builds fine for me on gcc and clang without this option set - no warning?

vanhauser-thc: It was the nice call that needed it. I commented it out again though. I would not mind keeping it. if you prefer to warn (again) remove it.

Dockerfile
vanhauser-thc: yeah maybe. I am not a docker expert. ok that might not need expert knowledge ... I am not a docker guy. the WORKDIR would reset the cwd for every new RUN instance? :)   well it works :) change it if you would prefer it cleaner

custom_mutators/honggfuzz/honggfuzz.c
vanhauser-thc: yes it can grow (a mutator that cannot grow would be a bad mutator ... i guess)

examples/afl_untracer/afl-untracer.c
vanhauser-thc: what is the difference of foo() to foo(void)?

src/afl-fuzz-queue.c
vanhauser-thc: uh right! fixing this and the others

examples/afl_untracer/afl-untracer.c
domenukk: `foo()` can take any amount of parameters, `foo(void)` can take none.

src/afl-fuzz-redqueen.c
vanhauser-thc: yup

src/afl-fuzz-queue.c
vanhauser-thc: ah ... uhm ... if you are bored  feel free :)

examples/afl_untracer/afl-untracer.c
vanhauser-thc: learned something ... oh my ...

Dockerfile
domenukk: `WORKDIR` is like a docker-wide `cd` so it's convenient if you do multiple `RUN` commands that should be in the same folder. Not really worth the effort to change it if it works though, especially as these are only single commands. :)

vanhauser-thc:
we are green! float splitting seems to be fixed! :)

custom_mutators/honggfuzz/honggfuzz.c
```diff
@@ -0,0 +1,143 @@
...
+                                const uint8_t *filename_new_queue,
+                                const uint8_t *filename_orig_queue) {
+
+  if (run.global->mutate.dictionaryCnt >= 1024) return;
+
+  while (data->extras_cnt < data->afl->extras_cnt &&
+         run.global->mutate.dictionaryCnt < 1024) {
+
+    memcpy(run.global->mutate.dictionary[run.global->mutate.dictionaryCnt].val,
+           data->afl->extras[data->extras_cnt].data,
```
domenukk, line 77: In my eyes, custom mutators should not access the (internal) `afl_state_t`. We will need to add a stable API for all use-cases. Else we get a mess - for example mutators need to be recompiled each time we change anything

vanhauser-thc: I can understand your thinking, however custom modules must be able to get that information. either we have a ton of function that export a single parameter each or we go the easy road and people have to recompile.

vanhauser-thc: we can limit the problem if we always add new parameters at the end of the structure, not in the middle.

domenukk: how about adding a public afl struct in the beginning, that is exposed to the header file also, then private parameters?

vanhauser-thc: we could do that yes. sounds like a good idea

docs/binaryonly_fuzzing.md
```diff
@@ -8,12 +8,17 @@
...
   The following is a description of how these binaries can be fuzzed with afl++
 
+
 ## TL;DR:
 
   qemu_mode in persistent mode is the fastest - if the stability is
   high enough. Otherwise try retrowrite, afl-dyninst and if these
   fail too then standard qemu_mode with AFL_ENTRYPOINT to where you need it.
 
+  If your a target is library use examples/afl_frida/.
```
hexcoder-: If your a target is library
->
If your target is a library

examples/afl_frida/libtestinstr.c
```diff
@@ -0,0 +1,35 @@
+/*
+   american fuzzy lop++ - a trivial program to test the build
```
hexcoder-, line 2: Disclaimer is not accurate, I guess. Zalewski and Google had nothing to do with it?

examples/afl_untracer/afl-untracer.c
```diff
@@ -56,6 +56,7 @@
 #include <sys/shm.h>
 #include <sys/wait.h>
 #include <sys/types.h>
+#include <sys/personality.h>
```
hexcoder-, line 59: It is Linux-only, right?

examples/afl_frida/Makefile
```diff
@@ -0,0 +1,23 @@
+ifdef DEBUG
```
hexcoder-: it is GMake specific so,
mv Makefile GNUmakefile
please.

examples/defork/Makefile
```diff
@@ -0,0 +1,64 @@
+#
```
hexcoder-, line 1: mv Makefile GNUmakefile
please

examples/defork/forking_target.c
```diff
@@ -0,0 +1,48 @@
+#include <stdio.h>
```
hexcoder-, line 1: The binary `forking_target` should not be in the repo, IMHO. (Can't comment there, so I did it here)

src/afl-fuzz.c
```diff
@@ -129,18 +131,23 @@ static void usage(afl_state_t *afl, u8 *argv0, int more_help) {
...
       "found\n"
       //"  -B bitmap.txt - mutate a specific test case, use the out/fuzz_bitmap
       //" "file\n"
       "  -C            - crash exploration mode (the peruvian rabbit thing)\n"
+      "  -b cpu_id     - bind the fuzzing process to the specified CPU core "
+      "(0-...)\n"
       "  -e ext        - file extension for the fuzz test input file (if "
       "needed)\n\n",
-      argv0, EXEC_TIMEOUT, MEM_LIMIT);
+      argv0, EXEC_TIMEOUT, MEM_LIMIT, FOREIGN_SYNCS_MAX);
```
hexcoder-, line 150: Please add new env. vars `AFL_CYCLE_SCHEDULES` and `AFL_EXPAND_HAVOC_NOW` to -hh output.

hexcoder- (approved):
Impressive work!

vanhauser-thc:
@hexcoder- you are really strong finding the spelling and grammar errors - could you please fix them instead of reporting? same time on your side, however no time required for anyone else :)
EDIT: I mean for future reviews.
Also - can you have an extra look on the new big section in our main README.md?

examples/afl_frida/libtestinstr.c
vanhauser-thc: it is test-instr with the change that it is test(buf,len) instead of main(...) so I think it is fair

examples/afl_untracer/afl-untracer.c
vanhauser-thc: it is using ptrace so yes it is linux only

examples/defork/Makefile
vanhauser-thc: done!

examples/defork/forking_target.c
vanhauser-thc: I agree, I guess dominik accidentally checked it in. removing it

src/afl-fuzz.c
vanhauser-thc: done

docs/binaryonly_fuzzing.md
vanhauser-thc: fixed

examples/afl_frida/Makefile
vanhauser-thc: done

vanhauser-thc:
@hexcoder- float splitting still fails occasionally. can you make the testcase at bit easier to solve?

src/afl-fuzz-init.c
```diff
@@ -53,25 +50,60 @@ void bind_to_free_cpu(afl_state_t *afl) {
 
   }
 
+  u8  cpu_used[4096] = {0}, lockfile[PATH_MAX] = "";
+  u32 i;
+
+  if (afl->cpu_to_bind != -1) {
+
+    i = afl->cpu_to_bind;
```
domenukk: The goto spaghetti code will go back to `try` if the current cpu cannot be bound and continue with the next one or crash or something.
set_cpu should instead be a function, returning a flag if binding worked, the normal code may then loop, `-b` should FATAL if the core cannot be bound.

hexcoder-:
> @hexcoder- float splitting still fails occasionally. can you make the testcase at bit easier to solve?

Yes, will do.

hexcoder-:
> @hexcoder- you are really strong finding the spelling and grammar errors - could you please fix them instead of reporting? same time on your side, however no time required for anyone else :)
> EDIT: I mean for future reviews.
> Also - can you have an extra look on the new big section in our main README.md?

ok with me.

src/afl-fuzz-init.c
domenukk: Also metaquestion: Why not make this an env var?

vanhauser-thc: because this is a PR to google afl and we would loose compatibility if it is not the same command line option.

the goto was there before and I used it to make the code simpler. if the bind fails but cpu_to_bind is != -1 then there is the FATAL. I introduced the feature with as little lines of code changed as possible.

yes set_cpu() would be nicer, but I hadn't had the time.

vanhauser-thc: also the code is expected to try the next cpu if a bind fails (except for -b, then it FATALs)

andreafioraldi (approved):

custom_mutators/honggfuzz/honggfuzz.c
domenukk: Should I try to do that today before the pr gets merged?

vanhauser-thc: how about we merge and then you do that? it would need to be tested and could be a source of failure? maybe?

vanhauser-thc:
lets merge this for now.
known issues: redqueen + lto issue, lto autodict not picking up some strings, afl state access to custom modules

PR #481, opened 2020-07-31
Haiku set explicitly to performance.

devnexen:
No command line to set through afl-system-config (the only one is a GUI).


vanhauser-thc:
thanks :)

PR #490, opened 2020-08-04
makefile fix?

vanhauser-thc:


vanhauser-thc:
some gnu make didnt compile llvm_mode.

PR #492, opened 2020-08-05
Fix support for mips in qemu mode

dozernz:
I had trouble running mips (more specifically mipsbe) binaries in qemu mode using AFLplusplus. 

This patch taken from https://www.mail-archive.com/qemu-devel@nongnu.org/msg611477.html fixes the issue and allows for emulating mips binaries.


andreafioraldi:
Thank you!

andreafioraldi (approved):

PR #493, opened 2020-08-05
Merge mis-targeted PR to dev

domenukk:
Just looked through the branches and noticed a PR got accepted to master. This was very likely not intended and we should merge this to dev, then to stable.


vanhauser-thc:
@domenukk  If you check the conflict you will see that I added this misplaced PR last week by hand :)
good idea that you look too!

PR #497, opened 2020-08-08
Add support for specific custom mutator name

murx-:
Currently the stage_short is only set to "custom" for each custom mutator. This results in queue files where it is unclear which custom mutator was responsible for producing the file.

Now
```
id:001452,src:001329,time:23965354,op:custom,pos:0
```
will become
```
id:000158,src:000002,time:1717,op:honggfuzz.so,pos:0
```


vanhauser-thc:
I like this, thank you!

PR #499, opened 2020-08-12
important push to stable

vanhauser-thc:
this is not an urgent PR but one I do not want to get cluttered with lots of other commits, so this gets reviewed properly.

besides the llvm change that implements ctor + ifunc instrumentation skipping for all llvm there is this:

1. when fuzzing a binary in qemu that was instrumented, the map is cluttered with the target picking up the map information and also writing in there. So I added an internal env AFL_DISABLE_LLVM_INSTRUMENTATION that is set when a fork server is in qemu_mode and added functionality to afl-llvm-rt.o to skip the forkserver and getting the shmem if present

2. I would like to get fixed map support in LTO to work ... like everytime. for that is not only the ctor/ifunc skipping, but also changing how __afl_manual_init() works.
I have split it up into two functions. one _early that grabs the shmem as early as possible (prio 3), and the forkserver which is still in __afl_manual_init put has moved its prio from 5 to 65535 to start as late as possible.
Still fixed mem can fail if an other constructor with prios 1-2 calls functions that are instrumented and gets run before _early. but otherwise this should be failsafe ....

(fixed memory is that the instrumentation writes at a hardcoded address, so if AFL_LLVM_SKIP_NEVERZERO is set we have a single instruction instrumentation: `add 0x1000c, 0x01`)


llvm_mode/afl-llvm-common.cc
```diff
@@ -558,27 +629,25 @@ bool isInInstrumentList(llvm::Function *F) {
...
 
-    return true;
+    }
 
   }
 
-  return true;  // not reached
+  return return_default;
 
 }
```
hexcoder-, line 557: Some code duplication in `initInstrumentList()` for `allowlist` and `denylist` (not part of this change, I know).

Fixed some typos/wording directly.

docs/Changelog.md
```diff
@@ -29,9 +29,9 @@ sending a mail to <afl-users+subscribe@googlegroups.com>.
      - added neverzero counting to trace-pc/pcgard
      - fixes for laf-intel float splitting (thanks to mark-griffin for
        reporting)
+     - skipping ctors and ifuncs for instrumentation
```
hexcoder-, line 35: for the non-insiders (ctors -> constructors and ifuncs -> LLVM function aliases?)

src/afl-forkserver.c
```diff
@@ -58,6 +58,8 @@ static list_t fsrv_list = {.element_prealloc_count = 0};
 
 static void fsrv_exec_child(afl_forkserver_t *fsrv, char **argv) {
 
+  if (fsrv->qemu_mode) setenv("AFL_DISABLE_LLVM_INSTRUMENTATION", "1", 0);
```
hexcoder-, line 61: Please add new env var. to usage()  of afl-clang-fast and the env var. "spell checker".

hexcoder- (approved):

docs/Changelog.md
vanhauser-thc: people who dont know ctos wont now constructors either, or?

ifuncs are not function aliases, that is something different.

ifunc is a resolver functionality and program start which allows you to map a function foo to either foo1 or foo2 depending on ... what you decide in your resolver function. so this is code that is run before even constructor(0). just learned about this last week :)

llvm_mode/afl-llvm-common.cc
vanhauser-thc: your are right. refactored it. much better now.

src/afl-forkserver.c
vanhauser-thc: this is an internal env var, and not for users. so it belongs in neither, same as AFL_EINS_ZWEI_POLIZEI

vanhauser-thc:
@hexcoder- 
can you please check this (so checkout dev) with the oldest clang version that we support?

please set LLVM_CONFIG, and in ./llvm_mode/:
```
make clean
make
AFL_PATH=`pwd/..` AFL_LLVM_INSTRUMENT=PCGUARD ../afl-clang-fast -o t ../test-instr.c
echo 0 | ../afl-showmap -o - ./t
```
the important thing is this line:
```
[*] Target map size: 16
```
if you see 65536 or a crash, then I need to create a special test branch for you.

vanhauser-thc:
as sancov calls _pcguard_init at ctor prio 2 we can earliest do the shmem at 3 ... damn. that is for llvm 6.0 - 12. I have to make sure that older llvm do not run later than 2, hence my question to heiko

hexcoder-:
On Ubuntu trusty there is LLVM/clang 3.4:
```
...
clang++ -O3 -funroll-loops -fPIC -D_FORTIFY_SOURCE=2 -Wall -g -Wno-pointer-sign -I ../include/ -DAFL_PATH=\"/usr/local/lib/afl\" -DBIN_PATH=\"/usr/local/bin\" -DLLVM_BINDIR=\"/usr/lib/llvm-3.4/bin\" -DVERSION=\"++2.66d\" -DLLVM_LIBDIR=\"/usr/lib/llvm-3.4/lib\" -DLLVM_VERSION=\"3.4\" -DAFL_CLANG_FLTO=\"\" -DAFL_REAL_LD=\"\" -DAFL_CLANG_FUSELD=\"\" -DCLANG_BIN=\"clang\" -DCLANGPP_BIN=\"clang++\" -DUSE_BINDIR=0 -Wno-unused-function `/usr/bin/llvm-config-3.4 --cxxflags` -fno-rtti -fPIC -std=gnu++11 -c afl-llvm-common.cc -o afl-llvm-common.o 
afl-llvm-common.cc:263:8: error: unknown type name 'GlobalIFunc'
  for (GlobalIFunc &IF : M->ifuncs()) {
       ^
afl-llvm-common.cc:263:29: error: no member named 'ifuncs' in 'llvm::Module'
  for (GlobalIFunc &IF : M->ifuncs()) {
                         ~  ^
2 errors generated.
make: *** [afl-llvm-common.o] Error 1

```
and on Ubuntu xenial there is LLVM/clang 3.8.0. Same compilation error here.

vanhauser-thc:
all compiles no, even found a bug for llvm 4 support. the constructor level for the _easy function works.

@hexcoder- @domenukk @andreafioraldi 
I want to push to stable and even make a release. the LTO laf/redqueen thing is an issue that needs to be fixed.

PR #505, opened 2020-08-14
Reworked maybe_grow to take a single ptr

domenukk:
This PR changes maybe_grow to take a single pointer and adds a function to free it also.
I'd be glad to hear y'alls opinions.

### DIFF
--- a/afl-cmin
+++ b/afl-cmin
@@ -1,5 +1,6 @@
 #!/usr/bin/env sh
 export AFL_QUIET=1
+export ASAN_OPTIONS=detect_leaks=0
 THISPATH=`dirname ${0}`
 export PATH="${THISPATH}:$PATH"
 awk -f - -- ${@+"$@"} <<'EOF'
--- a/afl-cmin
+++ b/afl-cmin
@@ -1,6 +1,5 @@
 #!/usr/bin/env sh
 export AFL_QUIET=1
-export ASAN_OPTIONS=detect_leaks=0
 THISPATH=`dirname ${0}`
 export PATH="${THISPATH}:$PATH"
 awk -f - -- ${@+"$@"} <<'EOF'
--- a/examples/afl_network_proxy/afl-network-server.c
+++ b/examples/afl_network_proxy/afl-network-server.c
@@ -75,7 +75,6 @@ static u8 *buf2;
 
 static s32    in_len;
 static u32    map_size = MAP_SIZE;
-static size_t buf2_len;
 
 static volatile u8 stop_soon;          /* Ctrl-C pressed?                   */
 
@@ -272,7 +271,7 @@ static void set_up_environment(afl_forkserver_t *fsrv) {
 
       setenv("QEMU_SET_ENV", buf, 1);
 
-      ck_free(buf);
+      maybe_grow_free(buf);
 
     } else {
 
@@ -343,7 +342,7 @@ static void usage(u8 *argv0) {
 
 }
 
-int recv_testcase(int s, void **buf, size_t *max_len) {
+int recv_testcase(int s, void **buf) {
 
   u32    size;
   s32    ret;
@@ -358,7 +357,8 @@ int recv_testcase(int s, void **buf, size_t *max_len) {
 
   if ((size & 0xff000000) != 0xff000000) {
 
-    *buf = ck_maybe_grow(buf, max_len, size);
+    *buf = maybe_grow(buf, size);
+    if (unlikely(!buf)) {PFATAL("Alloc");}
     received = 0;
     // fprintf(stderr, "unCOMPRESS (%u)\n", size);
     while (received < size &&
@@ -370,7 +370,8 @@ int recv_testcase(int s, void **buf, size_t *max_len) {
 #ifdef USE_DEFLATE
     u32 clen;
     size -= 0xff000000;
-    *buf = ck_maybe_grow(buf, max_len, size);
+    *buf = maybe_grow(buf, size);
+    if (unlikely(!buf)) {PFATAL("Alloc");}
     received = 0;
     while (received < 4 &&
            (ret = recv(s, &clen + received, 4 - received, 0)) > 0)
@@ -379,14 +380,15 @@ int recv_testcase(int s, void **buf, size_t *max_len) {
     // fprintf(stderr, "received clen information of %d\n", clen);
     if (clen < 1)
       FATAL("did not receive valid compressed len information: %u", clen);
-    buf2 = ck_maybe_grow((void **)&buf2, &buf2_len, clen);
+    buf2 = maybe_grow((void **)&buf2, clen);
+    if (unlikely(!buf2)) {PFATAL("Alloc");}
     received = 0;
     while (received < clen &&
            (ret = recv(s, buf2 + received, clen - received, 0)) > 0)
       received += ret;
     if (received != clen) FATAL("did not receive compressed information");
     if (libdeflate_deflate_decompress(decompressor, buf2, clen, (char *)*buf,
-                                      *max_len,
+                                      size,
                                       &received) != LIBDEFLATE_SUCCESS)
       FATAL("decompression failed");
       // fprintf(stderr, "DECOMPRESS (%u->%u):\n", clen, received);
@@ -413,7 +415,6 @@ int recv_testcase(int s, void **buf, size_t *max_len) {
 int main(int argc, char **argv_orig, char **envp) {
 
   s32    opt, s, sock, on = 1, port = -1;
-  size_t max_len = 0;
   u8     mem_limit_given = 0, timeout_given = 0, unicorn_mode = 0, use_wine = 0;
   char **use_argv;
   struct sockaddr_in6 serveraddr, clientaddr;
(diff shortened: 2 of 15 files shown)


afl-cmin
```diff
@@ -1,6 +1,5 @@
 #!/usr/bin/env sh
 export AFL_QUIET=1
-export ASAN_OPTIONS=detect_leaks=0
```
vanhauser-thc: why did you remove this? otherwise targets compiled with asan produce errors.
that an asan compiled showmap does not have leaks we can test with alf-showmap directly

domenukk: How do you mean? We don't care about targets here as this is the aflpp makefile?

examples/afl_network_proxy/afl-network-server.c
```diff
@@ -379,14 +380,15 @@ int recv_testcase(int s, void **buf, size_t *max_len) {
...
+    buf2 = maybe_grow((void **)&buf2, clen);
+    if (unlikely(!buf2)) { PFATAL("Alloc"); }
     received = 0;
     while (received < clen &&
            (ret = recv(s, buf2 + received, clen - received, 0)) > 0)
       received += ret;
     if (received != clen) FATAL("did not receive compressed information");
     if (libdeflate_deflate_decompress(decompressor, buf2, clen, (char *)*buf,
-                                      *max_len,
+                                      maybe_grow_bufsize(buf2),
```
vanhauser-thc: this is a different parameter. either this is now wrong, or I had a bug in it. too tired atm to analyse this :(

domenukk: Yes I wondered why you were using the internal size here.

include/alloc-inl.h
```diff
@@ -673,78 +619,96 @@ static inline size_t next_pow2(size_t in) {
...
 
-/* This function makes sure *size is > size_needed after call.
- It will realloc *buf otherwise.
- *size will grow exponentially as per:
- https://blog.mozilla.org/nnethercote/2014/11/04/please-grow-your-buffers-exponentially/
- Will return NULL and free *buf if size_needed is <1 or realloc failed.
- @return For convenience, this function returns *buf.
- */
-static inline void *maybe_grow(void **buf, size_t *size, size_t size_needed) {
+struct maybe_grow_buf {
```
vanhauser-thc: this is now a function with just two variables and nothing else?

include/alloc-inl.h
```diff
@@ -673,78 +619,96 @@ static inline size_t next_pow2(size_t in) {
...
+};
 
-  /* grow exponentially */
-  size_t next_size = next_pow2(size_needed);
+#define SIZE_OFFSET offsetof(struct maybe_grow_buf, buf)
 
-  /* handle overflow and zero size_needed */
-  if (!next_size) { next_size = size_needed; }
+/* Returs the container element to this ptr */
+static inline struct maybe_grow_buf *maybe_grow_buf(void *buf) {
```
vanhauser-thc: this function returns the buffer pointer right? the name sounds rather like you want to grow a buffer.
recommendation: rename to maybe_grow_get_bufptr and the one below maybe_grow_get_bufsize

include/alloc-inl.h
domenukk: Aah no this is just the internal struct used by the actual function to calculate offsets in a less.obscure way :)

include/alloc-inl.h
```diff
@@ -673,78 +619,96 @@ static inline size_t next_pow2(size_t in) {
...
  It will realloc *buf otherwise.
  *size will grow exponentially as per:
  https://blog.mozilla.org/nnethercote/2014/11/04/please-grow-your-buffers-exponentially/
- Will FATAL if size_needed is <1.
+ Will return NULL and free *buf if size_needed is <1 or realloc failed.
  @return For convenience, this function returns *buf.
  */
-static inline void *ck_maybe_grow(void **buf, size_t *size,
-                                  size_t size_needed) {
+static inline void *maybe_grow(void **buf, size_t size_needed) {
```
vanhauser-thc: that this is the function that actually increases a buffer is not clear in contrast to the other names.
maybe rename this to maybe_grow_need or _enlarge, or ...

include/alloc-inl.h
```diff
@@ -673,78 +619,96 @@ static inline size_t next_pow2(size_t in) {
...
 
   /* No need to realloc */
-  if (likely(*size >= size_needed)) { return *buf; }
+  if (likely(size >= size_needed)) { return *buf; }
 
   /* No initial size was set */
   if (size_needed < INITIAL_GROWTH_SIZE) { size_needed = INITIAL_GROWTH_SIZE; }
 
   /* grow exponentially */
   size_t next_size = next_pow2(size_needed);
```
vanhauser-thc: you coult put this together:
if  (size_needed < INITIAL_GROWTH_SIZE) { size_needed = INITIAL_GROWTH_SIZE; next_size = INITIAL_GROWTH_SIZE << 1; }
else next_size = next_pow2(size_needed);
this way if <= this is precomputed

include/alloc-inl.h
domenukk: Thats what `grow` was supposed to say 😃
Probably we want a better naming scheme overall?

include/alloc-inl.h
```diff
@@ -673,78 +619,96 @@ static inline size_t next_pow2(size_t in) {
...
 
-  /* handle overflow */
+  /* handle overflow and zero size_needed */
   if (!next_size) { next_size = size_needed; }
```
vanhauser-thc: would this be more secure as `if (next_size < size_needed)` ?

include/alloc-inl.h
```diff
@@ -30,12 +30,13 @@
...
 #include <string.h>
+#include <stddef.h>
 
 #include "config.h"
 #include "types.h"
 #include "debug.h"
 
-/* Initial size used for ck_maybe_grow */
+/* Initial size used for maybe_grow */
 #define INITIAL_GROWTH_SIZE (64)
```
vanhauser-thc, line 41: the ->size value in the structure could either be the full size of the allocated memory, or the size available to the user.
currently it is the previous.

if ->size is just internal it does not matter and whatever results in the fewest computations required should be it.

if this is given to the user it should be the latter as otherwise a user has to know do deduct a pointer size from the value and forgetting that would lead to overruns.

src/afl-fuzz-one.c
```diff
@@ -364,7 +364,7 @@ static void locate_diffs(u8 *ptr1, u8 *ptr2, u32 len, s32 *first, s32 *last) {
 
 #endif                                                     /* !IGNORE_FINDS */
 
-#define BUF_PARAMS(name) (void **)&afl->name##_buf, &afl->name##_size
+#define BUF_PARAMS(name) (void **)&afl->name##_buf
```
vanhauser-thc: if we would call this AFL_BUF_PARAM we could put it in types.h and never would need to define + undefine it.

vanhauser-thc: and the S at the end can be removed anyway as it not only the buf param, and not also the size

src/afl-fuzz-one.c
```diff
@@ -4339,7 +4351,8 @@ static u8 mopt_common_fuzzing(afl_state_t *afl, MOpt_globals_t MOpt_globals) {
 
               if (temp_len + extra_len >= MAX_FILE) break;
 
-              out_buf = ck_maybe_grow(BUF_PARAMS(out), temp_len + extra_len);
+              out_buf = maybe_grow(BUF_PARAMS(out), temp_len + extra_len);
+              if (unlikely(!out_buf)) { PFATAL("alloc"); }
```
vanhauser-thc, line 4354: if we move the PFATAL into the .h it cannot be forgotten and less code lines overall

vanhauser-thc:
done :)

src/afl-fuzz-one.c
domenukk: For multithreaded fuzzing we won't be able to use PFATAL anymore, so I suggest slowly reworking the error handling

afl-cmin
vanhauser-thc: we dont care about leaks in targets is what I means :) and with your removal you enable printing+erroring on them.
its in the afl-cmin script

include/alloc-inl.h
vanhauser-thc: could also be a different naming scheme. having grow in every function name might be misleading :)

src/afl-fuzz-one.c
vanhauser-thc: ah good point, yes then not moving FATAL.
however we have a disadvantage. we only need a check if a realloc was done. however the caller doesnt know if one was done or not and therefore always has to check. but whelp I guess that is the disadvanted of threading

domenukk: As maybe_grow is inlined the compiler can figure that out, but I can double check

afl-cmin
domenukk: ooh oops that was a fail :)

include/alloc-inl.h
domenukk: This is impossible given the current next_pow2 output. I added a unittests for that just in case.

domenukk:
Renamed `maybe_grow` to `afl_realloc`

PR #510, opened 2020-08-18
qemu/unicorn build fix for MUSL based linux distros.

devnexen:


vanhauser-thc:
what is the issue with musl?
could it be fixed? on qemu side or musl side?

devnexen:
it might be a temporary fix but for both unicorn and qemu, (thus more in qemu side to fix) it just does not build with linux and musl (missing headers, missing constants etc)

devnexen:
Any toughts @vanhauser-thc ? :-)

domenukk:
There seems to be qemu for alpine which is a musl-based distro:
https://alpine.pkgs.org/edge/alpine-community-x86_64/qemu-img-5.1.0-r0.apk.html
However it does list glib in its dependencies, would that be an additional glibc to build qemu? I have 0 clue about alpine :D
If they cannot build it in musl either I would say this PR is fine.

devnexen:
I ve made this PR as I tested in Alpine in fact but I m not sure it s glibc but glib isn't it ? And since it is the future 5.1.0 version maybe they have fixed things for musl since this old 3.1.1 ? Just guess tough

vanhauser-thc:
how about instead of error -> warn and let it run through? if the compilation fails it fails, but so a newer musl variant, or a parallel installed glibc could still work out.

devnexen:
In fact I knowledgeably did this PR as temporary fix until newer qemu version is used in here, rather than waiting musl to "catch up" with glibc feature-wise. But as you prefer folks no pb.

domenukk:
So once @andreafioraldi is done with his qemu5 port, this part of the problem will be gone. For unicorn I'd just say no here like the pr says.

andreafioraldi:
on qemu 5 i'm working on snapshot, I will update it when all the features are ready, ETA no more than few weeks, so IMO we don't need to fix this particular use case now for qemu 3.

PR #519, opened 2020-08-21
Added "-g" option for maximum number of user-specified tokens

antonio-morales:
Sometimes I've needed to make use of bigger dictionaries (with more than 200 items). Until now, the only option was to modify MAX_DET_EXTRAS in config.h.

I think it would be a good idea to be able to modify this option at runtime.

### DIFF
--- a/src/afl-fuzz-extras.c
+++ b/src/afl-fuzz-extras.c
@@ -248,10 +248,10 @@ static void extras_check_and_sort(afl_state_t *afl, u32 min_len, u32 max_len,
 
   }
 
-  if (afl->extras_cnt > MAX_DET_EXTRAS) {
+  if (afl->extras_cnt > afl->max_dict_entries) {
 
     WARNF("More than %d tokens - will use them probabilistically.",
-          MAX_DET_EXTRAS);
+    		afl->max_dict_entries);
 
   }
 
@@ -403,10 +403,10 @@ void add_extra(afl_state_t *afl, u8 *mem, u32 len) {
 
   /* We only want to print this once */
 
-  if (afl->extras_cnt == MAX_DET_EXTRAS + 1) {
+  if (afl->extras_cnt == afl->max_dict_entries + 1) {
 
     WARNF("More than %d tokens - will use them probabilistically.",
-          MAX_DET_EXTRAS);
+    		afl->max_dict_entries);
 
   }
 
--- a/src/afl-fuzz-one.c
+++ b/src/afl-fuzz-one.c
@@ -1509,13 +1509,13 @@ skip_interest:
 
     for (j = 0; j < afl->extras_cnt; ++j) {
 
-      /* Skip extras probabilistically if afl->extras_cnt > MAX_DET_EXTRAS. Also
+      /* Skip extras probabilistically if afl->extras_cnt > afl->max_dict_entries. Also
          skip them if there's no room to insert the payload, if the token
          is redundant, or if its entire span has no bytes set in the effector
          map. */
 
-      if ((afl->extras_cnt > MAX_DET_EXTRAS &&
-           rand_below(afl, afl->extras_cnt) >= MAX_DET_EXTRAS) ||
+      if ((afl->extras_cnt > afl->max_dict_entries &&
+           rand_below(afl, afl->extras_cnt) >= afl->max_dict_entries) ||
           afl->extras[j].len > len - i ||
           !memcmp(afl->extras[j].data, out_buf + i, afl->extras[j].len) ||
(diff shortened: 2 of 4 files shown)


domenukk:
Thanks for the pr.
Two things:
a) did you benchmark it? I would guess probabilistic use of tokes makes sense for a large number of elements
b) maybe make this an env variable instead of a commandline switch? Sth like `AFL_MAX_DET_EXRAS`

vanhauser-thc:
I thought about this and I propose something different. the extras dictionary grows to as many entries as which are to be added, no limit. IMHO it doesnt make sense to have a limit.
remember that there is not only -x file, but also -x directory (adding all files in that directory), plus the LTO autodictionary feature.
and checking before how many entries that would be and then change config.h or using -g is cumbersome.
if there are entries that the user wants to have added - they should be added.

domenukk:
As far as I understand (@antonio-morales correct me if I'm wrong), you can add an infinite amount of dict entries, but if you pass this threshold of currently 200, they are sometimes skipped, probabilistically
The important lines are:
```c
      /* Skip extras probabilistically if afl->extras_cnt > MAX_DET_EXTRAS. Also
         skip them if there's no room to insert the payload, if the token
         is redundant, or if its entire span has no bytes set in the effector
         map. */

      if ((afl->extras_cnt > MAX_DET_EXTRAS &&
           rand_below(afl, afl->extras_cnt) >= MAX_DET_EXTRAS) ||
...
```
This seems like an optimization so other, non-dict things also get a chance.
Making the threshold configurable via env makes sense, but you may actually not change it too much, even if you have more tokens

antonio-morales:
@domenukk @vanhauser-thc Yes, I have used bigger dictionaries a few occasions in the past (modifying MAX_DET_EXTRAS constant and rebuilding the code). I haven't' benchmarked it but, if I'm not mistaken, it's a linear increase in time or O(n). 
For example, if we have a file of size 512 bytes:
- 200 dictionary entries => 512 x 200 entries x 2 (over & insert) = 204800 iterations aprox
- 400 dictionary entries => 512 x 400 entries x 2 (over & insert) = 409,600 iterations approx

I think that dictionaries with more than 200 entries are useful, for example, if you're fuzzing complex interpreters/compilers.

The variable AFL_MAX_DET_EXRAS looks like a good idea to me. And it allows you to save "-g" flag for a future feature 😄 

And I've kept the probabilistic functionality because I think that it can be interesting in certain cases or scenarios.
There are 2 examples:
- You have created a dictionary with 400 entries, and you want to use all of them. So you can set AFL_MAX_DET_EXRAS=500, and AFL will use all the tokens.
- You use an automatic tool for creating dictionaries, and this tool has generated a dictionary with 5000 entries. However, you think not worth using all of these entries in all the iterations. So you set AFL_MAX_DET_EXRAS=1000, and AFL will use the entries in a probabilistic way.

domenukk:
Can you take a look if 1301552 is a good solution to your problem?

vanhauser-thc:
@domenukk this means the user has to make an extra effort in first needing to check how many dictionary entries there are to be loaded. With lto autodict this is additionally cumbersome. We should simply load all supplied. Expect the user to know what he/she is doing.

domenukk:
Ah you mean like a `AFL_EXTRAS_DET_ALWAYS`?

domenukk:
Just as a clarification, the important loop here:
https://github.com/AFLplusplus/AFLplusplus/blob/572944d7267e3612ef8da17a180bc3d8f1a958a7/src/afl-fuzz-one.c#L1510
doesn't just skip random entries *always*, but skips random entries for each `fuzz_one`. They will get picked up eventually.
This makes a lot of sense to me - if I have more than 200 dict entries, I don't want each `fuzz_one` to pick all of them up for each mutation, I want afl to pick random ones.
I would argue using this env variable should only be used very occasionally - by people who know what they are doing.
Most of the time this will deteriorate fuzzing.

vanhauser-thc:
no I mean currently there is a limit on entries that afl-fuzz reads in - and aborts otherwise.
try   afl-fuzz -x dictionaries/ -i in -o out ./test-instr

domenukk:
Ah @vanhauser-thc gotcha. How about #523 ?

vanhauser-thc:
If #523 is merged is this issue/feature done or is there anything still open?

antonio-morales:
@domenukk @vanhauser-thc I think https://github.com/AFLplusplus/AFLplusplus/commit/1301552101af899557a93a7535d8a57874fe6edf and #523 work well, so this issue can be closed now 👍

PR #523, opened 2020-08-23
Allow Large Extras

domenukk:
This PR allows large extras, making fuzzing with huge dictionaries possible (for people that want to)...

### DIFF
--- a/src/afl-fuzz-extras.c
+++ b/src/afl-fuzz-extras.c
@@ -319,8 +319,8 @@ void load_extras(afl_state_t *afl, u8 *dir) {
 
     if (st.st_size > MAX_DICT_FILE) {
 
-      FATAL(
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
Is that all that is needed? Was the extras duct already growing as big as required besides the fatal?

vanhauser-thc:
(Not at Home or Laptop to check myself)

domenukk:
Yes it's dynamically allocated. 
However if extras get larger than the max testcase size we'll run into further problems, but that's not a large issue

vanhauser-thc:
@domenukk max testcase size being MAX_FILE = 1MB by default? good to know but I do not think this is likely to happen.
all dictionaries we have in afl++ are ~100kb altogether :)

LGTM - IMHO you can merge

antonio-morales:
@vanhauser-thc 1MB is more than enough xddd

PR #524, opened 2020-08-24
Added out_file value when using stdio

wizche:
As discussed with @vanhauser-thc it makes more sense if the out_file is always populated.
This way if something goes wrong (ck_write/ck_read aborts) we get the `.cur_input` instead of the (null).
I changed the logic in the `afl_fsrv_write_to_testcase` but I am not sure if this breaks something.


domenukk:
Looks good, thanks! :)

PR #526, opened 2020-08-27
Prevent afl-fuzz from modifying stage_max during fuzzing

h1994st:
If `afl_custom_fuzz_count` is provided by the custom mutator, `afl-fuzz` should not modify `stage_max` anymore during fuzzing


vanhauser-thc:
makes sense! :) hadnt though about that

PR #528, opened 2020-08-28
Fix the /usr/bin/afl-clang and afl-clang++ symlinks

rhertzog:
Right now they are created pointing to '../afl-clang-fast' instead of 'afl-clang-fast'. Given that all the binaries are in the same directory, the symlinks are effectively broken.

This was fixed following this Debian report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=968263


vanhauser-thc:
LGTM, thanks!

PR #529, opened 2020-08-28
Manual pages related changes

rhertzog:
In the Debian packaging, I wanted to install your generated manual pages but I had to fix the installation path which is not in the expected /usr/share/man/ and lintian generated me lots of bad-whatis-entry warnings (https://lintian.debian.org/tags/bad-whatis-entry.html) so I actually improved your build system to include the one-line summary in the NAME section.

I also changed the list of manual pages to also cover afl-g++ (even though it's just a symlink, it's best that it has its own manual page too).


vanhauser-thc:
seems to be fine, thank you!

PR #530, opened 2020-08-28
Pass CPPFLAGS to all calls of the C compiler

rhertzog:
This variable is a standard way to inject options for the C
preprocessor. It's respected by the implicit rules of make
and autoconf/automake.

Debian sets this variable during package build to inject
`-D_FORTIFY_SOURCE=2` and we would like afl++ to respect it.

Note that this commit also adds $(CFLAGS) in the build of
afl-performance.o where it was missing. It might have been
on purpose but we want to keep CFLAGS everywhere as well
since Debian injects various options through that variable
(for hardening and reproducibility).

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -368,78 +368,78 @@ ready:
 	@echo "[+] Everything seems to be working, ready to compile."
 
 afl-gcc: src/afl-gcc.c $(COMM_HDR) | test_x86
-	$(CC) $(CFLAGS) src/$@.c -o $@ $(LDFLAGS)
+	$(CC) $(CFLAGS) $(CPPFLAGS) src/$@.c -o $@ $(LDFLAGS)
 	set -e; for i in afl-g++ afl-clang afl-clang++; do ln -sf afl-gcc $$i; done
 
 afl-as: src/afl-as.c include/afl-as.h $(COMM_HDR) | test_x86
-	$(CC) $(CFLAGS) src/$@.c -o $@ $(LDFLAGS)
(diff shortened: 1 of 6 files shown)


hexcoder-:
There are dependencies with `-D_FORTIFY_SOURCE=2`. When using the `DEBUG=1` setting, which switches off optimization (sets `-O0`), this conflicts with `-D_FORTIFY_SOURCE=2` and afl++ would not compile.

Currently we already use `-D_FORTIFY_SOURCE=2` in `CFLAGS_OPT` and consequently in `CFLAGS`, when not in the `DEBUG` setting.

Using `CPPFLAGS` unconditionally would circumvent our settings (if we want to switch off `_FORTIFY_SOURCE` and you enable it via `CPPFLAGS`).

So, if we accept `CPPFLAGS`, `_FORTIFY_SOURCE=2` should not be set in it IMO.

rhertzog:
> So, if we accept `CPPFLAGS`, `_FORTIFY_SOURCE=2` should not be set in it IMO.

That's fine. It doesn't need to be set by you (and this patch doesn't give any default value to that variable). For Debian (which doesn't do debug builds), it's set through an environment variable. And that variable can possibly be customized by the package builder.

In general, we want the upstream build system to not override the value passed in the environment but just to extend it with whatever custom option you need.

vanhauser-thc:
Originally I thought CPPFLAGS would just be an old name for CXXFLAGS ... looked it up and no, it is for flags that should be passed to both the cc and c++ compiler. learned something new.
so yes, the patch is fine and needed.

rhertzog:
Actually I was not sure if it was desirable for calls to the C++ compiler so my patch doesn't add it there. But maybe it should be added there too.

PR #531, opened 2020-08-28
Add -fdebug-prefix-map to CFLAGS_SAFE

rhertzog:
In Debian, we override CFLAGS to include -fdebug-prefix-map to avoid
hardcoding the build path in any generated debug information. This is
to help with getting the package to build reproducibly.

However you seem to voluntarily not honor CFLAGS but only CFLAGS_SAFE
for a limited number of source files. This resulted in a lintian warning
on Debian's side (https://lintian.debian.org/tags/file-references-package-build-path.html)
pointing to /usr/lib/afl/afl-llvm-rt-64.o and /usr/lib/afl/afl-llvm-rt.o.

With this commit, I'm manually adding -fdebug-prefix-map as a safe
build flag to CFLAGS_SAFE.

### DIFF
--- a/llvm_mode/GNUmakefile
+++ b/llvm_mode/GNUmakefile
@@ -224,7 +224,8 @@ CFLAGS_SAFE     := -Wall -g -Wno-pointer-sign -I ../include/ \
                    -DAFL_REAL_LD=\"$(AFL_REAL_LD)\" \
                    -DAFL_CLANG_LDPATH=\"$(AFL_CLANG_LDPATH)\" \
                    -DAFL_CLANG_FUSELD=\"$(AFL_CLANG_FUSELD)\" \
-                   -DCLANG_BIN=\"$(CLANG_BIN)\" -DCLANGPP_BIN=\"$(CLANGPP_BIN)\" -DUSE_BINDIR=$(USE_BINDIR) -Wno-unused-function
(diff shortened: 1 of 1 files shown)


andreafioraldi:
The build system ppl will be off until monday IIRC (@vanhauser-thc at least), LGTM to me but we will wait for their approval.

vanhauser-thc:
does not like it will hurt anything so I am fine :)

We do honor CFLAGS, but we have to ensure that specific flags cannot be overridden as otherwise compilation fails, features are not working or the performance impact is huge (e.g. afl-performance.c).

rhertzog:
> We do honor CFLAGS, but we have to ensure that specific flags cannot be overridden as otherwise compilation fails, features are not working or the performance impact is huge (e.g. afl-performance.c).

I understand this. But it would have been nice to document which flags are hurting performance in the case of afl-performance.c. The other patch for CPPFLAGS that you merged actually added the default CFLAGS to compilation line for afl-performance.o (I noted this in the commit description to make you aware of it).

I was not sure if ensuring that CFLAGS_SAFE is added after CFLAGS was sufficient to ensure that any problematic flags is overriden back to its safe value (which would be ideal), or if it was really important that only CFLAGS_SAFE is present... but this option would make it impossible to inject any custom flag like Debian wants.

vanhauser-thc:
lets see if it has unwanted side effects, if so this will take another workaround. :)

PR #534, opened 2020-08-31
push to stable (asap)

vanhauser-thc:
to support the GSOC releases (especially the grammar mutator) we should push to stable so that the code works.
We should do that today.


vanhauser-thc:
@domenukk 
```
+AFL_MAX_DET_EXTRAS: if the dict/extras file contains more tokens than this threshold,
+                    the tokens will sometimes be skipped during fuzzing.
```
I do not understand what I can configure here :) can you update the description please?
also you did not add the description into docs/env_variables.md where I might have read more text about it and understand it :)

andreafioraldi (approved):

docs/custom_mutators.md
```diff
@@ -88,6 +92,11 @@ def queue_new_entry(filename_new_queue, filename_orig_queue):
     This method determines whether the custom fuzzer should fuzz the current
     queue entry or not
 
+- `fuzz_count` (optional):
+
+    This method can be used to instruct afl-fuzz how often to perform a fuzz
```
domenukk: I don't really get this and I doubt the users would: what is a fuzzing attempt? what is 'this input data'? Maybe we should give an example?

domenukk (approved):

docs/custom_mutators.md
vanhauser-thc: better?

docs/env_variables.md
```diff
@@ -369,6 +384,16 @@ checks or alter some of the more exotic semantics of the tool:
     Note that this setting inhibits some of the user-friendly diagnostics
     normally done when starting up the forkserver and causes a pretty
     significant performance drop.
+  
+  - Setting AFL_MAX_DET_EXTRAS changes the count of dictionary enties/extras
```
hexcoder-: duplicate of line 372?

docs/env_variables.md
```diff
@@ -361,6 +369,13 @@ checks or alter some of the more exotic semantics of the tool:
     for an existing out folder, even if a different `-i` was provided.
     Without this setting, afl-fuzz will refuse execution for a long-fuzzed out dir.
 
+  - Setting AFL_MAX_DET_EXRAS will change the threshold at what number of elements
```
hexcoder-: typo?
AFL_MAX_DET_EXRAS
->
AFL_MAX_DET_EXTRAS
but  maybe duplicate (line 388)

llvm_mode/LLVMInsTrim.so.cc
```diff
@@ -243,7 +243,7 @@ struct InsTrim : public ModulePass {
     for (unsigned I = 0; I < PrevLocSize - 1; ++I)
       PrevLocShuffle.push_back(ConstantInt::get(Int32Ty, I));
 
-    for (unsigned I = PrevLocSize; I < PrevLocVecSize; ++I)
+    for (int I = PrevLocSize; I < PrevLocVecSize; ++I)
```
hexcoder-, line 246: Do not understand, what was wrong with it? Line 243 is using `unsigned` as well.

hexcoder- (approved):

domenukk:
just noticed that `AFL_FORKSRV_INIT_TMOUT` is used in showmap and tmin, although not listed in the help output. should we add it there?

llvm_mode/LLVMInsTrim.so.cc
domenukk: `PrevLocVecSize` is using a signed int, so this compares signed to unsigned, potentially a cause for bugs.
`PrevLocSize` however (compared to in line 243) is a unsigned int as well.
Maybe `PrevLocVecSize` should be `unsinged` instead?

vanhauser-thc:
> just noticed that `AFL_FORKSRV_INIT_TMOUT` is used in showmap and tmin, although not listed in the help output. should we add it there?

yes please!

llvm_mode/LLVMInsTrim.so.cc
vanhauser-thc: they could all be unsigned I think. I think I was the one doing that change because it triggered a compiler warning. dunno why no other line triggered ...

PR #537, opened 2020-08-31
Fixes missing variable declaration afl network example

bitwave:
I couldn't get the project get compiled using `make distrib`. Fixing the compiler warning by adding the missing variable declaration solved it for me.


domenukk:
@bitwave thanks!
@vanhauser-thc this is a leftover of the possible wrong usage in the deflate function.
I don't think this PR fixes the underlying issue (the wrong variable is passed to the deflate function).
Do you have an idea here? 
Also, is `make distrib` not part of our CI?

vanhauser-thc:
dunno why this happened. weird. specifying the the variable is not enough. fixed in dev now, thanks for reporting!

PR #539, opened 2020-08-31
Fixed stack use-after-return bug in strntoll

hazimeh:
When `str` holds a numerical value that cannot be represented by the numerical type, `strtol` returns the minimum or maximum value for that type, but it correctly sets `*endptr` to point to the first invalid character (or '\0').

Before this fix, `strntoll` did not correctly set `*endptr`, but left it to point somewhere within the stack buffer `char buf[64]`.
After returning from `strtoll`, `cmp_extend_encoding` uses the `*endptr` value to calculate the length of the number in the buffer. 

When the number is not representable, `*endptr` is still pointing somewhere in the previous stack frame, and is used to calculate `old_len = endptr - buf_8`, which turns out to be a huge number (because the stack is allocated at a high memory address).

`cmp_extend_encoding` then performs the following call:
```
memcpy(new_buf + idx + num_len, buf_8 + old_len, len - idx - old_len);
                                                           ^
                                                           `-- integer underflow (unless len-idx is large enough)
```
And the memcpy SEGFAULTs.

This was discovered when evaluating AFLplusplus against [Magma](https://github.com/HexHive/magma). One of the test cases of the `php` target had a 30-character repeated sequence of 1's, and the fuzzer kept crashing when fuzzing it. The crash was triaged with the help of ASan and UBSan.

Note: the AFL++ code base is riddled with undefined behavior for integers. I'm not sure if they affect the correctness or the effectiveness of the fuzzing process, but at least they don't crash the fuzzer yet (except for this bug).

### DIFF
--- a/src/afl-fuzz-redqueen.c
+++ b/src/afl-fuzz-redqueen.c
@@ -282,7 +282,6 @@ static long long strntoll(const char *str, size_t sz, char **end, int base) {
   memcpy(buf, beg, sz);
   buf[sz] = '\0';
   ret = strtoll(buf, end, base);
-  if (ret == LLONG_MIN || ret == LLONG_MAX) return ret;
   if (end) *end = (char *)beg + (*end - buf);
   return ret;
 


vanhauser-thc:
uh good find, thank you!
@andreafioraldi this is your code, please have a look to be sure this is the right way to fix it

vanhauser-thc:
> Note: the AFL++ code base is riddled with undefined behavior for integers. I'm not sure if they affect the correctness or the effectiveness of the fuzzing process, but at least they don't crash the fuzzer yet (except for this bug).

do you mean potential issues or where ubsan triggers and says there is a wrapping or an assignment that has a real loss and not only a theoretical loss?

hazimeh:
I didn't check them all, so I wouldn't know if they have any practical effect on fuzzer behavior.
A few are due to the SWAP_{16,32} macros, which implicitly cast the input to u{16,32}, but I don't think those are too dangerous.
However, there are some cases in the RedQueen implementation where the addition overflows. I saw one such example in the code where it tries to skip loop comparisons, and it performs a +1 to a UINT_MAX and it overflows. The variable names were too obfuscated and I didn't investigate further.

andreafioraldi:
ah yes, this happens then you copy strntoll from stackoverflow. textual redqueen is a new feature, ty for the report.

vanhauser-thc:
@andreafioraldi  I assume the fix is fine and no further action is required? (so I merged)

also - I do not think this is exploitable as although this is from target data, it is in our own memory area where this happens (meaning a target has very little control there)?
or should be label this as security issue and release an update?

andreafioraldi:
No the fix is incomplete (in term of the correctness of the mutator, but fix the memory corruption) that's why i didn't merged.
Exploitable means nothing here, the data comes from the execution of the target, if an attacker controls the binary that is fuzzed, he already has code execution.

andreafioraldi:
The main issue here is that I don't have a testcase to reproduce the bug to see if the fix works and build also a regression test. Cannot use PHP ofc in the test suite.

andreafioraldi:
Regards the other UBs, @hazimeh you mean this code?

```c
      if (s_v0 != o->v0) { s_v0_fixed = 0; }
      if (s_v1 != o->v1) { s_v1_fixed = 0; }
      if (s_v0 + 1 != o->v0) { s_v0_inc = 0; }
      if (s_v1 + 1 != o->v1) { s_v1_inc = 0; }
      if (s_v0 - 1 != o->v0) { s_v0_dec = 0; }
      if (s_v1 - 1 != o->v1) { s_v1_dec = 0; }
      s_v0 = o->v0;
      s_v1 = o->v1;
```

Most of the mutator's code do integers weird tricks, is expected that ubsan complains about them. In this case this just checks if v0 or v1 is a counter (uint64).

vanhauser-thc:
> No the fix is incomplete (in term of the correctness of the mutator, but fix the memory corruption) that's why i didn't merged.
> Exploitable means nothing here, the data comes from the execution of the target, if an attacker controls the binary that is fuzzed, he already has code execution.

so there will be a commit for this?
I thought about pushing this fix to stable now, but then this should rather wait :)

vanhauser-thc:
ah there is your commit already ;) just tell me when you are satisfied and sure it is fine (as far as it is without having a test case) as I will then merge to stable.

andreafioraldi:
Yes i'm compiling php just now

hazimeh:
[poc.tar.gz](https://github.com/AFLplusplus/AFLplusplus/files/5152028/poc.tar.gz)
Here's a MWE. The `parser_{afl,cmp}` binaries are compiled from the v1.0 branch on the Magma repo.
The `afl-fuzz` binary is compiled from the AFLplusplus repo, checked out at 986af28df27016813abdfdde8bdedda1f571703c, and compiled with `-fsanitize=integer,address`.
The `in` directory contains the seed file causing the culprit.

I tried to run this on another machine outside the docker container it was in, but the forkserver failed to spin up for some reason. However, I hope it has enough data in it to let you reproduce (I think the seed file should be enough).

andreafioraldi:
@hazimeh after how much time you experience the bug? I'm using php-fuzz-parser as target, it is right?

hazimeh:
After around a minute. And yes, `php-fuzz-parser` is the target I'm using.

andreafioraldi:
Ok ty reproduced, and my fix works.

andreafioraldi:
@vanhauser-thc all ok here, I tested the correctness of the mutator after the fix too (text redqueen alone found 17 inputs in the first iteration), push to stable.

PR #542, opened 2020-09-02
push to stable

vanhauser-thc:
Push before a new release, and the last 2.x release before 3.0


vanhauser-thc:
the only two important things are the redqueen heap corruption fix and the debian changes.
LGTM

src/afl-fuzz-redqueen.c
```diff
@@ -265,24 +265,21 @@ static u8 its_fuzz(afl_state_t *afl, u8 *buf, u32 len, u8 *status) {
...
 
 static int strntoll(const char *str, size_t sz, char **end, int base,
-                    long long* out) {
+                    long long *out) {
 
   char        buf[64];
   long long   ret;
   const char *beg = str;
 
   for (; beg && sz && *beg == ' '; beg++, sz--) {};
```
domenukk, line 274: this wants to be a while loop so badly 🙃

domenukk (approved):

andreafioraldi (approved):

hexcoder- (commented):
IMO it is not a good idea to widen the type to negative values, since this is not necessary.
So I reverted the u32-> s32 change and eliminated s32 for `argc` and `i`.

hexcoder- (approved):

PR #544, opened 2020-09-04
Export set afl_environment_variables to stats

ThomasTNO:
Logging the AFL environment variables is useful to improve reproducibility of fuzzing campaigns and allow for proper analysis.

The logging of these variables, wheLogging the AFL environment variables is useful to improve reproducibility of fuzzing campaigns and allow for proper analysis.

The logging of these variables, when set, is now integrated in `afl-stats`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1035,6 +1035,7 @@ without feedback, bug reports, or patches from:
   Andrea Biondo                         Vincent Le Garrec
   Khaled Yakdan                         Kuang-che Wu
   Josephine Calliotte                   Konrad Welc
+  Thomas Rooijakkers
 ```
 
 Thank you!
--- a/include/afl-fuzz.h
+++ b/include/afl-fuzz.h
@@ -945,12 +945,11 @@ void destroy_extras(afl_state_t *);
 
 /* Stats */
 
-FILE *open_file(const char *);
-void  write_fuzzer_config_file(afl_state_t *);
-void  write_stats_file(afl_state_t *, double, double, double);
-void  maybe_update_plot_file(afl_state_t *, double, double);
-void  show_stats(afl_state_t *);
-void  show_init_stats(afl_state_t *);
+void write_fuzzer_config_file(afl_state_t *);
+void write_stats_file(afl_state_t *, double, double, double);
+void maybe_update_plot_file(afl_state_t *, double, double);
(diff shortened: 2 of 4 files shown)


vanhauser-thc:
Thank your for the idea and patch.

this is a bit of an overhead, doing this for every write of fuzzer_stats, especially as this is something does not change during a run.
Also using afl_get_env() triggers ->  OKF("Loaded environment variable %s with value %s", env, val); .... so getenv() is the better choice (afl_get_env is for program startups so it is printed what is accessed. sometimes people forget they still had an env set in that shell ....)

I propose the following changes:
  1. use getenv()
  2.  move the feature to an own function e.g. "write_fuzzer_setup(afl), and write to a different, new file, e.g. "fuzzer_setup"
  3. call that function within this command block in afl-fuzz
```
  write_stats_file(afl, 0, 0, 0);
  maybe_update_plot_file(afl, 0, 0);
  save_auto(afl);
```
in the future more information can be written there (and maybe moved away from fuzzer_stats).

If you can do this today this will end up in the 2.68c release today/tomorrow, otherwise it will land in 3.0 at some future release.

vanhauser-thc:
Also for a start maybe write it like an env file so for the time being people could do just a ". out/fuzzer_config" and would have that env set.

so
```
AFL_FOO=1
AFL_BAR=/tmp/input
...
```

domenukk:
Why not write them to the commandline file?

vanhauser-thc:
this is also an option. however then we would
a) need to change the cmdline file  (currently it is one line per parameter) to be a command list
b) it would be a change for 3.0 as this could break existing user scripts.

ThomasTNO:
> 
> 
> Also for a start maybe write it like an env file so for the time being people could do just a ". out/fuzzer_config" and would have that env set.
> 
> so
> 
> ```
> AFL_FOO=1
> AFL_BAR=/tmp/input
> ...
> ```

I Like this suggestion, give me a minute.

vanhauser-thc:
if you create a new function for refactoring then add the "inline" tag please :) (open_file)

vanhauser-thc:
I think this needs a few more tweaks (maybe the out-dir cleaning fails), but I will take care of that.
thanks!

PR #551, opened 2020-09-08
rewrite gcc plugin

ButcherAdaCore:
When we started using AFL, it did not have an integrated GCC plugin.
There was one proposed by Austin Seipp, but for various reasons we
ended up using some of its infrastructure (runtime and wrapper), but
writing the GCC plugin proper from scratch.

With AFL++'s renewed interest in a GCC plugin, we rebased ours, with
some features that are or were missing in the one that was integrated:

* efficient, fully-functional inline and out-of-line instrumentation

Inline instrumentation was work in progress in the original plugin.
Controlled by AFL_GCC_OUT_OF_LINE.

* reproducible instrumentation

Obey -frandom-seed for pseudorandom number generation.

* licensing clarity and strict compliance

GPLv3+ for the plugin, that uses GCC internals; add a copy of the
license, as required.

* allow/deny list support

Copied and adjusted from the LLVM plugin implementation.

* neverZero support

Not as compact as the asm-wrapper version, but likely more efficient.
Both are quite thread-unsafe, with different caveats.
Controlled with AFL_GCC_SKIP_NEVERZERO.


vanhauser-thc:
Thank you! merging is for now and trying to integrate it properly :)

PR #553, opened 2020-09-09
Warn if non deterministic mode and dictionary is used.

murx-:
This commit adds a warning if a non deterministic mode (`-S` or `-d`) and a dictionary (`-x`) is used together. Dictionary mode is only used in the deterministic mode this it makes no sense to use it in non deterministic mode. 

I decided to print a warning and not terminate as it might break some peoples scripts.

### DIFF
--- a/src/afl-fuzz.c
+++ b/src/afl-fuzz.c
@@ -456,6 +456,10 @@ int main(int argc, char **argv_orig, char **envp) {
         }
 
         extras_dir[extras_dir_cnt++] = optarg;
+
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
that is actually not true.
the dictionary is used in havoc mode as well.

murx-:
thanks for clarifying. I was confused by the status screen and did not look at the havoc implementation.

vanhauser-thc:
all good, afl++ is quite a complex beast :)

PR #556, opened 2020-09-13
Haiku  build fix

devnexen:


domenukk:
Thanks :)
Should this maybe use use `USEMMAP` env instead, like it's the case in afl_sharedmem.c? (Or better yet, use the libafl impl)

hexcoder-:
For shared memory Haiku currently does only support some proprietary api with arenas, which we currently do not support.
Neither SysV shared memory nor `mmap`, right? So no `USEMMAP`, AFL++ can only run in dumb mode here :-(.

PR #558, opened 2020-09-14
Compiler plugins build fix proposal. NetBSD (and possibly some other …

devnexen:
…systems

 ) does not have type command.


domenukk:
How about `command -v`? I've seen that used in some places around the codebase. Think that works on Darwin, too? cc @hexcoder-

hexcoder-:
Thanks, @devnexen! There are some issues with NetBSD. I am on it.

hexcoder-:
I used command -v instead and fixed some minor compiler warnings.

PR #560, opened 2020-09-17
Fix broken links to READMEs in llvm_mode directory

jedevc:


vanhauser-thc:
please only ever push to dev, never to stable.
and as you can see this is fixed/changed in dev already :)

PR #564, opened 2020-09-23
Fix potential endless loop in custom_mutator_stage

wakolzin:
1) To be honest, I tried to understand why some splicing steps (like choice of random entry from queue) is used in custom mutator stage - and I didn't understand this.
In this PR "splicing-steps" were removed from custom mutator stage.
Please, explain the sense of this steps in custom mutator stage if I'm not right.

2) The current realization of custom mutator stage on dev branch permits infinite loop (with begin at label 'retry_external_pick') when following conditions are met:
* we use custom mutators;
* queue contains at least 4 entries;
* only one entry in this queue has len>=2 (others have len=1);
* current entry (with index `afl->current_entry`) has len >= 2.

### DIFF
--- a/include/afl-fuzz.h
+++ b/include/afl-fuzz.h
@@ -657,6 +657,9 @@ typedef struct afl_state {
    * they do not call another function */
   u8 *map_tmp_buf;
 
+  /* queue entries ready for splicing count (len > 1) */
+  u32 ready_for_splicing_count;
+
 } afl_state_t;
 
 struct custom_mutator {
--- a/src/afl-fuzz-one.c
+++ b/src/afl-fuzz-one.c
@@ -1699,6 +1699,9 @@ custom_mutator_stage:
           u8 *                new_buf = NULL;
           u32                 target_len = 0;
 
+          /* check if splicing is possible (if the only entry has len > 1
+           * check it is not current entry)
+           */
           if (afl->ready_for_splicing_count > 1 ||
               (afl->ready_for_splicing_count == 1 &&
                afl->queue_cur->len == 1)) {
@@ -2769,7 +2772,7 @@ static u8 mopt_common_fuzzing(afl_state_t *afl, MOpt_globals_t MOpt_globals) {
 
     len = afl->queue_cur->len;
 
-    /* maybe current entry stop being ready for splicing */
+    /* maybe current entry is not ready for splicing anymore */
     if (old_len > 1 && afl->queue_cur->len == 1)
       afl->ready_for_splicing_count--;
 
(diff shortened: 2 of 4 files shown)


vanhauser-thc:
If you do not understand it - why do you want to remove it for everyone?
If we put it there it might be in there for a reason, right?

splicing is one of the most successful mutation stategies.
hence a custom mutator should support it and that is why a custom mutator gets an extra random test case.
We are thinking however about making this optional as not all mutators do splicing.

good catch about the potential endless loop though.

If you want you can rewrite your PR to just fix the potential loop.

wakolzin:
Yes, I'll fix endless loop.

And few more questions about custom mutators and splicing. When we first time do custom mutator stage do we really need to choose random entry from queue or just mutate current entry?

I mean that we anyway further in splicing stage will choose random entry. Let's consider these lines in custom_mutator_stage (when is is executed after goto from retry_splicing):
```
          size_t mutated_size =
              el->afl_custom_fuzz(el->data, out_buf, len, &mutated_buf, new_buf,
target->len, max_seed_size);
```

`out_buf` - is got in retry_splicing stage from random entry;
`new_buf`- is got in custom mutator stage from random entry.

So we make splicing with 2 random entries, but not with current entry and random entry. Is it as expected?

vanhauser-thc:
That is up to the mutator to decide if they want to use the extra splicing data or not. we want to provide the option to do so.
I dont understand what you mean with 2 random entries. It only receives two queue entries. one is the one selection for mutation from the scheduler (out_buf), the other for splicing (new_buf) - if the mutator does splicing.

p13l13d13:
Just a little update to clarify this issue:

There is no need for the current entry length to be >=2. It will loop when all non-cur entries have len < 2, despite the length of the current one.

Anyway, if it's up to the mutator to decide what to do with optional data, there should be no limitations on data lenght. (let's say  someone wanna write some weird mutator and additionaly use ANY other entry from queue, regardless of its length).

But it's up to you to decide what to do with limitations, although I think this is the easiest way to fix the situation

vanhauser-thc:
hmm this check is done every time which is not very effective. And it is done even if there is no custom mutator.
you can either create a variable in afl struct in afl-fuzz.h to use once it is ready for splicing or just have a counter for the trying to find a splicing target and if none are found after 3-4 attempts then set it to NULL.

vanhauser-thc:
thanks for the changes.
I will accept this now although have various issues with inefficiency. I will fix this by hand though.

wakolzin:
Thank you for explanation and review.

PR #571, opened 2020-10-05
Statsd implem

Edznux:
Add configurable, in process, StatsD client to AFL++.

Notes:
- I'm definitely not quite fluent in C as in other languages.
- This started mostly for myself, so I didn't create any issue or anything before writing this. If you have anything design/conception/quality/whatever-wise feel free. If this PR is not something you're interested in at all, that's totally fine as well :)

## What

This PR adds a StatsD client to AFL++.
StatsD is a daemon that collects statistics based on a simple textual protocol (`<metricname>:<value>|<type>`). The protocol is implemented by a lot of clients, sometimes with extensions.

The code supports metric tags (fixed to `banner` and `afl_version` for now), with the `|#tag:value` notation.
Since there is a no standards for tags, I only implemented DogstatsD format, since that it seems to be the "first" one [based on this](https://github.com/netdata/netdata/issues/3813#issuecomment-399276423) (I didn't verified) and the one that I will be using / testing. If desired, I can add support for other through a config (`#define USE_STATSD_INFLUXDB` for example) in another PR.
For more information about different format, see [prometheus/statsd_exporter](https://github.com/prometheus/statsd_exporter#tagging-extensions) or [statsd issue #619](https://github.com/statsd/statsd/issues/619)

This PR only support UDP.

## Why

This allows AFL++ to send metrics directly to a StatsD server to monitor and manages the fuzzer instances more easily.

Depending on the StatsD platform used, one could setup monitors and alerts on metrics values or even automate actions based on those.
For example, triggering notifications on "slower than expected fuzzing", "times since last finding > x", "number of crashes above x threshold" ... and mark the CI job as failed could be one scenario.
The current way of doing it today is by adding another process on the fuzzing server, that parses and sends the data read from `fuzzer_stats`. However it has some drawbacks:
- Requires to deploy & run another program.
- Requires to poll the `fuzzer_stats` file instead of just sending the metrics when they are ready (% x amount of time).

In any case, it provide an "out of the box" way of monitoring the fuzzing process, with a standard technology.

Here's an example dashboard to visualize:
Screenshot_2020-10-05 AFL++ Fuzzing Datadog
[also a link, if you want to try it live, but it might not work at some point in the future](https://p.datadoghq.com/sb/aq3i11bf7rj6u6u1-6e0872d9eb3f13ef2da38d8ff071b06c)

## Doubt And Uncertainties

- Using global var for the udp socket (`statds_sock`) to avoid recreating the socket every time looks good IMO, but not sure what's the best practices there...
- Not sure how to check performance penalties.
- Should there be any check on the env var (`AFL_STATSD_PORT` & `AFL_STATSD_HOST`) or is it considered as a "You know what you are doing" setting ?
- This is currently sending one UDP packet with all the metrics about each `STATSD_UPDATE_SEC` seconds. It may be interesting to send metrics only when they changes (with a lower limit to not send metrics on every single increment of some metrics...). It will lead to sending smaller packets, and only when needed (modulo `STATSD_UPDATE_SEC`).
- Not sure if this should be enabled only by macro (`USE_STATSD`), by env var (let's say `AFL_USE_STATSD`) or both. We could also make the current env var the toggle (setting it up would enable the feature, living it empty would disable it).
- I plugged the entry point of this into the `afl-fuzz-stats.c`, not really sure that's the correct way of doing it. (might be better in [src/afl-fuzz-run.c:804](https://github.com/AFLplusplus/AFLplusplus/blob/dev/src/afl-fuzz-run.c#L804) ?)
- It might be interesting to add support for custom tags value, but I don't know how to do that cleanly tbh.

I guess a git squash/rebase will be needed if accepted.

Thanks!


domenukk:
Oh, this looks pretty sweet! :)
A few initial thoughts:
- Using `afl-fuzz-stats.c` as entrypoint sounds good, I would even go one step further and put the statsd-specific checks inside `if (cur_ms - afl->stats_last_plot_ms > PLOT_UPDATE_SEC * 1000) {` for a little extra speed (or do we ever need more stats packets than we need UI redraws?)
- We generally try to avoid global variables, for future multiprocessing, librarification, etc. - simply adding the statsd specific variables to the `afl_state` struct should work. Initialize your port when afl-fuzz starts.
- We have an odd-ish way to deal with env varibals (to allow for warnings on mistyped vars, documentation, etc.), see for example commit https://github.com/AFLplusplus/AFLplusplus/commit/53e63e9ded202b63de19590bb3d265d2f01ee929 on how to add them.

I don't have any experience with statsd (the screenshots look useful though), anybody else from the team?

Edznux:
> Using afl-fuzz-stats.c as entrypoint sounds good, I would even go one step further and put the statsd-specific checks inside if (cur_ms - afl->stats_last_plot_ms > PLOT_UPDATE_SEC * 1000) { for a little extra speed (or do we ever need more stats packets than we need UI redraws?)

I first thought about doing it this way, but I would assume the perf gain was insignificant compared to sending the packet itself. The ability to change this variable without interfering with the other would on the other hand allows us to set a very low UI refresh rate while still getting the stats (and thus the UI indirectly) fast enough with statsd.

>We generally try to avoid global variables, for future multiprocessing, librarification, etc. - simply adding the statsd specific variables to the afl_state struct should work. Initialize your port when afl-fuzz starts.

Will look into it, thanks.

>We have an odd-ish way to deal with env varibals (to allow for warnings on mistyped vars, documentation, etc.), see for example commit 53e63e9 on how to add them.

I added them to `src/afl-fuzz.c` and `include/envs.h` but I don't think it's relevant to `afl-cmin`, `src/afl-showmap.c` or `src/afl-tmin.c` since the metrics only works on `alf-fuzz`. I can add them if that's supposed to be there as well.

andreafioraldi:
One visualization fix, for the edges over time graph, use logscale or gamozo will get angry ;)

domenukk:
> 
> I added them to `src/afl-fuzz.c` and `include/envs.h` but I don't think it's relevant to `afl-cmin`, `src/afl-showmap.c` or `src/afl-tmin.c` since the metrics only works on `alf-fuzz`. I can add them if that's supposed to be there as well.

Aah sorry wrong hash, I meant something like
1301552101af899557a93a7535d8a57874fe6edf
or
425908a00cdcaa4d49a513d283431b8b6eed0486

/edit: that being said, I can also do the plumbing

domenukk:
> One visualization fix, for the edges over time graph, use logscale or gamozo will get angry ;)

maybe both?

@Edznux btw what happens if you add multiple secondary fuzzers (main with `-M main`, then secondaries with `-S s[1..n]`) to the statsd dashboard?

Edznux:
> > I added them to `src/afl-fuzz.c` and `include/envs.h` but I don't think it's relevant to `afl-cmin`, `src/afl-showmap.c` or `src/afl-tmin.c` since the metrics only works on `alf-fuzz`. I can add them if that's supposed to be there as well.
> 
> Aah sorry wrong hash, I meant something like
> [1301552](https://github.com/AFLplusplus/AFLplusplus/commit/1301552101af899557a93a7535d8a57874fe6edf)
> or
> [425908a](https://github.com/AFLplusplus/AFLplusplus/commit/425908a00cdcaa4d49a513d283431b8b6eed0486)
> 
> /edit: that being said, I can also do the plumbing

Ah yes, will add the `} else if (!strncmp(env,` in  `src/afl-fuzz-state.c`.
No worries, I can do it :)

> One visualization fix, for the edges over time graph, use logscale or gamozo will get angry ;)

Done on the live dashboard! (not really a visible change since my example fuzzer only has a flat line for this graph so I'm not going to update the screenshot :p)

>maybe both?

That's something everyone can change on there side. It will depends on every "graph provider" you are using.

> @Edznux btw what happens if you add multiple secondary fuzzers (main with -M main, then secondaries with -S s[1..n]) to the statsd dashboard?

That's the multiple line you see in some graphs, I named my `-M` instance `fuzzer01` and -S `fuzzer02`. (In case you didn't check the "live" dashboard, go ahead, you will understand). Every fuzzer instances has its own `banner` and this get injected in the tags which allow to "group by". In your example you will have a line for "main"  and "s[1..n]".

domenukk:
Ooh nice. Also just read up a bit on statsd some more, it seems to be a pretty widely supported format, definitely nice to have!

src/afl-fuzz-statsd.c
```diff
@@ -0,0 +1,158 @@
...
+  return 0;
+
+}
+
+int statsd_format_metric(afl_state_t *afl, char *buff, size_t bufflen) {
+
+/* Metric format:
+<some.namespaced.name>:<value>|<type>
+*/
+#ifdef USE_DOGSTATSD_TAGS
```
domenukk: Are these tags only used by Datadog or supported by other StatsD providers?
Not a big fan of a for-profit kinda-proprietary format here, maybe we find a way to make this generalizable/configurable?

domenukk: Ah the linked comment, https://github.com/netdata/netdata/issues/3813#issuecomment-399276423, kinda answers my question. Maybe we don't need the ifdef at all, other consumers could just ignore the tag?

Edznux: I agree, and it is possible to add other format as well.

(from the PR text directly, the links are working there, not in this quote)
> The code supports metric tags (fixed to banner and afl_version for now), with the |#tag:value notation.
Since there is a no standards for tags, I only implemented DogstatsD format, since that it seems to be the "first" one based on this (I didn't verified) and the one that I will be using / testing. If desired, I can add support for other through a config (#define USE_STATSD_INFLUXDB for example) in another PR.
For more information about different format, see prometheus/statsd_exporter or statsd issue #619

I just wanted to push the PR first, before implementing other format if the PR wasn't "accepted".

From what I saw, there is total  incompatibility between each provider, so will need to add a few common ones. Not sure which should be the default one though.

You can see a list [there](https://github.com/prometheus/statsd_exporter#tagging-extensions)

Edit: I don't have any other providers set up, so can't really test if they (some or all) ignores it or handle it properly or else.

domenukk: Okay, the way I see it it's mostly the seperator that differs, the content stays the same?
So we can easily set an sprintf string in the config (or via env?) the content stays the same, right?

domenukk: Okay, the way I see it it's mostly the seperator that differs, the content stays the same?
So we can easily set an sprintf string in the config (or via env?) the content stays the same, right?
/edit:
Or do we just want to add the 4 or 5 supported ones and have a `AFL_STATSD_TAG_FLAVOR` env
(there is also the grahite flavor missing from the list: https://graphite.readthedocs.io/en/latest/tags.html )

Edznux: Yep, that's what I'm thinking too.

So, changing the #define by an env var and then doing something along the lines of 
```c
else if (AFL_STATSD_TAG_FLAVOR == USE_LIBRATO_TAGS) {
 char tags[MAX_TAG_LEN * 2] = {0};
  snprintf(tags, MAX_TAG_LEN * 2, "#banner=%s,afl_version=%s", afl->use_banner,
           VERSION);
} else if  (AFL_STATSD_TAG_FLAVOR == USE_INFLUXDB_TAGS) {
 char tags[MAX_TAG_LEN * 2] = {0};
  snprintf(tags, MAX_TAG_LEN * 2, ",banner=%s,afl_version=%s", afl->use_banner,
           VERSION);
} else if  (AFL_STATSD_TAG_FLAVOR == USE_SIGNALFX_TAGS) {
 char tags[MAX_TAG_LEN * 2] = {0};
  snprintf(tags, MAX_TAG_LEN * 2, "[banner=%s,afl_version=%s]", afl->use_banner,
           VERSION);
}
```

The annoyance is that they aren't injected in the same position in the metrics.
DogstatsD is at the end of the line while most of the other are just after the metric name.
But that's not a big deal, for sure.

domenukk: ooh different locatons? Well that's just mean.
I was thinking along the lines of (during init)
```c
if (AFL_STATSD_TAG_FLAVOR == LIBRATO) {
  afl_state->statsd_format = "#banner=%s,afl_version=%s"
} else ...
```
And then later using `afl_state->statsd_format`in `snprintf` directly...

That being said, if the location of the tag differs anyway, you could also have individual complete snprintfs (the whole packet), hidden behind the if of your example above - and play around with some preprocessor magic to build the string and params? could be a bit define heavy  but quick - up to you :)

vanhauser-thc: > From what I saw, there is total incompatibility between each provider, so will need to add a few common ones. Not sure which should be the default one though.

the default should be a non-commercial one.
Also it is fine if you add support for other variants if you want. we will accept the PR :)

src/afl-fuzz-statsd.c
```diff
@@ -87,7 +188,7 @@ int statsd_send_metric(afl_state_t *afl) {
              (struct sockaddr *)&afl->statsd_server,
              sizeof(afl->statsd_server)) == -1) {
 
-    if (!close(afl->statsd_sock)) { perror("Cannot close socket"); }
+    if (!close(afl->statsd_sock)) { FATAL("Cannot close socket"); }
```
domenukk: You can use PFATAL for these errors, it'll print the errno information on top

Edznux: Done, thanks.

src/afl-fuzz-statsd.c
Edznux: Ok, so I made a few modification but it should be good now. I set the default to no tag if there is no/empty`AFL_STATSD_TAGS_FLAVOR` provided.

Quite unsure if the init should be done in `afl-fuzz.c:895`, but that looked a good place for it.
I inspected all the packets from the different types of tags type (with Wireshark, I don't have any other platform set up):
(truncated to first 4 lines each)

`AFL_STATSD_TAGS_FLAVOR=`:
```
fuzzing.cycle_done:0|g
fuzzing.cycles_wo_finds:0|g
fuzzing.execs_done:95077|g
fuzzing.execs_per_sec:4128.22|g
...
```

`AFL_STATSD_TAGS_FLAVOR=librato`:
```
fuzzing.cycle_done#banner=f1,afl_version=++3.00a:0|g
fuzzing.cycles_wo_finds#banner=f1,afl_version=++3.00a:0|g
fuzzing.execs_done#banner=f1,afl_version=++3.00a:9|g
fuzzing.execs_per_sec#banner=f1,afl_version=++3.00a:9000.00|g
...
```

`AFL_STATSD_TAGS_FLAVOR=dogstatsd`:
```
fuzzing.cycle_done:0|g|#banner:f1,afl_version:++3.00a
fuzzing.cycles_wo_finds:0|g|#banner:f1,afl_version:++3.00a
fuzzing.execs_done:77557|g|#banner:f1,afl_version:++3.00a
fuzzing.execs_per_sec:4136.37|g|#banner:f1,afl_version:++3.00a
...
```

`AFL_STATSD_TAGS_FLAVOR=signalfx`:
```
fuzzing.cycle_done[banner=f1,afl_version=++3.00a]:0|g
fuzzing.cycles_wo_finds[banner=f1,afl_version=++3.00a]:0|g
fuzzing.execs_done[banner=f1,afl_version=++3.00a]:4177|g
fuzzing.execs_per_sec[banner=f1,afl_version=++3.00a]:4135.64|g
...
```

`AFL_STATSD_TAGS_FLAVOR=influxdb`:
```
fuzzing.cycle_done,banner=f1,afl_version=++3.00a:0|g
fuzzing.cycles_wo_finds,banner=f1,afl_version=++3.00a:0|g
fuzzing.execs_done,banner=f1,afl_version=++3.00a:4164|g
fuzzing.execs_per_sec,banner=f1,afl_version=++3.00a:4147.41|g
...
```

Edit: well, the CI disagree haha
Edit 2: Fixed, I missed a check before read.

Edznux:
I think I fixed everything asked in the threads.

The last question is should I add a env var for enabling this (or disabling this) or is the `#define` enough ?
Other than that, I'm pretty happy with the result. Let me know if there's anything else :)

Thanks!
Edit: The dashboard is using the last version, with the flavor set to `dogstatsd`

vanhauser-thc:
I will have a thorough look at it on Friday. Thanks so far!

vanhauser-thc:
IMHO:
making this a compile option means it will hardly be ever used.
And when the compile option is set, it always send packets - even if no statsd is present. also not good.

so I would propose removing the ifdef, and requiring for statsd information to be sent that the _HOST parameter is required.
and before every call there is a `if (unlikely(afl->env->afl_statsd_host))` in front.
The performance impact is very little and is just in the stats check function.

what do you think?
@domenukk @Edznux

domenukk:
I agree. Compiletime won't be used. Using the the host or just a AFL_STATSD=1 env are both good options.

Edznux:
Yea agree, what about doing all options as a toggle then ?

Like the following pseudo code
```c
if (unlikely(AFL_STATSD != 0) || (AFL_STATSD_HOST != NULL) || (AFL_STATSD_PORT != NULL))  {
  if (cur_ms - afl->statsd_last_send_ms > STATSD_UPDATE_SEC * 1000) {

    /* reset counter, even if send failed. */
    afl->statsd_last_send_ms = cur_ms;
    if (statsd_send_metric(afl)) { WARNF("coundln't send statsd metric."); }
  }
}
```

vanhauser-thc:
Only check for one thing, checking for more just increases the CPU time.
Just set this one thing to disable if not everything that is needed is present.

Also note that the WARNF will not really show because the UI reprints the screen every second. only with AFL_NO_UI it would be visible.

Edznux:
OK, I will add the `AFL_STATSD=1` env var then. :)

And sure, it will not be really shown while the UI is still refreshing but I'm not sure where to warn for it in this case ? I definitely don't want to `exit()` on a failure to send the metric...

Also, since using StatsD and the UI is a bit redundant, I feel like people that will use the StatsD metrics will "often" disable the UI.

src/afl-fuzz.c
```diff
@@ -889,6 +894,8 @@ int main(int argc, char **argv_orig, char **envp) {
 
   }
 
+  if (unlikely(afl->afl_env.afl_statsd == 1)) { statsd_setup_format(afl); }
```
vanhauser-thc: I'd prefer a simple
`if (unlikely(afl->afl_env.afl_statsd)) { statsd_setup_format(afl); }`
for readability :)

src/afl-fuzz-stats.c
```diff
@@ -423,6 +423,18 @@ void show_stats(afl_state_t *afl) {
 
   }
 
+  if (unlikely(afl->afl_env.afl_statsd == 1)) {
+
+    if (cur_ms - afl->statsd_last_send_ms > STATSD_UPDATE_SEC * 1000) {
+
+      /* reset counter, even if send failed. */
+      afl->statsd_last_send_ms = cur_ms;
+      if (statsd_send_metric(afl)) { WARNF("coundln't send statsd metric."); }
```
vanhauser-thc, line 432: also here please remove the ==1 in the if above, plus fix the typo in the WARNF please

vanhauser-thc:
LGTM, just fix the typo and two nit picks and I am happy to merge. thank you! great effort and very useful!

Edznux:
Done!
Should be good to go now! 
Will you "Squash & Merge"  or do you want me to clean some things ?

Thanks a lot !

vanhauser-thc:
I will squash. I missed fixing the typo, will do this after the merge. thanks!

vanhauser-thc:
ah if you want you can add some documentation to docs/status_screen.md about the feature. totally forgot that. also adding the env vars to docs/env_variables.md

PR #574, opened 2020-10-06
Testcase Cache

domenukk:
This introduces a testcase cache, it seems to work well. Input appreciated

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -301,7 +301,7 @@ Decrease if RAM usage is high. */
 
 #if TESTCASE_CACHE_SIZE < 4
   #error \
-      "Dangerously low cache size: Set TESTCASE_CACHE_SIZE to 4 or more in config.h!
+      "Dangerously low cache size: Set TESTCASE_CACHE_SIZE to 4 or more in config.h!"
 #endif
 
 /* Maximum line length passed from GCC to 'as' and used for parsing
--- a/include/config.h
+++ b/include/config.h
@@ -297,7 +297,7 @@
 
 /* The amount of entries in the testcase cache, held in memory.
 Decrease if RAM usage is high. */
-#define TESTCASE_CACHE_SIZE 2048
+#define TESTCASE_CACHE_SIZE 3072
 
 #if TESTCASE_CACHE_SIZE < 4
   #error \
--- a/include/config.h
+++ b/include/config.h
@@ -300,7 +300,8 @@ Decrease if RAM usage is high. */
 #define TESTCASE_CACHE_SIZE 2048
 
 #if TESTCASE_CACHE_SIZE < 4
-  #error "Dangerously low cache size: Set TESTCASE_CACHE_SIZE to 4 or more in config.h!
+  #error \
+      "Dangerously low cache size: Set TESTCASE_CACHE_SIZE to 4 or more in config.h!
 #endif
 
 /* Maximum line length passed from GCC to 'as' and used for parsing
--- a/include/config.h
+++ b/include/config.h
@@ -295,6 +295,14 @@
 
 #define RESEED_RNG 100000
 
+/* The amount of entries in the testcase cache, held in memory.
+Decrease if RAM usage is high. */
+#define TESTCASE_CACHE_SIZE 2048
+
+#if TESTCASE_CACHE_SIZE < 4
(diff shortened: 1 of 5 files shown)


vanhauser-thc:
instead of merging I would propose we first test it on fuzzbench what the impact is.
is it ready? or does it still need changes?

domenukk:
It works as far as my tests go, so yes, ready

include/config.h
```diff
@@ -295,6 +295,15 @@
 
 #define RESEED_RNG 100000
 
+/* The amount of entries in the testcase cache, held in memory.
+Decrease if RAM usage is high. */
+#define TESTCASE_CACHE_SIZE 2048
```
vanhauser-thc: if we add this, this should be an env (and setting the env to any size should enable the feature). or?

src/afl-fuzz-queue.c
```diff
@@ -767,3 +768,89 @@ u32 calculate_score(afl_state_t *afl, struct queue_entry *q) {
...
+/* Returns the testcase buf from the file behind this queue entry.
+  Increases the refcount. */
+u8 *queue_testcase_take(afl_state_t *afl, struct queue_entry *q) {
+
+  if (!q->testcase_buf) {
+
+    u32 tid = 0;
+    /* Buf not cached, let's do that now */
+
+    if (likely(afl->q_testcase_cache_count == TESTCASE_CACHE_SIZE)) {
```
vanhauser-thc, line 794: for current afl-fuzz this is a bad approach. better: as afl-fuzz selects queue entries from start to finish. the ones should be evicted that have just been completed. The removes the need for `do { rand } while;` and instead you go backwards from where you currently are (or switch to end of queue if begin of queue is reached) to evict one entry

vanhauser-thc:
this also needs a fix for trimming, which would need to release and re-take here:
 * https://github.com/AFLplusplus/AFLplusplus/blob/572944d7267e3612ef8da17a180bc3d8f1a958a7/src/afl-fuzz-run.c#L709
 * https://github.com/AFLplusplus/AFLplusplus/blob/572944d7267e3612ef8da17a180bc3d8f1a958a7/src/afl-fuzz-run.c#L842

vanhauser-thc:
besides - I like the idea. 

for a real benchmark on fuzzbench it needs the trim fixes, and increasing the memcache to 3GB.

and the proposed change algo which item to evict would also slightly increase performance.

I think the current test run on fuzzbench has a big disadvantage as untrimmed test cases are used until they are evicted ...

src/afl-fuzz-queue.c
domenukk: I don't really agree here: most of the time the cache will be used for splicing, and splicing takes a random entry each time, so random eviction is fine. Going backwards also needs additional logic (granted, it's likely as fast or faster than rand, though)

domenukk: The while will still be needed, as the one before may already be in use from fuzzing during the splicing

vanhauser-thc: yes the while is needed. and as I said elsewhere it is a small performance thing. statistically it will be better. but that effort should only be done after it shows on fuzzbench that it helps performance. it should. fuzzbench shows otherwise, but this is because of the trim issue (I guess and hope). and we can increase the cache it will also help.
the env var also is something that would be for after it has proven to be useful and not now for another benchmark

vanhauser-thc:
if you do not have time for the trim fix I can do it on the weekend.

vanhauser-thc:
the trimming is actually not a big thing. because the length is updated afterwards with len = q->len afterwards. so the sole impact is that the testcases are larger in memory than they would need to be. so ... no clue why this does not improve fuzzing.

domenukk:
Trimming can also remove bytes in the middle, this won't work here, correct?
Alternatively I may have introduced some bugs with the buffer pointers

vanhauser-thc:
Trim only removes bytes at the end of the file unless you use a custom mutator trim

vanhauser-thc:
the fuzzbench run https://www.fuzzbench.com/reports/experimental/2020-10-09/index.html shows that the current implementation is a disadvantage. makes no sense to me to make such a difference, so my guess is there is a bug in here.
@domenukk can you please revise the code once you have time for it? afterwards I will review it too.

PR #577, opened 2020-10-07
fix persistent_demo_new compilation with afl-gcc

fuzzah:
this changes __AFL_LOOP macro defined in persistent_demo_new.c to fix compilation when there is no afl-clang/-fast/-lto available


vanhauser-thc:
persistent mode does not work with afl-gcc (the necessary __afl_... functions are in afl-compiler-rt.o).
maybe linking the .o would actually work, never tried.

fuzzah:
@vanhauser-thc you're right, but in persistent_demo_new.c there are defines that allow same code to be compiled with both supported and not supported persistent mode. But it seemed to have a typo or some sort of bad copy-paste error (question mark in the end of redefined __AFL_LOOP macro), which caused fail during compilation with afl-gcc

vanhauser-thc:
I fixed it in dev. your PR does not do as it should be working :)

PR #581, opened 2020-10-13
configurable testcache with malloc

vanhauser-thc:
https://www.fuzzbench.com/reports/experimental/2020-10-13/index.html

"aflplusplus_memcache_inf" is "aflplusplus" with testcache based on malloc. this is what this PR implements.
("aflplusplus_memcache" is based on mmap and has a horrible performance.)

After review/fix/merge this needs updating env_vars.md and README.md


vanhauser-thc:
(dominik did this, I just improved on it from fuzzbench insights as he currently has no time)

src/afl-fuzz-one.c
```diff
@@ -2376,15 +2337,16 @@ u8 fuzz_one_original(afl_state_t *afl) {
 
             } else {
 
+              /* insert mode */
+
               u32 clone_from, clone_to, clone_len;
 
               clone_len = choose_block_len(afl, new_len);
               clone_from = rand_below(afl, new_len - clone_len + 1);
+              clone_to = rand_below(afl, temp_len + 1);
```
vanhauser-thc, line 2330: the +1 here means at end of the data, IMHO this was missing. WDYT?

src/afl-fuzz-one.c
```diff
@@ -453,16 +453,7 @@ u8 fuzz_one_original(afl_state_t *afl) {
...
-  if (unlikely(fd < 0)) {
-
-    PFATAL("Unable to open '%s'", afl->queue_cur->fname);
-
-  }
-
+  orig_in = in_buf = queue_testcase_get(afl, afl->queue_cur);
   len = afl->queue_cur->len;
 
   orig_in = in_buf = mmap(0, len, PROT_READ | PROT_WRITE, MAP_PRIVATE, fd, 0);
```
vanhauser-thc: here are lines that need to be deleted after merge, otherwise it fails. the merge was fucked because of collisions and this one somehow slipped through

domenukk: Why after merge and not now?

src/afl-fuzz-one.c
```diff
@@ -1741,6 +1722,8 @@ u8 fuzz_one_original(afl_state_t *afl) {
               el->afl_custom_fuzz(el->data, out_buf, len, &mutated_buf, new_buf,
                                   target_len, max_seed_size);
 
+          if (new_buf) { new_buf = NULL; }
```
domenukk: The if isn't needed

src/afl-fuzz-queue.c
```diff
@@ -860,3 +861,130 @@ u32 calculate_score(afl_state_t *afl, struct queue_entry *q) {
...
+
+/* Returns the testcase buf from the file behind this queue entry.
+  Increases the refcount. */
+inline u8 *queue_testcase_get(afl_state_t *afl, struct queue_entry *q) {
+
+  u32 len = q->len;
+
+  /* first handle if no testcase cache is configured */
+
+  if (likely(!afl->q_testcase_cache_size)) {
```
domenukk: Why not default to cache? At least 100 megs or so?
And make a min cache size of 2x textcase size or 10x, that should be enough for all targets, gets rid of extra code complexity.

include/afl-fuzz.h
```diff
@@ -686,6 +691,22 @@ typedef struct afl_state {
   /* queue entries ready for splicing count (len > 4) */
   u32 ready_for_splicing_count;
 
+  /* How much current of the testcase cache is used so far */
```
hexcoder-: Comment seems a bit misleading. How about
`/* This is the user specified maximum size to use for the testcase cache. */`
?

src/afl-fuzz-queue.c
```diff
@@ -860,3 +861,130 @@ u32 calculate_score(afl_state_t *afl, struct queue_entry *q) {
...
+  }
+
+  /* now handle testcase cache */
+
+  if (unlikely(!q->testcase_buf)) {
+
+    /* Buf not cached, let's load it */
+    u32 tid = 0;
+
+    while (unlikely(afl->q_testcase_cache_size + len >=
```
domenukk, line 943: Once the cache is full, this is likely. I'd just leave the annotation away completely.

src/afl-fuzz-queue.c
```diff
@@ -860,3 +861,130 @@ u32 calculate_score(afl_state_t *afl, struct queue_entry *q) {
...
+                        afl->q_testcase_max_cache_size ||
+                    afl->q_testcase_cache_count >= TESTCASE_ENTRIES - 1)) {
+
+      /* Cache full. We neet to evict one to map one.
+         Get a random one which is not in use */
+
+      do {
+
+        tid = rand_below(afl, afl->q_testcase_max_cache_count);
```
domenukk, line 952: evicting something constant could be a speedup (last one - 1) as you suggested Marc

src/afl-fuzz-queue.c
```diff
@@ -31,11 +31,12 @@
...
-  u32 s = rand_below(afl, afl->queued_paths);
+  u32    s = rand_below(afl, afl->queued_paths);
   double p = rand_next_percent(afl);
   /*
   fprintf(stderr, "select: p=%f s=%u ... p < prob[s]=%f ? s=%u : alias[%u]=%u"
-  " ==> %u\n", p, s, afl->alias_probability[s], s, s, afl->alias_table[s], p < afl->alias_probability[s] ? s : afl->alias_table[s]);
+  " ==> %u\n", p, s, afl->alias_probability[s], s, s, afl->alias_table[s], p <
+  afl->alias_probability[s] ? s : afl->alias_table[s]);
   */
   return (p < afl->alias_probability[s] ? s : afl->alias_table[s]);
```
domenukk, line 41: unrelated, but how about setting probabilities as 0 - 100000 int instead of doubles?

src/afl-fuzz-run.c
```diff
@@ -709,6 +711,8 @@ u8 trim_case(afl_state_t *afl, struct queue_entry *q, u8 *in_buf) {
 
     });
 
+    if (orig_len != q->len) queue_testcase_retake(afl, q, orig_len);
```
domenukk: personal preference would be {} around if blocks :)

domenukk (approved):

src/afl-fuzz-one.c
vanhauser-thc: Heiko fixed it. Dunno I could not see the lines ...

src/afl-fuzz-one.c
vanhauser-thc: the while line is unneeded as the var is created in the for loop and not accessed anymore after this line

src/afl-fuzz-queue.c
vanhauser-thc: a full cache is not very good for performance. a too small configured cache hurts performance because most of the time something needs to be evicted.. a good configured cache fits all testcases and that should be honored IMHO.

src/afl-fuzz-queue.c
vanhauser-thc: see my other comment. a cache that only fits < 50% of the testcases hurts performance because of the evictions. so a default cache is not a good idea IMHO. and we need code for the option that someone wants that disabled (e.g. fuzzing on embedded that does not have a lot of ram). so thats why I propose to leave the default to no cache. however if none is set a comment about this is printed on startup.

src/afl-fuzz-queue.c
vanhauser-thc: that was a good idea for the original seed selection algo. with the new selection strategy this is not good because this current entry was likely selected because it has a good perf_score, so it is likely that the adjected seeds also have a good perf_score (because new finds vs old finds).
so in the current seed selection a better eviction strategy would be to throw out what has a low perf_score. As I dont know an efficient way to do that I use random. do you have an idea how to do that very fast on the perf_score? I will also think about it ...

src/afl-fuzz-run.c
vanhauser-thc: done

src/afl-fuzz-queue.c
vanhauser-thc: good idea I will look into that

vanhauser-thc: that is quite some complexity to change. it is currently so new I do not want to touch it :) future maybe

src/afl-fuzz-one.c
domenukk: I tried to understand it, but wouldn't that mean if `clone_from` and `clone_to` both result in max, we will write out of bounds?

src/afl-fuzz-queue.c
domenukk: I disagree, but we should maybe test it: Evictions are quasi free, a few extra isns, no runs to the kernel.

src/afl-fuzz-queue.c
domenukk: Right now, we evict in 100% of the times. Any bit of caching helps, as far as I'm concerned. See above comment.
We are single-threaded, so on a good allocator, frees should be cheap, and we can even replace that in the future.

src/afl-fuzz-one.c
vanhauser-thc: no because the tail copy is "temp_len - clone_to" so if the rand results in temp_len, then 0 bytes are copied. there iis already a +1 extra on the realloc which is unnecessary. at least that is how I read the code :)

src/afl-fuzz-queue.c
vanhauser-thc: yeah lets test this on fuzzbench after this PR has landed. one variant with 0MB, one with 1MB and another with 1024MB.

src/afl-fuzz-queue.c
vanhauser-thc: What do you mean by "right now we evict 100% of the time"? if the cache is large enough you never evict, hence "unlikely"

vanhauser-thc:
I think this is done now. ran several testcases for long time and it seems to work

domenukk:
Default cache size 50?

src/afl-fuzz-queue.c
domenukk: Once it is saturated, we evict, hence neither likely nor unlikely

vanhauser-thc: It is such a minimal difference, though I want people with more ram giving to this functionality a little advantage :)

vanhauser-thc:
> Default cache size 50?

lets do first the fuzzbench test run. also to check that nothing breaks.

hexcoder- (commented):
Nice work!

PR #586, opened 2020-10-20
qemu argv index was still wrong

b1gr3db:
Copy and paste error from previous "fix" resulted in wrong index. Should be the last one, not next to last.


andreafioraldi:
https://github.com/AFLplusplus/AFLplusplus/commit/8e809d8593d9230c123aa22c8cd0b695e54d7c68 this commit haunts u @domenukk :) 

Btw I agree, this should be the proper fix. 

In addition, should this line https://github.com/AFLplusplus/AFLplusplus/blob/73c0e1357fa33276ca0a0ee63aacc933aa9ceac5/src/afl-common.c#L145 alloc argc+3 instead?

PR #592, opened 2020-10-28
add gnuplot and afl-dyninst shared libs to Dockerfile

rhtenhove:
I've added `gnuplot-nox` to be able to use `afl-plot` in the container. Adds some overhead, but actually makes the plotter usable.

I've also added two (now failing) instructions to COPY `libdyninstAPI_RT.so` and `libAflDyninst.so`, so any binaries instrumented using afl-dyninst can be fuzzed from the container. 
I see this as common enough to not have the user copy those `.so` files into a mounted dir/container, and having to set `LD_LIBRARY_PATH` for every run.
> Also dyninst seems to add a duplicate dependency of `libdyninstAPI_RT.so` with the full path (i.e. `/usr/local/lib/libdyninstAPI_RT`) to the binary, so it really needs to be in that location as well.

To make this work, there needs to be an afl-dyninst image available, which can be setup next to the build of the AFL++ container. I've set it to aflplusplus/afl-dyninst as that seemed a proper name for it.

Also added a bit to the README.md to suggest using tmpfs for input files to lighten the load for local storage devices.

And finally made some minor tweaks to the Dockerfile to clean it up. Commit messages should be self explanatory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,10 +26,10 @@ RUN apt-get update && \
     gnuplot-nox \
     && rm -rf /var/lib/apt/lists/*
 
-RUN echo deb http://apt.llvm.org/focal/ llvm-toolchain-focal-11 main >> /etc/apt/sources.list && \
-    wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - 
-  
-RUN echo deb http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu focal main >> /etc/apt/sources.list && \
+RUN echo "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-11 main" >> /etc/apt/sources.list && \
+    wget -qO - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+
+RUN echo "deb http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu focal main" >> /etc/apt/sources.list && \
     apt-key adv --recv-keys --keyserver keyserver.ubuntu.com 1E9377A2BA9EF27F
 
 RUN apt-get update && apt-get full-upgrade -y && \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -62,3 +61,6 @@ RUN export REAL_CXX=g++-10 && export CC=gcc-10 && \
 RUN echo 'alias joe="jupp --wordwrap"' >> ~/.bashrc
 RUN echo 'export PS1="[afl++]$PS1"' >> ~/.bashrc
 ENV IS_DOCKER="1"
+
+COPY --from=aflplusplus/afl-dyninst /usr/local/lib/libdyninstAPI_RT.so /usr/local/lib/libdyninstAPI_RT.so
+COPY --from=aflplusplus/afl-dyninst /afl-dyninst/libAflDyninst.so /usr/local/lib/libAflDyninst.so
```
vanhauser-thc, line 66: Does this actually work? Wont this pull another container? I am not a docker expert ...

rhtenhove: See [here](https://docs.docker.com/engine/reference/builder/#copy) for how `--from` works. During the build it indeed pulls that other image, but after the 2 files are added it is no longer required and only the 2 files become part of the build.
So anyone pulling `aflplusplus/aflplusplus` won't have to pull, in this case, `aflplusplus/afl-dyninst`.

vanhauser-thc: OK. in this case please add copying afl-dyninst plus the two .sh helper scripts, otherwise people cannot use it.

rhtenhove: You propose to add afl-dyninst entirely into the container? Adding the two `.so` files should be enough to run any binaries instrumented by afl-dyninst. In that way the two images clearly have a different purpose, and afl++ never needs afl-dyninst during runtime as far as I can tell, right?

vanhauser-thc: my thinking is that people then do not need an own afl-dyninst container anymore, but that thinking was flawed as lots of other dyninst libraries are then required as well.

vanhauser-thc:
LGTM, thanks!

Dockerfile
rhtenhove: Awesome! Still, to have this succeed we need the `aflplusplus/afl-dyninst` image to be built from https://github.com/vanhauser-thc/afl-dyninst

I'd add it as an issue there, but issues are disabled in that repo.

Issue #562, opened 2020-09-18
unable to create '(null)'


ioxera:
When i trying to run "afl-cmin -i ./in -o ./out -- ../readelf -a @@", I encountered the following error：
![图片](https://user-images.githubusercontent.com/22150551/93594122-d684a400-f9e7-11ea-9a4f-e9a1f46fd02a.png)
but when i trying run "afl-cmin -i ./in -o ./out -f test -- ../readelf -a", The program has no errors, but it will prompt "[!] warning: all test cases, had the same traces, check syntax!" every time, and only one file will be obtained.
I don't know why the use of “afl_cmin" is different from that described in README.

ioxera:
I had the same problem when I tried "afl_showmap".

ButcherAdaCore:
I'm also seeing this issue when trying to run both afl-cmin and afl-showmap.

domenukk:
What versions and what compilation mode does this occur on?

vanhauser-thc:
Use the dev branch, this should be fixed already.

ioxera:
@vanhauser-thc yes，I saw it #524 ，But I still had this problem when I tried to use the dev version，I just pulled the update and tried again

vanhauser-thc:
The error message is different then though :)
Dunno why this should happen for you. Maybe limits or a filesystem thing

ioxera:
@vanhauser-thc OK，I use archlinux and Btrfs file systems, I will try other file systems, thank you for your answer

vanhauser-thc:
if switching to a different filesystem works please let us know.

mykter:
I'm getting the same error with afl-tmin, on the dev branch.
```
afl-tmin++3.00a by Michal Zalewski

[+] Read 36 bytes from 'out/crashes/id:000000,sig:06,src:001563,time:1000380,op:havoc,rep:2'.
[*] Spinning up the fork server...
[+] All right - fork server is up.
[*] Target map size: 44496
[*] Performing dry run (mem limit = 0 MB, timeout = 1000 ms)...

[-]  SYSTEM ERROR : Unable to create '(null)'
    Stop location : afl_fsrv_write_to_testcase(), src/afl-forkserver.c:984
       OS message : Bad address
```

vanhauser-thc:
@mykter what is the filesystem the -o out directory is on?

mykter:
aufs. I switched Docker to use overlay2, and no longer get the error.

[...]

PR #594, opened 2020-10-29
push to stable

vanhauser-thc:
This is the first push to stable for 3.0
There were a lot of changes with afl-cc how afl-fuzz works, etc and I tried to document these changes visibly in our main README.md at the top.
I expect this will flush up several issues we still might have, which is important before we do an official release.
I would propose aiming for a 1st December to 11th January for the release so that there is enough testing in between.

Not sure how much sense it makes to review this PR as it is ... huuuuuge. but well. best effort. :)


src/afl-fuzz-extras.c
```diff
@@ -370,16 +387,74 @@ static inline u8 memcmp_nocase(u8 *m1, u8 *m2, u32 len) {
...
+        goto restart_dedup;  // restart if several duplicates are in a row
+
+      }
+
+    }
+
+  }
+
+  if (afl->extras_cnt != orig_cnt)
+    afl->extras = afl_realloc((void **)&afl->extras,
```
domenukk: Note that `afl_realloc` will never shrink the buffer size. This call has no effect.
(The idea behind `afl_realloc` is, that we reuse buffers as much as possible only reallocate if we really need to -> if the new buffer is larger than what we saw to this day).
Even if extras could grow in this function, the `!=` check would be redundant as it's done in realloc, too.

src/afl-fuzz-extras.c
```diff
@@ -370,16 +387,74 @@ static inline u8 memcmp_nocase(u8 *m1, u8 *m2, u32 len) {
...
+      if (found) break;
+
+    }
+
+  }
 
   if (len > MAX_DICT_FILE) {
 
-    WARNF("Extra '%.*s' is very big (%s, limit is %s)", (int)len, mem,
+    WARNF("Extra '%.*s' is too big (%s, limit is %s)", (int)len, mem,
```
domenukk: Maybe add a "... skipping" note?

src/afl-fuzz-init.c
```diff
@@ -697,11 +713,9 @@ void read_testcases(afl_state_t *afl) {
 
     if (st.st_size > MAX_FILE) {
 
-      WARNF("Test case '%s' is too big (%s, limit is %s), skipping", fn2,
+      WARNF("Test case '%s' is too big (%s, limit is %s), partial reading", fn2,
```
domenukk, line 716: Did we ever check if we have issues with sharedmap input, mutations, splicing, ... when testcases are `MAX_FILE` sized?

src/afl-fuzz-init.c
```diff
@@ -985,6 +1035,83 @@ void perform_dry_run(afl_state_t *afl) {
...
+
+  q = afl->queue;
+  struct queue_entry *p, *prev = NULL;
+  int                 duplicates = 0;
+
+restart_outer_cull_loop:
+
+  while (q) {
+
+    if (q->cal_failed || !q->exec_cksum) { goto next_entry; }
```
domenukk, line 1048: so many gotos... hard to follow and may be problematic for the compiler

domenukk (approved):
Oh my...
Quite a lot of code :)
Seems to work, so let's do it?

src/afl-fuzz-extras.c
vanhauser-thc: uh true. so what is the correct way?

src/afl-fuzz-extras.c
vanhauser-thc: done

src/afl-fuzz-init.c
vanhauser-thc: two gotos, and IMHO easy to read. also works like a charm. not using goto will make this much more complex.

src/afl-fuzz-init.c
vanhauser-thc: me not explicitly. I know some fuzzbench inputs are larger than that.
but this needs a target that actually uses that amount - otherwise it will be trimmed ...
we could test with perl/php/json targets or pcap

src/afl-fuzz-extras.c
domenukk: I would just not resize it here, we may add extras later and need to re-resize, right? Alternatively, switch to normal malloc. I can also add an `afl_realloc_shrinking` if you want?

vanhauser-thc: or afl_realloc_exact ... yes I think that would be a good thing to have

docs/FAQ.md
```diff
@@ -182,31 +180,31 @@ afl-clang-fast PCGUARD and afl-clang-lto LTO instrumentation.
...
-     skip for instrumentation. Note that optimization might inline functions!
+     skip instrumenting. Note that optimization might inline functions!
 
-     Simply follow this document on how to do this: [llvm_mode/README.instrument_list.md](llvm_mode/README.instrument_list.md)
+     Simply follow this document on how to do this: [instrumentation/README.instrument_list.md](instrumentation/README.instrument_list.md)
      If PCGUARD is used, then you need to follow this guide (needs llvm 12+!):
      [http://clang.llvm.org/docs/SanitizerCoverage.html#partially-disabling-instrumentation](http://clang.llvm.org/docs/SanitizerCoverage.html#partially-disabling-instrumentation)
 
-     Only exclude those functions from instrumentation that provide no value
+     Only deny those functions from instrumentation that provide no value
```
hexcoder-: I don't understand why all of my editing was thrown away. Merge error?

vanhauser-thc: me neither ... dunno why that happened

src/afl-fuzz-extras.c
vanhauser-thc: -> a2739ef ?

vanhauser-thc: plus ...  b33306c

docs/FAQ.md
hexcoder-: Can you resurrect my editing, please?

vanhauser-thc: no time, sorry, two customer projects

vanhauser-thc:
IMHO @andreafioraldi needs to document the new qemu environment variables before we can merge into stable

docs/FAQ.md
vanhauser-thc: done

vanhauser-thc:
dont merge yet, still testing if a fix is working

vanhauser-thc:
and we should fix #562 ... @domenukk / @hexcoder-  ?

hexcoder-:
> and we should fix #562 ... @domenukk / @hexcoder- ?

ok, i will have a look.

vanhauser-thc:
@hexcoder- @domenukk @andreafioraldi IMHO we can merge now?

andreafioraldi:
Still debugging the stability issue of the persistent hook. But IMO yes, it is a minor feature, when i'll figure out it we can do another PR

andreafioraldi (approved):

hexcoder- (approved):
Nice work!

vanhauser-thc:
#597 - if this is really a bug on our side - affects both stable and dev (@andreafioraldi ), so this is not a blocker
#562 is IMHO fixed but I doubt 2 of the 3 reporters will retest
the other issues do not apply or are fixed IMHO.
so I am merging ...

PR #601, opened 2020-11-11
Fixing symcc custom mutator

Rumata888:
Changed symcc custom mutator to work both with stdin and file-input targets and fixed several errors in the mutator. Added two example targets for easy tests in the future. Added printing fo enqueued items (otherwise afl seems stuck during initial processing of huge input corpus). Fixed the problem where symcc overwrote previous testcases.


custom_mutators/symcc/symcc.c
```diff
@@ -95,40 +98,66 @@ void afl_custom_queue_new_entry(my_mutator_t * data,
                                 const uint8_t *filename_new_queue,
                                 const uint8_t *filename_orig_queue) {
 
+  int pipefd[2];
+  struct stat st;
+  ACTF("Queueing to symcc: %s", filename_new_queue);
+  u8 *fn = alloc_printf("%s", filename_new_queue);
+  if (!(stat(fn, &st) == 0 && S_ISREG(st.st_mode) && st.st_size)) {
+    PFATAL("Couldn't find enqueued file: %s",fn);
```
hexcoder-: mem leak for `fn`

custom_mutators/symcc/symcc.c
```diff
@@ -95,40 +98,66 @@ void afl_custom_queue_new_entry(my_mutator_t * data,
...
+  }
+   
+  if (afl_struct->fsrv.use_stdin){
+    if (pipe(pipefd)==-1)
+    {
+      exit(-1);
```
hexcoder-: better use PFATAL and tell why we have to abort.

custom_mutators/symcc/symcc.c
```diff
@@ -95,40 +98,66 @@ void afl_custom_queue_new_entry(my_mutator_t * data,
...
         if (r <= 0) return;
-        close(0);
-        ck_write(0, data->mutator_buf, r, fn);
-        ck_free(fn);
-
+        close(fd);
+        if (r>fcntl(pipefd[1],F_GETPIPE_SZ)) fcntl(pipefd[1],F_SETPIPE_SZ,MAX_FILE);
+        ck_write(pipefd[1], data->mutator_buf, r, filename_new_queue);
+      } else {
+        PFATAL("Something happened to the enqueued file before sending its contents to symcc binary");
```
hexcoder-: mem leak for `fn`

hexcoder-:
Thanks for your PR!

Rumata888:
Ok, I'll fix the memleaks and resubmit it. Thanks for auditing it.

Rumata888:
Added the fixes.

vanhauser-thc:
IMHO this looks fine. @hexcoder- ?

custom_mutators/symcc/symcc.c
```diff
@@ -135,6 +141,7 @@ void afl_custom_queue_new_entry(my_mutator_t * data,
 
         ssize_t r = read(fd, data->mutator_buf, MAX_FILE);
         DBG("fn=%s, fd=%d, size=%ld\n", fn, fd, r);
+        ck_free(fn);
         if (r <= 0) return;
         close(fd);
```
hexcoder-, line 145: better move close(fd) one line up before the return to avoid leaking fd.

vanhauser-thc: good catch!

hexcoder-: If we can get this fix as well, I am happy.

Rumata888:
Ok. Do you notice anything else or is that it?

Rumata888:
Updated it (also closed pipe before return and PFATAL)

vanhauser-thc:
thanks!
the pipe close before the fatal is unnecessary as fatal exits and then closes all file descriptors anyway, but it doesnt hurt :)

PR #604, opened 2020-11-15
Actually make python 'fuzz' method optional

virtuald:
The documentation states that the fuzz method is optional, but it crashes if you don't specify one.

Perhaps originally `mutator->afl_custom_fuzz` was not allowed to be NULL? However, that doesn't seem to be the case anymore. Thanks for all your hard work on this!


vanhauser-thc:
correct, it is optional, thanks for fixing!

PR #605, opened 2020-11-15
Use buffer protocol to retrieve result from python post_process

virtuald:
Saves an extra copy, gives post processing functions more flexibility. Haven't benchmarked it.

This could be added to the other python functions, but I'm only using the post processor at the moment so I don't want to implement something I can't easily test. :)

There's also some potential to pass in a memoryview or reusable buffer object if the custom mutator API is changed.

### DIFF
--- a/docs/custom_mutators.md
+++ b/docs/custom_mutators.md
@@ -130,6 +130,9 @@ def introspection():
     `post_process` function. This function is then transforming the data into the
     format expected by the API before executing the target.
 
+    This can return any python object that implements the buffer protocol and
+    supports PyBUF_SIMPLE. These include bytes, bytearray, etc.
+
 - `queue_new_entry` (optional):
 
     This methods is called after adding a new test case to the queue.
--- a/include/afl-fuzz.h
+++ b/include/afl-fuzz.h
@@ -325,8 +325,7 @@ typedef struct py_mutator {
   u8 *   fuzz_buf;
   size_t fuzz_size;
 
-  u8 *   post_process_buf;
-  size_t post_process_size;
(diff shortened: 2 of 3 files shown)


vanhauser-thc:
IMHO this looks good by I am not a python guy. @domenukk @andreafioraldi ?

domenukk:
@virtuald thaks for the pr! I think we are currently at a position (moving to 3.0) where changing the API wolud be possible. That being said, the current design may be easier to use from the python side, what do you think?
I guess I should also adapt all other functions, if possible?

domenukk:
Well, merging for now. Feel free to still answer @virtuald

virtuald:
@domenukk yes, it would be good to adapt the other functions in the same way. I just don't have anything setup to test it, so I didn't want to do those changes myself.

I'm still contemplating the right way to make these operations more efficient. `PyMemoryView_FromMemory` is what one would use to wrap an C array of bytes, but I suspect we'd want a useful way to reallocate the buffer and update the view. Using a  `io.BytesIO` object instead could be an alternative since you can get access to the underlying buffer. But I guess there are already some copies being done elsewhere, so I'm not sure how big of a difference it would make -- I suppose though that copying into the same buffer over and over again is faster than an allocation and a copy?

domenukk:
@virtuald the way custom mutators work in C right now is: we hand in a buffer, the mutator can then choose to return that pointer (and alter data in-place) or return a new buffer. Not sure if python would allow something similar.
However, if you really need that sort of perf, maybe python isn't the way to go in the first place.
It's more important to have the API convenient to use than squeeze every inch of performance out of it.

PR #616, opened 2020-12-02
Enable afl-cc to find helper obj from argv[0]

ButcherAdaCore:
Installed wrapper scripts in BIN_PATH, that re-execute the programs with a resolved path in HELPER_PATH, so that they can then use argv[0]'s dirname to find helper files.

This allows the afl-cc script installed in BIN_PATH to be replaced with a symlink to $(BIN2HELPER)/afl-cc to avoid the tiny overhead of the wrapper, at the expense of having to set AFL_PATH to HELPER_PATH in case the install tree is relocated.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -22,6 +22,7 @@ HASH=\#
 PREFIX     ?= /usr/local
 BIN_PATH    = $(PREFIX)/bin
 HELPER_PATH = $(PREFIX)/lib/afl
+BIN2HELPER  = ../lib/afl
 DOC_PATH    = $(PREFIX)/share/doc/afl
 MISC_PATH   = $(PREFIX)/share/afl
 MAN_PATH    = $(PREFIX)/man/man8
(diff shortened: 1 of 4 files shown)


vanhauser-thc:
I wanted to accept this and then thought about that this could break build systems that do not allow the compiler to be a shell script.

In the end I rewrote our function to find the runtime libraries to be very, very complete: 1890d7b9cf5a32368c0700ef790f97087b948e1f

It does not cover OS X, Solaris and FreeBSD without procfs for the exepath information but otherwise goes way further than what is in this PR.

WDYT?

hexcoder-:
@vanhauser-thc: I fixed the proc search in your commit. It would not compile for FreeBSD and DragonFly and would not work for Linux, Android and NetBSD, see https://github.com/AFLplusplus/AFLplusplus/commit/8f79116a15e53f0ecd4889ebf8980cc0cd85a494

vanhauser-thc:
if the afl-cmin/afl-cmin.bash changes dont work ping me.

PR #617, opened 2020-12-02
push to stable

vanhauser-thc:
Might be the last major push to stable before a 3.0 release :)


instrumentation/afl-compiler-rt.o.c
```diff
@@ -793,6 +810,7 @@ static void __afl_start_forkserver(void) {
         //(void)nice(-20);
 
         signal(SIGCHLD, old_sigchld_handler);
+        signal(SIGTERM, old_sigterm_handler);
```
vanhauser-thc, line 833: this is something I am not happy with. an extra signal handler change after every fork. but I have no clue what a better solution could be. ideas?

afl-cmin
```diff
@@ -113,14 +113,15 @@ function usage() {
...
-"For additional tips, please consult docs/README.md\n" \
+"For additional tips, please consult README.md\n" \
 "\n" \
 "Environment variables used:\n" \
+"AFL_ALLOW_TMP: allow unsafe use of input/output directories under {/var}/tmp\n" \
+"AFL_CRASH_EXITCODE: optional child exit code to be interpreted as crash\n" \
+"AFL_FORKSRV_INIT_TMOUT: time the fuzzer waits for the target to come up, initially\n" \
 "AFL_KEEP_TRACES: leave the temporary <out_dir>/.traces directory\n" \
 "AFL_PATH: path for the afl-showmap binary\n" \
 "AFL_SKIP_BIN_CHECK: skip check for target binary\n" \
```
hexcoder-: last line should NOT end with `\`, will fix it myself.

include/debug.h
```diff
@@ -270,6 +270,16 @@
                            \
   } while (0)
 
+/* Show a prefixed debug output. */
+
+#define DEBUGF(x...)                         \
+  do {                                       \
+                                             \
```
hexcoder-, line 277: we could simply the afl code if we put the condition ( whether to output something ) once in the macro:
```
 if (debug) {
 ...
 }
```
or some such.

instrumentation/README.lto.md
```diff
@@ -73,11 +78,11 @@ then add the pgp key of llvm and install the packages:
...
 wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - 
 apt-get update && apt-get upgrade -y
-apt-get install -y clang-11 clang-tools-11 libc++1-11 libc++-11-dev \
-    libc++abi1-11 libc++abi-11-dev libclang1-11 libclang-11-dev \
-    libclang-common-11-dev libclang-cpp11 libclang-cpp11-dev liblld-11 \
-    liblld-11-dev liblldb-11 liblldb-11-dev libllvm11 libomp-11-dev \
-    libomp5-11 lld-11 lldb-11 llvm-11 llvm-11-dev llvm-11-runtime llvm-11-tools
+apt-get install -y clang-12 clang-tools-12 libc++1-12 libc++-12-dev \
+    libc++abi1-12 libc++abi-12-dev libclang1-12 libclang-12-dev \
+    libclang-common-12-dev libclang-cpp11 libclang-cpp11-dev liblld-12 \
```
hexcoder-: how about `libclang-cpp11 libclang-cpp11-dev libllvm11`?

src/afl-cc.c
```diff
@@ -48,6 +48,7 @@
 static u8 * obj_path;                  /* Path to runtime libraries         */
 static u8 **cc_params;                 /* Parameters passed to the real CC  */
 static u32  cc_par_cnt = 1;            /* Param count, including argv0      */
+static u8   clang_mode;                /* Invoked as afl-clang*?            */
```
hexcoder-, line 51: now should be `bool` probably

src/afl-fuzz-init.c
```diff
@@ -954,7 +966,18 @@ void perform_dry_run(afl_state_t *afl) {
 #undef MSG_ULIMIT_USAGE
 #undef MSG_FORK_ON_APPLE
 
-        WARNF("Test case '%s' results in a crash, skipping", fn);
+        if (afl->fsrv.uses_crash_exitcode) {
+
+          WARNF(
+              "Test case '%s' results in a crash or AFL_CRASH_EXITCODE %d, "
+              "skipping",
```
hexcoder-, line 973: shouldn't that say 

> kept for splicing

instead of

> skipping

like in line 875? Different behavior should give different texts?

src/afl-showmap.c
```diff
@@ -1090,6 +1092,23 @@ int main(int argc, char **argv_orig, char **envp) {
 
     }
 
+    if (getenv("AFL_CRASH_EXITCODE")) {
```
hexcoder-, line 1097: I thought about the range of the exit codes a bit. In general the exit code has type `int`and apps can use that range. The old interface functions `wait()` and `waitpid()` are limited to 8 bits process exit codes, but the newer standard function `waitid()` uses `siginfo_t` which has a `si_status` field with the complete exit code (size `int`).
[waitid function opengroup](https://pubs.opengroup.org/onlinepubs/9699919799/functions/waitid.html)
[siginfo_t structure opengroup](https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_13)

So sometime after the next release we could switch to `waitid()` and support the full range of exit codes.

test/test-pre.sh
```diff
@@ -106,14 +106,7 @@ export AFL_LLVM_INSTRUMENT=AFL
...
   export PATH="/usr/local/bin:${PATH}"
 }
-# on MacOS X we prefer afl-clang over afl-gcc, because
-# afl-gcc does not work there
-test `uname -s` = 'Darwin' -o `uname -s` = 'FreeBSD' && {
-  AFL_GCC=afl-clang
-} || {
-  AFL_GCC=afl-gcc
-}
-command -v gcc >/dev/null 2>&1 || AFL_GCC=afl-clang
```
hexcoder-: with the reintroduction of afl-clang I will revert this change.

instrumentation/afl-compiler-rt.o.c
hexcoder-: If there was no way to avoid signal(SIGCHLD) handling here, I doubt there is one for signal(SIGTERM) handling...

hexcoder- (approved):
+1 for introducing the `bool` type!

hexcoder- (approved):

README.md
```diff
@@ -55,6 +55,7 @@ behaviours and defaults:
     * a caching of testcases can now be performed and can be modified by
       editing config.h for TESTCASE_CACHE or by specifying the env variable
       `AFL_TESTCACHE_SIZE` (in MB). Good values are between 50-500 (default: 50).
+  * utils/ got renamed to utils/
```
domenukk: This one got renamed once too often

src/afl-showmap.c
domenukk: Oh nice, didn't even know about the `waitid` API. I think the use-case is obscure enough that we don't have to spend too much time on it. There are still enough returncodes to chose from

src/afl-cc.c
```diff
@@ -113,15 +113,15 @@ u8 *getthecwd() {
...
     a) we also check ../lib/afl
   4. if 3. failed we check /proc (only Linux, Android, NetBSD, DragonFly, and
-     FreeBSD with procfs
+     FreeBSD with procfs)
     a) and check here in ../lib/afl too
   5. we look into the AFL_PATH define (usually /usr/local/lib/afl)
   6. we finally try the current directory
 
-  if this all fail - we fail.
+  if all this fails - we fail.
```
domenukk: This sentence still looks strange. "all of this fails"?

README.md
vanhauser-thc: fixed :)

src/afl-cc.c
vanhauser-thc: fixed

instrumentation/README.lto.md
vanhauser-thc: good catch! fixed

src/afl-cc.c
vanhauser-thc: yes. many could be bool. but IMHO it is not really worth (at least my) time to check them all and change them.

src/afl-fuzz-init.c
vanhauser-thc: arent both kept for splicing?

src/afl-cc.c
domenukk: I mainly introduced bool to mark the distinction between u8 and flags in the afl state. I don't think we have to catch them all. When you spot a bool u8 and work on that code anyway, why not change it, but don't spend more time on it than necessary

vanhauser-thc:
I am merging now so we have a few more "testers" who clone from stable :)

PR #623, opened 2020-12-07
Fix Grammar Mutator Submodule

rchildre3:
Fixes: 651ad18e2179 ("added the grammar mutator as a git submodule ...")

* Project pointer never pushed
* Reduces dirctory complexity
* Building and dependencies for the subproject should be isolated to the
  subproject's documentation
* Fix broken link to README
* Use `--init` for `git submodule`


vanhauser-thc:
thanks for the PR!
The enhancements for the git stuff is good.
However script download + install is a must as an alternative. some people get a .tar.get and dont have git on a target system hence both mechanisms must exist and work.

domenukk:
We should probably replicate the way unicorn and qemu-mode are being built right now? I'll see if I can do this easily. Accepting the PR now, and going from there, thanks

domenukk:
Please have a look at eda0687

rchildre3:
@vanhauser-thc 
> dont have git on a target system

Interesting design requirement. Should this be documented somewhere like in `CONTRIBUTING.md`?

@domenukk 
eda0687 doesn't remove the need for `git`. Shouldn't the scripts download and unpack a tarball of the source to comply with the guidance that `git` not be required to build

domenukk:
The requirement is rather "didn't clone this repo from git".
We want to be able to distribute it as tarball, or check it out via svn.
Having `git` installed for the actual build is fine imho.

vanhauser-thc:
yes thank you domenukk - that is what I meant :)

PR #624, opened 2020-12-07
Sampling next seed by weight (hit_count, bitmap_size, exec_us)

mboehme:
This commit extends the weight-based sampling by assigning weights based on how often the seed's path is exercised, the number of branches it covers, and the time it takes to execute.

### DIFF
--- a/src/afl-fuzz-one.c
+++ b/src/afl-fuzz-one.c
@@ -445,8 +445,10 @@ u8 fuzz_one_original(afl_state_t *afl) {
 
   if (unlikely(afl->not_on_tty)) {
 
-    ACTF("Fuzzing test case #%u (%u total, %llu uniq crashes found)...",
-         afl->current_entry, afl->queued_paths, afl->unique_crashes);
+    ACTF("Fuzzing test case #%u (%u total, %llu uniq crashes found, perf_score=%0.0f, exec_us=%llu, hits=%u, map=%u)...",
+         afl->current_entry, afl->queued_paths, afl->unique_crashes,
+         afl->queue_cur->perf_score, afl->queue_cur->exec_us,
+         afl->n_fuzz[afl->queue_cur->n_fuzz_entry], afl->queue_cur->bitmap_size);
     fflush(stdout);
 
   }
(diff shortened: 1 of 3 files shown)


src/afl-fuzz-one.c
```diff
@@ -445,8 +445,10 @@ u8 fuzz_one_original(afl_state_t *afl) {
 
   if (unlikely(afl->not_on_tty)) {
 
-    ACTF("Fuzzing test case #%u (%u total, %llu uniq crashes found)...",
-         afl->current_entry, afl->queued_paths, afl->unique_crashes);
+    ACTF("Fuzzing test case #%u (%u total, %llu uniq crashes found, perf_score=%0.0f, exec_us=%llu, hits=%u, map=%u)...",
+         afl->current_entry, afl->queued_paths, afl->unique_crashes,
+         afl->queue_cur->perf_score, afl->queue_cur->exec_us,
+         afl->n_fuzz[afl->queue_cur->n_fuzz_entry], afl->queue_cur->bitmap_size);
```
vanhauser-thc, line 451: n_fuzz is NULL if it is not a aflfast schedule, so this would crash

src/afl-fuzz-queue.c
```diff
@@ -65,25 +80,35 @@ void create_alias_table(afl_state_t *afl) {
...
-
-    sum += q->perf_score;
+    avg_exec_us += q->exec_us;
+    avg_bitmap_size += log(q->bitmap_size);
 
   }
+  avg_exec_us /= afl->queued_paths;
+  avg_bitmap_size /= afl->queued_paths;
 
+  double sum = 0;
```
vanhauser-thc, line 95: here you modify the creation of the weightings in the alias table for all schedules, not only the aflfast schedules.

for explore/exploit I don't mind if this increases its performance (does it? have you tested that?).

however for the RARE and SEEK schedule they explicitly should not have exec timings in their weighting.

vanhauser-thc:
thanks!
please see my comments. :)

vanhauser-thc:
I am a bit under pressure with my time today, so I accept this and fix it by hand, please have a look afterwards

PR #629, opened 2020-12-13
Cleanup QEMU configuration options

rchildre3:
Take into account `STATIC`, `DEBUG`, and `PROFILING`

Turn off most options

Remove non-exsistant config options (e.g. bluez)


andreafioraldi:
LGTM, thanks for the contribution!

PR #637, opened 2020-12-16
minor push to stable

vanhauser-thc:
just a few fixes


domenukk (approved):

domenukk:
Sooo... Looking at the ci output, is Travis really no longer working? We should add github actions or some other new ci, right?

vanhauser-thc:
yes we should switch to github actions. I can have a look at that.

I think this PR can now be reviewed and tested - can some of you test that is compiles for 32 bit, FreeBSD and ARM64?

PR #647, opened 2020-12-20
Remove binary-only target's depend on source tools

rchildre3:
The `binary-only` target currently depends on the `all` target which
always build the source tools, such as afl-cc and afl-as.  This is
unnecessary if the user specifically is asking for only binary fuzzing
tools.


domenukk:
That makes sense in my eyes, but I'll let @vanhauser-thc decide. In any case, thanks! :)

vanhauser-thc:
you are right!
can you please add "test_shm test_python" to the binary-only target before PROGS?
thank you

vanhauser-thc:
thank you!

PR #665, opened 2020-12-29
Handle disabled seeds properly in distribution-based schedule

mboehme:
Set weight and perf_score to 0 when seed is disabled.


hexcoder-:
Good spotting, that makes sense, thanks a alot!

PR #666, opened 2020-12-29
push to stable

vanhauser-thc:


hexcoder- (approved):
Look good to me

GNUmakefile
```diff
@@ -606,7 +606,7 @@ source-only: all
 	-$(MAKE) -f GNUmakefile.gcc_plugin
 	$(MAKE) -C utils/libdislocator
 	$(MAKE) -C utils/libtokencap
-	$(MAKE) -C utils/aflpp_driver
+	-$(MAKE) -C utils/aflpp_driver
```
domenukk, line 609: For me as makefile noob: what does the `-` do?

include/afl-fuzz.h
```diff
@@ -734,6 +739,7 @@ typedef struct afl_state {
   char  mutation[8072];
   char  m_tmp[4096];
   FILE *introspection_file;
+  u32   bitsmap_size;
```
domenukk, line 742: `bitmap`, without the s?

src/afl-fuzz-queue.c
```diff
@@ -45,25 +45,19 @@ inline u32 select_next_queue_entry(afl_state_t *afl) {
 double compute_weight(afl_state_t *afl, struct queue_entry *q,
                       double avg_exec_us, double avg_bitmap_size) {
 
-  u32 hits;
+  double weight = 1.0;
```
domenukk, line 48: Maybe `float` is enough here? I always expect doubles to be super slow. (Gut feeling though, didn't benchmark it)

GNUmakefile
hexcoder-: Then `make` ignores the exit code of this command. This is needed when I run `make tests` for a non supported platform for utilos/aflpp_driver. Otherwise `make tests` would stop here with an error and won't reach the tests, that i am interested in.

hexcoder-:
> @andreafioraldi SHAPE_BYTES does (x + 1), so that commented out evaluation is always true.
Unless there is unsigned overflow...

> was that an oversight or should that be a 2, ... ?
The case for 2 is already there (see some lines above)

include/afl-fuzz.h
vanhauser-thc: yes, that is for counting the bits before classifying them. just for introspection compile

src/afl-fuzz-queue.c
vanhauser-thc: is it? we should check in godbolt

PR #670, opened 2020-12-31
libdislocator, ptr algo warning fix on clang/Xcode (m1).

devnexen:


vanhauser-thc:
LGTM :)

btw - on the mailing list someone was asking about IlluminOS or something like that. does afl++ compile there as well? do you know?

PR #522, opened 2020-08-22
Option for specifying forkserver initialization timeout via environment variable

mariusmue:
This PR introduces a new environment variable which allows to specify the timespan AFL should wait for initial contact with the
forkserver. (Currently, this is set to `timeout * FORK_WAIT_MULT`).
    
This addition is useful for fuzz-targets requiring a rather long setup time before the actual fuzzing starts (e.g., unicorn).

Additionally, I included vim-generated *.swp files to the .gitignore.

Cheers,
Marius

PR #678, opened 2021-01-07
User defined kill signal value

v-p-b:
This adds support for the AFL_KILL_SIGNAL environment variable discussed here:

https://github.com/intel/kernel-fuzzer-for-xen-project/pull/24#issuecomment-755271472

I need some guidance regarding:

* The additions made to support hang detection when running with KF/x, see:
https://github.com/AFLplusplus/AFLplusplus/compare/stable...v-p-b:env_kill?expand=1#diff-2d26035103ef709cb5d8fe28c0ed22a9554333e1dedf3f08dff919038488ec11R1146
https://github.com/intel/kernel-fuzzer-for-xen-project/pull/24#issuecomment-755711776
* Whether the environment variable is handled the right way (`getenv()` vs. `get_afl_env()`, `fsrv` struct extension...)

I will merge the commits from bleeding edge (this is currently forked from 3.0c), and squash my modifications if requested.


vanhauser-thc:
Hi! Thanks for the PR.
we only accept PRs to dev, never to stable, so that stable never has breaking code. I therefore switched the PR to point to dev.
I will now review the changes and make comments if necessary.

src/afl-forkserver.c
```diff
@@ -95,6 +95,10 @@ void afl_fsrv_init(afl_forkserver_t *fsrv) {
   fsrv->uses_asan = false;
 
   fsrv->init_child_func = fsrv_exec_child;
+  fsrv->kill_signal = SIGKILL;
+  if (get_afl_env("AFL_KILL_SIGNAL")){
+    fsrv->kill_signal = atoi(get_afl_env("AFL_KILL_SIGNAL"));
+  }
```
vanhauser-thc, line 120: this needs error handling in the case that e.g. ` AFL_KILL_SIGNAL=foobar`  or an invalid signal number. also people are used to use the signal name instead of the signal number.

src/afl-fuzz-state.c
```diff
@@ -507,8 +507,8 @@ void afl_states_stop(void) {
 
   LIST_FOREACH(&afl_states, afl_state_t, {
 
-    if (el->fsrv.child_pid > 0) kill(el->fsrv.child_pid, SIGKILL);
-    if (el->fsrv.fsrv_pid > 0) kill(el->fsrv.fsrv_pid, SIGKILL);
+    if (el->fsrv.child_pid > 0) kill(el->fsrv.child_pid, el->fsrv.kill_signal);
```
vanhauser-thc, line 534: I think it needs a second variable `AFL_DO_NOT_KILL_CHILD` or something like that? I think I remember @tklengyel writing that the child may not be killed for Xen mode.

src/afl-forkserver.c
vanhauser-thc: I recommend to reset the env var with an ensured correct value here so the same code is not needed in `at_exit()`

domenukk: Using a number was my initial idea, is there any way to get a mapping of string -> id? Personally, I'm rather used to doing kill 9

domenukk: Answering my own question, this is how kill does it:
https://github.com/coreutils/coreutils/blob/00ea4bacf6063ccc125209d5186f8f2382c6f0d4/src/operand2sig.c#L36
But I would argue users can look up the table instead

domenukk:
WRT adding new env variables, it's a bit more involved, see https://github.com/AFLplusplus/AFLplusplus/pull/522/files

(Surprising, given how many env vars we have :P)

/edit: the linked pr is still missing env var parsing for showmap and tmin...

src/afl-forkserver.c
vanhauser-thc: ok, however it has still to be checked that the env var is a valid number and FATAL if not.

src/afl-fuzz-state.c
domenukk: This was resolved on xen mode side

src/afl-forkserver.c
tklengyel: I would be fine with using just a number.

src/afl-fuzz-state.c
tklengyel: Yes, this is not an issue anymore, we report the forkserver's pid as the child's pid and that just works. The forkserver receives the signal twice here but that doesn't matter.

v-p-b:
I pushed two commits with long messages to describe my intentions - I hope they resolve at least some of the issues you raised! 

Tests run fine on my Ubuntu 20.04 - can someone take a look at what can be the problem with the CI?

src/afl-forkserver.c
```diff
@@ -1137,6 +1162,15 @@ fsrv_run_result_t afl_fsrv_run_target(afl_forkserver_t *fsrv, u32 timeout,
 
   /* Report outcome to caller. */
 
+  /* TODO We use SIGTERM here as an indicator of Xen mode,
+     although it's not equivalent! */
+  if (fsrv->kill_signal == SIGTERM && !*stop_soon_p &&
```
domenukk, line 1167: Pretty sure we can just always do the `last_run_timed_out` check here, and drop it below, any reason not to?
Is there a race condition here the recent code could catch?

Also, we can wrap it in a `unlikely`

src/afl-forkserver.c
domenukk: We try not to parse env variables here, but instead set the kill signal when parsing, before starting the forkserver.

domenukk (approved):
I think getting rid of the wifsignaled check before timeouts completely is fine, what do y'all think?
Apart from that, thanks a lot @v-p-b I can take it from here :)

src/afl-forkserver.c
domenukk: Ah I guess it is one (usually) unneeded jump? Well we don't save that jump by adding another condition, either :)

domenukk:
Cleaned it up in https://github.com/AFLplusplus/AFLplusplus/commit/a06b25538fd7b6eef6755094aa4678c2cb5333fd maybe somebody can take a look

PR #688, opened 2021-01-14
Update documentation to reflect new default power schedule

murx-:
afl++ v3.0 changed the default power schedule from explore to default. This has not been updated everywhere. 
Furthermore the  status bar of the status screen contains the CPU id which was not mentioned earlier.


vanhauser-thc:
thank you!

PR #689, opened 2021-01-18
cmplog merge

vanhauser-thc:
This is not for merging - yet - rather to discuss implementation decisions and how to make the code more readable or effective - if necessary.

So - what are the changes?

## instrumentation (cmplog-instructions-pass.so and afl-compiler-rt.o)

* we now also instrument non-u8/u16/u32/u64, namely floating points (from float 40 bit to 128 bit), u128 and _ExtInt() (anything from 8 to 128 bit, e.g. u111)
* we also log the compare type (equal flag, lesser flag, greater flag) -> attribute variable
* if the two items compared for instrumentation are the same, then they are not logged.

to allow this I had to change cmplog.h to support the new attribute variable, so old cmplog compiled targets will not work with this new afl-fuzz!
the u65-u128 support is implemented by adding a second u64 for arg1 and arg2. this is combined in afl-fuzz to u128.

## Colorization

Colorization is a technique to identify which bytes in the queue input do *not* change the trace_bits.

* Now the exact byte locations that can be changed are saved and used for the redqueen stage (called `taint`)
* previously up to 1000 execution attempts were performed in this phase. this works perfectly up to 500 bytes of input, anything larger depends if successful or not (and there was a bug, if colorization failed it would be retried every time the queue entry was re-selected for fuzzing, resuling in the same fail) - now colorization is always completed.
* previously random data was used to detect bytes that can be changed. now the values stay in the same type, e.g. uppercase stays uppercase - just a different one. same with numbers, potential hex chars, etc. this ensures we pass eg "if(isdigit())" etc. paths in a target. this is a little bit more expensive
* previously this random data was done for each single coloring step. now we create a fully changed copy once and copy from that, saving CPU cycles this way.
* another small bugfix - prevously the rng to replace was not checking if the rng happened to be the same character that would be replaced

The result is either "no byte can be changed without impacting the trace bits" => mark the queue entry as fully done and skip it, or in a taint map which bytes can be changed.

## Redqueen

In the redqueen phase the logged comparisons are written into the data and sent to the target at various locations, hopefully passing the comparison

* the collected taint information is used to only attempt replacements where the value fits, so u64 is only inserted if the tested location allowed for at least 8 bytes of data.
* previously only 8 insertion attempts were performed. now it is inserted everywhere where possible
* ExtInt, FP, u128 support
* fixed ascii2integer (never worked before)
* a few more tests to see if the values are actually worth (e.g. are we actually able to influence the values) to insert and skip otherwise
*  arithmetic math `if (a < b)` - if lesser or greater is detected, then cmplog will only solve one path. so now we add +1 and subtract -1 where this is the case to be able to find both
* a final combined path is added. say in one cmplog run 5 new paths are found. this is nice, but maybe it is possible to trigger all 5 new paths one one single input - if the changes were combined in one new input. for this to work the virgin bits have to be reset to pre-cmplog run, and after this step recombined.

For the combined data run and the arithmetic solving - I am unsure if this is really helpful. fuzzbench will show this.

## Triggering cmplog

more data is supported, more insertion per data - cmplog gets more expensive.
so there are 3 cmplog levels now. default is level 1, and after a few cycles without any finds the cmplog level is increased until it hits 3.
cmplog level 1 = only integer + string insertions
cmplog level 2 = floating point insertions
cmplog level 3 = arithmetics + ascii2integer

additionally with each empty cycle the size up to which cmplog is run on is increased.

there is more, e.g. cmplog lvl 1 only triggers if additionally there were 5 seconds without finds.
cmplog lvl2 adds cmplog for entries that are the favorite entry for at least one edge.
cmplog lvl3 -> any queue entry

## So ...

there is more, but these are the major items.
It is still not perfect, and if or how much it is better is still to be seen. so no merging yet, rather discussing and collecting further improvments...

Please review!


andreafioraldi:
all the changes LGTM

> there is more, e.g. cmplog lvl 1 only triggers if additionally there were 5 seconds without finds.

5 seconds is what I suggested because I use it in weizz, did u try to find a better time as redqueen is quite different?

IMO we should wait for the results in FuzzBench before merging, the changes should improve a lot the performance, especially the 3 levels of cmplog.

vanhauser-thc:
> > there is more, e.g. cmplog lvl 1 only triggers if additionally there were 5 seconds without finds.
> 
> 5 seconds is what I suggested because I use it in weizz, did u try to find a better time as redqueen is quite different?

no not yet. this is the basic, fine-tuning is then the next step, but I would do that after the merge.

but for a merge it first has to be better what we have. so waiting for fuzzbench results ...
but in the meantime - code review please @domenukk @hexcoder-  :)

domenukk:
What's the deal with the failing testcases?

vanhauser-thc:
ah right, I modified afl-showmap to ignore a map[0] == 1 value (because it is not always set ot 1, depends on the use case) and hence the tuples became one less, breaking a few tests.

vanhauser-thc:
fuzzbench results:
https://www.fuzzbench.com/reports/experimental/2021-01-19-aflpp/index.html (older state)
https://www.fuzzbench.com/reports/experimental/2021-01-20-aflpp/index.html (newer state, but weird general results)

both show that this changes to cmplog are beneficial - and that having all solving on by default would be a disaster (-l 3).

I will do some minor adjustments today and then merge.

afterwards I will start with the finetuning.

PR #704, opened 2021-01-26
Fix pending sub overflow displayed as 4.29G

joeyjiaojg:


vanhauser-thc:
is this happening on normal usage for you? or have you made changes?
because this should not happen, so the bug is somewhere else - that pending_not_fuzzed is not increased or that a fuzzed one was not marked as done.

joeyjiaojg:
I'm using afl_frida to fuzz, when pending paths reduced to 0, and then it becomes to 4.29G which is u32_max.
I think it's sub overflow?

vanhauser-thc:
> I'm using afl_frida to fuzz, when pending paths reduced to 0, and then it becomes to 4.29G which is u32_max.
> I think it's sub overflow?

sounds like it is an overflow at the location you provide a fix.
but the bug is not that, that is where the real bug impacts.
I will try to to hunt down the real bug on Thursday/Friday, and if I cannot find it then I will accept the PR.
thanks for pointing this issue out!

joeyjiaojg:
ok, thx

Zherya:
Hello! I would like to note that I also face with that bug periodically. For me it was firstly discovered at 3.00c version, as far as I remember. I use Linux OS and a target compiled with GCC and GCC plugin instrumentation. Nothing specific is used or any changes are made. I can try to catch that bug with some additional debug information, if required.

vanhauser-thc:
can you please check if the current dev state still has the same bug behaviour of if it is fixed?
it is not a good fix (if it is fixed), but then helps to refine it.

Zherya:
Sure, will be back with feedback soon

Zherya:
I've restarted (with '-i -' flag and all previous nodes termination) previous fuzzing session in which one of the fuzzing nodes faced with that bug, and got the next error during seed corpus processing (sorry, I'm not able to provide a screenshot):

[-] PROGRAM ABORT : disabled entry? this should not happen, please report!
                      Location : perform_dry_run(), src/afl-fuzz-init.c:1138

vanhauser-thc:
> I've restarted (with '-i -' flag and all previous nodes termination) previous fuzzing session in which one of the fuzzing nodes faced with that bug, and got the next error during seed corpus processing (sorry, I'm not able to provide a screenshot):
> 
> [-] PROGRAM ABORT : disabled entry? this should not happen, please report!
> Location : perform_dry_run(), src/afl-fuzz-init.c:1138

perfect, thats what I hoped :)
I will put a real fix in in a few hours, in the mean time just revert to the previous commit

vanhauser-thc:
@JoeyJiao @Zherya  can you please test the real fix that is in now?

Zherya:
@vanhauser-thc Everything works perfect now, thank you!

joeyjiaojg:
@vanhauser-thc 
but still not work for me.

(gdb) p/x afl->pending_not_fuzzed
$2 = 0xffffffdb

joeyjiaojg:
btw, your fix is only for perform_dry_run, but my case happens after 22 hrs running, so the problem should be exist during fuzz.

vanhauser-thc:
the fix also adds a check in another location but seems to be incomplete. weird.
I have to pinpoint the issue ... 

what are your commandline options and environment variables you use? 
basically the output of `cat out/default/fuzzer_setup`

vanhauser-thc:
@joeyjiaojg I found one more location I missed ... I dont see a possibility how this could happy anymore :) please test

Issue #703, opened 2021-01-25
sync_fuzzers doesn't preserve the 'src' attribute


ampanasiuk:
The queue elements that are introduced with the sync stage have their src attribute garbled. As can be witnessed by comparing the file sizes below:

```
-rw------- 1 root root 2.4K Jan 25 03:04 f01/queue/id:037773,sync:f09,src:000007,+cov
-rw------- 1 root root  15K Jan 25 03:04 f01/queue/id:037774,sync:f09,src:000008,+cov
-rw------- 1 root root  39K Jan 25 03:05 f01/queue/id:037775,sync:f08b,src:038692
-rw------- 1 root root 1.4K Jan 25 03:05 f01/queue/id:037776,sync:f07,src:038727
-rw------- 1 root root  310 Jan 25 03:05 f01/queue/id:037777,sync:f02,src:038029
-rw------- 1 root root  310 Jan 25 03:23 f07/queue/id:038729,sync:f01,src:037771
-rw------- 1 root root  39K Jan 25 03:23 f07/queue/id:038730,sync:f01,src:037773
-rw------- 1 root root  15K Jan 25 03:23 f07/queue/id:038731,sync:f01,src:037774,+cov
-rw------- 1 root root 2.4K Jan 25 03:24 f07/queue/id:038732,sync:f01,src:037775,+cov
```

the names were mapped like:

```
f01 id:037773 -> f07 id:038732 src:037775 
f01 id:037774 -> f07 id:038731 src:037774
f01 id:037775 -> f07 id:038730 src:037773
f01 id:037777 -> f07 id:038729 src:037771
```

where the lhs id should equal the rhs src, but instead, we have lhs_id + rhs_src constant for the particular syncing stage.

I think this code in `sync_fuzzers` is responsible for this:

```
      while (o >= m) {
          ...
          snprintf(path, sizeof(path), "%s/%s", qd_path, namelist[o]->d_name);
          afl->syncing_case = next_min_accept;
          next_min_accept++;
          o--;
```
where `next_min_accept` increases and `o` decreses with each loop iteration.

vanhauser-thc:
yes this looks wrong.
can you please send a pull request with a fix?
I dont have the time to fix this until Thursday/Friday.

ampanasiuk:
Hi,
I'll take a look.

PR #707, opened 2021-01-27
Bug 703 wrong src attribute in sync stage filenames

ampanasiuk:
The src attribute on sync stage filenames is wrong, pr fixes this, bug #703


vanhauser-thc:
in the middle of something else so I am not sure I understand this without context and time, but it looks ok to me :)
thanks!

PR #710, opened 2021-01-28
Fix dev branch for android

joeyjiaojg:


domenukk:
Looks good to me, thanks!

domenukk:
@vanhauser-thc not sure how important 32 bit support on Android is these days, but could the 128 bit ints bite us somewhere else?

joeyjiaojg:
128 bit is recently added in AFLpp code for some cmplog function, it's not in Android. but also I always work on 64 bit Android, so I don't think 32bit is a matter at least for me.

domenukk:
Other targets may also not compile with 128bit data types so we may want to hide them behind appropriate defines. :)

vanhauser-thc:
@joeyjiaojg I have a check that only includes 128 bit support on 64 bit platforms.
I thought __x86_x64__ is only set when compiling for 64 bit, but that assumption seems to be wrong then

vanhauser-thc:
No I was correct `__x86_64__` is only defined if the compiler is set to 64 bit compilations.
So I do not understand why an error was produced.

@joeyjiaojg can you please produce an error output for that?

joeyjiaojg:
@vanhauser-thc You are right, I found on android it's controlled by __aarch64__, after digging further, there is never 32 bit afl-fuzz compiled before, so I don't see any further change is needed.

PR #712, opened 2021-01-28
Fix getopt arg string

zodf0055980:
Is afl-fuzz '-P' case here?


vanhauser-thc:
uh well spotted, thanks!
I resorted the getopt string and inserted that mistake. if you could check that there are no other mistakes of mine that would be great!

zodf0055980:
I check the other getopt string, I think it all correct now. Thanks.

vanhauser-thc:
thank you :)

PR #724, opened 2021-02-02
afl-untracer: Mac M1 build update proposal.

devnexen:


vanhauser-thc:
thanks! did you try it? does it actually work? I coded that without having a Mac ;)

devnexen:
Tried works better on Intel gear rather than ARM due to memory protection enforced in the latter.

vanhauser-thc:
cool thank you :)

PR #726, opened 2021-02-03
Fix for off by one in libqasan's memmem

b1gr3db:
Off by one error resulted in memmem calling memcmp where h + needle_len is one past the end. libqasan then detected and flagged itself in the case I hit.


andreafioraldi:
ty!

Issue #274, opened 2020-03-26
[ideas] Dashboard, exit confirmation, better resuming


fuzzah:
Just lost my status screens with few billions of execs :(
So, my ideas:
1. Can we store all the fuzzer stats as shown in status screen, and then restore it when resuming (with both AFL_AUTORESUME and '-i' option) ? If it somehow affects or depends on inner state of AFL++ instance, can we restore this inner state too?
2. Can we prevent random Ctrl+C presses for example, by asking to confirm? You know, there may be lots of opened windows, tty multiplexer sessions and so on, anything can happen.
3. What if AFL++ had some sort of dashboard / control center available via web interface? It could mimic afl-whatsup and afl-plot in some ways. It could be used for much better monitoring of current fuzzing job (and maybe also show some analysis of previous runs too). It could also be used for showing separate status screens, restarting instances with provided command lines, changing input test cases, etc .. Most importantly, it would provide a way to access all the stats from any device capable of web surfing, so that you could monitor & control your fuzzing jobs while walking your dog :D

vanhauser-thc:
well the fuzzer_stats could be loaded on resume and variables set accordingly. This is however nothing important. you lost something that looked cool, but you lost nothing of real value :)
If you would like to have that in, send a PR, I think everybody else has something with higher priority on their plate

tushargoyl123:
Hi this is Tushar , I want to work on this issue , I am new to open source Can u guide me through??

vanhauser-thc:
@tushargoyl123 do have you a strong experience in coding C/C++?

tushargoyl123:
> @tushargoyl123 do have you a strong experience in coding C/C++?

Yes I know it very well...

vanhauser-thc:
@tushargoyl123 the only relevant feature request for this repo is the first one IMHO (2 is undesired, 3 is out of scope and rather an own project):

```
Can we store all the fuzzer stats as shown in status screen, and then restore it when resuming (with both AFL_AUTORESUME and '-i' option) ? If it somehow affects or depends on inner state of AFL++ instance, can we restore this inner state too?
```

basically what has to be done is that if `afl-fuzz` is started in resume mode, then the existing "$out_dir/fuzzer_stats" file has to be loaded, the values read in and assigned to the internal variables. pretty easy and fast to do - if you know what you are doing.

[...]

PR #728, opened 2021-02-04
load existing stats file when in AFL_AUTORESUME or -i - mode.

vj-27:
Hi, this is my first PR here :) 
It loads some of the variables in the stats file when using AFL_AUTORESUME or "-i -" . ( issue #274  )
It can load the following fields on status screen:
Overall Results:
- cycles done
- total paths
- uniq crashes
- uniq hangs
Findings in Depth:
- favored paths
Path Geometry:
- levels
- pending
- pend fav
- own finds
- imported

### DIFF
--- a/src/afl-fuzz-stats.c
+++ b/src/afl-fuzz-stats.c
@@ -90,20 +90,20 @@ void write_setup_file(afl_state_t *afl, u32 argc, char **argv) {
 }
 
 /* load some of the existing stats file when resuming.*/
-void load_stats_file(afl_state_t *afl) {
+u32 load_stats_file(afl_state_t *afl) {
 
   FILE *f;
   u8    buf[MAX_LINE];
   u8 *  lptr;
   u8    fn[PATH_MAX];
   u32   lineno = 0;
-
+  u32   prev_run_time = 0;
   snprintf(fn, PATH_MAX, "%s/fuzzer_stats", afl->out_dir);
   f = fopen(fn, "r");
   if (!f) {
 
     WARNF("Unable to load stats file '%s'", fn);
-    return;
+    return prev_run_time;
 
   }
 
@@ -134,6 +134,15 @@ void load_stats_file(afl_state_t *afl) {
       char *nptr;
       switch (lineno) {
 
+        case 3:
+          if (!strcmp(keystring, "run_time          ")) {
+
+            prev_run_time = 1000 * strtoull(lptr, &nptr, 10);
+            afl->start_time -= prev_run_time;
+
+          }
+
+          break;
         case 5:
           if (!strcmp(keystring, "cycles_done       "))
             afl->queue_cycle =
@@ -147,10 +156,6 @@ void load_stats_file(afl_state_t *afl) {
           if (!strcmp(keystring, "paths_total       "))
             afl->queued_paths = strtoul(lptr, &nptr, 10);
           break;
-        case 11:
-          if (!strcmp(keystring, "paths_favored     "))
-            afl->queued_favored = strtoul(lptr, &nptr, 10);
-          break;
         case 12:
           if (!strcmp(keystring, "paths_found       "))
             afl->queued_discovered = strtoul(lptr, &nptr, 10);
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
at a first look this seems to be good!

if you import total execs done, then you will also need to load the run_time as otherwise the exec calculations will be wrong.

then - do not load any pending_* or *_favored - this is internally assessed.

vj-27:
@vanhauser-thc I have a doubt.
if we load run_time, which variable to set? In the calculations, cur_time - start_time is used and the cur_time is set by calling gettimeofday(). or can i make a new variable prev_run_time in the state and use it for calculations (**prev_run_time +cur_time - start_time**) ?

vanhauser-thc:
that would one option. easier would be just to subtract that value from start_time. the only disadvantage would be that start_time would be incorrect in the new fuzzer_stats but IMHO that is fine

domenukk:
One other thing, how about an `AFL_RESUME_NOSTATS` or similar for people that want the old behavior?
Well thinking about it: they can just rm the stats file 🤔

vj-27:
@domenukk sorry i didn't understand. :confused:

vj-27:
> Why not handle the `pref_run_time` inline, like all the other vars?
> We just need to make sure the `afl->start_time` is set before calling this function

Can you check the latest commit?

src/afl-fuzz-stats.c
```diff
@@ -89,6 +89,106 @@ void write_setup_file(afl_state_t *afl, u32 argc, char **argv) {
...
 
+/* load some of the existing stats file when resuming.*/
+void load_stats_file(afl_state_t *afl) {
+
+  FILE *f;
+  u8    buf[MAX_LINE];
+  u8 *  lptr;
+  u8    fn[PATH_MAX];
+  u32   lineno = 0;
+  afl->prev_run_time = 0;
```
vanhauser-thc: the structure is null initialized, not needed

src/afl-fuzz-stats.c
```diff
@@ -89,6 +89,106 @@ void write_setup_file(afl_state_t *afl, u32 argc, char **argv) {
...
+      strcpy(keystring, lstartptr);
+      lptr++;
+      char *nptr;
+      switch (lineno) {
+
+        case 3:
+          if (!strcmp(keystring, "run_time          ")) {
+
+            afl->prev_run_time = 1000 * strtoull(lptr, &nptr, 10);
+            afl->start_time -= afl->prev_run_time;
```
vanhauser-thc: this makes the env var pointless, remove this line, then see other comment

src/afl-fuzz.c
```diff
@@ -1700,6 +1703,8 @@ int main(int argc, char **argv_orig, char **envp) {
   // (void)nice(-20);  // does not improve the speed
   // real start time, we reset, so this works correctly with -V
   afl->start_time = get_cur_time();
+  if (afl->in_place_resume || afl->afl_env.afl_autoresume)
+    afl->start_time -= afl->prev_run_time;
```
vanhauser-thc: remove this line, this would make the var pointless

vanhauser-thc:
what is missing is checking where start_time is used and plying the afl->prev_run_time var there, e.g.

```
afl-fuzz-stats.c:          "start_time        : %llu\n"
afl-fuzz-stats.c:          afl->start_time / 1000, cur_time / 1000,
```
would need to be
```
afl-fuzz-stats.c:          (afl->start_time - afl->prev_run_time var)  / 1000, cur_time / 1000,
```

but you have to check how start_time is used, this cannot be blindly replaced globally.

dominik is right that it needs afl->prev_run_time, because the value of start_time itself may not be modified.

vj-27:
@vanhauser-thc thanks for the review
I have made the changes so that it affects only the stats screen and the stats file. 
And in the status screen when resuming, should the "run time " start from 0 or from prev_run_time ?

vanhauser-thc:
there are a few more locations where prev_run_time has to be applied, but I take it from here.
Thank you for this PR!

PR #742, opened 2021-02-14
Improve LLVM source build instructions

rchildre3:
* Expose llvm-config flags on make to help debug
* Dynamic rtti given what gets compiled
* Detect underlinking and harden dynamic linking of llvm shared objects
* Reduce LLVM extraneous projects unneded for instrumentation
* Build only for the host target
* Use cmake '--build' to support Ninja and other build tools

### DIFF
--- a/GNUmakefile.llvm
+++ b/GNUmakefile.llvm
@@ -273,7 +273,8 @@ endif
 CLANG_CPPFL  =  $(shell $(LLVM_CONFIG) --cxxflags) -fPIC $(CXXFLAGS)
 CLANG_LFL    =  $(shell $(LLVM_CONFIG) --ldflags --libs passes --system-libs) $(LDFLAGS)
 
-ifeq "$(shell $(LLVM_CONFIG) --has-rtti) 2> /dev/null" "YES"
+ifeq "$(shell $(LLVM_CONFIG) --has-rtti 2> /dev/null)" "YES"
+  $(info [*] LLVM compiled containing RTTI info, compiling passes with -frtti)
   CLANG_CPPFL += -frtti
 else
   CLANG_CPPFL += -fno-rtti
@@ -385,25 +386,25 @@ endif
 endif
 
 instrumentation/afl-llvm-common.o: instrumentation/afl-llvm-common.cc instrumentation/afl-llvm-common.h
-	$(CXX) $(CFLAGS) $(CPPFLAGS) `$(LLVM_CONFIG) --cxxflags` -fno-rtti -fPIC -std=$(LLVM_STDCXX) -c $< -o $@ 
+	$(CXX) $(CLANG_CPPFL) -c $< -o $@
 
 ./libLLVMInsTrim.so: instrumentation/LLVMInsTrim.so.cc instrumentation/MarkNodes.cc instrumentation/afl-llvm-common.o | test_deps
-	-$(CXX) $(CLANG_CPPFL) -DLLVMInsTrim_EXPORTS -fno-rtti -fPIC -std=$(LLVM_STDCXX) -shared $< instrumentation/MarkNodes.cc -o $@ $(CLANG_LFL) instrumentation/afl-llvm-common.o
+	$(CXX) $(CLANG_CPPFL) -DLLVMInsTrim_EXPORTS -shared $< instrumentation/MarkNodes.cc -o $@ $(CLANG_LFL) instrumentation/afl-llvm-common.o
 
 ./afl-llvm-pass.so: instrumentation/afl-llvm-pass.so.cc instrumentation/afl-llvm-common.o | test_deps
 ifeq "$(LLVM_MIN_4_0_1)" "0"
 	$(info [!] N-gram branch coverage instrumentation is not available for llvm version $(LLVMVER))
 endif
-	$(CXX) $(CLANG_CPPFL) -DLLVMInsTrim_EXPORTS -fno-rtti -fPIC -std=$(LLVM_STDCXX) -shared $< -o $@ $(CLANG_LFL) instrumentation/afl-llvm-common.o
+	$(CXX) $(CLANG_CPPFL) -DLLVMInsTrim_EXPORTS -shared $< -o $@ $(CLANG_LFL) instrumentation/afl-llvm-common.o
 
 ./SanitizerCoveragePCGUARD.so: instrumentation/SanitizerCoveragePCGUARD.so.cc instrumentation/afl-llvm-common.o | test_deps
 ifeq "$(LLVM_10_OK)" "1"
-	-$(CXX) $(CLANG_CPPFL) -fno-rtti -fPIC -std=$(LLVM_STDCXX) -shared $< -o $@ $(CLANG_LFL) instrumentation/afl-llvm-common.o
+	-$(CXX) $(CLANG_CPPFL) -shared $< -o $@ $(CLANG_LFL) instrumentation/afl-llvm-common.o
 endif
 
 ./afl-llvm-lto-instrumentlist.so: instrumentation/afl-llvm-lto-instrumentlist.so.cc instrumentation/afl-llvm-common.o
 ifeq "$(LLVM_LTO)" "1"
-	$(CXX) $(CLANG_CPPFL) -fno-rtti -fPIC -std=$(LLVM_STDCXX) -shared $< -o $@ $(CLANG_LFL) instrumentation/afl-llvm-common.o
+	$(CXX) $(CLANG_CPPFL) -shared $< -o $@ $(CLANG_LFL) instrumentation/afl-llvm-common.o
 endif
 
 ./afl-ld-lto: src/afl-ld-lto.c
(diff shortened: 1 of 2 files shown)


GNUmakefile.llvm
```diff
@@ -270,15 +270,20 @@ endif
 ifneq "$(LLVM_CONFIG)" ""
   CLANG_CFL += -I$(shell dirname $(LLVM_CONFIG))/../include
 endif
-CLANG_CPPFL  = `$(LLVM_CONFIG) --cxxflags` -fno-rtti -fPIC $(CXXFLAGS) -Wno-deprecated-declarations
-CLANG_LFL    = `$(LLVM_CONFIG) --ldflags` $(LDFLAGS)
+CLANG_CPPFL  =  $(shell $(LLVM_CONFIG) --cxxflags) -fPIC $(CXXFLAGS)
+CLANG_LFL    =  $(shell $(LLVM_CONFIG) --ldflags --libs passes --system-libs) $(LDFLAGS)
 
+ifeq "$(shell $(LLVM_CONFIG) --has-rtti) 2> /dev/null" "YES"
+  CLANG_CPPFL += -frtti
```
vanhauser-thc, line 278: why do you add rtti? I have never experienced it being needed.

instrumentation/README.lto.md
```diff
@@ -88,16 +88,35 @@ apt-get install -y clang-12 clang-tools-12 libc++1-12 libc++-12-dev \
 ### Building llvm yourself (version 12)
 
 Building llvm from github takes quite some long time and is not painless:
-```
+```sh
 sudo apt install binutils-dev  # this is *essential*!
 git clone https://github.com/llvm/llvm-project
```
vanhauser-thc: my oversight - can you please add `--depth=1`? I dont want to push that and then your commit might conflict

instrumentation/README.lto.md
```diff
@@ -88,16 +88,35 @@ apt-get install -y clang-12 clang-tools-12 libc++1-12 libc++-12-dev \
...
 cd build
-cmake -DLLVM_ENABLE_PROJECTS='clang;clang-tools-extra;compiler-rt;libclc;libcxx;libcxxabi;libunwind;lld' -DCMAKE_BUILD_TYPE=Release -DLLVM_BINUTILS_INCDIR=/usr/include/ ../llvm/
-make -j $(nproc)
-export PATH=`pwd`/bin:$PATH
-export LLVM_CONFIG=`pwd`/bin/llvm-config
+
+# Add -G Ninja if ninja-build installed
+cmake \
+    -DCLANG_INCLUDE_DOCS="OFF" \
+    -DCLANG_PLUGIN_SUPPORT="OFF" \
```
vanhauser-thc: this would break afl++ ...

instrumentation/README.lto.md
```diff
@@ -88,16 +88,35 @@ apt-get install -y clang-12 clang-tools-12 libc++1-12 libc++-12-dev \
...
+    -DCMAKE_BUILD_TYPE=Release \
+    -DLLVM_BINUTILS_INCDIR=/usr/include/ \
+    -DLLVM_BUILD_LLVM_DYLIB="ON" \
+    -DLLVM_ENABLE_BINDINGS="OFF" \
```
vanhauser-thc, line 106: why this?

instrumentation/README.lto.md
```diff
@@ -88,16 +88,35 @@ apt-get install -y clang-12 clang-tools-12 libc++1-12 libc++-12-dev \
...
+    -DLLVM_INCLUDE_EXAMPLES="OFF" \
+    -DLLVM_INCLUDE_TESTS="OFF" \
+    -DLLVM_LINK_LLVM_DYLIB="ON" \
+    -DLLVM_TARGETS_TO_BUILD="host" \
+    ../llvm/
+cmake --build . --parallel
+
+# Ensure that the linker finds your libLLVM.so in /usr/local/lib over a
+# preinstalled one in /usr/lib or /lib
+sudo cmake --install .
```
vanhauser-thc: not everyone wants this system-wide installed. or can even.
you can pose this as an easier alternative.
however this section must set LLVM_CONFIG - likely there is another one in /usr/bin and then this compile would never be used.
(that is about the part you deleted)

vanhauser-thc:
thank you so far, I left some comments

instrumentation/README.lto.md
rchildre3: AFL++ does not use the [OCaml or Golang] bindings to llvm

[OCaml or Golang]: https://github.com/llvm/llvm-project/blob/llvmorg-12.0.0-rc1/llvm/docs/CMake.rst#llvm-specific-variables

vanhauser-thc: which would be ok if you build it just for afl++. but what if people do "make install" - can this then fuck up other tools who call the default clang - which then do not have the bindings anymore?

rchildre3: I am investigating options to revert the `cmake --install` option

GNUmakefile.llvm
rchildre3: While the LLVM project "[does not use exceptions or RTTI]", [LLVM_ENABLE_RTTI] is still a valid build option which would require `-frtti` as a codegen option for compiling the AFL++ passes as "[Mixing code compiled with -frtti with that compiled with -fno-rtti may not work]".  The other branch of this conditional should almost always be hit with most LLVM binary configurations in the wild, this is a safety check for non-default configurations.

I tested LLVM configs with both rtti and no rtti and while the Makefile without this patch worked, `llvm-config` exposes this option for good cause in preventing bugs and it should be leveraged in building plugins to its infrastructure. The [Boost] project also warn of mixing such code. 

@vanhauser-thc Did you test with a LLVM binary configured with RTTI Enabled?

[does not use exceptions or RTTI]: https://github.com/llvm/llvm-project/blob/llvmorg-12.0.0-rc1/llvm/docs/CodingStandards.rst#do-not-use-rtti-or-exceptions

[LLVM_ENABLE_RTTI]: https://github.com/llvm/llvm-project/blob/llvmorg-12.0.0-rc1/llvm/docs/CMake.rst#llvm-specific-variables

[Mixing code compiled with -frtti with that compiled with -fno-rtti may not work]: https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html

[Boost]: https://www.boost.org/doc/libs/1_75_0/doc/html/boost_typeindex/mixing_sources_with_rtti_on_and_.html

instrumentation/README.lto.md
rchildre3: fixed in e9849855

instrumentation/README.lto.md
rchildre3: I have instrumented binaries with `afl-clang-lto` with this exact configuration on both `release/11.x` and `release/12.x` of `llvm-project`. What breakages did you observe?

instrumentation/README.lto.md
rchildre3: Reverted in e9849855

instrumentation/README.lto.md
rchildre3: > or can even

Given the `sudo make install` command in the `AFLplusplus` directory I though that sudoer capabilities were assumed. But reverted in e9849855

GNUmakefile.llvm
vanhauser-thc: passes are just their own .cc and afl-llvm-common.c, compiled together with the same compiler with the same options at the same time.
instrumenting a binary with a pass does not compile that pass into the target.
hence rtti or not has zero impact. it is just not needed. I dont mind if it is there, it just doesnt make a difference.

instrumentation/README.lto.md
vanhauser-thc: for afl-clang-lto the linker (ld.lld) loads the pass. in afl-clang-fast however clang loads the pass and disabling loading that would fail. also cmplog and laf for lto should fail because that too is done with clang

rchildre3: Ah, my understanding is this build process was only for building only LTO mode, (i.e. LAF mode out of scope, hence not testing it).  Reverted in df75f62e334a081db868c5b32a73d19877a68c33

GNUmakefile.llvm
rchildre3: I understand that pass code and target code are not linked together. However, pass code and LLVM / Clang code are linked when the pass is invoked. The `CLANG_CPPFL` variable compiles AFL++ code which will be linked against LLVM / Clang code, correct? If so then, the `rtti` of the code should then match how the LLVM / Clang was compiled (i.e. `llvm-config --has-rtti`)

GNUmakefile.llvm
```diff
@@ -380,25 +386,25 @@ endif
 endif
 
 instrumentation/afl-llvm-common.o: instrumentation/afl-llvm-common.cc instrumentation/afl-llvm-common.h
-	$(CXX) $(CFLAGS) $(CPPFLAGS) `$(LLVM_CONFIG) --cxxflags` -fno-rtti -fPIC -std=$(LLVM_STDCXX) -c $< -o $@ 
+	$(CXX) $(CLANG_CPPFL) -c $< -o $@
```
vanhauser-thc, line 389: and why are you removing LLVM_STDCXX and -fPIC?
the plugins compile from llvm 3.7 onwards (except lto), and they must still work with your changes. 
same with your removal of -Wno-writable-strings ... your llvm version might not warn, but that flag is there for a reason.
llvm is quite volatile in its internal changes etc. little changes can break it.
so I can only merge if it either fixes a break - or improves and does not break something.

GNUmakefile.llvm
vanhauser-thc: ah right, that makes sense!

rchildre3:
Will close this and split into two separate PRs, I made this one too unwieldy

PR #746, opened 2021-02-15
CMPLOG llvm pass: use CreateIntCast to cast to the right integer type

f0rki:
In the current dev branch I get errors while compiling a target with cmplog (afl-clang-lto++). My target uses bigints, which leads to trouble for the cmplog instrumentation. This PR fixes the issues (at least for me). Currently I get the following error:

```
Invalid bitcast
  %10 = bitcast i80 %9 to i128, !dbg !29860
in function _ZSt7signbite
LLVM ERROR: Broken function found, compilation aborted!
```

I think bitcast `i80` to `i128` doesn't work since bitcast only works with types of the exact same bit size. So we need to use `zext` instead to extend to `i128` in the case above. I believe that `CreateIntCast` on the `IRBuilder` is the right function for this job (except we need a bitcast to an integer type first in case we are dealing with floating point values).

I think that previously 128 bit integer types were also implicitly truncated due to the use of `uint64_t val = cint->getZExtValue();` to create a new constant integer. I haven't tested this, but this should be fixed now too.

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -49,7 +49,6 @@ sending a mail to <afl-users+subscribe@googlegroups.com>.
       CLANG for old afl-clang
     - fixed a potential crash in the LAF feature
     - workaround for llvm 13
-    - workaround for llvm internal lto bug that lets not bitcast from _ExtInt()
   - qemuafl
     - QASan (address sanitizer for Qemu) ported to qemuafl!
       See qemu_mode/libqasan/README.md
--- a/include/envs.h
+++ b/include/envs.h
@@ -16,7 +16,6 @@ static char *afl_environment_deprecated[] = {
 
 static char *afl_environment_variables[] = {
 
-    "_AFL_LTO_COMPILE",
     "AFL_ALIGNED_ALLOC",
     "AFL_ALLOW_TMP",
     "AFL_ANALYZE_HEX",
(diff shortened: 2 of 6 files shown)


vanhauser-thc:
are you using afl-clang-lto?
because I implemented a fix for this like 4 hours ago :)

f0rki:
lol. apparently afl++ dev speed is too fast for me. yes I am using clang LTO. yes it seems fixed in current dev. 

But I am wondering whether I would loose instrumentation of some compare instructions with the current workaround in dev?

> I think that previously 128 bit integer types were also implicitly truncated due to the use of uint64_t val = cint->getZExtValue(); to create a new constant integer.

Does this part still apply?

vanhauser-thc:
I like your solution better than mine, in LTO I just skip this crashing bitcast. can you rework your PR to fit to the changed dev state?

f0rki:
sure. I'll rebase and revert your workaround (edit: ~~but probably tomorrow morning~~). Do you have another project where this issue comes up that I can use for testing?

vanhauser-thc:
> > I think that previously 128 bit integer types were also implicitly truncated due to the use of uint64_t val = cint->getZExtValue(); to create a new constant integer.
> 
> Does this part still apply?

yes:
https://llvm.org/doxygen/classllvm_1_1ConstantInt.html#ac09a21c371a9c535cbc13e8f82503aec

f0rki:
@vanhauser-thc done. rebased and reverted workaround.

vanhauser-thc:
thank you!

PR #752, opened 2021-02-18
custom mutator rust support

### LINKED ISSUE #701: Custom Mutator Rust Support
Dear Community,

I have created some basic rust bindings for the custom mutator API. I've polished it a bit now and pushed it to a separate repository [here](https://github.com/julihoh/aflpp_custom_mutator_rust) for inspection.

Has anybody else tried this yet? Feedback would be greatly appreciated. Of course, if you are interested, it would probably be cool to integrate this into AFL++ properly, but I'm not sure yet what the best way to do this would be.

Here are some notes about it:
* it tries to be a somewhat safe abstraction, but this if my first experience doing FFI in rust and surely contains many safety issues :)
* I have written some documentation for it, check it out by running `cargo doc -p aflpp_custom_mutator --open`
* it references AFL++ as a git submodule in order to generate the bindings required to access `afl_state`, so don't forget to pull that.
* there is a trivial example in the `example` directory, I have pasted the meat of it at the end of this issue.
* it covers the basic mutator api, but no trimming, postprocessing or havoc support

Please do have a look at the documentation, it describes in more detail what this binding can do  :)

An example mutator:
```rust
use aflpp_custom_mutator::{CustomMutator, FuzzResult, afl_state, export_mutator};

struct ExampleMutator;

impl CustomMutator for ExampleMutator {
    fn init(afl: &'static afl_state, seed: std::os::raw::c_uint) -> Self
    where
        Self: Sized {
            Self
    }

    fn fuzz(&mut self, buffer: &mut [u8], add_buff: Option<&[u8]>, max_size: usize) -> FuzzResult {
        buffer.reverse();
        FuzzResult::InPlace
    }
}

export_mutator!(ExampleMutator);
```

Cheers

julihoh:
Fixes #701


julihoh:
I wasn't really sure about the directory layout, please feel free to nit.

For the documentation I put a short notice into `docs/custom_mutators.md`. Since the binding is documented in rustdoc, I don't really want to duplicate all documentation into `custom_mutators.md`. I hope that's okay.

custom_mutators/rust/custom_mutator/src/fallible.rs
```diff
@@ -0,0 +1,159 @@
...
+            Err(e) => {
+                Self::handle_err(e);
+                panic!("Error in afl_custom_init")
+            }
+        }
+    }
+
+    fn fuzz_count(&mut self, buffer: &[u8]) -> u32 {
+        match self.fuzz_count(buffer) {
+            Ok(r) => r,
```
domenukk: Is this be an infinite recursion, leading to a stack overflow, or some rust I don't know?

custom_mutators/rust/custom_mutator/src/lib.rs
```diff
@@ -0,0 +1,490 @@
...
+use std::{ffi::CStr, os::raw::c_uint};
+
+#[cfg(feature = "afl_internals")]
+#[doc(hidden)]
+pub use custom_mutator_sys::afl_state;
+
+/// The result of a call to [`CustomMutator::fuzz`].
+#[derive(Debug)]
+pub enum FuzzResult<'l> {
+    /// Returned when the given buffer was modified in place. Modifying the buffer in place to be of a different length is not supported currently.
```
domenukk: Can't we, in general, check the slice size before returning?

custom_mutators/rust/custom_mutator/src/lib.rs
```diff
@@ -0,0 +1,490 @@
...
+    InPlace,
+    /// Returned when a buffer that is managed by the custom mutator should be returned.
+    NewBuffer(&'l [u8]),
+    /// Returned when the custom mutator fails. This will make AFL++ fail fast (ie crash) and corresponds to returning a NULL buffer and 0 length.
```
domenukk: Not sure if this is true for all functions, I remember there was a PR to allow 0 sized returns for custom mutators, they will be skipped then (do you remember, @vanhauser-thc?)

domenukk (commented):
Really cool, thanks! :)

docs/custom_mutators.md
```diff
@@ -4,6 +4,9 @@ This file describes how you can implement custom mutations to be used in AFL.
 For now, we support C/C++ library and Python module, collectivelly named as the
 custom mutator.
 
+There is also experimental rust support in `custom_mutators/rust`.
+Please refer to this directory for documentation.
```
vanhauser-thc: "... by executing `cargo doc`..." or something like this. some people might want to pick up rust and not know exactly how rustdoc works. like me :)

custom_mutators/rust/custom_mutator/src/lib.rs
vanhauser-thc: cant remember would have to check the source too. if it is not documented in docs/ it should be added

custom_mutators/rust/custom_mutator/src/fallible.rs
julihoh: Its a trait implementation (`impl Trait for Type`), so `self` refers to `M: FallibleCustomMutator`, not `CustomMutator`.

docs/custom_mutators.md
julihoh: Added a note about it! Thanks!

julihoh:
Oh man.. I had just written up a whole long explanation for why the fuzz method has to return this weird FuzzResult type. FuzzResults is weird, because ideally, the return type of `fuzz` would simply be `&[u8]`. 
As always, explaining why I did it this way made me realise how wrong I was 😬 .

Anyways, here are two options, I'm not sure which one is better.
1. Keep the fuzz method as is. Its limitation currently is that a slice of the input buffer can not be returned.
2. Change the fuzz method signature to the following:
```rust
    fn fuzz<'b, 's: 'b>(
        &'s mut self,
        buffer: &'b mut [u8],
        add_buff: Option<&[u8]>,
        max_size: usize,
    ) -> Option<&'b [u8]>;
```
This captures the lifetime requirements for the buffer.
Demonstration:
```rust
    fn fuzz<'b, 's: 'b>(
        &'s mut self,
        buffer: &'b mut [u8],
        add_buff: Option<&[u8]>,
        max_size: usize,
    ) -> Option<&'b [u8]> {
        // you can return any slice from this function
        match buffer.len() {
            // return some data out of our own buffer
            1 => Some(&self.my_own_buffer.as_slice()),
            // mutate the given buffer and return it
            2 => {
                buffer.reverse();
                Some(buffer)
            }
            // return a slice of the given buffer
            3 => Some(&buffer[0..1]),
            4 => {
                // the following line does not compile, as we would escape a reference to AFL++ owned buffer that could be used later (which it must not).
                //self.r = Some(buffer);
                Some(buffer)
            }
            // return None to let AFL++ skip this mutation
            _ => None,
        }
    }
```

The second option is nice, because it gives maximal flexibility, but it requires lifetime annotations, which is usually quite scary to newcomers (it sure was for me...).

Some notes on why simpler approaches don't work:
* leaving out all lifetime annotations doesn't work, because it would disallow returning the input buffer 
* using just a single lifetime for `self`, `buffer` and the return type does not work, because it would let us escape the buffer reference into `self`.

vanhauser-thc:
@julihoh can you please write it in simple terms so that non-rust ppl can understand it? :)

custom_mutators/rust/custom_mutator/src/lib.rs
julihoh: AFAICT:
* 0 sized return will skip the mutation
* null buffer return will `FATAL` in AFL++.

I think a FATAL in AFL++ should never be the goals from inside a custom mutator. if a custom mutator needs to abort, it can do so itself (ie. by `panic`ing). 
In my latest refactoring ([comment below](https://github.com/AFLplusplus/AFLplusplus/pull/752#issuecomment-781956576)), the behaviour would be to allow the custom mutator to return a 0 size to skip the mutation, but there would be no way to return a null pointer.

domenukk:
In my opinion, the lifetime reflects the actual lifetime of the buf (defined by the caller).
Returning your own, new buffer will be painful though, right? We need a way to do both, return the new one, or mutate the old.
As for error handling, instead of Panicing, why not return a `Result<.., TErr>` (and maybe rename TErr to Error?)

julihoh:
> @julihoh can you please write it in simple terms so that non-rust ppl can understand it? :)

I can try later, but its difficult :/



> In my opinion, the lifetime reflects the actual lifetime of the buf (defined by the caller).

you mean the lifetime annotation I proposed?

> Returning your own, new buffer will be painful though, right? We need a way to do both, return the new one, or mutate the old.

No, simple:
```rust
struct OwnBufferExampleMutator {
    own_buffer: Vec<u8>,
}

impl CustomMutator for OwnBufferExampleMutator {
    fn init(seed: c_uint) -> Self {
        Self {
            own_buffer: Vec::new(),
        }
    }

    fn fuzz<'b, 's: 'b>(
        &'s mut self,
        buffer: &'b mut [u8],
        add_buff: Option<&[u8]>,
        max_size: usize,
    ) -> Option<&'b [u8]> {
        self.own_buffer.reverse();
        Some(self.own_buffer.as_slice())
    }
```

> As for error handling, instead of Panicing, why not return a `Result<.., TErr>` (and maybe rename TErr to Error?)

Thats what FallibleCustomMutator is for :) I think Option makes it clear, that returning a None is entirely fine. With Result, I feel like something should be done about the error case (which again, FallibleCustomMutator can be used for).
I agree with the naming though

domenukk:
What's the benefit of an additional Fallible trait, instead of just always returning a `Result`? It saves one `Ok(())` but apart from that, I would just always use the fallible?

julihoh:
I think its not entirely clear, what to do in case of the error variant. I see two sensible options: 1. panic, 2. return _something_ to AFL as to not crash the process.

Usually I would discard option 2, but given that it is quite annoying to have a long running fuzzing campaign crash when your asleep, I think its quite reasonable.

It also requires specifying the Error type upfront, which is a non-trivial choice IMHO.

The current design cleanly splits out error handling from the rest of the code.

domenukk:
> 
> 
> I think its not entirely clear, what to do in case of the error variant. I see two sensible options: 1. panic, 2. return _something_ to AFL as to not crash the process.
> 
> Usually I would discard option 2, but given that it is quite annoying to have a long running fuzzing campaign crash when your asleep, I think its quite reasonable.

Returning 0 and/or `NULL` should be fine for most cases, then AFL++ can decide if it wants to continue, cancel this specific test case, or `FATAL` internally.
You could even check if a `DEBUG` env is is set, and print the stacktrace accordingly, what do you think?

julihoh:
> Returning 0 and/or `NULL` should be fine for most cases, then AFL++ can decide if it wants to continue, cancel this specific test case, or `FATAL` internally.

Thats exactly what FallibleCustomMutator does, if it's `handle_err` doesn't panic. :)

> You could even check if a `DEBUG` env is is set, and print the stacktrace accordingly, what do you think?

[exactly](https://github.com/AFLplusplus/AFLplusplus/pull/752/files#diff-c293c9c16491f996e001ddcd7b505eae16ab85972f4221fe6088ed8a02b340f2R17) :) 

```rust
struct FallibleMutatorExample;

impl FallibleCustomMutator for FallibleMutatorExample {
    type Error = i32;

    fn init(seed: c_uint) -> Result<Self, i32> {
        Ok(Self)
    }

    fn fuzz<'b, 's: 'b>(
        &'s mut self,
        buffer: &'b mut [u8],
        add_buff: Option<&[u8]>,
        max_size: usize,
    ) -> Result<Option<&'b [u8]>, i32> {
        Err(1)
    }

    fn handle_err(err: i32) {
        if std::env::var("DEBUG").is_ok() {
            eprintln!("Error in custom mutator: {}", err);
            // might panic in debug mode
        }
        // not panicing here -> continue execution
    }
}
```

strack traces for panics should work regardless. Results don't have stack traces, but using something like [anyhow](https://docs.rs/anyhow/1.0.38/anyhow/index.html) you can get a "error trace" (ie. which error lead to which).
This is how I used the binding in my project.

In fact, the FallibleCustomMutator thing was originally part of my project's internals and I decided to also put it into the reusable binding because I thought it could be useful.

In General, I was just trying to be very spartan with the binding as to not prescribe any specific behaviours and just focus on providing a safe interface, but nothing else. I do realise that FallibleCustomMutator isn't that.

andreafioraldi:
If you want to provide a concrete example of a Rust mutator, I suggest to build a small example using lain (https://crates.io/crates/lain). It is really cool and similar to protobuf mutator, but easier IMO.

domenukk:
> In General, I was just trying to be very spartan with the binding as to not prescribe any specific behaviors and just focus on providing a safe interface, but nothing else. I do realise that FallibleCustomMutator isn't that.

I think the Fallible one should be the default/only one, to be honest. There is no real reason not to use it? Worst case, it saves the API user a single OK, but makes the use more convoluted ("which one should I pick?")

custom_mutators/rust/example/src/lib.rs
```diff
@@ -43,4 +43,31 @@ impl CustomMutator for OwnBufferExampleMutator {
...
+        &'s mut self,
+        buffer: &'b mut [u8],
+        add_buff: Option<&[u8]>,
+        max_size: usize,
+    ) -> Result<Option<&'b [u8]>, i32> {
+        Err(1)
+    }
+
+    fn handle_err(err: i32) {
+        if std::env::var("DEBUG").is_ok() {
```
domenukk: This should just be default behavior inside the bindings, people can handle their own errors if they really need to :)

julihoh: done!

julihoh:
> I think the Fallible one should be the default/only one, to be honest. There is no real reason not to use it? Worst case, it saves the API user a single OK, but makes the use more convoluted ("which one should I pick?")

I just pushed a refactoring that makes FallibleCustomMutator the default. Internally it keeps the split, as it makes the implementation cleaner, but from the outside, there is just CustomMutator.

it also implements error handling by default, but the Error type still needs to be specified by the user.

julihoh:
> If you want to provide a concrete example of a Rust mutator, I suggest to build a small example using lain (https://crates.io/crates/lain). It is really cool and similar to protobuf mutator, but easier IMO.

done :)

domenukk:
Thanks! That's super useful. I'll merged and will fix some things like move examples to `[[example]]`. Also sent you an invite to the rust libafl if you're interested @julihoh

julihoh:
Awesome :)

Never accepted an invitation so fast ;) really exciting!

julihoh:
Also: would be great if you could tag me as soon as something goes wrong with the bindings in the future!

domenukk:
@julihoh Actually I have questions/suggestions for the bindings going forward:
a) should we maybe always return `str` instead of `Cstr` to the target?
b) rust mutators could, in theory, `describe` themselves according to their struct name, as default impl, right?
c) Same as a), we could take `str` as returns,  auto-converting in the bindings, that could be easier to use
what do you think?

julihoh:
> @julihoh Actually I have questions/suggestions for the bindings going forward:
> a) should we maybe always return `str` instead of `Cstr` to the target?

That would certainly be more convenient. However, converting between `str` and `CStr` is non-trivial:
1. the conversion can fail (rust strings can contain null-bytes, whereas c strings can not)
2. the conversion incurs at least a memcpy(-ish) 

This is the reason I had not implemented it before.

If the performance cost is to be neglected (which, given how all methods returning strings should be called rather infrequently, could be reasonable), I would suggest the following:
* We need a buffer to store the cstring into (essentially: we need to store the `CString` somewhere). We could use the FFIContext for that. This would hide the implementation details from the user and captures the semantics well IMHO.
* Ideally, the user would be able to return owned (`String`) and unowned (`&str`) strings from the respective methods. We could enable this by making the return types generic using the impl-Trait-syntax: `fn fun() -> impl AsRef<str>`.

We would need to decide on a error handling strategy. I would propose to just swallow any characters that are not compatible with c strings and probably it may also be wise to not allow unicode.

> b) rust mutators could, in theory, `describe` themselves according to their struct name, as default impl, right?

Yup, sounds like a good idea. I think we could use `std::any::type_name`, but it also includes the namespace, which could get a bit long.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7efca3261d789bd0e4e732abf26cfc71

> c) Same as a), we could take `str` as returns, auto-converting in the bindings, that could be easier to use
> what do you think?

See above.

I can create PRs for both suggestions once we converge on a design :)

julihoh:
BTW, on a similar note: 
`queue_new_entry` and `queue_get` both receive `&CStr` arguments where these arguments are representing file paths. There is whole load of trickiness around handling those kinds of strings in Rust. Ideally, the user may want to use these strings to do file operations (like reading from the file). In order to do that, they need an [`OsStr`, which is neither a `str` nor a `CStr`](https://doc.rust-lang.org/std/ffi/struct.OsString.html). 

If we just assume, that those strings are valid file paths, we could just convert them to OsStr in the bindings using the same buffer technique as for the conversion between CString and String (ie. storing in FFIContext).

vanhauser-thc:
But isn’t cstring wrong? It’s binary data

julihoh:
What exactly is binary data?

vanhauser-thc:
I am not sure you understand me or I understand you :)

fuzzing data, the data that is being mutated and returned to afl-fuzz, is not a cstring, but a binary data blob, which would translate to an u8 vector. not String, not CString.

Filenames are CString in C, and should be converted to OsStr in rust. they are guaranteed to be valid as otherwise afl-fuzz would hand over filenames that cannot exist.

julihoh:
yeah sorry, I didn't get what you meant :)

> Filenames are CString in C, and should be converted to OsStr in rust. they are guaranteed to be valid as otherwise afl-fuzz would hand over filenames that cannot exist.

Agreed. I'll cook up a PR soon!

domenukk:
> That would certainly be more convenient. However, converting between `str` and `CStr` is non-trivial:
> 
> 1. the conversion can fail (rust strings can contain null-bytes, whereas c strings can not)

Unlikely, Rust Strings are UTF8 internally, and utf8 doesn't allow 0 bytes.

> 2. the conversion incurs at least a memcpy(-ish)

The user would likely still have to do it, though

> > c) Same as a), we could take `str` as returns, auto-converting in the bindings, that could be easier to use
> > what do you think?
Sounds good I think :)

julihoh:
> > 1. the conversion can fail (rust strings can contain null-bytes, whereas c strings can not)
> 
> Unlikely, Rust Strings are UTF8 internally, and utf8 doesn't allow 0 bytes.

I wish I'd be making that shit up :D 

> [In Rust, strings are always valid UTF-8, which may contain zeros.](https://doc.rust-lang.org/std/ffi/struct.OsString.html)

Also, apparently there _is_ [a variant of UTF-8, which can't have null-bytes](https://en.wikipedia.org/wiki/UTF-8#Modified_UTF-8), but it seems like its not used in Rust.

> > 1. the conversion incurs at least a memcpy(-ish)
> 
> The user would likely still have to do it, though

Agreed

PR #755, opened 2021-02-21
Improve LLVM build instructions

rchildre3:
* Enable shell highlighting on code block
* Shallow clone of source due to extensive history
* Line break and sort the CMake options for visibility
* Disable most extraneous options (e.g. docs, tests, benchmarks,
  clang-tools-extra, OpenCL interface)
* Only build for the host architecture by default
* Support other sub-make interfaces, like the recommended Ninja Build
  System
* Harden against paths with spaces
* Prefer linking against the newly built LLVM libraries by prepending to
  LD_LIBRARY_PATH


vanhauser-thc:
yes that was a good choice. LGTM

PR #770, opened 2021-02-28
Fix printf specifiers of (s)size_t in QAsan hooks

rchildre3:
* size_t specifier is %zu or %zx
* ssize_t specifier is %zd

Helpful for cross compiling


vanhauser-thc:
thank you!

PR #772, opened 2021-03-01
WIP ergonomic improvements for rust custom mutators

julihoh:
Context: #752

- [x] allow returning of native rust `str` instead of `CStr`
- [x] receive file paths as `OsStr` instead of `CStr`
- [x] provide default implementation for `CustomMutator::describe` [like this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7efca3261d789bd0e4e732abf26cfc71)
- [x] decide on how to handle incompatibilities between `CStr` and `str`. Specifically, how to handle [this error](https://doc.rust-lang.org/std/ffi/struct.NulError.html)

# Optional
- [ ] allow returning owned strings from `str`-returning methods


julihoh:
okay, I verified everything works so far. I'd still need some input on the last question regarding the error handling @domenukk

domenukk:
In that case, it's fine to panic, even though it may theoretically be legal, it effectively still means your code is broken, right?

julihoh:
Yeah, I agree the error case is somewhat weird. I can't find a legit reason to be using the null character in utf-8. I updated the comment accordingly, as the code was already written to be panicking anyways :) 

This should be ready to review then!

domenukk:
Looks great, thanks! For path names, we could consider a `Path` instead of `OSStr`? Or is there anything useful users can do with OSString?
I guess the issue would be that it may internally move around memory again?

julihoh:
Good catch, I didn't think of that...

I think using paths as opposed to OsStr should be zero cost. I'll commit the necessary changes soon

domenukk:
Sweet! :)
Should we maybe add a github action that runs build and test for the binding, too?
Shall I merge without the owned str returns?

julihoh:
> Sweet! :)
> Should we maybe add a github action that runs build and test for the binding, too?

I added a GitHub action.
It runs `cargo check` and `cargo test` on the bindings. It also tests the feature flag configuration (afl_internals).
I haven't worked much with actions before, so I'm not sure I'm doing it right...

> Shall I merge without the owned str returns?

Yeah, I think maybe its a bit premature to add something like that and it would require yet another generic parameter. I think the API is complicated enough for now. I'd suggest we see whether it comes up in natural use. Adding it then shouldn't be too much of an issue.

domenukk:
It it easy enough to use? Do the return string borrows have to be `'static` now? Or live as long as the mutator? Or does it work through some rust magic?
Sorry if that's a stupid question 😀

julihoh:
TBH, I haven't really thought about it 😬 

here's what I think (without checking):
all &str returning methods have only one reference parameter (a reference to self). Therefore, the (elided) lifetime of the returned reference is the (elided) lifetime of self. [reference](https://doc.rust-lang.org/reference/lifetime-elision.html). Therefore, returning anything from 'self is allowed (eg: storing a String in self and returning a reference to it). 'static is allowed regardless ('static outlives any lifetime).

To be concrete (all pseudocode):
Should work:
```rust
struct Mutator {
     introspection: String,
}

impl mutator for Mutator {
      fn introspect(&mut self) -> &str {
           self.introspection = format!("Oh hi {}!", "Mark");
           return &self.introspection;
      }
}
```
Should work:
```rust
struct Mutator;

impl mutator for Mutator {
      fn describe(&mut self) -> &str {
           return "MyAwesomeMutator";
      }
}
```
Should not work:
```rust
struct Mutator;

impl mutator for Mutator {
      fn describe(&mut self) -> &str {
           return "MyAwesomeMutator".to_string();
      }
}
```
because the string is owned and returning a reference from it is illegal, because the string will be dropped by the end of the function.

julihoh:
my initial idea to solving this issue was to return a generic from the function that can be turned into a reference of type of str via the AsRef trait. This doesn't change the behaviour for returning &str, but it allows returning String (the String will move into the caller, where it will be dropped). This may be more ergonomic, but has 2 downsides:
1. it introduces yet another generic (I'm not a fan of going ham with generics)
2. it encourages a severely inefficient pattern, because dropping a string will deallocate it (implying it had to be allocated before). Allocating a fresh string for every call is not particularly efficient. The current situation forces the String to be stored in the struct, making reuse much easier.

Issue #18, opened 2019-07-16
FairFuzz (afl-rb) integration 


dkasak:
Has there been any discussion on incorporation of the [FairFuzz](https://github.com/carolemieux/afl-rb/) project by Caroline Lemieux into afl++? In my tests, its approach was very effective, certainly comparable in value to the aflfast schedules. 

That said, the code was slightly messy in a few places the last time I looked so perhaps it wouldn't be trivially mergeable and would require some manual porting.

What do you think?

vanhauser-thc:
there are so many afl spinoffs its hard to find those which are really making it better, so any recommendation is helpful :)
I see that trunk was last updated in january and then there is the testcomp branch which has a lot of changes.
Do you know which one of the two would be better to use?

dkasak:
From a quick look, I think the testcomp branch is just stuff related to the Test-Comp 2019 competition and isn't useful to us. I was using the master branch.

hexcoder-:
Hi,
hexcoder- here,

no discussion yet, but I have been just looking into it.

My opinion is :
the byte mask to protect input structure
is a bit like light weight taint tracking
and I can imagine positive effects.

However in my earlier tests, I saw afl-rb stalled frequently,
and I don’t know if this has been resolved.
Do you?

Thanks, hexcoder-

Von: Denis Kasak <notifications@github.com>
Gesendet: Mittwoch, 17. Juli 2019 00:27
An: vanhauser-thc/AFLplusplus <AFLplusplus@noreply.github.com>
Cc: Subscribed <subscribed@noreply.github.com>
Betreff: [vanhauser-thc/AFLplusplus] FairFuzz (afl-rb) integration (#18)


Has there been any discussion on incorporation of the FairFuzz<https://github.com/carolemieux/afl-rb/> project by Caroline Lemieux into afl++? In my tests, its approach was very effective, certainly comparable in value to the aflfast schedules.

That said, the code was slightly messy in a few places the last time I looked so perhaps it wouldn't be mergeable trivially cleanly and would require some manual porting of patches (particularly so because afl++ continued to evolve from afl, on which both were based).

What do you think?

dkasak:
[...]

Issue #16, opened 2019-07-16
llvm_mode/README.instrim missing


dkasak:
The file `llvm_mode/README.instrim` is mentioned [here](https://github.com/vanhauser-thc/AFLplusplus/blob/0af9f664db5b68e1eb8daa5257410f2e9c400076/docs/env_variables.txt#L123) but it appears it's not committed.

vanhauser-thc:
correct, that got lost when a commit fuckup happened. luckily I still have it, its in now.

vanhauser-thc:
thanks!

Issue #15, opened 2019-07-15
MOpt-AFL mutations selection optimization


andreafioraldi:
```MOpt-AFL is a AFL-based fuzzer that utilizes a customized Particle Swarm Optimization (PSO) algorithm to find the optimal selection probability distribution of operators with respect to fuzzing effectiveness.```

https://github.com/puppet-meteor/MOpt-AFL

This seems very interesting and the results showed in the paper confirm the validity of this technique. It does not change the way about how AFL works and maybe can be integrated here with a bit of source code refactoring.

andreafioraldi:
A good idea is to ask directly to the authors if they want merge they changes to AFL directly here.

hexcoder-:
I don't understand the issue here. MOpt is already included, first in branch MOpt and then merged to master...

andreafioraldi:
Lol I missed it, GG for the rapid merge

vanhauser-thc:
yes I think @andreafioraldi missed it :)

I did a lot of testing and verified that MOpt is really good and then added it

also instrim is now in trunk/master. makes llvm_mode much, much faster (but slightly lower coverage)

vanhauser-thc:
@andreafioraldi but if you ever see something interested, integrate it or create an issue. I always look and @hexcoder- too, but of course we can miss stuff :)

Issue #10, opened 2019-07-02
Inefficient llvm_mode generated code for neverZero_counters branch


vanhauser-thc:
the shared memory map between afl-fuzz and the target contains 65536 one-byte buckets.

In the current afl implementation these can wrap and in rare occasions result in a 0 byte value which afl-fuzz ignores.

in tests an implementation that saturates (leaves at max 255 value) is bad for performance and bad for path discovery.

An implementation that skips the 0 when wrapping (and goes directly to 1) improves path discovery instead, so this is the road to go.

*Problem*: the code in llvm_mode for this is inefficient.

This is what is generated :
```
mov     al, [r8+r11]
mov     ebx, eax
add     bl, 1
adc     al, 1
mov     [r8+r11], al
```
instead of what would be efficient:
```
mov     al, [r8+r11]
add     al, 1
adc     al, 0
mov     [r8+r11], al
```

Code can be found in https://github.com/vanhauser-thc/AFLplusplus/tree/neverZero_counters (and then in llvm_mode/afl-llvm-pass.so.cc

vegard:
Couldn't you even do something like this to save an instruction?

    mov     al, [r8+r11]
    add     al, 1
    adc     [r8+r11], al

vegard:
Second try :-) Why not do it directly with just 2 instructions:

    add [r8+r11], 1
    adc [r8+r11], 0

hexcoder-:
Yes, I would love to test/benchmark that, but we talk llvm IR here and I could not yet coerce llvm IR into producing that assembler output. Any hints on that?

vanhauser-thc:
Ok this has been resolved. its an issue an llvm and now has a fix:
https://reviews.llvm.org/D64190#change-e3PF3j20UdNS

not sure for which llvm versions that fix will be in - but its nothing we will be able to expect on a system in the next year at a minimum.

@hexcoder- 
Proposal: keep the afl-gcc implementation as the performance impact is minimal and the feature beneficial, and for llvm_mode we keep the env var.
I will rewrite the pass.

and then we merge?

hexcoder-:
ok, fine with me. I will update and then push the portability patch (sed -> tr and version check llvm<->clang)

vanhauser-thc:
done :)

PR #775, opened 2021-03-02
Filter out std string functions taking non-char character types

stbergmann:
This had caused an assert SIGABRT with LibreOffice (see <https://git.libreoffice.org/core/+/e2c9ac71cec0f205b1d4864538e8158c22558296%5E%21> "ofz#30767 Build-Failure") at

> #3  0x00007ffff7a07026 in __GI___assert_fail (assertion=0x7ffff79c3b98 "isString() && \"Not a string\"", file=0x7ffff79c3800 "~/llvm/inst/include/llvm/IR/Constants.h", line=661, function=0x7ffff79c3bb5 "llvm::StringRef llvm::ConstantDataSequential::getAsString() const") at /usr/src/debug/glibc-2.32-37-g760e1d2878/assert/assert.c:101
> #4  0x00007ffff79b9dd1 in llvm::ConstantDataSequential::getAsString (this=0xcb75f90) at ~/llvm/inst/include/llvm/IR/Constants.h:661
> #5  0x00007ffff79b8645 in (anonymous namespace)::AFLdict2filePass::runOnModule (this=0xd175d50, M=...) at ~/AFLplusplus/instrumentation/afl-llvm-dict2file.so.cc:406
> #6  0x000000000550fb63 in (anonymous namespace)::MPPassManager::runOnModule (M=..., this=<optimized out>) at ~/llvm/llvm-project/llvm/lib/IR/LegacyPassManager.cpp:1550
> #7  llvm::legacy::PassManagerImpl::run (this=0x9925a90, M=...) at ~/llvm/llvm-project/llvm/lib/IR/LegacyPassManager.cpp:541
> #8  0x000000000550feb9 in llvm::legacy::PassManager::run (this=this@entry=0x7fffffff91c0, M=...) at ~/llvm/llvm-project/llvm/lib/IR/LegacyPassManager.cpp:1677
> #9  0x000000000653efb3 in (anonymous namespace)::EmitAssemblyHelper::EmitAssembly (this=this@entry=0x7fffffff9670, Action=Action@entry=clang::Backend_EmitObj, OS=std::unique_ptr<llvm::raw_pwrite_stream> = {...}) at ~/llvm/llvm-project/clang/lib/CodeGen/BackendUtil.cpp:1015
> #10 0x0000000006540856 in clang::EmitBackendOutput (Diags=..., HeaderOpts=..., CGOpts=..., TOpts=..., LOpts=..., TDesc=..., M=0x944b6f0, Action=<optimized out>, OS=...) at /usr/include/c++/10/bits/move.h:76
> #11 0x000000000689383c in clang::BackendConsumer::HandleTranslationUnit (this=0x944a210, C=...) at ~/llvm/llvm-project/clang/include/clang/Basic/TargetInfo.h:1076
> #12 0x00000000078fe1c9 in clang::ParseAST (S=..., PrintStats=<optimized out>, SkipFunctionBodies=<optimized out>) at ~/llvm/llvm-project/clang/lib/Parse/ParseAST.cpp:171
> #13 0x00000000067b9729 in clang::FrontendAction::Execute (this=this@entry=0x941b1a0) at ~/llvm/llvm-project/clang/lib/Frontend/FrontendAction.cpp:949
> #14 0x00000000066f6586 in clang::CompilerInstance::ExecuteAction (this=this@entry=0x940f390, Act=...) at ~/llvm/llvm-project/clang/lib/Frontend/CompilerInstance.cpp:949
> #15 0x000000000686ecfb in clang::ExecuteCompilerInvocation (Clang=Clang@entry=0x940f390) at ~/llvm/llvm-project/clang/lib/FrontendTool/ExecuteCompilerInvocation.cpp:278
> #16 0x00000000039f6f04 in cc1_main (Argv=..., Argv0=0x7fffffffcc0c "~/llvm/inst/bin/clang-13", MainAddr=MainAddr@entry=0x39f0a60 <GetExecutablePath[abi:cxx11](char const*, bool)>) at ~/llvm/llvm-project/clang/tools/driver/cc1_main.cpp:246
> #17 0x00000000039f054d in ExecuteCC1Tool (ArgV=...) at ~/llvm/llvm-project/clang/tools/driver/driver.cpp:330
> #18 0x00000000039f25c5 in main (argc_=<optimized out>, argc_@entry=145, argv_=<optimized out>, argv_@entry=0x7fffffffc3d8) at ~/llvm/llvm-project/clang/tools/driver/driver.cpp:407

when (in frame #5) FuncName is "_ZNKSt17basic_string_viewIDsSt11char_traitsIDsEE4findEPKDsm" (i.e.,

> std::basic_string_view<char16_t, std::char_traits<char16_t> >::find(char16_t const*, unsigned long) const

) and thus isStdString is true.


vanhauser-thc:
Thanks for the notification!
Please note we only accept PRs to dev so I switched the base.

If I understand the issue correctly, the instrumentation stumbles over unicode?
then your fix would not be good, because those data bytes would be missed.
so instead of "continue" it needs different functionality to extract the data, hmmm.

(note to self: we have similar code in SanitizeCoverageLTO.cpp which also needs to be fixed.)

stbergmann:
> If I understand the issue correctly, the instrumentation stumbles over unicode?

yes, see the "char16_t" in the demangled FuncName

vanhauser-thc:
I am deep in a different project, can you quickly help me with a piece of code that lets me replicate the issue?
```
#include <stdio.h>
#include <string>
#include <string_view>

int main() {
    const std::u16string_view deg = u"deg";
    const std::u16string_view grad = u"grad";
    const std::u16string_view rad = u"rad";
    std::u16string rString = u"grad";

    if (std::u16string_view::npos != rString.find(deg))
    {
        printf("x\n");
    }
    else if (std::u16string_view::npos != rString.find(grad))
    {
        printf("y\n");
    }
    else if (std::u16string_view::npos != rString.find(rad))
    {
        printf("z\n");
    }
    return 0;
}
```
tells me
```
test.c:6:16: error: ‘u16string_view’ in namespace ‘std’ does not name a type; did you mean ‘u16string’?
    6 |     const std::u16string_view deg = u"deg";
```

stbergmann:
> ```
> test.c:6:16: error: ‘u16string_view’ in namespace ‘std’ does not name a type; did you mean ‘u16string’?
>     6 |     const std::u16string_view deg = u"deg";
> ```

You presumably compile as C++14 or older, but `std::u16string_view` was only introduced in C++17.

vanhauser-thc:
@stbergmann 
my example code does not crash and somehow seems to behave very different to yours. I blind-coded a fix, but sadly cannot test if it works (by not crashing and still obtaining the values).
Can you test it somehow? -> c269c3977ccb96710a2488b19c72bae51832a827

stbergmann:
> Can you test it somehow? -> [c269c39](https://github.com/AFLplusplus/AFLplusplus/commit/c269c3977ccb96710a2488b19c72bae51832a827)

While that does not crash, I don't think it makes much sense:  When run on LibreOffice's `sax/source/tools/converter.cxx`, it generates the same `$AFL_LLVM_DICT2FILE` output as my patch here, and that output does not contain any of the three relevant UTF-16 string literals `u"deg"`, `u"grad"`, `u"rad"`.  When
```
Str2 = Array->getRawDataValues().str();
```
(in `AFLdict2filePass::runOnModule`) now forces the content of such a UTF-16 string literal into a `std::string`, depending on target endianness, either `Str2[0]` or `Str2[1]` will be `NUL`, so the later
```
if (optLen < 3)  // too short? skip
  continue;
``` 
(also in `AFLdict2filePass::runOnModule`) will ignore it anyway.

stbergmann:
(Re "I don't think it makes much sense":  Or maybe you're fine with that, producing endianness-dependent output for just those UTF-16 and UTF-32 string literals in the source that happen to not have a null byte among the leading bytes of their `getRawDataValues()` representation.  I know too little what all this code is actually about.)

vanhauser-thc:
> > Can you test it somehow? -> [c269c39](https://github.com/AFLplusplus/AFLplusplus/commit/c269c3977ccb96710a2488b19c72bae51832a827)
> 
> While that does not crash, I don't think it makes much sense: When run on LibreOffice's `sax/source/tools/converter.cxx`, it generates the same `$AFL_LLVM_DICT2FILE` output as my patch here, and that output does not contain any of the three relevant UTF-16 string literals `u"deg"`, `u"grad"`, `u"rad"`. When
> 
> ```
> Str2 = Array->getRawDataValues().str();
> ```
> 
> (in `AFLdict2filePass::runOnModule`) now forces the content of such a UTF-16 string literal into a `std::string`, depending on target endianness, either `Str2[0]` or `Str2[1]` will be `NUL`, so the later
> 
> ```
> if (optLen < 3)  // too short? skip
>   continue;
> ```

but this is not how optLen is generated:
```
          if (HasStr1)
            thestring = Str1;
          else
            thestring = Str2;

          optLen = thestring.length()
```
so 8 characters of unicode 16 = std::string with a length of 16 bytes, null bytes are just normal characters.

```
(Re "I don't think it makes much sense": Or maybe you're fine with that, producing endianness-dependent output for just those UTF-16 and UTF-32 string literals in the source that happen to not have a null byte among the leading bytes of their getRawDataValues() representation. I know too little what all this code is actually about.)
```
that is totally fine. you cannot run the instrumented binary on a different endian platform either :)

When I try to compile libreoffice/sax/source/tools/converter.cxx it does fail because an include is not found that does not exist in the checkout, I guess it is generated by the build system ... 
sigh ... looks like I have to look further how to get those unicode strings ...

stbergmann:
> but this is not how optLen is generated:
> 
> ```
>           if (HasStr1)
>             thestring = Str1;
>           else
>             thestring = Str2;
> 
>           optLen = thestring.length()
> ```
> 
> so 8 characters of unicode 16 = std::string with a length of 16 bytes, null bytes are just normal characters.

From just having scanned through the code, I thought it was
```
// ensure we do not have garbage
size_t offset = thestring.find('\0', 0);
if (offset + 1 < optLen) optLen = offset + 1;
thestring = thestring.substr(0, optLen);
```
that would shrink both `optLen` and `thestring` to anything up to the first embedded null byte.

> When I try to compile libreoffice/sax/source/tools/converter.cxx it does fail because an include is not found that does not exist in the checkout, I guess it is generated by the build system ...
> sigh ... looks like I have to look further how to get those unicode strings ...

A very simple reproducer I came up with is
```
#include <string>
void f(std::u16string s) {
    s.find(u"foobar");
    s.find(u"\u0123\u4567\u89AB\uCDEF");
}
```
With your patch, it drops `u"foobar"` (see above) but includes something like `"#\x01gE\xab\x89\xef\xcd\x00"` for `u"\u0123\u4567\u89AB\uCDEF"` in the output.

vanhauser-thc:
> From just having scanned through the code, I thought it was
> 
> ```
> // ensure we do not have garbage
> size_t offset = thestring.find('\0', 0);
> if (offset + 1 < optLen) optLen = offset + 1;
> thestring = thestring.substr(0, optLen);
> ```
> 
> that would shrink both `optLen` and `thestring` to anything up to the first embedded null byte.

that what happens when you try to fix code while in deep in something else :) good catch.
I fixed this now all works perfectly, thank you!
will push the new commit id to oss-fuzz now.

> A very simple reproducer I came up with is
> ```
> #include <string>
> void f(std::u16string s) {
>     s.find(u"foobar");
>     s.find(u"\u0123\u4567\u89AB\uCDEF");
> }
> ```

that helped a lot!

> With your patch, it drops `u"foobar"` (see above) but includes something like `"#\x01gE\xab\x89\xef\xcd\x00"` for `u"\u0123\u4567\u89AB\uCDEF"` in the output.

`"\u0123\u4567\u89AB\uCDEF"` => `"#\x01gE\xab\x89\xef\xcd\x00"`
is correct, only binary characters are encoded. e.g. \u0123 is \x23 \x01 and \x23 is `#`. `\u4567` becomes `gE` etc.

stbergmann:
> I fixed this now all works perfectly, thank you!
> will push the new commit id to oss-fuzz now.

So lets abandon this one then.  Thanks for your quick reaction!

PR #776, opened 2021-03-02
Add AFL_QEMU_EXCLUDE_RANGES

realmadsci:
This environment variable allows rejection of
specific regions from instrumentation.

It takes priority over AFL_INST_LIBS and AFL_QEMU_INST_RANGES,
so it can be used to poke a "hole" in previously included sections.

This requires the qemuafl PR first, and merely adds readme and afl_environment_variables support to that:
https://github.com/AFLplusplus/qemuafl/pull/16


andreafioraldi:
LGTM, I will test it a bit tomorrow

Issue #692, opened 2021-01-19
Bump Submodule pseudo-pointers


rchildre3:
`qemuafl` and `unicornafl` submodule pointers pointers moved in 292f91a55f6af5e06b41b429e55f65b7df4d8d16, but the pseudo-pointers (`QEMUAFL_VERSION` and `UNICORNAFL_VERSION`) have not been changed

domenukk:
Great catch, thanks. Fixed in 473b3e5
@vanhauser-thc use `./unicorn_mode/update_uc_ref.sh` and `./qemu_mode/update_ref.sh` if you ever need to do it (only works in git for obvious reasons)

vanhauser-thc:
I did that for the cmplog_merge PR. what rchildred3 reported was not my doing :)

PR #777, opened 2021-03-03
Use a pre-commit hook to keep submodules in-sync

### LINKED ISSUE #773: qemuafl submodule pointers out-of-sync
**Describe the bug**
qemuafl submodule pointer and pseudo pointer out-of-sync on dev as of ad7a7fcf075c617e09cb516da000b244be161093

**Expected behavior**
qemuafl submodule pointer and pseudo pointer are in-sync

**Recommendation**
Add rule in pre-commit hook that verifies that `git submodule status` and pseudo pointer file match

rchildre3:
Will prevent thing like #692 and #773 is enabled by the developers

Closes #773


vanhauser-thc:
I dont think that is what @andreafioraldi wants ...
otherwise we could just force a fresh sync to the repo in the build script.
The argument back then was that the latest commit is not necessarily one that works.
also @domenukk for unicornafl

andreafioraldi:
yep

rchildre3:
> force a fresh sync

Possibly, though I have already lost work due to this line of code.  I would just ask for not syncing if there is an uncommited (i.e. dirty) submodule
https://github.com/AFLplusplus/AFLplusplus/blob/07884e00549b247d59c6289e58808aae401eeb79/qemu_mode/build_qemu_support.sh#L135

> latest commit is not necessarily one that works

`HEAD` is not the latest commit, it is just the current commit (i.e. wherever the working directory's file's came from)

vanhauser-thc:
It is not that easy.
We support installations that are not a git checkout, but e.g. an afl++ .tar.gz download.
again for people who are on git, we might need a git stash so we can git pull a potentially new version.
The current qemuafl does not patch anything anymore AFAIK - that was different in the original qemu_mode we had, where the we hat a few .diff that we applied.
maybe we can get rid of the git stash @andreafioraldi ?

andreafioraldi:
i guess yes, btw with git stash you don't loose yout work, you can do git stash pop and rebase it

domenukk:
Especially the stash drop is a foot gun, maybe we can get rid of that at least?

andreafioraldi:
yes we can safely remove sh -c 'git stash && git stash drop' 1>/dev/null 2>/dev/null

i forgot that it was also doing drop

andreafioraldi:
@domenukk that code is copied from unicornafl, so you are the author problably 😅

rchildre3:
@vanhauser-thc 
> We support installations that are not a git checkout

How are tarball installations affected by any of the changes in this MR?

@andreafioraldi 
Agreed `git stash push`s are understandable, but `git stash drop`s are not

rchildre3:
I do not have a preference as to what solution looks like, but it appears that given the assumptions that

1. The submoulde pointers and pseudo pointer files need a human to enforce that they are in sync to run `update_ref.sh`
2. #692 and #773 are a result of that not happening

The sync's potentially should be programmatically enforced. Would the developers be amenable to other solutions?

vanhauser-thc:
we only update the reference (by hand) when we are sure that the target repo is stable and not WIP.
I removed the stash in 29cccede77d55c528212c8a2c319a938cbd82977

andreafioraldi:
stash drop removed, and now you can set NO_CHECKOUT=1 to skip the git checkout when invoking the build script.

PR #778, opened 2021-03-04
This fixes 3 different crash issues

vanhauser-thc:


instrumentation/afl-compiler-rt.o.c
```diff
@@ -350,17 +369,18 @@ static void __afl_map_shm(void) {
 
     }
 
+    close(shm_fd);
+
```
hexcoder-, line 374: suggest also moving:
`shm_fd = -1;`
from line 376

src/afl-fuzz-queue.c
```diff
@@ -198,34 +198,35 @@ void create_alias_table(afl_state_t *afl) {
   while (nS)
     afl->alias_probability[S[--nS]] = 1;
 
-#ifdef INTROSPECTION
-  u8 fn[PATH_MAX];
-  snprintf(fn, PATH_MAX, "%s/introspection_corpus.txt", afl->out_dir);
-  FILE *f = fopen(fn, "a");
-  if (f) {
+  /*
```
hexcoder-, line 201: problems with introspection?

hexcoder- (approved):
LGTM

instrumentation/afl-compiler-rt.o.c
vanhauser-thc: moved it, however should rather remove it, it is not used afterwards (if, it would be a bug)

src/afl-fuzz-queue.c
vanhauser-thc: generated 1GB data in just a few minutes for some targets ... have to do that introspection different ...

PR #779, opened 2021-03-05
Restore correct CTX behaviour and enable K-CTX coverage for LLVM Classic

andreafioraldi:
The old CTX is now renamed as CALLER, and CTX is now true context-sensitive branch coverage.

As this kind of feedback causes many collisions I introduced the K-CTX branch coverage that takes into account only the top-K functions in the calling context.

### DIFF
--- a/include/envs.h
+++ b/include/envs.h
@@ -81,6 +81,7 @@ static char *afl_environment_variables[] = {
     "AFL_LLVM_CMPLOG",
     "AFL_LLVM_INSTRIM",
     "AFL_LLVM_CTX",
+    "AFL_LLVM_CTX_K",
     "AFL_LLVM_DICT2FILE",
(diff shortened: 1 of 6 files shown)


andreafioraldi:
Any considerations before merging?

vanhauser-thc:
ctx-1 is the same as caller - but I guess slower? If so maybe only allow values 2+? if ctx-1 is faster or same speed, then maybe getting rid of caller instead?

andreafioraldi:
when k=1 ctx-1 is automatically converted to caller, that is faster.
What we can do is to distinguish them in the llvm pass but not in afl-cc, having just ctx and k-ctx

vanhauser-thc:
> when k=1 ctx-1 is automatically converted to caller, that is faster.

cool then there is nothing to do IMHO :)

PR #785, opened 2021-03-09
doc: Add lld to required dependencies

maflcko:
Without this dependency LTO build will fail: `GNUmakefile.llvm:216: ld.lld not found, cannot enable LTO mode`.

Also, remove duplicate `clang`.


vanhauser-thc:
note that we only accept PRs to the dev branch, I changed the destination.

very true, thank you.

PR #786, opened 2021-03-09
push to stable

vanhauser-thc:
pre-release


afl-cmin
```diff
@@ -358,12 +360,12 @@ BEGIN {
...
   # Make sure that we're not dealing with a directory.
 
-  if (0 == system("test -d "in_dir"/"first_file)) {
+  if (0 == system("test -d ""\""in_dir"/"first_file"\"")) {
     print "[-] Error: The input directory is empty or contains subdirectories - please fix." > "/dev/stderr"
     exit 1
   }
 
-  if (0 == system("ln "in_dir"/"first_file" "trace_dir"/.link_test")) {
+  if (0 == system("ln \""in_dir"/"first_file"\" "trace_dir"/.link_test")) {
```
domenukk, line 368: What if `"` is part of the file name? Shouldn't this be something like `'` and be properly escaped?

src/afl-cc.c
```diff
@@ -1613,6 +1693,7 @@ int main(int argc, char **argv, char **envp) {
           "  AFL_DONT_OPTIMIZE: disable optimization instead of -O3\n"
           "  AFL_NO_BUILTIN: no builtins for string compare functions (for "
           "libtokencap.so)\n"
+          "  AFL_NOOP: behave lik a normal compiler (to pass configure tests)\n"
```
domenukk: lik*e*

domenukk (approved):

afl-cmin
domenukk: @hexcoder- ^ what do you think?

instrumentation/afl-compiler-rt.o.c
```diff
@@ -457,6 +459,17 @@ static void __afl_map_shm(void) {
 
     if (!__afl_area_ptr) { __afl_area_ptr = __afl_area_ptr_dummy; }
 
+    if (__afl_debug)
```
domenukk: Some brackets here would be greatly appreciated. Reduces wtf/line :)

PR #787, opened 2021-03-09
setting general schedule priority for performance on Mac.

devnexen:


vanhauser-thc:
isnt user interactive more performant setting?

devnexen:
I admit I struggled choosing between those two but felt more fit for the nature of AFL++.

vanhauser-thc:
traditionally afl tries to grab as most resources as possible to be as fast as possible with fuzzing.
I am not a Mac user  - I guess it cannot be run with the Desktop?
in that case I would propose that if a Mac desktop is used then the current prio is set, if a Mac server is used, then the interactive user prio?

devnexen:
oh it works even with non server edition, just tried a fuzz run with interactive setting seems ok.

vanhauser-thc:
thank you!

PR #796, opened 2021-03-10
Revert to proper installation path of manual pages

wykwit:
`MAN_PATH` changed in #529 has been overwritten in 4b3ad5f which may cause issues with packaging.


domenukk:
I think @hexcoder- could know more

vanhauser-thc:
git blame says it was me, but that is nothing I would normally change ... so a mistake.
but fixed it directly because I saw this PR after the the comment in the commit, sorry

PR #802, opened 2021-03-10
doc: Clarify memory limits in Readme

maflcko:
See https://github.com/AFLplusplus/AFLplusplus/blob/3bb12b456f0d43bb5e2d9d8f570b7c837666fa09/include/config.h#L121


vanhauser-thc:
thank you!

PR #813, opened 2021-03-13
 doc: Clarify default value of N iterations in aflpp_driver 

maflcko:
See

https://github.com/AFLplusplus/AFLplusplus/blob/e5bdba4b9f22c0f2e4ff60ffc9bfa8fbaeb586d2/utils/aflpp_driver/aflpp_driver.c#L228


vanhauser-thc:
ty

PR #817, opened 2021-03-15
Make @@ handling more flexible

realmadsci:
Several of the AFL utilities currently require the @@ command line argument to be _its own_ argument rather than just a part of an argument. The documentation doesn't mention this limitation, so this goes a little bit contrary to what a user would expect.

Also, several of the tools already accept @@ as a partial argument and correctly replace the text with the input filename chosen by AFL++. So making them all consistent would be nice.

These changes fix a few of the utils to accept @@ as part of a command line argument rather than requiring it to be a separate argument. This is handy for targets that accept a filename in a format like `--file=@@` or for allowing the user to quote the file in case there are spaces in the path like `-i "@@"`.

### DIFF
--- a/src/afl-showmap.c
+++ b/src/afl-showmap.c
@@ -72,8 +72,7 @@ static u8 *in_data,                    /* Input data                        */
 static u64 total;                      /* tuple content information         */
 static u32 tcnt, highest;              /* tuple content information         */
 
-static u32 in_len,                     /* Input data length                 */
-    arg_offset;                        /* Total number of execs             */
+static u32 in_len;                     /* Input data length                 */
 
 static u32 map_size = MAP_SIZE;
 
@@ -738,7 +737,7 @@ int main(int argc, char **argv_orig, char **envp) {
 
   // TODO: u64 mem_limit = MEM_LIMIT;                  /* Memory limit (MB) */
 
-  s32    opt, i;
+  s32    opt;
   u8     mem_limit_given = 0, timeout_given = 0, unicorn_mode = 0, use_wine = 0;
   char **use_argv;
 
@@ -986,7 +985,7 @@ int main(int argc, char **argv_orig, char **envp) {
   if (getenv("AFL_DEBUG")) {
 
     DEBUGF("");
-    for (i = 0; i < argc; i++)
+    for (int i = 0; i < argc; i++)
       SAYF(" %s", argv[i]);
     SAYF("\n");
 
@@ -1012,12 +1011,30 @@ int main(int argc, char **argv_orig, char **envp) {
 
   }
 
+
   if (in_dir) {
 
-    detect_file_args(argv + optind, "", &fsrv->use_stdin);
+    /* If we don't have a file name chosen yet, use a safe default. */
+    u8 *use_dir = ".";
+
+    if (access(use_dir, R_OK | W_OK | X_OK)) {
+
+      use_dir = get_afl_env("TMPDIR");
+      if (!use_dir) { use_dir = "/tmp"; }
+
+    }
+
+    stdin_file = at_file ? strdup(at_file)
+                         : (char *)alloc_printf("%s/.afl-showmap-temp-%u",
+                                                use_dir, (u32)getpid());
+    unlink(stdin_file);
+
+    // If @@ are in the target args, replace them and also set use_stdin=false.
+    detect_file_args(argv + optind, stdin_file, &fsrv->use_stdin);
 
   } else {
 
+    // If @@ are in the target args, replace them and also set use_stdin=false.
     detect_file_args(argv + optind, at_file, &fsrv->use_stdin);
 
   }
@@ -1039,14 +1056,6 @@ int main(int argc, char **argv_orig, char **envp) {
   } else {
 
     use_argv = argv + optind;
-
-  }
-
-  i = 0;
-  while (use_argv[i] != NULL && !arg_offset) {
-
-    if (strcmp(use_argv[i], "@@") == 0) { arg_offset = i; }
-    i++;
 
   }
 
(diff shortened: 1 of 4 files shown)


utils/crash_triage/triage_crashes.sh
```diff
@@ -90,8 +90,9 @@ for crash in $DIR/crashes/id:*; do
 
   for a in $@; do
 
-    if [ "$a" = "@@" ] ; then
-      use_args="$use_args $crash"
+    if echo "$a" | grep -qF '@@'; then
```
vanhauser-thc, line 93: I think this can go wrong if $a is e.g. -e... as this then interpreted as an option to echo?

src/afl-showmap.c
```diff
@@ -986,7 +985,7 @@ int main(int argc, char **argv_orig, char **envp) {
   if (getenv("AFL_DEBUG")) {
 
     DEBUGF("");
-    for (i = 0; i < argc; i++)
+    for (int i = 0; i < argc; i++)
```
vanhauser-thc: some compilers need a special command line option to support this and fail otherwise (I think this is C99)

src/afl-showmap.c
```diff
@@ -1012,12 +1011,30 @@ int main(int argc, char **argv_orig, char **envp) {
 
   }
 
+
```
vanhauser-thc, line 958: below here you move when what is happening. why?

vanhauser-thc:
I see this as necessary for programs that have an `--infile=@@` like option and thinking about that I am surprised that nobody else needed this so far.
In a few days there will be a new release, as this has a potential bricking impact I will wait with accepting this PR until a) the remarks I have made are resolved (and others might have some too) and b) the 3.11c release is out.
thank you!

utils/crash_triage/triage_crashes.sh
realmadsci: $a is quoted here to prevent that?
I took the line almost verbatim from afl-cmin.bash https://github.com/AFLplusplus/AFLplusplus/blob/e5bdba4b9f22c0f2e4ff60ffc9bfa8fbaeb586d2/afl-cmin.bash#L173

src/afl-showmap.c
realmadsci: The `detect_file_args()` call is a very side-effect-heavy function. It both replaces the "@@" symbols in the argv with its second arg (`stdin_file`) and also sets the boolean in its third arg (`fsrv->use_stdin`) if it finds any.

The previous code was sending "" for the second arg to get the third arg side effect without replacing the @@ symbols, and then it would manually replace the symbols itself later (former line 1048). This required `detect_file_args()` to have yet another special option (which I removed) to tolerate a blank "" for arg 2, and it also got the @@ substitution wrong (or at least different than what `detect_file_args()` would have done) by only substituting whole indexes from argv.

I moved the "determine `stdin_file`" code up here so that `detect_file_args()` could be called in the "normal" way that it is used in other afl utils and get both "@@ replacement" and "stdin detection" effects in one pass.

src/afl-showmap.c
realmadsci: Fair point - that does require C99 or a compiler with extensions.
Then again, so does `//` comments.
Is there a minimum target compiler documentation somewhere I can review?

realmadsci: In the meantime, we can undo the changes to 741 and 989?

utils/crash_triage/triage_crashes.sh
realmadsci: On second thought you're right - both of these are probably wrong.
If we were allowed to assume bash, a safer solution would be to do `grep -qF '@@' <<< "$a"` but I'm not really sure how to do this in plain old `sh`.

realmadsci: Upon further testing, I think it is safe due to the quote. It at least takes the correct branch in the if statement:

```sh
$ if echo "-e@@" | grep -qF "@@"; then echo "found @@"; else echo "Nope"; fi
found @@
$ if echo "-e @@" | grep -qF "@@"; then echo "found @@"; else echo "Nope"; fi
found @@
$ if echo "-e" | grep -qF "@@"; then echo "found @@"; else echo "Nope"; fi
Nope
$ if echo "-n" | grep -qF "@@"; then echo "found @@"; else echo "Nope"; fi
Nope
$ if echo "--version" | grep -qF "@@"; then echo "found @@"; else echo "Nope"; fi
Nope
```

hexcoder-: POSIX `/bin/echo` and the more portable `/bin/printf` do not support any options. But GNU `/bin/echo` and the built-in command `echo` in bash both support `-n`, `-e`, `-E` making it non trivial to output a string with `-[neE]+`, e.g. `-nennen`.
So quoting `-e` does not prevent `echo` from consuming it instead of echoing it, which would spoil the targets argument.
Also echo adds an unwanted newline.
How about this instead?
```
  for a in $@; do

    case "$a" in
      *@@*)
      unset use_stdio
      use_args="$use_args `printf %s "$a" | sed -e 's<@@<'$crash'<g'`"
      ;;
      *)
      use_args="$use_args $a"
      ;;
    esac

  done

```

hexcoder- (changes requested):

utils/crash_triage/triage_crashes.sh
realmadsci: I like the use of `case` to do the pattern matching!
I'm also a fan of assuming files don't have `<` in their names, although technically linux allows anything except slash and null.
Since you mentioned printf, maybe we can use `printf %q "$crash"` to escape the filename and pass that in rather than the crazy sed slash escape that I made?

Regardless, I have shown in my earlier comment that the original code in the PR actually does what is expected. It accurately determines that "-e" and friends are _not @@_ and sends them to the other path. This code also matches what is done in cmin.bash:
https://github.com/AFLplusplus/AFLplusplus/blob/e5bdba4b9f22c0f2e4ff60ffc9bfa8fbaeb586d2/afl-cmin.bash#L173

My recommendation is to take this PR as-is and then (if desired) make another one to address potential clean-ups with both triage_crashes.sh and cmin.bash.

vanhauser-thc:
@hexcoder- can this be merged?

PR #819, opened 2021-03-15
Remove AFL_PRELOAD and AFL_USE_QASAN handlers

realmadsci:
**NOTE: This is coupled with https://github.com/AFLplusplus/qemuafl/pull/17**

These are now processed in afl-qemu-trace so that the "copy+paste" code
that is in all of the other AFL tools can be removed.

This also allows the AFL_USE_QASAN flag to work the same when used
with tools like afl-fuzz as it does with afl-qemu-trace. This is
important in situations where loading the QASAN library changes
the address of your desired entrypoint, or for crash validation
using the same environment that afl-fuzz was using.

With this change, the same set of environment variables can be used
in exactly the same way between afl-fuzz, afl-showmap, and
afl-qemu-trace, and you will get exactly the same guest environment.


realmadsci:
You will likely want to defer this one until after a release...

Issue #774, opened 2021-03-02
AFL_LIBRARY_PATH


fuzzah:
When compiling complex targets consisting of binary and multiple libraries, I'd like to fuzz everything that I have compiled: this exact binary and all the libraries bundled with it. Usually such libraries are saved in build tree in directories like "build/.libs/". But when I run target it may use libraries that available in OS itself (e.g. bzip2, libxml2). Of course my OS libraries are not same: no sanitizers, no AFL instrumentation, another version.
 
There exist `AFL_PRELOAD` which loads few manually specified libraries, but there is no **AFL_LIBRARY_PATH** which would set binary env to `LD_LIBRARY_PATH` to path that contains MANY libraries wanted by target.
 
We can set `LD_LIBRARY_PATH` for AFL++, this seems to work, but we eventually may load something into AFL++ itself which can't be good. Also we can run our target via `env LD_LIBRARY_PATH=...` with `AFL_SKIP_BIN_CHECK` which is not convenient and seems to slow down fuzzing process.
 
P.S. At first I thought this is super easy to implement and make a PR but then I realized there's actually lots of AFL++ code for AFL_PRELOAD :(

vanhauser-thc:
a PR for this should rather be easy.
just edit afl-forkserver.c and before the exec..() call you do   char *p; if ((p = afl_get_env("AFL_LIB..."))) setenv("LD_LIB..", p, 1);

fuzzah:
@vanhauser-thc sorry for being late with reply :(
Won't it break things? AFL_PRELOAD logic seems to be kinda different from what you suggest to do.

vanhauser-thc:
what would that break?

fuzzah:
> 
> 
> what would that break?

I've put some more attention, you're right about it not breaking things.
I've created PR #824 that covers this issue (though FATAL/PFATAL in afl-forkserver.c resulted in afl-fuzz not being able to communicate with forkserver on incorrect values of env var, so I've put this logic in afl-fuzz itself).

realmadsci:
I typically would just use the "env" tool to do environment variable injection in the afl-fuzz command line. Something like:
`AFL_SKIP_BIN_CHECK=1 afl-fuzz -i input -o output -- /usr/bin/env LD_LIBRARY_PATH=/path/to/libs my_target --my-args`

This is now possible with QEMU mode as well as of 3.12a:
[...]

PR #824, opened 2021-03-16
add AFL_TARGET_ENV to afl-fuzz

fuzzah:
This covers cases like specifying LD_LIBRARY_PATH only for the target. This also allows fuzzing targets that require AFL-specific env variables (e.g. AFL++ itself with AFL_BENCH_JUST_ONE).
I have fuzzed parsing function **extract_and_set_env** for ~2 hours on 6 cores and got hundreds of millions of execs and 100% coverage so it should be secure enough.
 
AFL_TARGET_ENV supports cases like
A=1
B='1 2 3'
C="x y z"
So quotes supported and should match. Outside quotes separator is ' ' (space).
Keys (vars) without values are not allowed and keys starting with numbers are not allowed either, so basically AFL_TARGET_ENV tries to follow UNIX rules. Keys with lowercase are allowed because some applications use them.
 
~~Length limit for value of AFL_TARGET_ENV is hardcoded to be 30000 characters.~~
upd: not hardcoded anymore


fuzzah:
Also this fixes #774

src/afl-common.c
```diff
@@ -706,6 +706,105 @@ char *get_afl_env(char *env) {
 
 }
 
+int extract_and_set_env(const char *env_str) {
+
+  size_t len = strnlen(env_str, 30000);
+  char * p = strndup(env_str, len);
```
vanhauser-thc: please just use `u8 *p = ck_strdup(env_str);` instead, PFATAL can go too.

src/afl-common.c
```diff
@@ -706,6 +706,105 @@ char *get_afl_env(char *env) {
...
+    num_pairs += 1;
+
+    setenv(key, val, 1);
+
+  }
+
+  if (num_pairs > 0) { ret_val = 1; }
+
+free_and_return:
+  free(p);
```
vanhauser-thc: because of ck_strdup, this then needs to be ck_free

src/afl-common.c
vanhauser-thc: ah and `if (!env_str) { return 0; }` before that

src/afl-common.c
```diff
@@ -706,6 +706,105 @@ char *get_afl_env(char *env) {
...
+
+  size_t len = strnlen(env_str, 30000);
+  char * p = strndup(env_str, len);
+
+  if (!p) { PFATAL("Unable to copy environment variables (OOM?)"); }
+
+  int ret_val = 0;  // return false by default
+
+  char *end = p + len;
+  *end = '\0';  // ensure null present
```
vanhauser-thc: also please switch all `int` to `u32` and `char` to `u8`

vanhauser-thc:
just a few nits, fixing the conflict and then it can be merged :)

fuzzah:
> 
> 
> just a few nits, fixing the conflict and then it can be merged :)

Thank you for reviewing this.
Fixed & squashed.
Also fuzzed for 500M execs just in case :)

fuzzah:
changed the last sneaky char to u8..
@vanhauser-thc i guess i shouldn't have squashed and force-pushed :( sorry for that

vanhauser-thc:
LGTM, thank you!

Issue #828, opened 2021-03-18
In place AFL resume keeps fuzzer_stats but not plot_data 


richinseattle:
When resuming a job in place, it is desirable to be able to keep continuous statistics. This has been implemented in the case of fuzzer_stats but not for plot_data as shown below: 

```
  if (!afl->in_place_resume) {

    fn = alloc_printf("%s/fuzzer_stats", afl->out_dir);
    if (unlink(fn) && errno != ENOENT) { goto dir_cleanup_failed; }
    ck_free(fn);

  }

    fn = alloc_printf("%s/plot_data", afl->out_dir);
    if (unlink(fn) && errno != ENOENT) { goto dir_cleanup_failed; }
    ck_free(fn);
```

If this change is not desired, then please add an environment variable to keep plot_data

vanhauser-thc:
implemented by @richinseattle  :)

PR #830, opened 2021-03-18
preserve plot_data for in-place resume

richinseattle:
This pull request checks for in place resume before attempting to unlink plot_data to fix https://github.com/AFLplusplus/AFLplusplus/issues/828


vanhauser-thc:
thanks richard!
I think this maybe needs more. the timestamp in the plot data will be the current time, so a big gap will be displayed in the timeline.
on the other hand if you apply the start up time delta (like in the display for runtime in the UI) then crashes etc. will not fit the timestamp displayed.
WDYT?

richinseattle:
So the question comes down to the value of the actual timestamp to the end user. 

In the case of plot_data, the unix_time is used mostly as an incrementing id so a delta should be just as effective given a start time. A comment could be added to the file with the unix time whenever a job is created/resumed to retain that information, and otherwise use a delta since last recorded time to keep things relative (this would need to be recovered from the last line on resume).

For the crashes file name, the create time is already captured by the filesystem, so the naming convention could be changed to a time delta to match plot_data. This way both values are preserved and if an end user wants to add the file create time back to their file names when they archive their crashes, that can be done easily on the command line. 

Will this break existing scripts that process crashes? I'll check afl-collect and crashwalk.

If that sounds suitable and doesn't break the triage tools, I will work on a new PR.

vanhauser-thc:
just add to this pr.

I think all you need is adding this in a few placed: `afl->prev_run_time + get_cur_time() - afl->start_time) / 1000` for total seconds of runtime

richinseattle:
On closer inspection, the file names were already using relative timestamps

Files are initialized with time:0
https://github.com/AFLplusplus/AFLplusplus/blob/dev/src/afl-fuzz-init.c#L1317-L1318

This is the file name being generated when new coverage is found which is done similar to the code you pasted above without the division (that's already handled in get_cur_time())
https://github.com/AFLplusplus/AFLplusplus/blob/dev/src/afl-fuzz-bitmap.c#L328-L329

Updated PR changes afl-fuzz-stats to use relative timestamps and afl-fuzz-init to handle opening and seeking to the end of an already existing afl_plot file when in-place resume happens.

src/afl-fuzz-init.c
```diff
@@ -2008,17 +2012,34 @@ void setup_dirs_fds(afl_state_t *afl) {
...
+    if (!afl->fsrv.plot_file) { PFATAL("fdopen() failed"); }
+
+    fprintf(afl->fsrv.plot_file,
           "# unix_time, cycles_done, cur_path, paths_total, "
           "pending_total, pending_favs, map_size, unique_crashes, "
           "unique_hangs, max_depth, execs_per_sec, total_execs, edges_found\n");
+
+  } else {
+
+    int fd = open(tmp, O_WRONLY | O_CREAT, 0600);
```
vanhauser-thc, line 2032: IMHO this also needs O_APPEND ... ?

vanhauser-thc:
LGTM, thanks!

domenukk:
> On closer inspection, the file names were already using relative timestamps
> 
> Files are initialized with time:0
> https://github.com/AFLplusplus/AFLplusplus/blob/dev/src/afl-fuzz-init.c#L1317-L1318
> 
> This is the file name being generated when new coverage is found which is done similar to the code you pasted above without the division (that's already handled in get_cur_time())
> https://github.com/AFLplusplus/AFLplusplus/blob/dev/src/afl-fuzz-bitmap.c#L328-L329
> 
> Updated PR changes afl-fuzz-stats to use relative timestamps and afl-fuzz-init to handle opening and seeking to the end of an already existing afl_plot file when in-place resume happens.

Can we have the file timestamps also load the last offset on resume? Or is that already the case?

vanhauser-thc:
@domenukk if you re-read what you quoted - yes that was already there. plus he linked the code where you can see that :)

domenukk:
Great, sorry then 😀

PR #831, opened 2021-03-18
Push to stable

vanhauser-thc:


GNUmakefile
```diff
@@ -576,7 +576,11 @@ endif
 deepclean:	clean
 	rm -rf unicorn_mode/unicornafl
 	rm -rf qemu_mode/qemuafl
+ifeq "$(IN_REPO)" "1"
 # NEVER EVER ACTIVATE THAT!!!!! git reset --hard >/dev/null 2>&1 || true
+	git checkout unicorn_mode/unicornafl
+	git checkout qemu_mode/qemuafl
+endif
```
vanhauser-thc, line 583: @domenukk how is the right action for deepclean being "git checkout"? IMHO this should remove the checkout, but so that it can be recovered again by build_unicorn_...

domenukk: There is no such thing as "downloading a submodule" or "checking out a submodule", they are always part of the repository. They are just not initialized when you do a non-recursive `clone`. Once you initialized and updated the submodule, there is no way to "un-download", although we can reset it to the empty-looking state this way.

TL;DR: Delete changes the node to be deleted in git, not what we want. Checkout checks out the (non-deleted) state of the folder afterwards. It'll still be an empty folder.

The only use I see is for testing.

docs/Changelog.md
```diff
@@ -8,6 +8,15 @@
...
 
+### Version ++3.12a (dev)
+  - afl-cc:
+    - fix cmplog rtn (rare crash and not being able to gather ptr data)
+    - link runtime not to shared libs
+    - ensure shared libraries are properly built and instrumented
+  - qemu_mode (thanks @realmadsci):
+    - move AFL_PRELOAD and AFL_USE_QASAN logic inside afl-qemu-trace
+    - add AFL_QEMU_CUSTOM_BIN
+
```
domenukk, line 23: Should there be a note about fixing the falsely removed unicorn folder? It worked on tarballs, just not git checkouts.

include/envs.h
```diff
@@ -50,6 +50,7 @@ static char *afl_environment_variables[] = {
     "AFL_FAST_CAL",
     "AFL_FORCE_UI",
     "AFL_FUZZER_ARGS",  // oss-fuzz
+    "AFL_GDB",
```
domenukk, line 54: What does it do?

domenukk (approved):

GNUmakefile
vanhauser-thc: but wouldnt it rather be "make clean ; git stash" or "make clean ; git drop" rather than "git checkout"?

include/envs.h
vanhauser-thc: for aflpp_driver.c ... prints out the pid, waits a second and also dump the memory map from proc

andreafioraldi:
Dlopen seems still broken @vanhauser-thc , i get undefined symbol: __afl_area_ptr at dlopen time

andreafioraldi:
I'm pushing a testcase in the CI for it

andreafioraldi:
it is failing with AFL_LLVM_INSTRUEMENT=AFL, with PCGUARD works (as __afl_area_ptr is not needed)

andreafioraldi:
```
readelf -Ws test-dlopen.plain | grep area
   125: 0000000000408148     8 OBJECT  GLOBAL DEFAULT   24 __afl_area_ptr
   127: 0000000000408140     8 OBJECT  GLOBAL DEFAULT   24 __afl_area_ptr_dummy
   139: 00000000004081d0 0x200000 OBJECT  GLOBAL DEFAULT   25 __afl_area_initial
   140: 0000000000408150     8 OBJECT  GLOBAL DEFAULT   24 __afl_area_ptr_backup
```
makes no sense, `__afl_area_ptr` is exported as it should using the dynamic list

vanhauser-thc:
> ```
> readelf -Ws test-dlopen.plain | grep area
>    125: 0000000000408148     8 OBJECT  GLOBAL DEFAULT   24 __afl_area_ptr
>    127: 0000000000408140     8 OBJECT  GLOBAL DEFAULT   24 __afl_area_ptr_dummy
>    139: 00000000004081d0 0x200000 OBJECT  GLOBAL DEFAULT   25 __afl_area_initial
>    140: 0000000000408150     8 OBJECT  GLOBAL DEFAULT   24 __afl_area_ptr_backup
> ```
> 
> makes no sense, `__afl_area_ptr` is exported as it should using the dynamic list

did it ever work with a previous release? last 2.x, 3.00, 3.10, 3.11?

vanhauser-thc:
@andreafioraldi 
This works:
```
../afl-clang-fast -DTEST_SHARED_OBJECT=1 -z defs -fPIC -shared -o libtest.so ../test-instr.c
../afl-clang-fast -o test-dlopen2 test-dlopen.c -ldl -L. -ltest
echo 0 | ./test-dlopen2
```
but dlopen does not. maybe this needs something else- or it is simply not possible?

GNUmakefile
domenukk: What do we want to achieve with deepclean?
If the folders should be empty afterwards(to test the scripts or something?) this is correct. If we want to go back to the original state of the repo, the commented out `reset hard` would have worked.

Make clean and git stash  inside the folders would be a good alternative as the folder contents are not cleared, but what's the point of deepclean then?

vanhauser-thc: originally deepclean was meant to get back to the checkout state, so no checkout of qemuafl and unicorn. (like e.g. you want to create a tar ball from the current repo state). for subversion this was needed for me. but wasnt the "git reset hard" also affecting the whole afl++ repo? that should not be the case I think.

andreafioraldi:
2.68c works well

andreafioraldi:
3.11c too

andreafioraldi:
Unrelated, but with LLVM 13 I get
```
instrumentation/SanitizerCoveragePCGUARD.so.cc:920:13: error: use of undeclared identifier 'GetOrCreateFunctionComdat'
            GetOrCreateFunctionComdat(F, TargetTriple, CurModuleUniqueId))
```

andreafioraldi:
The first bug is here should be !shared_linking
```c
if (shared_linking)
      cc_params[cc_par_cnt++] =
          alloc_printf("-Wl,--dynamic-list=%s/dynamic_list.txt", obj_path);
```

andreafioraldi:
The LLVM 13 error seems related to the fact that I use the LLVM ubuntu repo and I just did an upgrade this morning. Apparently they removed that function

andreafioraldi:
Nope this line of code is still in LLVM master https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Instrumentation/SanitizerCoverage.cpp#L680,
A missing header maybe? @vanhauser-thc usually how do you keep the pcguard pass updated with LLVM?

vanhauser-thc:
> Nope this line of code is still in LLVM master https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Instrumentation/SanitizerCoverage.cpp#L680,
> A missing header maybe? @vanhauser-thc usually how do you keep the pcguard pass updated with LLVM?

only if something breaks. there are so many additions in our version that is it not a copy-paste thing ...
@andreafioraldi I pushed a change, can you retry compiling? will take me 4+ hours to compile the current llvm 13 ...

vanhauser-thc:
ah, damn just adding the include breaks it ... working on it

vanhauser-thc:
@andreafioraldi now ...

GNUmakefile
domenukk: If you want to create a tarball, you'd want to do a `git reset hard` to get rid of all temp files and local changes ;)
But if you just want the folders of qemuafl and unicornafl to be emtpy, the current code is correct.

include/envs.h
domenukk: Should we add documentation for it?

vanhauser-thc: nah, it is rather niche, and I guess I am the only one using it ever

andreafioraldi:
i still have the error

andreafioraldi:
The header is #include "llvm/Transforms/Instrumentation.h" that was still there. Btw including the SanitizerCoverage.h header is a mistake I guess when compiling the pass using old llvm versions.

vanhauser-thc:
still compiling clang 13 .... 68% ... sigh

andreafioraldi:
You can just use the LLVM ubuntu repo lol https://apt.llvm.org/

vanhauser-thc:
fixed ...

vanhauser-thc:
@andreafioraldi can this be merged now?

andreafioraldi:
i'll test in some minutes

andreafioraldi:
go, the patch is working on my machine and test-llvm passes, we can merge IMO

Issue #659, opened 2020-12-26
Add build documentation for M1 Macs


Manouchehri:
Figured it'd make sense to have a central ticket for others to follow the state of AFL++ on the new ARM based Macs.

What works and doesn't (that I've seen so far):

- [x] Building the main afl++ binary
- [x] Building LLVM mode
- [x] Building Frida mode
- [x] Using afl-clang-fast
- [ ] QEMU mode (questionable usefulness to run on macOS)
- [ ] Unicorn mode (questionable usefulness to run on macOS)

vanhauser-thc:
building llvm mode is possible when you install a real clang, not the crippled xcode.
qemu itself is non of our business, and qemu_mode and unicorn are based on this, so nothing we can/will do.

afl-clang -integrated-as - does NOT work, as the assembly is for intel, not for arm ...
someone could add this - but why? install a real llvm or gcc and be happy.

so IMHO - nothing to do here for us. or?

vanhauser-thc:
and `afl-clang-fast -integrated-as` how did you get this idea? dont use  -integrated-as with afl-clang-fast :)

Manouchehri:
> building llvm mode is possible when you install a real clang, not the crippled xcode.

Cool, I'll try to get that working. =) (I don't think homebrew can install native clang yet.)

> qemu itself is non of our business, and qemu_mode and unicorn are based on this, so nothing we can/will do.

Makes sense. Also, it seems somewhat pointless to me. 

> so IMHO - nothing to do here for us. or?

Correct. Just documenting what I've tried for the sake of others. 😃 

> and afl-clang-fast -integrated-as how did you get this idea? dont use -integrated-as with afl-clang-fast.

Purely was just trying to get something running initially. Was not expecting it to work properly.

Manouchehri:
Got a bit further, still working on getting LLVM to work. Not sure why `/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/sys/cdefs.h` is still being included and erroring out.. (Pretty sure this isn't an afl++ problem.)

brew + LLVM installation:

```
sudo mkdir /opt/homebrew
sudo chown -R $(whoami) /opt/homebrew
cd /opt
curl -L https://github.com/Homebrew/brew/tarball/master | tar xz --strip 1 -C homebrew

/opt/homebrew/bin/brew update
/opt/homebrew/bin/brew install llvm
/opt/homebrew/bin/brew install make
```

```
[...]

PR #833, opened 2021-03-19
Frida

WorksButNotTested:
Hi. Grateful for any feedback you have on this.

### DIFF
--- a/frida_mode/Makefile
+++ b/frida_mode/Makefile
@@ -1,0 +1,310 @@
+PWD:=$(shell pwd)/
+INC_DIR:=$(PWD)inc/
+SRC_DIR:=$(PWD)src/
+INCLUDES:=$(wildcard $(INC_DIR)*.h)
+SOURCES:=$(wildcard $(SRC_DIR)*.c)
+BUILD_DIR:=$(PWD)build/
+CFLAGS:= $(CFLAGS) \
+		-fPIC \
+		-D_GNU_SOURCE
+
+FRIDA_BUILD_DIR:=$(BUILD_DIR)frida/
+FRIDA_TRACE:=$(FRIDA_BUILD_DIR)afl-frida-trace.so
+
+ARCH=$(shell uname -m)
+ifeq "$(ARCH)" "aarch64"
+ARCH:=arm64
+TESTINSTR_BASE:=0x0000aaaaaaaaa000
+endif
+
+ifeq "$(ARCH)" "x86_64"
+TESTINSTR_BASE:=0x0000555555554000
+endif
+
+ifeq "$(shell uname)" "Darwin"
+OS:=macos
+AFL_FRIDA_INST_RANGES=0x0000000000001000-0xFFFFFFFFFFFFFFFF
+CFLAGS:=$(CFLAGS) -Wno-deprecated-declarations
+TEST_LDFLAGS:=-undefined dynamic_lookup
+endif
+ifeq "$(shell uname)" "Linux"
+OS:=linux
+AFL_FRIDA_INST_RANGES=$(shell $(PWD)test/testinstr.py -f $(BUILD_DIR)testinstr -s .testinstr -b $(TESTINSTR_BASE))
+CFLAGS:=$(CFLAGS) -Wno-prio-ctor-dtor
+TEST_LDFLAGS:=
+endif
+
+ifndef OS
+$(error "Operating system unsupported")
+endif
+
+VERSION=14.2.13
+GUM_DEVKIT_FILENAME=frida-gum-devkit-$(VERSION)-$(OS)-$(ARCH).tar.xz
+GUM_DEVKIT_URL="https://github.com/frida/frida/releases/download/$(VERSION)/$(GUM_DEVKIT_FILENAME)"
+GUM_DEVKIT_TARBALL:=$(FRIDA_BUILD_DIR)$(GUM_DEVKIT_FILENAME)
+GUM_DEVIT_LIBRARY=$(FRIDA_BUILD_DIR)libfrida-gum.a
+GUM_DEVIT_HEADER=$(FRIDA_BUILD_DIR)frida-gum.h
+
+TEST_BUILD_DIR:=$(BUILD_DIR)test/
+
+LIBPNG_FILE:=$(TEST_BUILD_DIR)libpng-1.2.56.tar.gz
+LIBPNG_URL:=https://downloads.sourceforge.net/project/libpng/libpng12/older-releases/1.2.56/libpng-1.2.56.tar.gz
+LIBPNG_DIR:=$(TEST_BUILD_DIR)libpng-1.2.56/
+LIBPNG_MAKEFILE:=$(LIBPNG_DIR)Makefile
+LIBPNG_LIB:=$(LIBPNG_DIR).libs/libpng12.a
+
+HARNESS_FILE:=$(TEST_BUILD_DIR)StandaloneFuzzTargetMain.c
+HARNESS_OBJ:=$(TEST_BUILD_DIR)StandaloneFuzzTargetMain.o
+HARNESS_URL:="https://raw.githubusercontent.com/llvm/llvm-project/main/compiler-rt/lib/fuzzer/standalone/StandaloneFuzzTargetMain.c"
+
+PNGTEST_FILE:=$(TEST_BUILD_DIR)target.cc
+PNGTEST_OBJ:=$(TEST_BUILD_DIR)target.o
+PNGTEST_URL:="https://raw.githubusercontent.com/google/fuzzbench/master/benchmarks/libpng-1.2.56/target.cc"
+
+TEST_BIN:=$(TEST_BUILD_DIR)pngtest
+
+TESTINSTBIN:=$(BUILD_DIR)testinstr
+TESTINSTSRC:=$(PWD)test/testinstr.c
+
+TEST_DATA_DIR:=$(PWD)build/test/libpng-1.2.56/contrib/pngsuite/
+
+TESTINSTR_DATA_DIR:=$(BUILD_DIR)testinstr_in/
+TESTINSTR_DATA_FILE:=$(TESTINSTR_DATA_DIR)test.dat
+FRIDA_OUT:=$(PWD)frida_out
+QEMU_OUT:=$(PWD)qemu_out
+
+.PHONY: all frida test clean format test_frida test_qemu compare testinstr test_testinstr standalone
+
+all: $(FRIDA_TRACE)
+
+frida: $(FRIDA_TRACE)
+
+$(BUILD_DIR):
+	mkdir -p $(BUILD_DIR)
+
+############################# FRIDA ############################################
+$(FRIDA_BUILD_DIR): | $(BUILD_DIR)
+	mkdir -p $@
+
+$(GUM_DEVKIT_TARBALL): | $(FRIDA_BUILD_DIR)
+	wget -O $@ $(GUM_DEVKIT_URL)
+
+$(GUM_DEVIT_LIBRARY): | $(GUM_DEVKIT_TARBALL)
+	tar Jxvf $(GUM_DEVKIT_TARBALL) -C $(FRIDA_BUILD_DIR)
+
+$(GUM_DEVIT_HEADER): | $(GUM_DEVKIT_TARBALL)
+	tar Jxvf $(GUM_DEVKIT_TARBALL) -C $(FRIDA_BUILD_DIR)
+
+$(FRIDA_TRACE): $(GUM_DEVIT_LIBRARY) $(GUM_DEVIT_HEADER) $(SOURCES) Makefile | $(FRIDA_BUILD_DIR)
+	$(CC) -shared \
+		$(CFLAGS) \
+		-o $@ $(SOURCES) \
+		$(GUM_DEVIT_LIBRARY) \
+		-I $(FRIDA_BUILD_DIR) \
+		-I .. \
+		-I ../include \
+		-I $(INC_DIR) \
+		../instrumentation/afl-compiler-rt.o.c \
+		-lpthread -ldl -lresolv
+
+	cp -v $(FRIDA_TRACE) ../
+
+############################# TEST #############################################
+
+test: $(TEST_BIN)
+
+$(TEST_BUILD_DIR): $(BUILD_DIR)
+	mkdir -p $@
+
+$(HARNESS_FILE): | $(TEST_BUILD_DIR)
+	wget -O $@ $(HARNESS_URL)
+
+$(HARNESS_OBJ): $(HARNESS_FILE)
+	$(CC) -o $@ -c $<
+
+$(PNGTEST_FILE): | $(TEST_BUILD_DIR)
+	wget -O $@ $(PNGTEST_URL)
+
+$(PNGTEST_OBJ): $(PNGTEST_FILE) | $(LIBPNG_DIR)
+	$(CXX) -std=c++11 -I $(LIBPNG_DIR) -o $@ -c $<
+
+$(LIBPNG_FILE): | $(TEST_BUILD_DIR)
+	wget -O $@ $(LIBPNG_URL)
+
+$(LIBPNG_DIR): $(LIBPNG_FILE)
+	tar zxvf $(LIBPNG_FILE) -C $(TEST_BUILD_DIR)
+
+$(LIBPNG_MAKEFILE): | $(LIBPNG_DIR)
+	cd $(LIBPNG_DIR) && ./configure
+
+$(LIBPNG_LIB): $(LIBPNG_MAKEFILE)
+	make -C $(LIBPNG_DIR)
+
+$(TEST_BIN): $(HARNESS_OBJ) $(PNGTEST_OBJ) $(LIBPNG_LIB)
+	$(CXX) \
+		-o $@ \
+		$(HARNESS_OBJ) $(PNGTEST_OBJ) $(LIBPNG_LIB) \
+		-lz \
+		$(TEST_LDFLAGS)
+
+############################# TESTINSR #########################################
+$(TESTINSTR_DATA_DIR): | $(BUILD_DIR)
+	mkdir -p $@
+
+$(TESTINSTR_DATA_FILE): | $(TESTINSTR_DATA_DIR)
+	echo -n "000" > $@
+
+$(TESTINSTBIN): $(TESTINSTSRC) | $(BUILD_DIR)
+	$(CC) -o $@ $<
+
+testinstr: $(TESTINSTBIN)
+
+############################# CLEAN ############################################
+clean:
+	rm -rf $(BUILD_DIR)
+
+############################# FORMAT ###########################################
+format:
+	cd .. && echo $(SOURCES) | xargs -L1 ./.custom-format.py -i
+	cd .. && echo $(INCLUDES) | xargs -L1 ./.custom-format.py -i
+	cd .. && ./.custom-format.py -i $(TESTINSTSRC)
+
+############################# RUN #############################################
+
+# Add the environment variable AFL_DEBUG_CHILD=1 to show printf's from the target
+
+png_frida: $(FRIDA_TRACE) $(TEST_BIN)
+	make -C ..
+	cd .. && \
+		./afl-fuzz \
+			-O \
+			-i $(TEST_DATA_DIR) \
+			-o $(FRIDA_OUT) \
+			-- \
+				$(TEST_BIN) @@
+
+png_qemu: $(TEST_BIN)
+	make -C ..
+	cd .. && \
+		./afl-fuzz \
+			-Q \
+			-i $(TEST_DATA_DIR) \
+			-o $(QEMU_OUT) \
+			-- \
+				$(TEST_BIN) @@
+
+compare: $(FRIDA_TRACE) $(TEST_BIN)
+	cd .. && \
+		./afl-fuzz \
+			-V30 \
+			-O \
+			-i $(TEST_DATA_DIR) \
+			-o $(FRIDA_OUT) \
+			-- \
+				$(TEST_BIN) @@
+	cd .. && \
+		./afl-fuzz \
+			-V30 \
+			-Q \
+			-i $(TEST_DATA_DIR) \
+			-o $(QEMU_OUT) \
+				-- \
+					$(TEST_BIN) @@
+	cat frida_out/default/fuzzer_stats
+	cat qemu_out/default/fuzzer_stats
+
+testinstr_qemu: $(TESTINSTBIN) $(TESTINSTR_DATA_FILE)
+	make -C ..
+	cd .. && \
+			AFL_QEMU_INST_RANGES=$(AFL_FRIDA_INST_RANGES) \
+			./afl-fuzz \
+				-Q \
+				-i $(TESTINSTR_DATA_DIR) \
+				-o $(QEMU_OUT) \
+				-- \
+					$(TESTINSTBIN) @@
+
+testinstr_frida: $(FRIDA_TRACE) $(TESTINSTBIN) $(TESTINSTR_DATA_FILE)
+	make -C ..
+	cd .. && \
+			AFL_FRIDA_INST_RANGES=$(AFL_FRIDA_INST_RANGES) \
+			AFL_FRIDA_INST_NO_OPTIMIZE=1 \
+			AFL_FRIDA_INST_NO_PREFETCH=1 \
+			AFL_FRIDA_INST_STRICT=1 \
+			./afl-fuzz \
+				-O \
+				-i $(TESTINSTR_DATA_DIR) \
+				-o $(FRIDA_OUT) \
+				-- \
+					$(TESTINSTBIN) @@
+
+standalone: $(FRIDA_TRACE) $(TESTINSTBIN) $(TESTINSTR_DATA_FILE)
+	cd .. && \
+			AFL_FRIDA_INST_RANGES=$(AFL_FRIDA_INST_RANGES) \
+			AFL_DEBUG_CHILD=1 \
+			AFL_FRIDA_DEBUG_MAPS=1 \
+			AFL_FRIDA_INST_NO_OPTIMIZE=1 \
+			AFL_FRIDA_INST_NO_PREFETCH=1 \
+			AFL_FRIDA_INST_TRACE=1 \
+			AFL_FRIDA_INST_STRICT=1 \
+			LD_PRELOAD=$(FRIDA_TRACE) \
+			DYLD_INSERT_LIBRARIES=$(FRIDA_TRACE) \
+			$(TESTINSTBIN) $(TESTINSTR_DATA_FILE)
+
+tmin_qemu: $(TEST_BIN)
+	make -C ..
+	cd .. && \
+		./afl-tmin \
+			-Q \
+			-i $(TEST_DATA_DIR)basn0g01.png \
+			-o $(QEMU_OUT)/qemu-min-basn0g01.png \
+			-- \
+				$(TEST_BIN) @@
+
+tmin_frida: $(TEST_BIN)
+	make -C ..
+	cd .. && \
+		./afl-tmin \
+			-O \
+			-i $(TEST_DATA_DIR)basn0g01.png \
+			-o $(FRIDA_OUT)/qemu-min-basn0g01.png \
+			-- \
+				$(TEST_BIN)
+
+showmap_qemu: $(TEST_BIN)
+	make -C ..
+	cd .. && \
+		./afl-showmap \
+			-Q \
+			-i $(TEST_DATA_DIR) \
+			-o $(QEMU_OUT) \
+			-- \
+				$(TEST_BIN) @@
+
+showmap_frida: $(TEST_BIN)
+	make -C ..
+	cd .. && \
+		./afl-showmap \
+			-O \
+			-i $(TEST_DATA_DIR) \
+			-o $(FRIDA_OUT) \
+			-- \
+				$(TEST_BIN) @@
+
+analyze_qemu: $(TEST_BIN)
+	make -C ..
+	cd .. && \
+		./afl-analyze \
+			-Q \
+			-i $(TEST_DATA_DIR)basn0g01.png \
+			-- \
+				$(TEST_BIN) @@
+
+analyze_frida: $(TEST_BIN)
+	make -C ..
+	cd .. && \
+		./afl-analyze \
+			-O \
+			-i $(TEST_DATA_DIR)basn0g01.png \
+			-- \
+				$(TEST_BIN) @@
--- a/frida_mode/Makefile
+++ b/frida_mode/Makefile
@@ -308,3 +308,43 @@ analyze_frida: $(TEST_BIN)
 			-i $(TEST_DATA_DIR)basn0g01.png \
 			-- \
 				$(TEST_BIN) @@
+
+cmin_qemu: $(TEST_BIN)
+	make -C ..
+	cd .. && \
+		./afl-cmin \
+			-Q \
+			-i $(TEST_DATA_DIR) \
+			-o $(QEMU_OUT) \
+			-- \
+				$(TEST_BIN) @@
+
+cmin_frida: $(TEST_BIN)
+	make -C ..
+	cd .. && \
+		./afl-cmin \
+			-O \
+			-i $(TEST_DATA_DIR) \
+			-o $(FRIDA_OUT) \
+			-- \
+				$(TEST_BIN) @@
+
+cmin_bash_qemu: $(TEST_BIN)
+	make -C ..
+	cd .. && \
+		./afl-cmin.bash \
+			-Q \
+			-i $(TEST_DATA_DIR) \
+			-o $(QEMU_OUT) \
+			-- \
+				$(TEST_BIN) @@
+
+cmin_bash_frida: $(TEST_BIN)
+	make -C ..
+	cd .. && \
+		./afl-cmin.bash \
+			-O \
+			-i $(TEST_DATA_DIR) \
+			-o $(FRIDA_OUT) \
+			-- \
+				$(TEST_BIN) @@
--- a/frida_mode/README.md
+++ b/frida_mode/README.md
@@ -1,0 +1,48 @@
+# FRIDA MODE
+The purpose of FRIDA mode is to provide an alternative binary only fuzzer for AFL
+just like that provided by QEMU mode. The intention is to provide a very similar
+user experience, right down to the options provided through environment variables.
+
+Additionally, the intention is to be able to make a direct performance comparison
+between the two approaches. Hopefully, we should also be able to leverage the same
+approaches for adding features which QEMU uses, possibly even sharing code.
+
+## Limitations
+The current focus is on x64 support for Intel. Although parts may be architecturally
+dependent, the approach itself should remain architecture agnostic.
+
+## Usage
+FRIDA mode requires some small modifications to the afl-fuzz and similar tools in
+AFLplusplus. The intention is that it behaves identically to QEMU, but uses the 'O'
+switch rather than 'Q'.
+
+## Design
+AFL Frida works by means of a shared library injected into a binary program using
+LD_PRELOAD, similar to the way which other fuzzing features are injected into targets.
+
+## Testing
+Alongside the FRIDA mode, we also include a test program for fuzzing. This test
+program is built using the libpng benchmark from fuzz-bench and integrating the
+StandaloneFuzzTargetMain from the llvm project. This is built and linked without
+any special modifications to suit FRIDA or QEMU. However, at present we don't have
+a representative corpus.
+
+## Getting Started
+To build everything run `make`.
+
+To run the benchmark sample with qemu run `make test_qemu`.
+To run the benchmark sample with frida run `make test_frida`.
+
+# Configuration options
+* `AFL_FRIDA_DEBUG_MAPS` - See `AFL_QEMU_DEBUG_MAPS`
+* `AFL_FRIDA_EXCLUDE_RANGES` - See `AFL_QEMU_EXCLUDE_RANGES`
+* `AFL_FRIDA_INST_NO_OPTIMIZE` - Don't use optimized inline assembly coverage instrumentation (the default where available). Required to use `AFL_FRIDA_INST_TRACE`.
+* `AFL_FRIDA_INST_NO_PREFETCH` - Disable prefetching. By default the child will report instrumented blocks back to the parent so that it can also instrument them and they be inherited by the next child on fork.
+* `AFL_FRIDA_INST_RANGES` - See `AFL_QEMU_INST_RANGES`
+* `AFL_FRIDA_INST_STRICT` - Under certain conditions, Stalker may encroach into excluded regions and generate both instrumented blocks and coverage data (e.g. indirect calls on x86). The excluded block is generally honoured as soon as another function is called within the excluded region. The overhead of generating, running and instrumenting these few additional blocks is likely to be fairly small, but it may hinder you when checking that the correct number of paths are found for testing purposes or similar. There is a performance penatly for this option during block compilation where we check the block isn't in a list of excluded ranges.
+* `AFL_FRIDA_INST_TRACE` - Generate some logging when running instrumented code. Requires `AFL_FRIDA_INST_NO_OPTIMIZE`.
+
+# TODO
+* Add AARCH64 inline assembly optimization from libFuzz
+* Fix issues running on OSX
+* Identify cause of erroneous additional paths
--- a/frida_mode/README.md
+++ b/frida_mode/README.md
@@ -3,46 +3,133 @@ The purpose of FRIDA mode is to provide an alternative binary only fuzzer for AF
 just like that provided by QEMU mode. The intention is to provide a very similar
 user experience, right down to the options provided through environment variables.
 
-Additionally, the intention is to be able to make a direct performance comparison
-between the two approaches. Hopefully, we should also be able to leverage the same
-approaches for adding features which QEMU uses, possibly even sharing code.
+Whilst AFLplusplus already has some support for running on FRIDA [here](https://github.com/AFLplusplus/AFLplusplus/tree/stable/utils/afl_frida)
+this requires the code to be fuzzed to be provided as a shared library, it
+cannot be used to fuzz executables. Additionally, it requires the user to write
+a small harness around their target code of interest, FRIDA mode instead takes a
+different approach to avoid these limitations.
 
-## Limitations
-The current focus is on x64 support for Intel. Although parts may be architecturally
-dependent, the approach itself should remain architecture agnostic.
+# Current Progress
+As FRIDA mode is new, it is missing a lot of features. Most importantly,
+persistent mode. The design is such that it should be possible to add these
+features in a similar manner to QEMU mode and perhaps leverage some of its
+design and implementation.
 
-## Usage
-FRIDA mode requires some small modifications to the afl-fuzz and similar tools in
-AFLplusplus. The intention is that it behaves identically to QEMU, but uses the 'O'
-switch rather than 'Q'.
+  | Feature/Instrumentation  | frida-mode |
+  | -------------------------|:----------:|
+  | NeverZero                |            |
+  | Persistent Mode          |            |
+  | LAF-Intel / CompCov      |            |
+  | CmpLog                   |            |
+  | Selective Instrumentation|     x      |
+  | Non-Colliding Coverage   |            |
+  | Ngram prev_loc Coverage  |            |
+  | Context Coverage         |            |
+  | Auto Dictionary          |            |
+  | Snapshot LKM Support     |            |
 
-## Design
-AFL Frida works by means of a shared library injected into a binary program using
-LD_PRELOAD, similar to the way which other fuzzing features are injected into targets.
+# Compatibility
+Currently FRIDA mode supports Linux and macOS targets on both x86/x64
+architecture and aarch64. Later releases may add support for aarch32 and Windows
+targets as well as embedded linux environments.
 
-## Testing
-Alongside the FRIDA mode, we also include a test program for fuzzing. This test
-program is built using the libpng benchmark from fuzz-bench and integrating the
-StandaloneFuzzTargetMain from the llvm project. This is built and linked without
-any special modifications to suit FRIDA or QEMU. However, at present we don't have
-a representative corpus.
+FRIDA has been used on various embedded targets using both uClibc and musl C
+runtime libraries, so porting should be possible. However, the current build
+system does not support cross compilation.
 
 ## Getting Started
 To build everything run `make`.
 
-To run the benchmark sample with qemu run `make test_qemu`.
-To run the benchmark sample with frida run `make test_frida`.
+To run the benchmark sample with qemu run `make png_qemu`.
+To run the benchmark sample with frida run `make png_frida`.
 
-# Configuration options
+## Usage
+FRIDA mode requires some small modifications to the afl-fuzz and similar tools
+in AFLplusplus. The intention is that it behaves identically to QEMU, but uses
+the 'O' switch rather than 'Q'. Whilst the options 'f', 'F', 's' or 'S' may have
+made more sense for a mode powered by FRIDA Stalker, they were all taken, so
+instead we use 'O' in homage to the [author](https://github.com/oleavr) of
+FRIDA.
+
+Similarly, the intention is to mimic the use of environment variables used by
+QEMU where possible (although replacing `s/QEMU/FRIDA/g`). Accodingly, the
+following options are currently supported.
+
 * `AFL_FRIDA_DEBUG_MAPS` - See `AFL_QEMU_DEBUG_MAPS`
 * `AFL_FRIDA_EXCLUDE_RANGES` - See `AFL_QEMU_EXCLUDE_RANGES`
-* `AFL_FRIDA_INST_NO_OPTIMIZE` - Don't use optimized inline assembly coverage instrumentation (the default where available). Required to use `AFL_FRIDA_INST_TRACE`.
-* `AFL_FRIDA_INST_NO_PREFETCH` - Disable prefetching. By default the child will report instrumented blocks back to the parent so that it can also instrument them and they be inherited by the next child on fork.
 * `AFL_FRIDA_INST_RANGES` - See `AFL_QEMU_INST_RANGES`
-* `AFL_FRIDA_INST_STRICT` - Under certain conditions, Stalker may encroach into excluded regions and generate both instrumented blocks and coverage data (e.g. indirect calls on x86). The excluded block is generally honoured as soon as another function is called within the excluded region. The overhead of generating, running and instrumenting these few additional blocks is likely to be fairly small, but it may hinder you when checking that the correct number of paths are found for testing purposes or similar. There is a performance penatly for this option during block compilation where we check the block isn't in a list of excluded ranges.
-* `AFL_FRIDA_INST_TRACE` - Generate some logging when running instrumented code. Requires `AFL_FRIDA_INST_NO_OPTIMIZE`.
+
+# Performance
+
+Additionally, the intention is to be able to make a direct performance
+comparison between the two approaches. Accordingly, FRIDA mode includes a test
+target based on the [libpng](https://libpng.sourceforge.io/) benchmark used by
+[fuzzbench](https://google.github.io/fuzzbench/) and integrated with the
+[StandaloneFuzzTargetMain](https://raw.githubusercontent.com/llvm/llvm-project/main/compiler-rt/lib/fuzzer/standalone/StandaloneFuzzTargetMain.c)
+from the llvm project. This is built and linked without any special
+modifications to suit FRIDA or QEMU. We use the test data provided with libpng
+as our corpus.
+
+Whilst not much performance tuning has been completed to date, performance is
+around 30-50% of that of QEMU mode, however, this gap may reduce with the 
+introduction of persistent mode. Performance can be tested by running 
+`make compare`, albeit a longer time measurement may be required for move 
+accurate results. 
+
+Whilst [afl_frida](https://github.com/AFLplusplus/AFLplusplus/tree/stable/utils/afl_frida)
+claims a 5-10x performance increase over QEMU, it has not been possible to
+reproduce these claims. However, the number of executions per second can vary
+dramatically as a result of the randomization of the fuzzer input. Some inputs
+may traverse relatively few paths before being rejected as invalid whilst others
+may be valid inputs or be subject to much more processing before rejection.
+Accordingly, it is recommended that testing be carried out over prolongued
+periods to gather timings which are more than indicative.
+
+# Design
+FRIDA mode is supported by using `LD_PRELOAD` (`DYLD_INSERT_LIBRARIES` on macOS)
+to inject a shared library (`afl-frida-trace.so`) into the target. This shared
+library is built using the [frida-gum](https://github.com/frida/frida-gum)
+devkit from the [FRIDA](https://github.com/frida/frida) project. One of the
+components of frida-gum is [Stalker](https://medium.com/@oleavr/anatomy-of-a-code-tracer-b081aadb0df8),
+this allows the dynamic instrumentation of running code for AARCH32, AARCH64,
+x86 and x64 architectutes. Implementation details can be found
+[here](https://frida.re/docs/stalker/).
+
+Dynamic instrumentation is used to augment the target application with similar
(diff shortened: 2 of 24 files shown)


WorksButNotTested:
I see you have some FRIDA support here. https://github.com/AFLplusplus/AFLplusplus/tree/stable/utils/afl_frida
I have taken a bit of a different approach, hoping to get pure binary fuzzing support. I hope it doesn't step on anyone's toes.

I have done a little performance testing, although I am not certain it is perfect. I have built the libpng benchmark for fuzzbench and using the sample images in libpng for a corpus, this is the results I get with a 30 second run...

QEMU 4500 execs/sec
FRIDA 1500 execs/sec (with prefetch)
FRIDA 250 execs/sec (without prefetch)

NB. Prefetch relays instrumented blocks back from the forked child to the parent to be instrumented, hence saving them being compiled each time they are executed. It is the same approach taken by QEMU.

I'm particularly keen if anyone has any feedback on performance testing. I've not been able to replicate the 12x performance gain reported in utils/afl_frida, but I'm not sure if this is down to a different in testing methodology or something I have screwed up.

vanhauser-thc:
Hi Jon, thank you for this!
Will look into this, and I am sure @andreafioraldi and maybe  s1341 as well.
To discuss features, etc. you can do that on discord (https://discord.gg/DAGdcpz) or in this issue here. there is very little traffic on the mailing list in my experience.

src/afl-fuzz.c
```diff
@@ -1288,13 +1306,47 @@ int main(int argc, char **argv_orig, char **envp) {
...
+          snprintf(frida_afl_preload, frida_afl_preload_reqd + 1, "%s %s",
+                   afl_preload, "./afl-frida-trace");
+
+      if (frida_afl_preload_len != frida_afl_preload_reqd) {
+
+        FATAL("Invalid AFL_PRELOAD length: %d != %d", frida_afl_preload_len,
+              frida_afl_preload_reqd);
+
+      }
+
```
vanhauser-thc, line 1321: getenv can return NULL, afl-frida-trace is most likely not the in the current directory plus this is overly complicated.
simply do:
```
...
afl_preload = getenv("AFL_PRELOAD");
u8 *frida_binary = find_binary("afl-frida-trace");
if (afl_preload) { frida_afl_preload = alloc_printf("%s:%s", afl_preload, frida_binary); }
else { frida_afl_preload = alloc_printf("%s", frida_binary); }
ck_free(frida_binary);
OKF("Frida Mode setting LD_PRELOAD %s", frida_afl_preload);
...
```
All the error checkings are in the functions.

src/afl-fuzz.c
```diff
@@ -2068,6 +2122,8 @@ int main(int argc, char **argv_orig, char **envp) {
 
   }
 
+  if (frida_afl_preload) { free(frida_afl_preload); }
```
vanhauser-thc: this would be ck_free() then

src/afl-fuzz.c
WorksButNotTested: Yeah. Thanks, that code was a bit clumsy.

src/afl-fuzz.c
WorksButNotTested: Fixed.

src/afl-fuzz.c
vanhauser-thc: find_binary can now be switched to find_afl_binary if you pull into your repo

src/afl-fuzz.c
```diff
@@ -1288,13 +1304,39 @@ int main(int argc, char **argv_orig, char **envp) {
...
       setenv("DYLD_INSERT_LIBRARIES", getenv("AFL_PRELOAD"), 1);
 
     }
 
+  } else if (afl->fsrv.frida_mode) {
+
+    OKF("Frida Mode setting LD_PRELOAD ./afl-frida-trace");
+    setenv("LD_PRELOAD", "./afl-frida-trace", 1);
+    setenv("DYLD_INSERT_LIBRARIES", "./afl-frida-trace", 1);
+
```
vanhauser-thc, line 1340: this block needs to be updated similarly as well

frida_mode/Makefile
```diff
@@ -0,0 +1,137 @@
...
+		-Wno-prio-ctor-dtor \
+		-DWITH_PREFETCH \
+		-DWITH_INLINE_INSTR \
+		-o $@ $(SOURCES) \
+		$(GUM_DEVIT_LIBRARY) \
+		-I $(FRIDA_BUILD_DIR) -I .. -I ../include -I $(INC_DIR) \
+		../instrumentation/afl-compiler-rt.o.c \
+		-lpthread -ldl -lresolv
+
+	cp $(FRIDA_TRACE) ../
```
vanhauser-thc: please add -v to cp

vanhauser-thc:
I have a few minor nits.
IMHO this can then be merged as it cannot break existing functionality and then can easier be tested.
@andreafioraldi ?

andreafioraldi:
We should not merge until a release with the fixes for dlopen IMO @vanhauser-thc

frida_mode/Makefile
```diff
@@ -0,0 +1,137 @@
...
+INC_DIR:=$(PWD)inc/
+SRC_DIR:=$(PWD)src/
+INCLUDES:=$(wildcard $(INC_DIR)*.h)
+SOURCES:=$(wildcard $(SRC_DIR)*.c)
+BUILD_DIR:=$(PWD)build/
+
+FRIDA_BUILD_DIR:=$(BUILD_DIR)frida/
+FRIDA_TRACE:=$(FRIDA_BUILD_DIR)afl-frida-trace
+
+GUM_DEVKIT_ARCH=x86_64
```
domenukk: The devkit version should depend on the current arch, right?

frida_mode/README.md
```diff
@@ -0,0 +1,34 @@
...
+between the two approaches. Hopefully, we should also be able to leverage the same
+approaches for adding features which QEMU uses, possibly even sharing code.
+
+## Limitations
+The current focus is on x64 support for Intel. Although parts may be architecturally
+dependent, the approach itself should remain architecture agnostic.
+
+## Usage
+FRIDA mode requires some small modifications to the afl-fuzz and similar tools in
+AFLplusplus. The intention is that it behaves identically to QEMU, but uses the 'O'
```
domenukk: I would argue against adding a new switch, just calling the `frida-trace` binary is fine imho. This leaves greater flexibility to the user. Also `O` seems random.

frida_mode/src/instrument.c
```diff
@@ -0,0 +1,148 @@
...
+#include "config.h"
+#include "prefetch.h"
+
+extern uint8_t *        __afl_area_ptr;
+static __thread guint64 previous_pc;
+
+GumAddress current_log_impl;
+uint64_t   afl_prev_loc;
+
+#if GUM_NATIVE_CPU == GUM_CPU_AMD64 && defined(WITH_INLINE_INSTR)
```
domenukk: We can copy&paste aarch64 support from libafl Frida here

src/afl-fuzz.c
```diff
@@ -109,6 +109,7 @@ static void usage(u8 *argv0, int more_help) {
       "maximum.\n"
       "  -m megs       - memory limit for child process (%u MB, 0 = no limit "
       "[default])\n"
+      "  -O            - use binary-only instrumentation (FRIDA mode)\n"
```
domenukk, line 112: Again, let's not add a switch? What do you guys think?

frida_mode/README.md
vanhauser-thc: IMHO this is ok. Like -W wine, -Q qemu and -U unicorn. -F would be better but that it is taken, same as -f. if there is something more fitting than -O - sure.

andreafioraldi:
> I see you have some FRIDA support here. https://github.com/AFLplusplus/AFLplusplus/tree/stable/utils/afl_frida
> I have taken a bit of a different approach, hoping to get pure binary fuzzing support. I hope it doesn't step on anyone's toes.
> 
> I have done a little performance testing, although I am not certain it is perfect. I have built the libpng benchmark for fuzzbench and using the sample images in libpng for a corpus, this is the results I get with a 30 second run...
> 
> QEMU 4500 execs/sec
> FRIDA 1500 execs/sec (with prefetch)
> FRIDA 250 execs/sec (without prefetch)
> 
> NB. Prefetch relays instrumented blocks back from the forked child to the parent to be instrumented, hence saving them being compiled each time they are executed. It is the same approach taken by QEMU.
> 
> I'm particularly keen if anyone has any feedback on performance testing. I've not been able to replicate the 12x performance gain reported in utils/afl_frida, but I'm not sure if this is down to a different in testing methodology or something I have screwed up.

For fuzzbench QEMU runs in persistent mode, are you sure that you are doing the same with frida? Block caching in the parent improves the perf when using fork(), but it has a moderate impact when using persistent mode.

frida_mode/Makefile
andreafioraldi: Yes

andreafioraldi:
As frida as first class citizen in AFL++ is interesting especially for macOS binary fuzzing (which is not supported by QEMU), and I don't own a mac, we should ask someone to test these patches in macOS (ideally, both x64 and M1) before merging.

frida_mode/README.md
domenukk: This can just use -U and then `frida-trace`. It's something hidden from the user that has hardly any benefit imho.
Even -U I would probably scrap to be honest, and make SKIP_BINCHECK default

frida_mode/src/instrument.c
andreafioraldi: If @s1341 agree

src/afl-fuzz.c
andreafioraldi: Dunno, we should be consistent and I don't think we can remove -Q without breaking tons of user scripts built around AFL/AFL++ ...

src/afl-fuzz.c
WorksButNotTested: Fixed.

src/afl-fuzz.c
WorksButNotTested: I hope this is what you meant? (e.g. using find_afl_binary rather than a fixed relative path)

frida_mode/Makefile
WorksButNotTested: Fixed

frida_mode/Makefile
WorksButNotTested: Do you suggest I perhaps use the variable ARCH, document it as needing to be set in README.md and perhaps default to x86_64 if not set? Happy to go with whatever you think best.

frida_mode/README.md
vanhauser-thc: we can not set skip bincheck off default. this is what detects persistent mode and deferred forkserver.
and removing -U would just add complexity to using it.

vanhauser-thc:
> As frida as first class citizen in AFL++ is interesting especially for macOS binary fuzzing (which is not supported by QEMU), and I don't own a mac, we should ask someone to test these patches in macOS (ideally, both x64 and M1) before merging.

good idea.
@Manouchehri  and @iamvineeth23 - could you see if this works for you or where it needs tweaking?

frida_mode/README.md
WorksButNotTested: I'm not sure I follow? `afl-frida-trace` is a shared object rather than an EXE. I don't think I can make it an executable as and executable cannot dlopen another executable (even if it is PIE which effectively makes it a DSO anyway) as the dynamic loader complains.

f & s were the obvious choices (as it is supported by FRIDAs stalker engine), but they were taken. Kind of running out of alphabet! So I went with O in homage to [Ole](https://github.com/oleavr) (the author). Happy to change though if you have a preference?

frida_mode/src/instrument.c
WorksButNotTested: Absolutely. There's a good few more features to add yet too. Deferred fork mode and persistent mode. I just wanted to get something in as a starting point for discussion and to check everyone was happy before I wrote a whole load of code.

I'm happy to add it to the PR though if you'd like?

src/afl-fuzz.c
WorksButNotTested: As I mentioned above, I don't think that FRIDA mode can be a binary. I think it needs to be an LD_PRELOADED shared object. So there's a good few little logic changes in the behaviour of AFL which are gated on the `frida_mode` varaible.

WorksButNotTested:
> > I see you have some FRIDA support here. https://github.com/AFLplusplus/AFLplusplus/tree/stable/utils/afl_frida
> > I have taken a bit of a different approach, hoping to get pure binary fuzzing support. I hope it doesn't step on anyone's toes.
> > I have done a little performance testing, although I am not certain it is perfect. I have built the libpng benchmark for fuzzbench and using the sample images in libpng for a corpus, this is the results I get with a 30 second run...
> > QEMU 4500 execs/sec
> > FRIDA 1500 execs/sec (with prefetch)
> > FRIDA 250 execs/sec (without prefetch)
> > NB. Prefetch relays instrumented blocks back from the forked child to the parent to be instrumented, hence saving them being compiled each time they are executed. It is the same approach taken by QEMU.
> > I'm particularly keen if anyone has any feedback on performance testing. I've not been able to replicate the 12x performance gain reported in utils/afl_frida, but I'm not sure if this is down to a different in testing methodology or something I have screwed up.
> 
> For fuzzbench QEMU runs in persistent mode, are you sure that you are doing the same with frida? Block caching in the parent improves the perf when using fork(), but it has a moderate impact when using persistent mode.

I was using the `make compare` to perform my comparison. I didn't give any environment arguments to QEMU to pass it any other address to use for the deferred fork server, so I assume it is using `main`?

Similarly, in it's current form, `frida_mode` also starts the fork-server at `main`, although it should be possible to add support for `AFL_ENTRYPOINT` at a later point.

I just wanted something for a like-for-like comparison really. Obviously it would be good to extract as much performance as possible, but I rather suspect ` s1341`'s work integrating with libAFL will be the best avenue for that. I was just hoping performance might be within the same order of magnitude as QEMU for now.

WorksButNotTested:
> > As frida as first class citizen in AFL++ is interesting especially for macOS binary fuzzing (which is not supported by QEMU), and I don't own a mac, we should ask someone to test these patches in macOS (ideally, both x64 and M1) before merging.
> 
> good idea.
> @Manouchehri and @iamvineeth23 - could you see if this works for you or where it needs tweaking?

That would be awesome if someone has some time! Please feel free to hack away at things.

Equally if anyone wants to improve on anything or add new features please feel free. I'm hoping to get some time to keep adding to it though, so just drop me a message to de-conflict.

frida_mode/Makefile
vanhauser-thc: maybe just `$(shell uname -m)` ?

frida_mode/Makefile
```diff
@@ -0,0 +1,137 @@
...
+	cd .. && ./afl-fuzz -f frida_mode/frida_test.dat -O -i $(TEST_DATA_DIR) -o frida_mode/frida_out -- frida_mode/build/test/pngtest frida_mode/frida_test.dat
+
+test_qemu: $(TEST_BIN)
+	make -C ..
+	cd .. && ./afl-fuzz -f frida_mode/qemu_test.dat -Q -i $(TEST_DATA_DIR) -o frida_mode/qemu_out -- frida_mode/build/test/pngtest frida_mode/qemu_test.dat
+
+compare: $(FRIDA_TRACE) $(TEST_BIN)
+	cd .. && timeout -s INT 30 ./afl-fuzz -f frida_mode/frida_test.dat -O -i $(TEST_DATA_DIR) -o frida_mode/frida_out -- frida_mode/build/test/pngtest frida_mode/frida_test.dat || true
+	cd .. && timeout -s INT 30 ./afl-fuzz -f frida_mode/qemu_test.dat -Q -i $(TEST_DATA_DIR) -o frida_mode/qemu_out -- frida_mode/build/test/pngtest frida_mode/qemu_test.dat || true
+	cat frida_out/default/fuzzer_stats
```
vanhauser-thc, line 213: no need for timeout ... just add -V30 to afl-fuzz

vanhauser-thc:
@WorksButNotTested 
I checkout out your branch and tested a bit.
```
gcc -o t test-instr.c
./afl-fuzz -O -i in -o out -- ./t
```
this results in
```
total paths : 11
stability : 80.78%
```
whereas qemu mode (correctly) shows:
```
total paths: 3
stability: 100.0%
```
I saw the same issue with meme's hotwax setup.
utils/afl_frida -> `afl-fuzz -i in -o out -- ./afl-frida`  (loads test-instr.c as a shared library) <- has 100% stability, finds correctly 3 paths and is much much much faster than qemu.
EDIT: I should mention that afl_frida uses persistent mode, and that of course is part of what it makes much faster.

Manouchehri:
Sure, I can test this. @WorksButNotTested How are you testing any of this btw? =)

WorksButNotTested:
> Sure, I can test this. @WorksButNotTested How are you testing any of this btw? =)

At the minute, I am running `make test_frida`.

frida_mode/Makefile
WorksButNotTested: Fixed

Manouchehri:
Taken and modified from https://github.com/AFLplusplus/AFLplusplus/issues/659#issuecomment-751387565.

```sh
brew update
brew install llvm make python3 wget
git clone https://github.com/WorksButNotTested/AFLplusplus.git -b frida
cd AFLplusplus/                                                                                                                 
export PATH="/opt/homebrew/opt/make/libexec/gnubin:$PATH"
export PATH="/opt/homebrew/opt/llvm/bin:$PATH"
export CC="clang"
export CXX="clang++"
export LDFLAGS="-isysroot '$(xcrun --sdk macosx --show-sdk-path)' -arch arm64 -L/opt/homebrew/opt/llvm/lib -Wl,-rpath,/opt/homebrew/opt/llvm/lib"
export CPPFLAGS="-isysroot '$(xcrun --sdk macosx --show-sdk-path)' -arch arm64 -I/opt/homebrew/opt/llvm/include"
export CFLAGS="-isysroot '$(xcrun --sdk macosx --show-sdk-path)' -arch arm64  -I/opt/homebrew/opt/llvm/include"

make all
make -C frida_mode
```

```
dave@mini AFLplusplus % make -C frida_mode
make: Entering directory '/Users/dave/AFLplusplus/frida_mode'
gcc -shared \
        -fPIC \
        -Wno-prio-ctor-dtor \
        -DWITH_PREFETCH \
        -DWITH_INLINE_INSTR \
        -o /Users/dave/AFLplusplus/frida_mode/build/frida/afl-frida-trace /Users/dave/AFLplusplus/frida_mode/src/instrument.c /Users/dave/AFLplusplus/frida_mode/src/interceptor.c /Users/dave/AFLplusplus/frida_mode/src/main.c /Users/dave/AFLplusplus/frida_mode/src/prefetch.c \
        /Users/dave/AFLplusplus/frida_mode/build/frida/libfrida-gum.a \
        -I /Users/dave/AFLplusplus/frida_mode/build/frida/ -I .. -I ../include -I /Users/dave/AFLplusplus/frida_mode/inc/ \
        ../instrumentation/afl-compiler-rt.o.c \
        -lpthread -ldl -lresolv
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
1 warning generated.
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
/Users/dave/AFLplusplus/frida_mode/src/interceptor.c:11:7: warning: implicit conversion from enumeration type 'GumReplaceReturn' to different
      enumeration type 'GumAttachReturn' [-Wenum-conversion]
      gum_interceptor_replace(interceptor, address, replacement, user_data);
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 warnings generated.
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
/Users/dave/AFLplusplus/frida_mode/src/main.c:2:10: fatal error: 'sys/personality.h' file not found
#include <sys/personality.h>
         ^~~~~~~~~~~~~~~~~~~
1 warning and 1 error generated.
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
1 warning generated.
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
../instrumentation/afl-compiler-rt.o.c:1740:12: warning: 'syscall' is deprecated: first deprecated in macOS 10.12 - syscall(2) is unsupported;
      please switch to a supported interface. For SYS_kdebug_trace use kdebug_signpost(). [-Wdeprecated-declarations]
  long r = syscall(SYS_write, __afl_dummy_fd[1], ptr, len);
           ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/unistd.h:742:6: note: 'syscall' has
      been explicitly marked deprecated here
int      syscall(int, ...);
         ^
2 warnings generated.
make: *** [Makefile:62: /Users/dave/AFLplusplus/frida_mode/build/frida/afl-frida-trace] Error 1
make: Leaving directory '/Users/dave/AFLplusplus/frida_mode'

```

vanhauser-thc:
@WorksButNotTested I removed the persistent mode from utils/afl_frida/afl-frida.c - still 100% stability and 3 paths ... but it runs slower than your version.

Here is the code - I think if you merge the right thing into your solution it will run a bit slower but will have the correct paths and stability.
https://gist.github.com/vanhauser-thc/6913fa7c291ee685f3c63a049140a222

once this is done it is easy to add an AFL_FRIDA_PERSISTENT env feature that runs the main in a persistent loop (which will most of the time not work, but if, it will be much faster)

WorksButNotTested:
> @WorksButNotTested
> I checkout out your branch and tested a bit.
> 
> ```
> gcc -o t test-instr.c
> ./afl-fuzz -O -i in -o out -- ./t
> ```
> 
> this results in
> 
> ```
> total paths : 11
> stability : 80.78%
> ```
> 
> whereas qemu mode (correctly) shows:
> 
> ```
> total paths: 3
> stability: 100.0%
> ```
> 
> I saw the same issue with meme's hotwax setup.
> utils/afl_frida -> `afl-fuzz -i in -o out -- ./afl-frida` (loads test-instr.c as a shared library) <- has 100% stability, finds correctly 3 paths and is much much much faster than qemu.
> EDIT: I should mention that afl_frida uses persistent mode, and that of course is part of what it makes much faster.

Ok cool. I'll re-test with that target and see if I can account for the additional paths found. I suspect it might be because in the current form there's no support for `AFL_QEMU_INST_RANGES` or `AFL_QEMU_EXCLUDE_RANGES` whereas `afl_frida` has this [check](https://github.com/AFLplusplus/AFLplusplus/blob/23f7bee81c46ad4f0f65fa56d08064ab5f1e2e6f/utils/afl_frida/afl-frida.c#L177). So we will be counting the paths inside printf (and any allocators it uses too).

How are you running your performance comparisons between `afl_frida` and qemu mode at the moment?

vanhauser-thc:
@Manouchehri 
switch in this snippit:
```
#ifdef __APPLE__
  #include <mach/mach.h>
  #include <mach-o/dyld_images.h>
#else
  #include <sys/wait.h>
  #include <sys/personality.h>
#endif
```

Manouchehri:
```
From b0625d048e6ac166500ae0a9ab4e35bb0ba0722d Mon Sep 17 00:00:00 2001
From: David Manouchehri <david.manouchehri@riseup.net>
Date: Fri, 19 Mar 2021 10:53:27 -0500
Subject: [PATCH] Don't include on macOS.

---
 frida_mode/src/main.c | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/frida_mode/src/main.c b/frida_mode/src/main.c
index 2871b4af..7014aa43 100644
--- a/frida_mode/src/main.c
+++ b/frida_mode/src/main.c
@@ -1,5 +1,7 @@
 #include <unistd.h>
+#if !defined(__APPLE__)
 #include <sys/personality.h>
+#endif
 #include <sys/types.h>
 #include <sys/wait.h>
 
-- 
2.24.3 (Apple Git-128)
```

Still not quite working yet.

```
dave@mini AFLplusplus % make -C frida_mode                                       
make: Entering directory '/Users/dave/AFLplusplus/frida_mode'
gcc -shared \
        -fPIC \
        -Wno-prio-ctor-dtor \
        -DWITH_PREFETCH \
        -DWITH_INLINE_INSTR \
        -o /Users/dave/AFLplusplus/frida_mode/build/frida/afl-frida-trace /Users/dave/AFLplusplus/frida_mode/src/instrument.c /Users/dave/AFLplusplus/frida_mode/src/interceptor.c /Users/dave/AFLplusplus/frida_mode/src/main.c /Users/dave/AFLplusplus/frida_mode/src/prefetch.c \
        /Users/dave/AFLplusplus/frida_mode/build/frida/libfrida-gum.a \
        -I /Users/dave/AFLplusplus/frida_mode/build/frida/ -I .. -I ../include -I /Users/dave/AFLplusplus/frida_mode/inc/ \
        ../instrumentation/afl-compiler-rt.o.c \
        -lpthread -ldl -lresolv
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
1 warning generated.
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
/Users/dave/AFLplusplus/frida_mode/src/interceptor.c:11:7: warning: implicit conversion from enumeration type 'GumReplaceReturn' to different
      enumeration type 'GumAttachReturn' [-Wenum-conversion]
      gum_interceptor_replace(interceptor, address, replacement, user_data);
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 warnings generated.
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
1 warning generated.
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
1 warning generated.
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
../instrumentation/afl-compiler-rt.o.c:1740:12: warning: 'syscall' is deprecated: first deprecated in macOS 10.12 - syscall(2) is unsupported;
      please switch to a supported interface. For SYS_kdebug_trace use kdebug_signpost(). [-Wdeprecated-declarations]
  long r = syscall(SYS_write, __afl_dummy_fd[1], ptr, len);
           ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/unistd.h:742:6: note: 'syscall' has
      been explicitly marked deprecated here
int      syscall(int, ...);
         ^
2 warnings generated.
ld: warning: ignoring file /Users/dave/AFLplusplus/frida_mode/build/frida/libfrida-gum.a, building for macOS-arm64 but attempting to link with file built for unknown-unsupported file format ( 0x21 0x3C 0x61 0x72 0x63 0x68 0x3E 0x0A 0x2F 0x20 0x20 0x20 0x20 0x20 0x20 0x20 )
Undefined symbols for architecture arm64:
  "___libc_start_main", referenced from:
      _on_main in main-93cda9.o
...
  "_gum_x86_writer_put_push_reg", referenced from:
      _instr_basic_block in instrument-7fd932.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [Makefile:62: /Users/dave/AFLplusplus/frida_mode/build/frida/afl-frida-trace] Error 1
make: Leaving directory '/Users/dave/AFLplusplus/frida_mode'

```

frida_mode/Makefile
Manouchehri:
```
dave@mini AFLplusplus % echo $(shell uname -m)
zsh: command not found: shell

dave@mini AFLplusplus % echo $(uname -m) 
arm64
```

frida_mode/Makefile
```diff
@@ -0,0 +1,137 @@
...
+INCLUDES:=$(wildcard $(INC_DIR)*.h)
+SOURCES:=$(wildcard $(SRC_DIR)*.c)
+BUILD_DIR:=$(PWD)build/
+
+FRIDA_BUILD_DIR:=$(BUILD_DIR)frida/
+FRIDA_TRACE:=$(FRIDA_BUILD_DIR)afl-frida-trace
+
+GUM_DEVKIT_ARCH=x86_64
+GUM_DEVKIT_VERSION=14.2.13
+GUM_DEVKIT_FILENAME=frida-gum-devkit-$(GUM_DEVKIT_VERSION)-linux-$(GUM_DEVKIT_ARCH).tar.xz
```
Manouchehri: This shouldn't be hardcoded to Linux.

src/afl-fuzz.c
domenukk: All right then :) do we have any letters left that are actually part of Frida? Is R taken? Or should we just think of it as the @oleavr mode

frida_mode/Makefile
vanhauser-thc: maybe `$(shell uname -s|tr A-Z a-z)`?

Manouchehri: So the current URL for macOS is:
```
https://github.com/frida/frida/releases/download/14.2.13/frida-gum-devkit-14.2.13-macos-arm64.tar.xz
```
Not `darwin`. =(

frida_mode/Makefile
```diff
@@ -0,0 +1,137 @@
...
+	wget -O $@ $(GUM_DEVKIT_URL)
+
+$(GUM_DEVIT_LIBRARY): | $(GUM_DEVKIT_TARBALL)
+	tar Jxvf $(GUM_DEVKIT_TARBALL) -C $(FRIDA_BUILD_DIR)
+
+$(GUM_DEVIT_HEADER): | $(GUM_DEVKIT_TARBALL)
+	tar Jxvf $(GUM_DEVKIT_TARBALL) -C $(FRIDA_BUILD_DIR)
+
+$(FRIDA_TRACE): $(GUM_DEVIT_LIBRARY) $(GUM_DEVIT_HEADER) $(SOURCES) Makefile | $(FRIDA_BUILD_DIR)
+	gcc -shared \
```
Manouchehri: Could we _not_ hardcode this to gcc? =)

Manouchehri: `$(CC)` instead?

Manouchehri:
```
dave@mini AFLplusplus % make -C frida_mode         
make: Entering directory '/Users/dave/AFLplusplus/frida_mode'
clang -shared \
        -fPIC \
        -Wno-prio-ctor-dtor \
        -DWITH_PREFETCH \
        -DWITH_INLINE_INSTR \
        -o /Users/dave/AFLplusplus/frida_mode/build/frida/afl-frida-trace /Users/dave/AFLplusplus/frida_mode/src/instrument.c /Users/dave/AFLplusplus/frida_mode/src/interceptor.c /Users/dave/AFLplusplus/frida_mode/src/main.c /Users/dave/AFLplusplus/frida_mode/src/prefetch.c \
        /Users/dave/AFLplusplus/frida_mode/build/frida/libfrida-gum.a \
        -I /Users/dave/AFLplusplus/frida_mode/build/frida/ -I .. -I ../include -I /Users/dave/AFLplusplus/frida_mode/inc/ \
        ../instrumentation/afl-compiler-rt.o.c \
        -lpthread -ldl -lresolv
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
1 warning generated.
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
/Users/dave/AFLplusplus/frida_mode/src/interceptor.c:11:7: warning: implicit conversion from enumeration type 'GumReplaceReturn' to different enumeration type 'GumAttachReturn' [-Wenum-conversion]
      gum_interceptor_replace(interceptor, address, replacement, user_data);
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 warnings generated.
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
1 warning generated.
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
1 warning generated.
warning: unknown warning option '-Wno-prio-ctor-dtor'; did you mean '-Wno-reorder-ctor'? [-Wunknown-warning-option]
../instrumentation/afl-compiler-rt.o.c:1740:12: warning: 'syscall' is deprecated: first deprecated in macOS 10.12 - syscall(2) is unsupported; please switch to a supported interface. For SYS_kdebug_trace use kdebug_signpost(). [-Wdeprecated-declarations]
  long r = syscall(SYS_write, __afl_dummy_fd[1], ptr, len);
           ^
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/unistd.h:742:6: note: 'syscall' has been explicitly marked deprecated here
int      syscall(int, ...);
         ^
2 warnings generated.
Undefined symbols for architecture arm64:
  "___libc_start_main", referenced from:
      _on_main in main-47e4c0.o
      _on_libc_start_main in main-47e4c0.o
      _init in main-47e4c0.o
ld: symbol(s) not found for architecture arm64
clang-11: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [Makefile:62: /Users/dave/AFLplusplus/frida_mode/build/frida/afl-frida-trace] Error 1
make: Leaving directory '/Users/dave/AFLplusplus/frida_mode'
```

Manouchehri:
I tried adding this as a dirty hack as well, didn't make a difference.

```
#define ___libc_start_main __libc_start_main
```

WorksButNotTested:
Leave this with me for a bit. I'm looking into why I am seeing more paths than I should when running `testinstr` as a simple target. I have added support in another branch for `AFL_FRIDA_INST_RANGES` and `AFL_FRIDA_EXCLUDE_RANGES` to match QEMU, but I am still seeing more branches than I should even when I restrict FRIDA to only monitoring the `testinstr` function itself.
If I run the application standalone (with `afl-frida-trace` LD_PRELOAD'ed) and I add logging to `afl_maybe_log`, everything looks as expected. So I guess it must be an integration issue somewhere. Or perhaps AFL expects me to use a write barrier when writing to the shared memory?

frida_mode/Makefile
WorksButNotTested: Yeah. My bad. I’ll do that no worries.

frida_mode/Makefile
WorksButNotTested: I can still use uname for the platform detection and just throw in an if statement to convey what uname says to the right token in the file name?

vanhauser-thc:
@WorksButNotTested 
look here: https://gist.github.com/vanhauser-thc/6913fa7c291ee685f3c63a049140a222
no need to lock the map or anything.

what could be the issue is that previous_pc is not reset on fork. on a new fork this has to be reset to 0.

andreafioraldi:
> Leave this with me for a bit. I'm looking into why I am seeing more paths than I should when running `testinstr` as a simple target. I have added support in another branch for `AFL_FRIDA_INST_RANGES` and `AFL_FRIDA_EXCLUDE_RANGES` to match QEMU, but I am still seeing more branches than I should even when I restrict FRIDA to only monitoring the `testinstr` function itself.
> If I run the application standalone (with `afl-frida-trace` LD_PRELOAD'ed) and I add logging to `afl_maybe_log`, everything looks as expected. So I guess it must be an integration issue somewhere. Or perhaps AFL expects me to use a write barrier when writing to the shared memory?

That's why Staker, especially on x86, is imprecise and stalk parts of frida itself. If the number of edges is stable, it is fine.

frida_mode/src/main.c
```diff
@@ -0,0 +1,92 @@
...
+
+  gum_init_embedded();
+  if (!gum_stalker_is_supported()) {
+
+    gum_deinit_embedded();
+    FATAL("Failed to initialize embedded");
+
+  }
+
+  intercept(__libc_start_main, on_libc_start_main, NULL);
```
Manouchehri: `__libc_start_main` is a glibc specific thing, so we should probably use something like `gum_darwin_find_entrypoint`. Not sure what the platform agnostic API is to find the entrypoint.

vanhauser-thc:
> That's why Staker, especially on x86, is imprecise and stalk parts of frida itself. If the number of edges is stable, it is fine.

I do not think this is the issue here but who knows ... setting a debug in instr_basic_block that prints the instruction pointer could help to see if this is happening or not.

frida_mode/Makefile
vanhauser-thc: yes, that is the easy way.

vanhauser-thc: you can see that in the main afl++ GNUmakefile

vanhauser-thc:
Yes one issue is that the previous location is not resetted.

this should do that:
```
diff --git a/frida_mode/src/instrument.c b/frida_mode/src/instrument.c
index 2f4df483..14170b15 100644
--- a/frida_mode/src/instrument.c
+++ b/frida_mode/src/instrument.c
@@ -3,7 +3,7 @@
 #include "prefetch.h"
 
 extern uint8_t *        __afl_area_ptr;
-static __thread guint64 previous_pc;
+__thread guint64 previous_pc;
 
 GumAddress current_log_impl;
 uint64_t   afl_prev_loc;
diff --git a/frida_mode/src/main.c b/frida_mode/src/main.c
index 2871b4af..5f969ad0 100644
--- a/frida_mode/src/main.c
+++ b/frida_mode/src/main.c
@@ -21,6 +21,7 @@ int __libc_start_main(int *(main)(int, char **, char **), int argc,
                       void (*rtld_fini)(void), void(*stack_end));
 
 extern void __afl_manual_init();
+extern __thread guint64 previous_pc;
 
 static int on_fork() {
 
@@ -56,6 +57,7 @@ int *on_main(int argc, char **argv, char **envp) {
   __afl_manual_init();
 
   /* Child here */
+  previous_pc = 0;
   prefetch_start(stalker);
   return main_fn(argc, argv, envp);
 ```

but this is not all. I put a fprintf  prev_loc, curr_loc ) into afl_maybe_log and it showed a lot of hits - way more than there are in the small test-instr. binary.

WorksButNotTested:
> Yes one issue is that the previous location is not resetted.
> 
> this should do that:
> 
> ```
> diff --git a/frida_mode/src/instrument.c b/frida_mode/src/instrument.c
> index 2f4df483..14170b15 100644
> --- a/frida_mode/src/instrument.c
> +++ b/frida_mode/src/instrument.c
> @@ -3,7 +3,7 @@
>  #include "prefetch.h"
>  
>  extern uint8_t *        __afl_area_ptr;
> -static __thread guint64 previous_pc;
> +__thread guint64 previous_pc;
>  
>  GumAddress current_log_impl;
>  uint64_t   afl_prev_loc;
> diff --git a/frida_mode/src/main.c b/frida_mode/src/main.c
> index 2871b4af..5f969ad0 100644
> --- a/frida_mode/src/main.c
> +++ b/frida_mode/src/main.c
> @@ -21,6 +21,7 @@ int __libc_start_main(int *(main)(int, char **, char **), int argc,
>                        void (*rtld_fini)(void), void(*stack_end));
>  
>  extern void __afl_manual_init();
> +extern __thread guint64 previous_pc;
>  
>  static int on_fork() {
>  
> @@ -56,6 +57,7 @@ int *on_main(int argc, char **argv, char **envp) {
>    __afl_manual_init();
>  
>    /* Child here */
> +  previous_pc = 0;
>    prefetch_start(stalker);
>    return main_fn(argc, argv, envp);
> ```
> 
> but this is not all. I put a fprintf prev_loc, curr_loc ) into afl_maybe_log and it showed a lot of hits - way more than there are in the small test-instr. binary.

I think you could be right about resetting that previous_pc.

FRIDA will be following all of the code including main and all of the printfs by default. Check out this branch I was working in to check things out (I didn’t want to pollute the PR until I knew it would work).

https://github.com/WorksButNotTested/AFLplusplus/tree/ranges/frida_mode

As you can see in the bottom of the Makefile, I’ve built testinstr as an executable. I’ve put the testinstr function is it’s own section so you can work out where it is (readelf -SW) and implemented the environment variables like qemu mode for including and excluding ranges so you should be able to restrict exactly what is followed.

It uses the gum_stalker_exclude API, which should do the job, (and mean the original instrument code is run for those blocks, which should be faster) but I’ve put a guard in the instrumentation code for now (I’m hoping it can be removed).

I didn’t get chance to wrap things up fully yesterday as it got kinda late, but see what you make of it.

WorksButNotTested:
> > Leave this with me for a bit. I'm looking into why I am seeing more paths than I should when running `testinstr` as a simple target. I have added support in another branch for `AFL_FRIDA_INST_RANGES` and `AFL_FRIDA_EXCLUDE_RANGES` to match QEMU, but I am still seeing more branches than I should even when I restrict FRIDA to only monitoring the `testinstr` function itself.
> > If I run the application standalone (with `afl-frida-trace` LD_PRELOAD'ed) and I add logging to `afl_maybe_log`, everything looks as expected. So I guess it must be an integration issue somewhere. Or perhaps AFL expects me to use a write barrier when writing to the shared memory?
> 
> That's why Staker, especially on x86, is imprecise and stalk parts of frida itself. If the number of edges is stable, it is fine.

Stalker shouldn’t ever stalk itself. If I it does, you won’t end up with just a load of false edges, it will crash catastrophically (and likely be very very difficult to debug).

It will, however by default stalk everything else, that includes any libraries (right down until it hits a syscall and all the way back up again) and any hooks you install with an interceptor. Call outs are not stalked though.

frida_mode/src/instrument.c
s1341: Of course I agree! Note that my code is not yet optimized...

frida_mode/Makefile
WorksButNotTested: Fixed

src/afl-fuzz.c
WorksButNotTested: Here's what's taken...
https://github.com/AFLplusplus/AFLplusplus/blob/687dd9cb6759a0549a04c93d2f073d8cbfba397e/src/afl-fuzz.c#L366

frida_mode/Makefile
WorksButNotTested: Fixed

frida_mode/Makefile
WorksButNotTested: Done

WorksButNotTested:
I have changed some compilation flags to runtime environment variables to make things more flexible. I have also added support for restricting the ranges of the code which is stalked (`AFL_FRIDA_INST_RANGES` and `AFL_FRIDA_EXCLUDE_RANGES`) so that I can try to narrow down the cause of the additional paths. I have added `AFL_FRIDA_INST_STRICT` to avoid a few specific instances where stalker can encroach into excluded ranges (it checks the excluded ranges when instrumenting a block) and `AFL_FRIDA_INST_TRACE` (along with `AFL_FRIDA_INST_NO_OPTIMIZE`) to trace each instrumented block which is run. I have also added `AFL_FRIDA_INST_NO_PREFETCH` to rule out any prefetching issues. I have also fixed the use of the `personality` syscall so that ASLR is in fact disabled (you need to exec yourself to enable it). Lastly I included the change to reset `afl_prev_loc` on each fork. Oh and I fixed a few of the other little review comments.

The `testinstr` function is now within it's own section, and `testinstr.py` is added to parse the elf and idenfity the bounds of the section so that all other code can be excluded. I have also added `make standalone` to run `afl-frida-trace` outside of AFL against testinstr. Note that the base address is hard-coded in the makefile (if that is platform specific you may need to change it but it does support `AFL_FRIDA_DEBUG_MAPS` so you should be able to check easily).

Now running `make standalone` you can see the address of each block which is run which seems to tally with the disassembly in GHIDRA. However, running `make frida_testinstr` which runs the same target with the same options (minus a few debugging options) still shows about 18 paths for this [one method](https://github.com/WorksButNotTested/AFLplusplus/blob/75d431eb8799fc6e9e4c0a9d1278077996a32bfe/frida_mode/test/testinstr.c#L19-L32). However, the target `make qemu_testinstr` (which runs the same test in qemu mode) shows a similar number.

I'm a bit confused as to what is causing them to report such a high number, particularly when you say that util/afl-frida reports only 3? Is there something wrong with my test? Or have I made a wrong assumption somewhere along the line?

Fixes for the difference in libc on macOS are still outstanding, as is the addition of optimized assembly instrumentation on AARCH64. Please let me know whether you would like these addressed in the same PR, or whether you are happy to merge and then tackle these in a subsequent PR?

vanhauser-thc:
the main important thing is that the stability is 100%, not that the number of paths is 3. especially when you start tracing into libs if course there might be more paths.
how is the stability with the changes?

WorksButNotTested:
> the main important thing is that the stability is 100%, not that the number of paths is 3. especially when you start tracing into libs if course there might be more paths.
> how is the stability with the changes?

100% after 30s (40k execs) running `make testinstr_frida`. Looks like running without `AFL_FRIDA_INST_NO_PREFETCH` is causing the stability to drop below 50% though. So something with that feature is causing some issues. I'll have to take a look tomorrow.

vanhauser-thc:
@WorksButNotTested

with frida_mode/test/test-instr.c  I get 100% stability, however a few more paths than there should be. 
with the main test-instr.c from the root dir I get 100% stability and 3 paths. plus it is slightly faster than qemu :)
`./afl-fuzz -O -i in -o out -- ./test-instr`
so this looks good so far!

vanhauser-thc:
also pngtest is now at 100% stability!
there it is however half the speed of qemu.
If it would be possible to cache/keep the frida instrumentation that would increase the speed a lot. how that should be possible with a fork()ed child that is then destroyed ... 
maybe instrumenting everything and then installing the forkserver is possible? this would miss dlopen libs though, but the speed should be much better than qemu.

domenukk:
Maybe in the same way as qemu mode? Translating the same block in the parent, too, when it gets hit

vanhauser-thc:
When I set either `AFL_FRIDA_INST_NO_PREFETCH=1`or `AFL_FRIDA_INST_NO_OPTIMIZE=1` the speed increases to be slightly better than qemu for the pngtest! (while still at 100% stability)
Why is that the case? both options sound like they would decrease the performance but the opposite is happening. also setting both env vars does not further increase the speed.

WorksButNotTested:
> When I set either `AFL_FRIDA_INST_NO_PREFETCH=1`or `AFL_FRIDA_INST_NO_OPTIMIZE=1` the speed increases to be slightly better than qemu for the pngtest! (while still at 100% stability)
> Why is that the case? both options sound like they would decrease the performance but the opposite is happening. also setting both env vars does not further increase the speed.

That’s the purpose of prefetch. The parent creates a hash table of blocks that have been instrumented. This is inherited by the child. The child checks each block against the hash table when it instruments it, and if absent, writes it to a shared memory region. The fork call in the parent is then hooked by Frida and the blocks in the shared memory region are instrumented by calling prefetch in Frida and then added to the hash table. This the next child should inherit both the instrumented blocks and an updated hashtable.

I’m not sure why those options are slowing slowing things down. It seems counterintuitive, I’ll need to take a look tomorrow, but it seems likely I’ve screwed something up! If they do degrade performance though, I’ll take them out and we’ll end up with a faster and simpler implementation, which doesn’t sound too bad.

s1341:
Perhaps you are thrashing in the prefetch/update code because you are currently running under a fork-server, no? So every execution you spend a bunch of time pulling the prefetched blocks into the hash-table.

This might 'just go away' when you have persistent mode. 

Would it be possible to persist the data to the share/parent in such a way that it can immediately be interpreted as a hash-table without having to re-insert every block?

vanhauser-thc:
> Perhaps you are thrashing in the prefetch/update code because you are currently running under a fork-server, no? So every execution you spend a bunch of time pulling the prefetched blocks into the hash-table.
> 
> This might 'just go away' when you have persistent mode.
> 
> Would it be possible to persist the data to the share/parent in such a way that it can immediately be interpreted as a hash-table without having to re-insert every block?

it is not possible to have persistent mode here (at least as default)

WorksButNotTested:
> Perhaps you are thrashing in the prefetch/update code because you are currently running under a fork-server, no? So every execution you spend a bunch of time pulling the prefetched blocks into the hash-table.
> 
> This might 'just go away' when you have persistent mode.
> 
> Would it be possible to persist the data to the share/parent in such a way that it can immediately be interpreted as a hash-table without having to re-insert every block?

I have dropped the hash table all together now. It was superfluous! Without prefetch, on every fork, we would have to re-instrument every block we run. With prefetch, we instrument each block only twice, once in the child when we run it and once in the parent when we prefetch it. (Assuming we don't hit the limit of the number of new blocks per run we can fit in our shared memory region).

Having run the png test again, I can see that frida mode is about 1/2 the speed of qemu for me, and adding either of AFL_FRIDA_INST_NO_PREFETCH or AFL_FRIDA_INST_NO_OPTIMIZE degrades performance. Is someone able to confirm if they see the same results?

Persistent mode is the next feature I want to add. It should probably be the default, but for consistency should I see what QEMU mode does by default and match that?

vanhauser-thc:
@WorksButNotTested I see the same thing, half speed and even less with either of the two envs set.
but: this was different yesterday. same initial speed but the two envs double the speed. so whatever your did there made it worse.

WorksButNotTested:
> @WorksButNotTested I see the same thing, half speed and even less with either of the two envs set.
> but: this was different yesterday. same initial speed but the two envs double the speed. so whatever your did there made it worse.

Are you able to reproduce yesterday's results (by checking out an older commit)? I tried before I changed anything, but I didn't see the same results?

frida_mode/README.md
domenukk: > we can not set skip bincheck off default. this is what detects persistent mode and deferred forkserver.
> and removing -U would just add complexity to using it.

Persistent mode and deferred forkserver could be agreed upon done during forkserver handshake btw ;)

vanhauser-thc:
@WorksButNotTested  weird. I cannot reproduce the results, but I still have the commands etc. in my history .... now setting the envs degrade the performance - even with the state from yesterday that I used. clueless.

WorksButNotTested:
> @WorksButNotTested weird. I cannot reproduce the results, but I still have the commands etc. in my history .... now setting the envs degrade the performance - even with the state from yesterday that I used. clueless.

Maybe AFL just ended up fuzzing some paths which returned quickly (the input data is random after all) rather than going right into the depths of libpng? I guess we are performing reasonably short runs, so that could allow the results to be skewed by such things? Are you happy with how things are working now? Or do you have any other concerns?

I think the remaining things we have talked about carrying on with now are:
1. Fix support for macOS, looks like we should be able to use frida to find the main entry point, rather than using __libc_start_main.
2. Add inline assembly instrumentation for AARCH64 (from libAFL)
3. Add support for persistent mode.

Which would you like to see completed by this PR? And which would you like to come in a subsequent PR?

vanhauser-thc:
We will do another fix release (3.12) within the next days. I would want your patch after that as it is a bigger thing that that impacts several tools. and then we can call the following release 3.20 as it has good new stuff :)

IMHO from your list 1 and 2 is essential. and:
4. add the necessary support to afl-tmin, afl-showmap and afl-analyze

3. is a nice extra when everything works. plus anything that adds new features, performance, ...

WorksButNotTested:
No worries. No sense trying to merge before it is sufficiently mature. I’ll work on sorting the remaining issues. Just wanted to get the right balance between what’s essential and trying to solve everything at once.

vanhauser-thc:
just tested afl-showmap, works nicely!

vanhauser-thc:
some more thoughts before it can be merged:
  * afl-frida-trace is a shared object, not a program, so should be rather called afl-frida-trace.so
  * `make` should not download and build the jpeg test. only when a test... target is called this should be done :)

not for merge but later:
  * cross compiling should be possible (so specifying a compiler with CC= and specifying the target with ARCH= or something like that) (being able to run it with musl, diet, uclibc would be great but doesnt seem to be possible)

an issue I see is that it is specially interesting to run it on an embedded hardware. but this runs into glibc issues (not built with the same version, e.g. `/lib64/libc-2.25.so: version `GLIBC_2.27' not found (required by ./afl-frida-trace)`). but not sure how to solve this easily

WorksButNotTested:
> some more thoughts before it can be merged:
> 
>     * afl-frida-trace is a shared object, not a program, so should be rather called afl-frida-trace.so
> 
>     * `make` should not download and build the jpeg test. only when a test... target is called this should be done :)
> 
> 
> not for merge but later:
> 
>     * cross compiling should be possible (so specifying a compiler with CC= and specifying the target with ARCH= or something like that) (being able to run it with musl, diet, uclibc would be great but doesnt seem to be possible)
> 
> 
> an issue I see is that it is specially interesting to run it on an embedded hardware. but this runs into glibc issues (not built with the same version, e.g. `/lib64/libc-2.25.so: version `GLIBC_2.27' not found (required by ./afl-frida-trace)`). but not sure how to solve this easily

I can take care of the first two quite simply, thanks for highlighting them.

The issue of libc compatibility can be addressed by cross compiling FRIDA from source. I have used FRIDA in the past on both uClibc and musl targets (and indeed contributed some patches to fix a few little issues). However, as with all things embedded, it is dependent on first building a toolchain which is a very close match to that used on the target, e.g. using `crosstool-ng` or similar. The primary issue for supporting this would be testing, there are obviously a huge number of different embedded targets with different libc variants and differing versions. I suggest perhaps this might be better tackled by providing a README with instructions on how to cross-compile? Perhaps an example build script could be provided for a popular readily available target to use as a template?

vanhauser-thc:
whenever it is ready it can be merged :)

WorksButNotTested:
> whenever it is ready it can be merged :)

Awesome thanks. I have fixed the two minor issues above and added support for inline assembly on AARCH64, as well as fixing macOS support (I only have Intel to test). I want to improve the docs a bit, but hopefully I'm pretty much there as far as the implementation?

macOS doesn't have support for the `personality` syscall, so we can't disable ASLR dynamically there, but I gather there are tools to disable it by patching the flags in the binary. (This is needed to pass absolute addresses to `AFL_FRIDA_INST_RANGES` and `AFL_FRIDA_EXCLUDE_RANGES`). Otherwise, I guess we can look at supporting relative offsets (e.g. module+0xbadf00d) separate from the PR?

As we can't use `pyelftools` to parse a mac image, the `testinstr_X` build targets don't limit coverage on OSX. Also the `AFL_FRIDA_INST_RANGES` test on AARCH64 doesn't work properly since when FRIDA stalker executes an excluded range, coverage is stopped until you return, even if the excluded code calls a function which is not (this is inherent since when in excluded code we are running the original code, not instrumented code). Thus by having `main` excluded, and the `testinstr` function not, it will not work. Except by happy coincidence on x64 where indirect calls such as the one we make to `main` aren't checked for exclusion (an omission in Stalker which shouldn't cause much impact). None of the exclusion issues are cause by problems in the AFL FRIDA integration per-se, but either by limitations in FRIDA, or the test which I put together. However, I wonder if the `testinstr_X` targets have now served their purpose and can be removed?

Any thoughts on the last few bits which should be address before the merge?

vanhauser-thc:
I just checked in a fix in dev so afl-frida-trace.so/dylib does not need to be executable (because it should not be), please ensure it is not created with executable permissions.

I think then the only thing left is -O support for afl-cmin and afl-cmin.bash ... and that is is I think. then we can merge.

there are a few more things to do, but I will take care of them (adding to to install, distrib and binary-only targets, docs/envs.md, etc.)

WorksButNotTested:
> I just checked in a fix in dev so afl-frida-trace.so/dylib does not need to be executable (because it should not be), please ensure it is not created with executable permissions.
> 
> I think then the only thing left is -O support for afl-cmin and afl-cmin.bash ... and that is is I think. then we can merge.
> 
> there are a few more things to do, but I will take care of them (adding to to distrib and binary-only targets, docs/envs.md, etc.)

The shared object is built here...
https://github.com/AFLplusplus/AFLplusplus/blob/e1384b5086e918350426cd0ece7dbe9c451f771f/frida_mode/Makefile#L99

I assume the compiler should do the right thing?

Awesome, thanks I’m hoping I’ll get time to sort out the tmin things tomorrow. Hoping it will just be a couple of lines to change like the other tools.

Thanks for taking care of the last few housekeeping things for me!

WorksButNotTested:
Ok. I think that should be all the loose ends.

vanhauser-thc:
LGTM, thanks!

PR #835, opened 2021-03-20
closes fd when executing one by one.

wideglide:
Causes too many open files errors if this functionality is used.


domenukk:
Thanks

PR #845, opened 2021-03-25
Dev

joeyjiaojg:


instrumentation/afl-compiler-rt.o.c
```diff
@@ -14,9 +14,7 @@
 
 */
 
-#ifdef __ANDROID__
-  #include "android-ashmem.h"
-#endif
+#include "android-ashmem.h"
```
vanhauser-thc: why should android-ashmem.h also be included by non-android systems?

joeyjiaojg: instrumentation/afl-compiler-rt.o.c is used for all oses, not just android.
And inside android-ashmem.h there is 
```
#ifdef __ANDROID__
```
so no need outer ifdef.

vanhauser-thc: the ifdef in the include is good, but please keep the ifdef in the tools. yes they are not needed, but it surprises other people looking at the code. (because for all other includes we ifdef if they are platform specific)

joeyjiaojg: fixed

vanhauser-thc:
looks good to me. are you done and can this be merged?

joeyjiaojg:
yeah, it's done, please merge.

PR #852, opened 2021-03-30
Add network_proxy build targets to gitignore

rchildre3:
All other build targets in `utils/` are ignored except for these due to the lack of file extension.


domenukk:
Thanks! :)

PR #853, opened 2021-03-30
Fix grammar mutator build script

rchildre3:
Fixes: 6d2ac3e3140 ("fix grammar download script")

The git submodle entry point is "grammar_mutator" not "grammar-mutator"

The build script fails without this


vanhauser-thc:
worked fine on a non-git checkout (subversion) :)
thanks!

PR #854, opened 2021-03-31
fix if A == B always evalutes to true

joeyjiaojg:


joeyjiaojg:
Marco cannot be compared to enum in C as it always true.

vanhauser-thc:
I do not understand this.
`#if GUM_NATIVE_CPU == GUM_CPU_AMD64` why would this be always true?

joeyjiaojg:
Here's the explanation https://stackoverflow.com/questions/34677148/why-does-the-c-preprocessor-consider-enum-values-as-equal

vanhauser-thc:
ah ok, I understand now, thanks :)
I think the ifdef is wrong though, it should be `__x86_64__`

joeyjiaojg:
Fixed

vanhauser-thc:
thank you!

PR #855, opened 2021-04-02
 Add support for standalone leak-sanitizer

### LINKED ISSUE #825: LSAN not aborting in persistent mode
It seems that leaksanitizer (under ASAN) does not work in AFL++'s persistence mode.

The following C code produces a memory leak:
```
#include <stdio.h>
#include <string.h> 
#include <stdlib.h>
#include <unistd.h>
//#include <sanitizer/lsan_interface.h>

__AFL_FUZZ_INIT();
int main() {
__AFL_INIT();
unsigned char *b = __AFL_FUZZ_TESTCASE_BUF;
while(__AFL_LOOP(1000)) {
   size_t len = __AFL_FUZZ_TESTCASE_LEN;
   if(len >= 100 || len == 0) continue;
   char *leak = malloc(len+1);
   memcpy(leak,b,len);
   leak[len] = '\0';
   fprintf(stderr,"%s\n", leak);
   fflush(stderr);
}
}
```

`export ASAN_OPTIONS='abort_on_error=1:detect_leaks=1:symbolize=0' ; AFL_USE_ASAN=1 afl-clang-lto try.c ; printf "AAAAAA" | ./a.out`
produces the correct report:
```
afl-cc ++3.11a by Michal Zalewski, Laszlo Szekeres, Marc Heuse - mode: LLVM-LTO-PCGUARD
afl-llvm-lto++3.11a by Marc "vanHauser" Heuse <mh@mh-sec.de>
Info: Found constructor function asan.module_ctor with prio 1, we will not instrument this, putting it into a block list.
[+] Instrumented 21 locations with no collisions (on average 0 collisions would be in afl-gcc/vanilla AFL) (non-hardened, ASAN mode).
AAAAAA

=================================================================
==1536867==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 7 byte(s) in 1 object(s) allocated from:
    #0 0x2ca6dd  (/tmp/a.out+0x2ca6dd)
    #1 0x2feabd  (/tmp/a.out+0x2feabd)

SUMMARY: AddressSanitizer: 7 byte(s) leaked in 1 allocation(s).
Aborted
```

We then run the same program in afl-fuzz:
`mkdir test/ ; printf "AAAA" > test/1`
`afl-fuzz -i test/ -o test-out/ ./a.out`

but there are no crashes.

I tried to run add a call the __lsan_do_leak_check(); at the end of the loop, but this does not cause a crash either.

MegaManSec:
Add support for standalone leak-sanitizer, introducting the environment variable AFL_USE_LSAN.

AFL_USE_LSAN introduces the macro __AFL_LEAK_CHECK() which will check
for a memory leak when the macro is run. This is especially helpful
when using __AFL_LOOP().

If __AFL_LEAK_CHECK() is not used when AFL_USE_LSAN=1 is set,
the leak checker will run when the program exits.

Fixes #825.

### DIFF
--- a/README.md
+++ b/README.md
@@ -601,7 +601,7 @@ Every -M/-S entry needs a unique name (that can be whatever), however the same
 For every secondary fuzzer there should be a variation, e.g.:
  * one should fuzz the target that was compiled differently: with sanitizers
    activated (`export AFL_USE_ASAN=1 ; export AFL_USE_UBSAN=1 ;
-   export AFL_USE_CFISAN=1 ; `
+   export AFL_USE_CFISAN=1 ; export AFL_USE_LSAN`)
  * one should fuzz the target with CMPLOG/redqueen (see above)
  * one to three fuzzers should fuzz a target compiled with laf-intel/COMPCOV
    (see above). Important note: If you run more than one laf-intel/COMPCOV
--- a/README.md
+++ b/README.md
@@ -601,7 +601,7 @@ Every -M/-S entry needs a unique name (that can be whatever), however the same
 For every secondary fuzzer there should be a variation, e.g.:
  * one should fuzz the target that was compiled differently: with sanitizers
    activated (`export AFL_USE_ASAN=1 ; export AFL_USE_UBSAN=1 ;
-   export AFL_USE_CFISAN=1 ; export AFL_USE_LSAN`)
+   export AFL_USE_CFISAN=1 ; export AFL_USE_LSAN=1`)
  * one should fuzz the target with CMPLOG/redqueen (see above)
  * one to three fuzzers should fuzz a target compiled with laf-intel/COMPCOV
    (see above). Important note: If you run more than one laf-intel/COMPCOV
--- a/docs/env_variables.md
+++ b/docs/env_variables.md
@@ -55,7 +55,7 @@ make fairly broad use of environmental variables instead:
     overridden.
 
   - Setting `AFL_USE_ASAN` automatically enables ASAN, provided that your
-    compiler supports that. Note that fuzzing with ASAN is mildly challenging
+    compiler supports itt. Note that fuzzing with ASAN is mildly challenging
     - see [notes_for_asan.md](notes_for_asan.md).
 
     (You can also enable MSAN via `AFL_USE_MSAN`; ASAN and MSAN come with the
@@ -63,6 +63,13 @@ make fairly broad use of environmental variables instead:
     similarly by setting the environment variable `AFL_USE_UBSAN=1`. Finally
     there is the Control Flow Integrity sanitizer that can be activated by
     `AFL_USE_CFISAN=1`)
+
+  - Setting `AFL_USE_LSAN` automatically enables Leak-Sanitizer, provided
+    that your compiler supports it. To perform a leak check within your
+    program at a certain point (such as at the end of an __AFL_LOOP,
+    you can run the macro __AFL_CHECK_LEAK(); which will cause
+    an abort if any memory is leaked (you can combine this with the
+    LSAN_OPTIONS=suppressions option to supress some known leaks).
 
   - Setting `AFL_CC`, `AFL_CXX`, and `AFL_AS` lets you use alternate downstream
     compilation tools, rather than the default 'clang', 'gcc', or 'as' binaries
@@ -628,7 +635,12 @@ optimal values if not already present in the environment:
     msan_track_origins=0
     allocator_may_return_null=1
 ```
-  Be sure to include the first one when customizing anything, since some
-    MSAN versions don't call `abort()` on error, and we need a way to detect
-    faults.
+  - Similarly, the default `LSAN_OPTIONS` are set to:
+```
+    exit_code=86
+    fast_unwind_on_malloc=0
+````
+  Be sure to include the first ones for LSAN and MSAN when customizing
+     anything, since some MSAN and LSAN versions don't call `abort()` on
+     error, and we need a way to detect faults.
 
--- a/docs/env_variables.md
+++ b/docs/env_variables.md
@@ -55,7 +55,7 @@ make fairly broad use of environmental variables instead:
     overridden.
 
   - Setting `AFL_USE_ASAN` automatically enables ASAN, provided that your
-    compiler supports itt. Note that fuzzing with ASAN is mildly challenging
+    compiler supports it. Note that fuzzing with ASAN is mildly challenging
     - see [notes_for_asan.md](notes_for_asan.md).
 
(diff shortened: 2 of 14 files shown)


README.md
```diff
@@ -601,7 +601,7 @@ Every -M/-S entry needs a unique name (that can be whatever), however the same
 For every secondary fuzzer there should be a variation, e.g.:
  * one should fuzz the target that was compiled differently: with sanitizers
    activated (`export AFL_USE_ASAN=1 ; export AFL_USE_UBSAN=1 ;
-   export AFL_USE_CFISAN=1 ; `
+   export AFL_USE_CFISAN=1 ; export AFL_USE_LSAN`)
```
vanhauser-thc: please add `=1` to be consistent with the others

docs/env_variables.md
```diff
@@ -628,7 +635,12 @@ optimal values if not already present in the environment:
...
     allocator_may_return_null=1
 ```
-  Be sure to include the first one when customizing anything, since some
-    MSAN versions don't call `abort()` on error, and we need a way to detect
-    faults.
+  - Similarly, the default `LSAN_OPTIONS` are set to:
+```
+    exit_code=23
+    fast_unwind_on_malloc=0
+````
```
vanhauser-thc: one ` too much :)

src/afl-analyze.c
```diff
@@ -781,6 +781,19 @@ static void set_up_environment(void) {
...
 
+  x = get_afl_env("LSAN_OPTIONS");
+
+  if (x) {
+
+    if (!strstr(x, "exit_code=" STRINGIFY(LSAN_ERROR))) {
+
+      FATAL("Custom LSAN_OPTIONS set without exit_code=" STRINGIFY(
+          LSAN_ERROR) " - please fix!");
+
```
vanhauser-thc, line 791: isnt exit_code 23 the default? is this check necessary then? if lsan has symbolize - that you have to check to be 0 (if afl->debug is not set)

src/afl-cc.c
```diff
@@ -758,7 +758,7 @@ static void edit_params(u32 argc, char **argv, char **envp) {
     if (!strncmp(cur, "-fsanitize-coverage-", 20) && strstr(cur, "list="))
       have_instr_list = 1;
 
-    if (!strcmp(cur, "-fsanitize=address") || !strcmp(cur, "-fsanitize=memory"))
+    if (!(strcmp(cur, "-fsanitize=address") && strcmp(cur, "-fsanitize=memory")))
       asan_set = 1;
```
vanhauser-thc: the original line is correct. if either one is set then USE_ASAN/USE_MSAN is ignored as only one of them can be set.

src/afl-cc.c
```diff
@@ -914,6 +918,13 @@ static void edit_params(u32 argc, char **argv, char **envp) {
 
   }
 
+  if (getenv("AFL_USE_LSAN")) {
+    cc_params[cc_par_cnt++] = "-includesanitizer/lsan_interface.h";
```
vanhauser-thc: is this line correct? it looks .... weird :)

src/afl-forkserver.c
```diff
@@ -538,6 +538,14 @@ void afl_fsrv_start(afl_forkserver_t *fsrv, char **argv,
...
            1);
 
+    /* LSAN, too, does not support abort_on_error=1. */
+
+    if (!getenv("LSAN_OPTIONS"))
+     setenv("LSAN_OPTIONS",
+            "exitcode=" STRINGIFY(LSAN_ERROR) ":"
+            "fast_unwind_on_malloc=0",
+            1);
+
```
vanhauser-thc, line 550: have you checked for other defaults that might hurt performance and do not help detect issues? eg. symbolize

vanhauser-thc:
For the future please note that we only accepts PRs to dev, I changed the base now.

A few questions and issues, looks good otherwise!

src/afl-cc.c
MegaManSec: Yup it's correct, unless there's some other way afl does includes. To use __lsan_do_leak_check, the program needs to be built with sanitizer/lsan_interface.h

src/afl-cc.c
MegaManSec: well they are the same thing, but i had originally changed it because i had added a check for something else, then removed it. i will change it back though ;)

src/afl-forkserver.c
MegaManSec: Documentation for LSAN_OPTIONS is.. limited:-). But it seems everything is disabled by default, according to https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer

However, while LSAN_OPTIONS can be used to pass other options which asan may use, e.g. `handle_sigill`, I do not think this would really matter in an LSAN way.

I can of course add the ones which have been added for, let's say, UBSAN_OPTIONS, if you want though.

src/afl-cc.c
vanhauser-thc: I have never seen the option before :)

src/afl-analyze.c
```diff
@@ -818,6 +830,13 @@ static void set_up_environment(void) {
                          "handle_sigfpe=0:"
                          "handle_sigill=0", 0);
 
+   setenv("LSAN_OPTIONS",
+         "exitcode=" STRINGIFY(LSAN_ERROR) ":"
+         "fast_unwind_on_malloc=0:"
+         "symbolize=0",
+         0);
+
```
vanhauser-thc, line 839: can you please add `print_suppressions=0` ?

src/afl-cc.c
```diff
@@ -914,6 +918,13 @@ static void edit_params(u32 argc, char **argv, char **envp) {
...
+  }
+
+  cc_params[cc_par_cnt++] =
+      "-D__AFL_LEAK_CHECK()=__lsan_do_leak_check()";
+
```
vanhauser-thc: I now see that -D__AFL_LEAK_CHECK  is not part of if(AFL_USE_ASAN) above. IMHO it makes more sense to move it in there?

src/afl-cc.c
MegaManSec: I'm not sure how portable it is, but it should be supported by any compiler (aka just clang) which supports LSAN, anyways.

src/afl-cc.c
MegaManSec: I had originally put it here (forgetting to put if(..)) to collate all the definitions in one place, but I suppose it will fit fine in the if(..) above

vanhauser-thc:
LGTM, thank you!

MegaManSec:
Thank you for the assistance.

PR #856, opened 2021-04-03
Autodict-QL : Token generation for fuzzing with CodeQL

Microsvuln:
Hey guys.

This is the initial PR for Autodict-QL, a token generation plugin for AFL++ with using the power of CodeQL semantic analysis engine.  (Somehow integrating CodeQL usability to AFL++ :-) )

For now, there are 5 codeql scripts that extracts tokens from the CodeBase and use them as dictionaries for fuzzing. additional scripts can also be added and the integration is so easy .

Please let me know your comments :-)

Thanks
Arash


vanhauser-thc:
looks good to me, thank you!

DuckRui:
Hello, I have a question about Autodict-QL.I used it  generates some tokens. Do I must let AFL++ fuzzing in deterministic(set -D ), if I want these tokens usdful? Beacuse I see havoc stage will also use tokens to replace or insert. I want to know whether deterministic stage is must?

vanhauser-thc:
you do not need -D, havoc uses all mutation strategies.

DuckRui:
> you do not need -D, havoc uses all mutation strategies.

I get it. Thank you very much!

PR #860, opened 2021-04-07
android: disable sigaction inside debuggerd

joeyjiaojg:
check https://github.com/google/AFL/blob/master/docs/INSTALL#L173

### DIFF
--- a/instrumentation/afl-compiler-rt.o.c
+++ b/instrumentation/afl-compiler-rt.o.c
@@ -1144,6 +1144,18 @@ void __afl_manual_init(void) {
 
 __attribute__((constructor())) void __afl_auto_init(void) {
 
+#ifdef __ANDROID__
+  // Disable handlers in linker/debuggerd, check include/debuggerd/handler.h
+  signal(SIGABRT, SIG_DFL);
+  signal(SIGBUS, SIG_DFL);
+  signal(SIGFPE, SIG_DFL);
+  signal(SIGILL, SIG_DFL);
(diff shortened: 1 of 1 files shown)


hexcoder- (approved):
Thanks!

vanhauser-thc:
@hexcoder- that was too early! :)

@joeyjiaojg we were discussing this PR internally. When compiling with ASAN, this hijacks signals, so likely this patch will not work with asan on android? have you tested this?
If asan makes problems with this, you can try moving the setting to `__afl_auto_first()` which might be before asan.

andreafioraldi:
__afl_auto_first is nto before asan because sanitizers use the preinit array

joeyjiaojg:
Can you explain more why asan breaks?
ASAN for AFL++ on Android breaks even before this change, so I mainly use AFL++ only for afl-frida.

vanhauser-thc:
we are not saying "does break", only "could break". I recommend testing with asan and the signal handler reset if things still work for you.

joeyjiaojg:
> __afl_auto_first is nto before asan because sanitizers use the preinit array

right, .preinit_array initialized in linker before calling the constructor.
Anyhow, I haven't succeeded to bisect asan failure reason yet.

joeyjiaojg:
No matter if I enabled the signal patch or not, HWAddressSanitizer always crashes the target binary.

It breaks in this line
```
  if (getenv("AFL_DISABLE_LLVM_INSTRUMENTATION")) return;
```
Which seems not reasonable to me.

PR #864, opened 2021-04-07
android: rename to aflpp* to coexist with google/AFL

joeyjiaojg:


domenukk:
What is the use case you need this for? I don't think we want to rename any of the existing tools, maybe there is a better solution.

vanhauser-thc:
Same. with options like `-Z -p explore -D` you have basically vanilla afl (just more speed), so having old afl in parallel serves no purpose.
and vanilla afl should be renamed not afl++ :)

joeyjiaojg:
I don't agree google/AFL is the same as AFL++ w/ -Z -p explore -D, but I can keep the change in my repo only.

For me, both google/AFL and AFL++ are needed as I'm still using afl-smart and HWASAN binary. And I agree I can use AFL++ more frequency in future to see if it can replace google/AFL.

vanhauser-thc:
also you could have your own patches in vanilla that you want to use.
But the thing is - adding this to the official afl++ that renames the files to something different will be very surprising to users.
if you would push this patch to the vanilla afl repo - that would make sense though.

So - please switch them up so that the afl++ variants are their name here (afl-fuzz) and the vanilla afl installation something like afl-old-... afl-vanilla-.... afl-my-.... afl-private-... whatever.
thank you!

domenukk:
Closing this for now (sorry if it breaks your personal workflow ;) )

PR #880, opened 2021-04-16
Frida persistent

WorksButNotTested:
I have now added support on x64 for persistent mode (function call only so far), in-memory test cases and complog.
Grateful for any feedback.


vanhauser-thc:
This is for the first round:

1. please merge from dev backwards and do `make code-format` then push the changes. this will properly format your code :)

2. please move all test files like frida_parallel.sh etc. into the test/ subfolder - or delete them.
also put -D in all the afl-fuzz lines, this eliminates most of the randomness and is better for testing.
Also maybe put the test cases of the Makefile into the test/ subfolder so the main Makefile is easier to read.
And put a newline at the end of all files

too tired today to thoroughly look through the code now :)

frida_mode/Makefile
```diff
@@ -1,13 +1,17 @@
...
+PARENT:=$(shell realpath $(PWD)..)/
 INC_DIR:=$(PWD)include/
 SRC_DIR:=$(PWD)src/
 INCLUDES:=$(wildcard $(INC_DIR)*.h)
-SOURCES:=$(wildcard $(SRC_DIR)*.c)
+SOURCES:=$(wildcard $(SRC_DIR)**/*.c) $(wildcard $(SRC_DIR)*.c)
 BUILD_DIR:=$(PWD)build/
 CFLAGS+=-fPIC -D_GNU_SOURCE
+QEMU_INC_DIR:=$(PARENT)qemu_mode/qemuafl/qemuafl/
+QEMU_INC_API:=$(QEMU_INC_DIR)api.h
```
domenukk: Frida mode should probably not depend on qemuafl being checked out?

frida_mode/frida_parallel.sh
```diff
@@ -0,0 +1,33 @@
...
+export AFL_FRIDA_PERSISTENT_ADDR=$($PWD/test/get_symbol_addr.py -f $TEST_BIN -s main -b $TESTINSTR_BASE)
+export AFL_FRIDA_INST_RANGES=$($PWD/test/get_section_addrs.py -f $TEST_BIN -s .text -b $TESTINSTR_BASE)
+export AFL_FRIDA_INST_STRICT=1
+
+echo AFL_FRIDA_PERSISTENT_ADDR = $AFL_FRIDA_PERSISTENT_ADDR
+echo AFL_FRIDA_INST_RANGES = $AFL_FRIDA_INST_RANGES
+
+rm -rf $FRIDA_OUT
+
+tmux new-window -n:master  "AFL_FRIDA_PERSISTENT_ADDR=$AFL_FRIDA_PERSISTENT_ADDR AFL_FRIDA_INST_RANGES=$AFL_FRIDA_INST_RANGES AFL_FRIDA_INST_STRICT=1 $PARENT/afl-fuzz -O -V60 -i $TEST_DATA_DIR -o $FRIDA_OUT -D master  -- $TEST_BIN @@"
```
domenukk: `main` and `secondary` nodes, please :)

frida_mode/frida_parallel_hook.sh
```diff
@@ -0,0 +1,36 @@
...
+export AFL_FRIDA_INST_RANGES=$($PWD/test/get_section_addrs.py -f $TEST_BIN -s .text -b $TESTINSTR_BASE)
+export AFL_FRIDA_INST_STRICT=1
+
+echo AFL_FRIDA_PERSISTENT_ADDR = $AFL_FRIDA_PERSISTENT_ADDR
+echo AFL_FRIDA_INST_RANGES = $AFL_FRIDA_INST_RANGES
+echo AFL_FRIDA_PERSISTENT_HOOK = $AFL_FRIDA_PERSISTENT_HOOK
+
+rm -rf $FRIDA_OUT
+
+tmux new-window -n:master  "AFL_FRIDA_PERSISTENT_HOOK=$AFL_FRIDA_PERSISTENT_HOOK AFL_FRIDA_PERSISTENT_ADDR=$AFL_FRIDA_PERSISTENT_ADDR AFL_FRIDA_INST_RANGES=$AFL_FRIDA_INST_RANGES AFL_FRIDA_INST_STRICT=1 $PARENT/afl-fuzz -O -V60 -i $TEST_DATA_DIR -o $FRIDA_OUT -D master  -- $TEST_BIN $AFLPP_DRIVER_DUMMY_INPUT"
```
domenukk: Same

frida_mode/qemu_parallel.sh
```diff
@@ -0,0 +1,33 @@
...
+export AFL_QEMU_PERSISTENT_ADDR=$($PWD/test/get_symbol_addr.py -f $TEST_BIN -s main -b $TESTINSTR_BASE_QEMU)
+export AFL_QEMU_INST_RANGES=$($PWD/test/get_section_addrs.py -f $TEST_BIN -s .text -b $TESTINSTR_BASE_QEMU)
+export AFL_QEMU_PERSISTENT_GPR=1
+
+echo AFL_QEMU_PERSISTENT_ADDR = $AFL_QEMU_PERSISTENT_ADDR
+echo AFL_QEMU_INST_RANGES = $AFL_QEMU_INST_RANGES
+
+rm -rf $QEMU_OUT
+
+tmux new-window -n:master  "AFL_QEMU_PERSISTENT_ADDR=$AFL_QEMU_PERSISTENT_ADDR AFL_QEMU_INST_RANGES=$AFL_QEMU_INST_RANGES AFL_QEMU_PERSISTENT_GPR=1 $PARENT/afl-fuzz -V60 -Q -i $TEST_DATA_DIR -o $QEMU_OUT -D master  -- $TEST_BIN @@"
```
domenukk: 3rd time.
In any case, do we want these scripts?

frida_mode/src/complog/complog.c
```diff
@@ -0,0 +1,72 @@
...
+        range->base_address + range->size);
+
+  }
+
+}
+
+static gboolean complog_overlaps(GumAddress base1, GumAddress limit1,
+                                 GumAddress base2, GumAddress limit2) {
+
+  if (base1 < base2) return false;
```
domenukk: Could be simplified?

frida_mode/src/complog/complog_x64.c
```diff
@@ -0,0 +1,367 @@
...
+
+}
+
+static guint64 cmplog_get_operand_value(GumCpuContext *context,
+                                        complog_ctx_t *ctx) {
+
+  switch (ctx->type) {
+
+    case X86_OP_REG:
+      return complog_read_reg(context, ctx->reg);
```
domenukk, line 188: What about non-64 byte compares?

frida_mode/src/complog/complog_x64.c
```diff
@@ -0,0 +1,367 @@
...
+      FATAL("Invalid operand type: %d\n", operand->type);
+
+  }
+
+}
+
+static void complog_instrument_call_put_callout(GumStalkerIterator *iterator,
+                                                cs_x86_op *         operand) {
+
+  complog_ctx_t *ctx = g_malloc(sizeof(complog_ctx_t));
```
domenukk, line 235: unchecked malloc?

frida_mode/src/complog/complog_x64.c
```diff
@@ -0,0 +1,367 @@
...
+
+  complog_handle_cmp_sub(context, operand1, operand2, ctx->operand1.size);
+
+}
+
+static void complog_instrument_cmp_sub_put_callout(GumStalkerIterator *iterator,
+                                                   cs_x86_op *         operand1,
+                                                   cs_x86_op *operand2) {
+
+  complog_pair_ctx_t *ctx = g_malloc(sizeof(complog_pair_ctx_t));
```
domenukk, line 305: unchecked

frida_mode/src/complog/complog_x64.c
```diff
@@ -0,0 +1,367 @@
...
+
+  if (x86.op_count != 2) return;
+
+  operand1 = &x86.operands[0];
+  operand2 = &x86.operands[1];
+
+  if (operand1->type == X86_OP_INVALID) return;
+  if (operand2->type == X86_OP_INVALID) return;
+
+  if (operand1->type == X86_OP_MEM && operand1->mem.segment != X86_REG_INVALID)
```
domenukk: I personally perfer brakets around conditions, especially if they are multiline :)

frida_mode/Makefile
WorksButNotTested: I'm hoping to share much of the code for qemu asan down the line. There enough similarity that it doesn't make sense to re-implement it all. So I figured a dependency on QEMU mode was preferable to duplication?

frida_mode/frida_parallel.sh
WorksButNotTested: These scripts are gone now.

frida_mode/frida_parallel_hook.sh
WorksButNotTested: These scripts are gone now.

frida_mode/qemu_parallel.sh
WorksButNotTested: These scripts are gone now.

frida_mode/src/complog/complog.c
WorksButNotTested: Done.

frida_mode/src/complog/complog_x64.c
WorksButNotTested: The function `complog_read_reg` deals with comparisons of different sizes. Although a 64bit type is used to represent the value, it is masked and shifted as necessary.

frida_mode/src/complog/complog_x64.c
WorksButNotTested: Fixed.

frida_mode/src/complog/complog_x64.c
WorksButNotTested: Fixed

frida_mode/src/complog/complog_x64.c
WorksButNotTested: Done

WorksButNotTested:
I have now added support for `AFL_INST_LIBS`, restructured things to get all the test code out of the way and updated the docs. Let me know what you think.

frida_mode/Makefile
vanhauser-thc: The thing is that qemuafl is not necessarily checked out. I think it is fine if you copy the file and put it into the frida_mode include directory

frida_mode/Makefile
```diff
@@ -1,86 +1,58 @@
...
 endif
 
 ifndef OS
  $(error "Operating system unsupported")
 endif
 
-VERSION=14.2.13
-GUM_DEVKIT_FILENAME=frida-gum-devkit-$(VERSION)-$(OS)-$(ARCH).tar.xz
-GUM_DEVKIT_URL="https://github.com/frida/frida/releases/download/$(VERSION)/$(GUM_DEVKIT_FILENAME)"
+GUM_DEVKIT_VERSION=14.2.13
```
vanhauser-thc: current is 14.2.17
do you want to pin it to that specific version? or update the version from time to time?
you could also get the most current version automatically and if in the future this fails then its pinned by hand (or fixed)?

utils/qemu_persistent_hook/read_into_rdi.c
```diff
@@ -5,7 +5,7 @@
 
 void afl_persistent_hook(struct x86_64_regs *regs, uint64_t guest_base,
                          uint8_t *input_buf, uint32_t input_buf_len) {
-
+\
```
vanhauser-thc: wrong blackslash

utils/aflpp_driver/aflpp_qemu_driver_hook.c
```diff
@@ -5,7 +5,7 @@
 
 void afl_persistent_hook(struct x86_64_regs *regs, uint64_t guest_base,
                          uint8_t *input_buf, uint32_t input_buf_len) {
-
+\
```
vanhauser-thc: wrong blackslash

src/afl-fuzz-init.c
```diff
@@ -2774,6 +2774,14 @@ void check_binary(afl_state_t *afl, u8 *fname) {
 
     WARNF("AFL_PERSISTENT is no longer supported and may misbehave!");
 
+  } else if (getenv("AFL_FRIDA_PERSISTENT_ADDR")) {
+
+    OKF(cPIN "FRIDA Persistent mode configuration options detected.");
```
vanhauser-thc: does not need cPIN IMHO

vanhauser-thc:
please install clang-format-11, do a `make code-format` and push the changes.

besides the comments it looks good in my opinion.

frida_mode/Makefile
domenukk: Or we can pull it up to aflplusplus directly, and include it like we do with `types.h` and `config.h`?

frida_mode/Makefile
```diff
@@ -1,86 +1,58 @@
 PWD:=$(shell pwd)/
```
domenukk: The file should probably be a `GNUMakefile` (unless it really doesn't use GNU), we usually print a little info for non-gnu systems.

frida_mode/README.md
```diff
@@ -10,23 +10,23 @@ a small harness around their target code of interest, FRIDA mode instead takes a
...
-  | Snapshot LKM Support     |            |
+As FRIDA mode is new, it is missing a lot of features. The design is such that it
+should be possible to add these features in a similar manner to QEMU mode and
+perhaps leverage some of its design and implementation.
+
+  | Feature/Instrumentation  | frida-mode | Notes                                   |
+  | -------------------------|:----------:|:---------------------------------------:|
+  | NeverZero                |     x      |                                         |
+  | Persistent Mode          |     x      | (x64 only)(Only on function boundaries) |  
+  | LAF-Intel / CompCov      |     -      | (Superseded by CmpLog)                  |
```
domenukk, line 21: Wouldn't it be cool (and not to hard) to support both, for different use-cases?

frida_mode/include/stalker.h
```diff
@@ -0,0 +1,8 @@
+#include "frida-gum.h"
+
+void        stalker_init();
```
domenukk: This should be
```c
void stalker_init(void);
```

(also true for the following lines)

frida_mode/include/ranges.h
```diff
@@ -1,6 +1,6 @@
 #include "frida-gum.h"
 
-void ranges_init(GumStalker *stalker);
+void ranges_init();
```
domenukk: should be `void ranges_init(void);`

frida_mode/src/complog/complog.c
```diff
@@ -0,0 +1,72 @@
...
+}
+
+gint complog_sort(gconstpointer a, gconstpointer b) {
+
+  return ((GumMemoryRange *)b)->base_address -
+         ((GumMemoryRange *)a)->base_address;
+
+}
+
+void complog_init() {
```
domenukk: `void cmplog_init(void) {`

frida_mode/src/complog/complog.c
```diff
@@ -0,0 +1,72 @@
...
+
+static gboolean complog_range(const GumRangeDetails *details,
+                              gpointer               user_data) {
+
+  GumMemoryRange range = *details->range;
+  g_array_append_val(complog_ranges, range);
+
+}
+
+gint complog_sort(gconstpointer a, gconstpointer b) {
```
domenukk: I guess this could be `static`?

frida_mode/src/complog/complog_x64.c
```diff
@@ -0,0 +1,427 @@
...
+    case X86_REG_EBP:
+      return ctx->rbp & GUM_INT32_MASK;
+
+    case X86_REG_ESI:
+      return ctx->rsi & GUM_INT32_MASK;
+
+    case X86_REG_EDI:
+      return ctx->rdi & GUM_INT32_MASK;
+
+    case X86_REG_R8D:
```
domenukk: This duplicated code could be simplified/combined with `#define`s

frida_mode/src/instrument/instrument.c
```diff
@@ -0,0 +1,145 @@
...
+#include "stalker.h"
+
+static gboolean               tracing = false;
+static gboolean               optimize = false;
+static gboolean               strict = false;
+static GumStalkerTransformer *transformer = NULL;
+
+uint64_t __thread previous_pc = 0;
+
+static void on_basic_block(GumCpuContext *context, gpointer user_data) {
```
domenukk: you can try marking this with `__attribute__ ((hot))` and see if it improves speed (or not, who knows)

frida_mode/src/instrument/instrument.c
```diff
@@ -0,0 +1,145 @@
...
+   * the critical path with additional stack adjustments if tracing is not
+   * enabled. If tracing is enabled, then we're printing a load of diagnostic
+   * information so this overhead is unlikely to be noticeable.
+   */
+  static char buffer[200];
+  int         len;
+  guint64     current_pc = (guint64)user_data;
+  uint8_t *   cursor;
+  uint64_t    value;
+  if (tracing) {
```
domenukk: Maybe make this `unlikely(tracing)` if you have unlikely available, or define it?

frida_mode/src/instrument/instrument.c
```diff
@@ -0,0 +1,145 @@
...
+
+  }
+
+  current_pc = (current_pc >> 4) ^ (current_pc << 8);
+  current_pc &= MAP_SIZE - 1;
+
+  cursor = &__afl_area_ptr[current_pc ^ previous_pc];
+  value = *cursor;
+
+  if (value == 0xffffffffffffffff)
```
domenukk: This can never be true, as `*cursor` is of (unsigned) type `uint8_t`. Max value is 0xff.
Tested it here: https://godbolt.org/z/498zx1K87

P.S.: I personally prefer brackets around `if` statements, especially with subsequent `else`

frida_mode/Makefile
WorksButNotTested: Turns out it's only a single structure I need for the CPU regs, so I have just duplicated it. I only need it as I have reused the `utils/aflpp_driver/aflpp_qemu_driver_hook.so` and so need to pass the context in the same format.

Down the line, its [this](https://github.com/AFLplusplus/AFLplusplus/tree/stable/qemu_mode/libqasan) code I expect to be using, so that's not in qemuafl and so doesn't have the same problems.

frida_mode/Makefile
WorksButNotTested: I have bumped the version to the latest. I think I would rather hard-code the version so we have the opportunity to test new versions rather than potentially hit issues in the unlikely event that a new FRIDA release breaks something. Given the makefile is now very simple, I would expect someone wanting to use a newer version would be easily able to do so. In any case, I expect that people hacking on this would actually prefer to build FRIDA from source, but I didn't want to add it as submodules and incur all of their dependencies and additional compilation time.

utils/qemu_persistent_hook/read_into_rdi.c
WorksButNotTested: Removed, it looks like the auto formatting might be adding this in?

utils/aflpp_driver/aflpp_qemu_driver_hook.c
WorksButNotTested: Removed, it looks like the auto formatting might be adding this in?

src/afl-fuzz-init.c
WorksButNotTested: Removed

frida_mode/Makefile
WorksButNotTested: I've moved everything to using GNUmakefile now with a Makefile wrapper.

frida_mode/README.md
vanhauser-thc: compcov is very different to implement, much more work and less often effective than cmplog.
so putting compcov on the todo list but putting it somewhere on the bottom is the right thing IMHO :)

WorksButNotTested: Happy to add it somewhere down the line if there's a need. When I was discussing with @vanhauser-thc, he seemed to say that CompLog was much better? If there are use-cases where CompCov is better, or CompLog unsuitable, then I am happy to re-evaluate? CompLog support shouldn't be too hard, but it will take some time and would need implementing for each architecture.

frida_mode/include/stalker.h
WorksButNotTested: I have fixed this issue for all sources and headers now.

frida_mode/include/ranges.h
WorksButNotTested: I have fixed this issue for all sources and headers now.

frida_mode/src/complog/complog.c
WorksButNotTested: I have fixed this issue for all sources and headers now.

frida_mode/src/complog/complog.c
WorksButNotTested: Fixed

frida_mode/src/complog/complog_x64.c
WorksButNotTested: Fixed.

frida_mode/src/instrument/instrument.c
WorksButNotTested: It can't do any harm, so I've added it in. Should only take effect on x86 and aarch32 for now, or with AFL_FRIDA_NO_OPTIMISE

frida_mode/src/instrument/instrument.c
WorksButNotTested: Fixed

frida_mode/src/instrument/instrument.c
WorksButNotTested: Good spot. Weird the compiler didn't warn me! Fixed

WorksButNotTested:
> please install clang-format-11, do a `make code-format` and push the changes.
> 
> besides the comments it looks good in my opinion.

I've run the make code-format now. I also have a `make format` target of my own which just does the frida mode code. Like I said though, it seems keen to introduce those weird backslashes?

vanhauser-thc:
LGTM

frida_mode/README.md
domenukk: if you have decoded the instruction already, it should be straight forward to emit compcov coverage at that spot in frida.
I agree complog is better, though. Ignore it for now :)

PR #889, opened 2021-04-28
Integer overflow/underflow fixes in libdislocator

Zherya:
1. Adding the check of integer overflow of 'max_mem' variable when calculating bytes number from megabytes number; changing 'max_mem' type from 'u32' to 'size_t' to be able to store more than 4 GB value on 64-bit machines;

2. Consider the following scenario: multithreaded application, one thread allocates some data and the another thread frees that data. 'total_mem' variable is defined as thread local, so when the second thread frees the data, its 'total_mem' value may be underflowed (line 353):
https://github.com/AFLplusplus/AFLplusplus/blob/da65eef57279dae7b652c40bdb31bc6cd749f63d/utils/libdislocator/libdislocator.so.c#L341-L353 Also, in my opinion, there is no sense of making 'total_mem' value separate for each thread as long as all threads share the same heap.

### DIFF
--- a/utils/libdislocator/libdislocator.so.c
+++ b/utils/libdislocator/libdislocator.so.c
@@ -144,7 +144,7 @@ typedef struct {
 
 /* Configurable stuff (use AFL_LD_* to set): */
 
-static u32 max_mem = MAX_ALLOC;         /* Max heap usage to permit         */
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
the _Atomic might break platforms like Haiko or some *BSD ... lets see. thank you!

PR #891, opened 2021-04-28
Various fixes to work on aarch64 mode (copied to other architectures)

WorksButNotTested:


vanhauser-thc:
is this ready to be merged?

frida_mode/GNUmakefile
```diff
@@ -10,11 +10,19 @@ OBJS:=$(foreach src,$(SOURCES),$(OBJ_DIR)$(notdir $(patsubst %.c, %.o, $(src))))
 CFLAGS+=-fPIC \
 		-D_GNU_SOURCE \
 		-D_FORTIFY_SOURCE=2 \
-		-Wno-pointer-arith \
+		-Werror \
+		-Wall \
+		-Wextra \
+		-Wpointer-arith \
```
vanhauser-thc: please remove -Werror -Wall and -Wextra
this is great if you are developing, but it should not be in the branch.
also rather set -Wno-pointer-arith

frida_mode/GNUmakefile
```diff
@@ -31,7 +39,9 @@ endif
 
 ifeq "$(shell uname)" "Darwin"
  OS:=macos
- CFLAGS:=$(CFLAGS) -Wno-deprecated-declarations
+ RT_CFLAGS:=$(RT_CFLAGS) -Wno-deprecated-declarations
+else
+ RT_CFLAGS:=$(RT_CFLAGS) -Wno-prio-ctor-dtor
```
vanhauser-thc, line 50: this needs to be removed. some compilers do not now prio-ctor-dtor

vanhauser-thc:
if you rebase the comment about Werror etc. should be fixed

frida_mode/GNUmakefile
WorksButNotTested: Ok no worries. Are you ok if I keep them e.g. if you build with ‘DEV=1 make’ or something they are enabled. Otherwise they aren’t?

frida_mode/GNUmakefile
WorksButNotTested: Sure no problems. It may lead to warnings in the build though. I assume that’s preferable?

frida_mode/GNUmakefile
WorksButNotTested: Or maybe a separate (non-default) make target which enables the options?

vanhauser-thc: as you want "make debug" or "make DEBUG=1" or dev/DEV ... as you like

PR #895, opened 2021-04-30
Push to stable

vanhauser-thc:
There is a bug in afl-clang-fast pcguard (default for 10.0.1 and newer):

```
$ cat main.c
int main(void) { return 0; }
```
that would fail with `afl-clang-fast -pie main.c`

so we should push to stable.


.gitmodules
```diff
@@ -7,3 +7,7 @@
 [submodule "qemu_mode/qemuafl"]
 	path = qemu_mode/qemuafl
 	url = https://github.com/AFLplusplus/qemuafl
+[submodule "frida_mode/frida"]
+	path = frida_mode/frida
+	url = https://github.com/WorksButNotTested/frida.git
```
domenukk: Shouldn't we depend on upstream frida istead?

docs/Changelog.md
```diff
@@ -10,6 +10,7 @@ sending a mail to <afl-users+subscribe@googlegroups.com>.
 
 ### Version ++3.13a (development)
```
domenukk, line 11: Add changelog for grammar mutatior?

utils/libdislocator/libdislocator.so.c
```diff
@@ -500,19 +500,20 @@ size_t malloc_usable_size(const void *ptr) {
...
 
   if (tmp) {
 
-    u8 *tok;
-    s32 mmem = (s32)strtol((char *)tmp, (char **)&tok, 10);
-    if (*tok != '\0' || errno == ERANGE) FATAL("Bad value for AFL_LD_LIMIT_MB");
+    char *             tok;
+    unsigned long long mmem = strtoull(tmp, &tok, 10);
+    if (*tok != '\0' || errno == ERANGE || mmem > SIZE_MAX / 1024 / 1024)
+      FATAL("Bad value for AFL_LD_LIMIT_MB");
```
domenukk, line 510: I've seen a commit in the history recently called "Dominik loves brackets" :P

.gitmodules
vanhauser-thc: current upstream frida has a breaking bug. this is not for release but so we can push to stable for the moment.

docs/Changelog.md
vanhauser-thc: ok

utils/libdislocator/libdislocator.so.c
vanhauser-thc: but he doesn't love them so much he does the work ;) (this is from an external PR)

frida_mode/include/complog.h
```diff
@@ -0,0 +1,14 @@
+#ifndef _COMPLOG_H
```
hexcoder-: In Frida all occurences of "CompLog" should be renamed to our previously used "CmpLog" to avoid confusion, please!
That is: in files, functions and documentation

hexcoder- (approved):
Please remove binary files (which could be rebuild) from the repository:
frida_mode/src/lib/lib
frida_mode/test/fuzzbench/fuzzer
frida_mode/test/libxml/xml

frida_mode/include/complog.h
vanhauser-thc: good catch, removed files and renamed files etc.

vanhauser-thc:
please have a close look at 86452cc959bd4b0d5fe6e60d0eefbc7848fe38e2 and make a few tests.
a user stumbled upon a bug which seems that resulted in no trimming being performed against targets that read from stdin.

PR #899, opened 2021-04-30
Add fixed frida-gum-devkit to repo

WorksButNotTested:


domenukk:
Can't we reference the devkit at Frida directly?

WorksButNotTested:
> Can't we reference the devkit at Frida directly?

I’ve got this PR pending for frida-gum. I’m waiting for Ole to merge it for me. This is a possible stopgap until it makes it upstream.

https://github.com/frida/frida-gum/pull/549

vanhauser-thc:
@WorksButNotTested nah we keep it like it is for the moment.
and please take not to submit compiled binaries, you did that before and this PR also has a frida_mode/test/libxml/xml ... :)

WorksButNotTested:
No worries. Sorry about the binaries, you didn’t put a review comment, so I didn’t know it was causing you problems. Apologies.

They’re not artefacts left over from the build process, but rather targets to run Frida mode against to demonstrate a fix, or check for a regression. 

What’s the best way to manage this sort of thing without cluttering things up with a load of code or dependencies just to build a test case?

Uploading the binaries isn’t ideal either, because without the source and toolchain they’re not reproducible or maintainable and of course when someone clones the repo they’re getting binaries from a random project contributor!

I just couldn’t think of a better way to support testability.

Happy to go with whatever you feel will work best.

vanhauser-thc:
why should there be binaries in the repo that all users download? IMHO that makes no sense.
basically only you use them so keep them local in your file system but out of the repository. and compile them freshly when necessary.

domenukk:
FWIW the original unicorn-afl examples are also checked in. They're only a few kb, and fixed offsets in the binary are needed for the example harnesses in that case. The newer examples use `objdump` to find these addresses, and recreate the binaries, but that is more likely to fail with odd compilers and archs.

WorksButNotTested:
Obviously the related tests won’t work without those binaries, so should I simply remove those tests all together and carry them in a branch downstream?

domenukk:
Can we have a script that generates the test binaries, and sets up the test correctly (fixing offsets etc.?)

WorksButNotTested:
The tests in question are the r2 and libxml tests from fuzzbench. I guess it depends on how they can be recreated?

WorksButNotTested:
Thinking about it, one is covered by another test case, the other I guess I could recreate a minimal reproducer?

Obviously won’t be such a representative test, but should be easier for people to see what the issue was being fixed and gets around the need for binaries and complex dependencies?

PR #900, opened 2021-05-01
instrumentation further move to C++11

devnexen:


devnexen:
saw some old code path referring to old clang version whereas docs mentions 6.0 minimum but I did not dare to touch.

vanhauser-thc:
Hi David,
thanks for the PR.
The code is as close as possible to SanitizeCoverage.cpp from llvm current code state. This change would be more proper but makes a diff more cumbersome for me to follow changes.

The llvm 6.0 minimum is new, and is basically because if see no reason anyone would use anything older. It should still most compile even with 3.3 .... but we will keep the ifdef for some time. maybe after summer we will remove them, lets see.

devnexen:
once removed it might be possible to push the enveloppe beyond C++11 as clang 6 can go even to c++2a support.

vanhauser-thc:
I see there is also a qemuafl reference update in your PR?

devnexen:
ah it was unintended

Issue #881, opened 2021-04-16
Crash in AFL after afl_custom_trim returns larger than original input


realmadsci:
The grammar mutator can occasionally return inputs that are larger than the original during a trim operation.
This is known and was partially addressed by https://github.com/AFLplusplus/AFLplusplus/pull/463.

AFL no longer fails with an error in this case, but if the size growth it large enough it later experiences memory corruption and crashes.

I believe this is the at-fault line:
https://github.com/AFLplusplus/AFLplusplus/blob/dev/src/afl-fuzz-mutators.c#L401

This is a memcpy from the (now larger) `retbuf`/`retlen` into `in_buf` which does not appear to be resized to fit the larger "trimmed" test case.

vanhauser-thc:
can you please test if b6643a8 fixes this?

vanhauser-thc:
@realmadsci we want to merge to stable and waiting for your feedback :)

realmadsci:
That particular commit ID is a readme change, so I'll look and see if I can find which one you meant...

vanhauser-thc:
oops
just try the current dev branch

vanhauser-thc:
https://github.com/AFLplusplus/AFLplusplus/commit/86a8ef168dda766d2f25f15c15c4d3ecf21d0667

realmadsci:
Looks good to me, as long as afl_realloc() has a fast-exit for cases where it is already big enough.

vanhauser-thc:
can you please test it?

realmadsci:
Will do

realmadsci:
It works!
I was using an updated grammar_mutator based on https://github.com/AFLplusplus/Grammar-Mutator/commit/2abe1e3aa3212304baadd089d4e2e4e7ec16e494 (HEAD of dev branch) for my testing.

vanhauser-thc:
thank you!

realmadsci:
After further testing, this case still crashes - it just takes longer to get there.

The https://github.com/AFLplusplus/AFLplusplus/commit/86a8ef168dda766d2f25f15c15c4d3ecf21d0667 commit fixed the immediate memcpy issue but causes a double-free to occur later.

The buffer that `in_buf` is pointing at is "owned" by `afl->testcase_buf` which was allocated by `queue_testcase_get()` inside of `fuzz_one_original()`. When this local realloc runs it occasonally invalidates that pointer but nothing updates `afl->testcase_buf` to point at the new buffer.

[...]

PR #903, opened 2021-05-06
Fix memory errors when trim causes testcase growth (#881)

realmadsci:
This PR reverts previous attempts to reallocate `in_buf` and replaces them
with an approach that leaves `in_buf` alone and just updates the on-disk file
to avoid getting into "who owns `in_buf` now?" complications.

Modify trim_case_custom to avoid writing into in_buf because
some custom mutators can cause the testcase to grow rather than
shrink.

Instead of modifying in_buf directly, we write the update out
to the disk when trimming is complete, and then the caller is
responsible for refreshing the in-memory buffer from the file.

This is still a bit sketchy because it does need to modify q->len in
order to notify the upper layers that something changed, and it could
end up telling upper layer code that the q->len is *bigger* than
the buffer (q->testcase_buf) that contains it, which is asking
for trouble down the line somewhere...

### DIFF
--- a/src/afl-fuzz-one.c
+++ b/src/afl-fuzz-one.c
@@ -508,7 +508,7 @@ u8 fuzz_one_original(afl_state_t *afl) {
 
     u32 old_len = afl->queue_cur->len;
 
-    u8 res = trim_case(afl, afl->queue_cur, &in_buf);
+    u8 res = trim_case(afl, afl->queue_cur, in_buf);
     orig_in = in_buf = queue_testcase_get(afl, afl->queue_cur);
 
     if (unlikely(res == FSRV_RUN_ERROR)) {
@@ -3007,16 +3007,16 @@ static u8 mopt_common_fuzzing(afl_state_t *afl, MOpt_globals_t MOpt_globals) {
 
     u32 old_len = afl->queue_cur->len;
 
-    u8 res = trim_case(afl, afl->queue_cur, &in_buf);
+    u8 res = trim_case(afl, afl->queue_cur, in_buf);
     orig_in = in_buf = queue_testcase_get(afl, afl->queue_cur);
 
-    if (unlikely(res == FSRV_RUN_ERROR)) {
+    if (res == FSRV_RUN_ERROR) {
 
       FATAL("Unable to execute target application");
 
     }
 
-    if (unlikely(afl->stop_soon)) {
+    if (afl->stop_soon) {
 
       ++afl->cur_skipped_paths;
       goto abandon_entry;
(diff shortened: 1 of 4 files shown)


src/afl-fuzz-one.c
```diff
@@ -3007,16 +3007,16 @@ static u8 mopt_common_fuzzing(afl_state_t *afl, MOpt_globals_t MOpt_globals) {
 
     u32 old_len = afl->queue_cur->len;
 
-    u8 res = trim_case(afl, afl->queue_cur, &in_buf);
+    u8 res = trim_case(afl, afl->queue_cur, in_buf);
     orig_in = in_buf = queue_testcase_get(afl, afl->queue_cur);
 
-    if (unlikely(res == FSRV_RUN_ERROR)) {
+    if (res == FSRV_RUN_ERROR) {
```
vanhauser-thc: could you please recover this unlikely() and the one below? thanks!

domenukk:
Thanks!
However, I'm personally not totally conviced by this solution yet: it allocates and frees the `out_buf` for each call to trim.
During afl++'s execution, we try to reuse buffers as much as possible. Maybe we can pass a scratch buf to this function, instead?

domenukk:
For example, it looks like we have `out_buf` from the outer scope laying around unused, that we could pass in as `**` for us to use at this point

realmadsci:
>  it allocates and frees the `out_buf` for each call to trim

Actually it only allocates/frees in the cases where trim was successful, so it's not quite as bad as it looks?

domenukk:
I mean, probably for most use cases, but if we can get around the allocation completely, why not do it?
That being said, can we maybe even use the last successful `retbuf` somehow by change the control flow a little?

realmadsci:
Is the concern memory management or performance?
As it stands, this implementation will free once *per trim cycle of a testcase*, and allocate once *per trim cycle of the testcase* 
(plus however many times it grows beyond the buffer size, but that would happen anyway even if it was a recycled buffer).
For the grammar mutator there are easily 10k+ executions of the target and is almost always 1k executions per trimming cycle. A single alloc/free per 1k-10k+ loops through the target is barely noticeable in my test case, so I'm not very motivated to try and optimize it further for performance reasons.

I did try to look for a way to reuse `retbuf` for this job, but I couldn't figure out how to make the control flow such that it could know when it was going to be the "last successful" run without just writing the file every time.

My current concern was getting stability first and I'm fine with (somebody) optimizing later if it makes sense. As I mentioned in the commit comment, I'm not super satisfied with it architecturally either because it modifies `q->len` without resizing `in_buf` and therefore *requires* the upper level code to reload from disk or potentially crash. But fixing that issue could get a bit more involved than anybody wants to deal with. If you can come up with a way to fix that as well as avoiding the alloc/free, I'd be all for it!

I think such a fix would look like changing the signature of trim_case_custom to somthing like this:
`size_t trim_case_custom(afl_state_t *afl, u8 *in_buf, size_t in_buf_len, u64 in_buf_cksum, u8 **out_buf, struct custom_mutator *mutator)`
Then have `trim_case_custom` do `afl_realloc(out_buf, retlen)` and return the new length of the buffer. Then the upper layers can take responsibility for rewriting the file in the queue and updating their knowledge of its length. OTOH, it would have to return 0 or something if the trim didn't change anything to avoid rewriting the file or `in_buf` even if nothing was updated?

PR #904, opened 2021-05-06
Exit on time

SecNotice:
Added code to terminate fuzzing by timeout if no new paths are found.
The absence of new paths found may indicate fuzzing problems. It can be pointless to continue fuzzing.
The patch adds the ability to end fuzzing after a specified time.

### DIFF
--- a/docs/env_variables.md
+++ b/docs/env_variables.md
@@ -284,6 +284,10 @@ checks or alter some of the more exotic semantics of the tool:
     normally indicated by the cycle counter in the UI turning green. May be
     convenient for some types of automated jobs.
 
+  - `AFL_EXIT_ON_TIME` Causes afl-fuzz to terminate if no new paths were 
+    found within a specified period of time. May be convenient for some 
+    types of automated jobs.
+
   - `AFL_EXIT_ON_SEED_ISSUES` will restore the vanilla afl-fuzz behaviour
     which does not allow crashes or timeout seeds in the initial -i corpus.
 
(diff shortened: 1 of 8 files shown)


vanhauser-thc:
There is a similar but harsher env already: AFL_EXIT_WHEN_DONE - but that exits after 100 cycles without finds, which for a large corpus can take a long time.
I am not a big fan of either to be honest.

If a fuzzer has no finds (anymore) and you terminate it then the CPU is unused. you would need to review the setup or start a different fuzzer manually anyway, so why not terminating non-performing instances by hand too?

Also this can hurt you fuzzing campaign if there is a roadblock that a laf-intel instance is slowly chomping through, and while doing so, all other instances exit because they dont get further and once the laf-intel one has a solve others can work with they are gone.

maybe in your case this makes perfectly sense you propose this though. what do you think?

SecNotice:
Thanks for the clarification of the technical nuances!

Of course, the specific reasons for the lack of growth will need to be dealt with separately.

But my patch is more designed for those cases where fuzzing is integrated into CI. Without an in-depth, obligatory search for "to the bitter end" crashes, we need to fuzz a lot of goals with satisfactory quality. One of the quality criteria we have chosen is the absence of an increase in the found paths.

My organization's compliance requires consistent criteria for all types of fuzzing. Including for JavaScript, Python, etc. Time was chosen as such a criterion.
And my case is not an isolated one, such requirements are imposed on several teams with several products in each.

vanhauser-thc:
Thank you for the clarification!
In my experience with fuzzer CI integration there is always the time criterium, however it is about maximum total runtime, not runtime until no finds. Because otherwise it is possible that a CI test run runs for infinity.
To run a fuzz instance for a specific amount of time, there is the -V parameter (and that is what is used in the CI setups I have seen or helped).
If you still think this new env var is a good idea I will merge it.

SecNotice:
Yes, of course we use the "-V" option as a general process terminator.
Sure. The cases of "slow" growth of the found paths are the most "tedious" ones.

The patch I am proposing is the result of the consensus of our micro-community on what suits us all. After AFL ++ receives this patch, only a few difficult cases will remain for "manual" work.

Thank you!

PR #906, opened 2021-05-08
Changes to bump FRIDA version and to clone FRIDA repo in to build dir…

WorksButNotTested:
FRIDA doesn't like being built from a submodule as it can't find it's `.git` directory back at the repo root. Therefore we instead (if built with `FRIDA_SOURCE=1`) clone it into the build directory and build it from there. This also means that we don't have frida as a submobule in the git repo and have to worry about updating the commit we are referencing each time, instead we get the HEAD each time which is probably what we want if we are hacking on FRIDA features.

`$(FRIDA_GUM)` is marked as `.PHONY` which means that `make` will be run inside the cloned FRIDA repo on each build (to check if any sources have been updated e.g. to allow the developer to tweak stuff in FRIDA and build it into `frida_mode`). Again only if `frida_mode` is built with `FRIDA_SOURCE=1`. I have also changed the devkit in this case to compress with `xz -0` to speed up the build, since the tarball will never be distributed size shouldn't be an issue. Note that this has been updated to clone from the official FRIDA repo rather than my fork now that the fix has been merged.

Lastly, the frida version number has been bumped to `14.2.18` which includes the fix. As before when not building with `FRIDA_SOURCE=1` the tarball is downloaded from the official FRIDA releases.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,7 +7,3 @@
 [submodule "qemu_mode/qemuafl"]
 	path = qemu_mode/qemuafl
 	url = https://github.com/AFLplusplus/qemuafl
-[submodule "frida_mode/frida"]
-	path = frida_mode/frida
-	url = https://github.com/WorksButNotTested/frida.git
-	branch = x64_stalker_fix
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
Thank you!

PR #912, opened 2021-05-10
OSX-specific improvements

virtuald:
* Fix `afl-cc` to work correctly with Xcode's clang (yes, this sucks, but it's nice if you're just testing things and don't want to bother with an entire installation of LLVM)
* By default AFL++'s shared memory usage exceeds the default configuration on OSX, so add documentation for how to change it

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -103,6 +103,41 @@ The llvm instrumentation requires a fully-operational installation of clang. The
 comes with Xcode is missing some of the essential headers and helper tools.
 See README.llvm.md for advice on how to build the compiler from scratch.
 
+MacOS X supports SYSV shared memory used by AFL's instrumentation, but the
+default settings aren't usable with AFL++. The default settings on 10.14 seem
+to be:
+
+```bash
+$ ipcs -M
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
@virtuald so when I execute `gcc` on MacOS it is actually Xcode's clang? Is there also an Xcode `clang`

virtuald:
On my 10.14 mac:

```
$ gcc --version
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/c++/4.2.1
Apple LLVM version 10.0.1 (clang-1001.0.46.4)
Target: x86_64-apple-darwin18.7.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
$ clang --version
Apple LLVM version 10.0.1 (clang-1001.0.46.4)
Target: x86_64-apple-darwin18.7.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

PR #916, opened 2021-05-11
Fixes to memory operands for complog

WorksButNotTested:
Looks like when I encountered an isntruction like `cmp [rax], rcx`, when I was updating the bitmap, I was using the values of `rax` and `rcx`, whereas `rax` should have been dereferenced and its value used when updating the shared memory.


vanhauser-thc:
yes works now, thanks!

PR #918, opened 2021-05-13
FASAN Support

WorksButNotTested:
After looking at the implementation used by libAFL and QASAN. I opted for a different approach because:

* libAFL doesn't cater for load-widening
* QASAN has it's shadow memory in the host rather than the guest.

After many different failed attempts, the resulting implementation is actually very simple. The AddressSanitizer DSO is `LD_PRELOAD`ed into the target. This takes care of hooking all the various libc (and other standard library functions) to add the necessary instrumentation (this is the same approach used when compiling an application with `-fsanitize=address`), we therefore don't need to dynamically instrument these system libraries. If the use specifies `AFL_INST_LIB` then they will want to use the other environment variables to select only the proprietary libraries for instrumentation, otherwise they may encounter the load widening issues as well as perform a load of wasted work.

By preloading the DSO, it will take care of creating the shadow memory for us and all we need to do is dynamically instrument the remaining (non standard libraries). We do this by calculating the address of any memory operands and calling back into the DSO to validate the read/write operation.

The user must pass the path of `libclang_rt.asan-x86_64.so` as the first DSO in `AFL_PRELOAD`.

### DIFF
--- a/frida_mode/test/fasan/GNUmakefile
+++ b/frida_mode/test/fasan/GNUmakefile
@@ -1,0 +1,99 @@
+PWD:=$(shell pwd)/
+ROOT:=$(shell realpath $(PWD)../../..)/
+BUILD_DIR:=$(PWD)build/
+LIBASAN:=/usr/lib/llvm-10/lib/clang/10.0.0/lib/linux/libclang_rt.asan-x86_64.so
+# LIBASAN:=/usr/lib/x86_64-linux-gnu/libasan.so.6.0.0
+
+TEST_DATA_DIR:=$(BUILD_DIR)in/
+TEST_DATA_FILE:=$(TEST_DATA_DIR)in
+FRIDA_OUT:=$(BUILD_DIR)frida-out
+
+TEST_SRC:=$(PWD)/test.c
+TEST_BIN:=$(BUILD_DIR)test
+
+CFLAGS+=-fPIC \
+		-D_GNU_SOURCE \
+		-g \
+		-fno-omit-frame-pointer \
+		-Wno-stringop-overflow \
+
+LDFLAGS+=-ldl \
+
+ifdef DEBUG
+CFLAGS+=-Werror \
+		-Wall \
+		-Wextra \
+		-Wpointer-arith
+else
+CFLAGS+=-Wno-pointer-arith
+endif
+
+.PHONY: all clean format frida-noasan frida debug run
+
+############################## ALL #############################################
+
+all: $(TEST_BIN)
+
+$(TEST_BIN): $(TEST_SRC) GNUmakefile | $(BUILD_DIR)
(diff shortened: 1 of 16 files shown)


frida_mode/test/fasan/GNUmakefile
```diff
@@ -0,0 +1,99 @@
+PWD:=$(shell pwd)/
+ROOT:=$(shell realpath $(PWD)../../..)/
+BUILD_DIR:=$(PWD)build/
+LIBASAN:=/usr/lib/llvm-10/lib/clang/10.0.0/lib/linux/libclang_rt.asan-x86_64.so
```
domenukk: The hard-coded asan path could be an issue, this should probably use `llvm-config` to get the path (or something?)

vanhauser-thc: yes that will break on so many platforms. needs llvm-config (and require LLVM_CONFIG env set otherwise)

WorksButNotTested: Sure. No problems, do you have an example I can use as a reference? No worries if not, I’ll fire up my Google.

vanhauser-thc: just copy-paste what is done in GNUmakefile.llvm to get a working llvm-config

frida_mode/test/fasan/GNUmakefile
```diff
@@ -28,6 +26,65 @@ else
...
+endif
+
+# LIBASAN:=/usr/lib/llvm-10/lib/clang/10.0.0/lib/linux/libclang_rt.asan-x86_64.so
+# LIBASAN:=/usr/lib/x86_64-linux-gnu/libasan.so.6.0.0
+
+LLVM_CONFIG ?= llvm-config
+ifeq "$(shell test -e '$(shell which $(LLVM_CONFIG))' && echo 1)" "1"
+  $(info Found llvm-config: '$(shell which $(LLVM_CONFIG))')
+else
+  $(warning Cannot find llvm-config)
```
vanhauser-thc, line 60: isnt it better to $(errror ...) here as without llvm-config it wont work anyway? or is fasan just not available but everything else will be fine?

vanhauser-thc: ah you still try to work with clang below to find the lib ... this will only work for linux though, not MacOS or Android. but well ...

WorksButNotTested: The user can pass in the direct path to the library too if they like.

WorksButNotTested: The use can set `LIBASAN` manually: e.g. `LIBASAN=/path/to/my/dso make` So all is not lost at that point.

Issue #905, opened 2021-05-07
LoadLibraryA fails in Wine mode


0x4d5a-ctf:
**Describe the bug**
I'm using an harnesss executable to dynamically load a DLL via `LoadLibraryA`. Functions exported by this DLL are used and fuzzed by my harness. Since its a 32 bit PE file I wanted to use Wine Mode `-W`.
The first dry run works, as seen in the output below. But the following calls to `LoadLibraryA` in the forked process result in not loading the DLL. `GetErrorCode` returns `ERROR_INVALID_PARAMETER` (87). 

In the first dry run, you can observe the `Loaded DLL @ 10000000` output, while successive interations result in `Unable to load ReproDll.dll. Last Error: 87`.

**To Reproduce**
1. Download the attached VS 2019 solution or prebuild binaries
2. Use the following command line to run afl-fuzz in Wine mode and observe output. 
`AFL_DEBUG=1 AFL_WINE_PATH=/usr/lib/wine/wine afl-fuzz -f /root/fuzzing/fuzz/current.pdf -t10000+ -W -i in -o out -- ./ReproExe.exe @@`

Reproduce setup is as minimal as it can be: Executable `ReproExe.exe` calls `LoadLibraryA` on the `ReproDll.dll` in the working directory. Prints the loaded address on success, prints the error otherwise.

**Expected behavior**
Calls to `LoadLibraryA` should load the library in the forked process as well.

**Screen output/Screenshots**
```
root@v2202009128997126755:~/fuzzing/fuzz# AFL_DEBUG=1 AFL_WINE_PATH=/usr/lib/wine/wine afl-fuzz -f /root/fuzzing/fuzz/current.pdf -t10000+ -W -i in -o out -- ./ReproExe.exe @@
[+] Loaded environment variable AFL_DEBUG with value 1 
[+] Loaded environment variable AFL_PATH with value /root/fuzzing/AFLplusplus
afl-fuzz++3.01a based on afl by Michal Zalewski and a big online community
[+] afl++ is maintained by Marc "van Hauser" Heuse, Heiko "hexcoder" Eißfeldt, Andrea Fioraldi and Dominik Maier
[...]
[*] Validating target binary...
[*] Attempting dry run with 'id:000000,time:0,orig:dummy.pdf'...
[*] Spinning up the fork server...
[...]

PR #933, opened 2021-05-21
doc: Wine LoadLibrary workaround

0x4d5a-ctf:
Just a little bit of documentation:
- debugging output for wine in AFL++ wine mode
- description for a workaround when using `LoadLibrary` in a forked wine instance ( #905 )

I didn't know which path the `README.wine.md` should be in, so i placed it in the `qemu_mode` folder.


vanhauser-thc:
thank you!

PR #939, opened 2021-05-26
final push to stable (really?)

vanhauser-thc:


vanhauser-thc:
IMHO it is pretty safe to merge, but would like to have one person to give a look at say its ok :)

Issue #731, opened 2021-02-05
compiler flags instead of environment variables?


virtuald:
I've got some changes locally that allow you to pass flags to the clang-lto pass instead of using environment variables. I find this particularly useful for a cmake-based build system, which doesn't allow you to configure per-target environment variables as part of the build process. If you have multiple executables or shared libraries that have different settings only configurable via environment variable, this is useful.
 
I have not investigated how one might do this for gcc.

If you think you'll accept the changes, I can clean them up and submit a PR at some point. Let me know, thanks!

vanhauser-thc:
how does that look like?
you can already pass a few things to the compiler e.g. `afl-cc --afl-lto` but no options like cmplog or laf.
I dont mind to accept a PR adding command line parameter options for env var features.
note that all such options must start with `--afl-...`

virtuald:
You can register command line options in the pass. You can pass the option to the pass via a linker option... something like `-Wl,-mllvm,option=whatever`. I guess the afl-cc wrapper could add `--afl-...` options which it could convert to the underlying LLVM option.

vanhauser-thc:
most env options are for afl-cc to load specific plugins. very few are actually for plugins, and those are mostly for corner cases.
what specific options are you thinking about?

vanhauser-thc:
@virtuald is this something you still want to follow up?

maflcko:
I think compiler flags are also needed to make `ccache` work. Right now, if compiled with/without `AFL_USE_ASAN` will just result in the binary that was compiled first? The same probably holds for other compiler options such as `AFL_LLVM_LAF_ALL`, ...

capgelka:
I found this feature also useful for the same cmake related reasons.

I am building custom mutations libraries as a part of the whole project. So they are instrumented as well as the target binaries, which cause the error
```
[*] Loading custom mutator library from '/home/capgelka/my_project/lib/libafl_custom_mutator.so'...

[-] PROGRAM ABORT : /home/capgelka/my_project/lib/libafl_custom_mutator.so: undefined symbol: __afl_area_ptr
         Location : load_custom_mutator(), src/afl-fuzz-mutators.c:151
```

It's actual for LLVM instrumentation, and everything is ok for outdated afl-gcc instrumentation.

There is an option `AFL_NOOPT` to disable instrumenting of my library, but unfortunately cmake doesn't allow to set env vars per target.

[...]

PR #942, opened 2021-05-27
Add afl_noopt flag to afl-cc as alternative way to set passtrough

capgelka:
Solves https://github.com/AFLplusplus/AFLplusplus/issues/731#issuecomment-849461963

Strange blank lines are the result of `./.custom-format.py -i src/afl-cc.c ` command

### DIFF
--- a/src/afl-cc.c
+++ b/src/afl-cc.c
@@ -31,6 +31,7 @@
 #include <strings.h>
 #include <limits.h>
 #include <assert.h>
+#include <getopt.h>
 
 #if (LLVM_MAJOR - 0 == 0)
   #undef LLVM_MAJOR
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
IMHO this is more complex than it needs to be, as there is already --afl... parsing.
I implemented it here: 8e86f7a
let me know what you think.

capgelka:
I believe the ideal solution is to have a way to configure all this options in any way (envar, flag, value from config). Which requires to match time to implement.

As for current PR you are obviously right, I just have not noticed this --afl prefixed options parsing. Can't say it's easy to understand what's going there, and this doubled check look strange  https://github.com/AFLplusplus/AFLplusplus/commit/8e86f7ad803e571bcd275d2aca597997ab0e4d2c#diff-4d5524fb8c50110c0f8f6d155c6e605f1b2627e7f09bafa1e6f4a47ff173192eR1227 )`, but it's already exists so better to use it. I have checked your patch on my projects and everything works well.

So I am closing this PR. Thanks for this patch!

vanhauser-thc:
I totally get it, the code is quite complex and does many thing in many different locations ... :)

PR #945, opened 2021-05-28
Support writing Stalker stats

WorksButNotTested:
Update to:
- Collect some stats about blocks instrumented and blocks executed.
- Make debugging of instrumented blocks a runtime rather than compile time option.
- Support for re-directing the STDOUT and STDERR of the target application.
- Update the readme.


vanhauser-thc:
> Make debugging of instrumented blocks a runtime rather than compile time option.

does this have a performance impact?

frida_mode/src/instrument/instrument_debug.c
```diff
@@ -1,26 +1,31 @@
...
 static void instrument_debug(char *format, ...) {
 
   va_list ap;
   char    buffer[4096] = {0};
+  int     len = 0;
 
   va_start(ap, format);
 
-  vsnprintf(buffer, sizeof(buffer) - 1, format, ap);
+  len = vsnprintf(buffer, sizeof(buffer) - 1, format, ap);
```
vanhauser-thc: I think the return value of vsnprintf is undefined behaviour (because of different implemenation on different platforms. so strlen() afterwards it would need to be.

WorksButNotTested: Fixed. There's another one in `stats.c` too.

WorksButNotTested:
> > Make debugging of instrumented blocks a runtime rather than compile time option.
> 
> does this have a performance impact?

This results in the addition of 3 function calls, one at the start of each compiled block, one at the end and one for each instruction within in. All 3 fail fast (they check the `fd` to determine if debug is enabled and just return if not).

But in any case, the block compilation isn't the most performance critical part. Due to the prefetch mechanism, each block is generally only compiled twice, once in the child and once in the parent (although each child can only report 8191 new blocks on x64, or 16383 on x86). So the vast majority of time the fuzzer spends should be running those compiled blocks rather than compiling more. Even if a new path is found, it is often just a different arrangement of the existing discovered blocks.

My experimentation so far has shown it to be negligible.

PR #960, opened 2021-06-04
fix ui fuzzing stage index

yuawn:
UI use incorrect index `STAGE_PYTHON` instead of `STAGE_CUSTOM_MUTATOR` for custom mutator status.


vanhauser-thc:
thank you!

PR #962, opened 2021-06-07
push to stable

vanhauser-thc:


instrumentation/afl-compiler-rt.o.c
```diff
@@ -761,7 +796,12 @@ static void __afl_start_snapshots(void) {
     if (child_stopped && was_killed) {
 
       child_stopped = 0;
-      if (waitpid(child_pid, &status, 0) < 0) _exit(1);
+      if (waitpid(child_pid, &status, 0) < 0) {
+
+        write_error("child_stopped && was_killed");
```
hexcoder-, line 803: could we add "waitpid" to the description?

instrumentation/afl-compiler-rt.o.c
```diff
@@ -629,6 +630,30 @@ static void __afl_unmap_shm(void) {
...
+      fclose(f);
+
+    }
+
+  }
+
+  fprintf(stderr, "Error(%s): %s\n", text, e);
+
+}
+
```
hexcoder-, line 658: How about we use some macro like WARNF(), which also automatically tells the origin (file and line number). Then the analysis would be easier, I think.

hexcoder- (approved):

instrumentation/afl-compiler-rt.o.c
vanhauser-thc: there is already one with that text. I dont want duplicate messages

instrumentation/afl-compiler-rt.o.c
vanhauser-thc: the WARNF location would be in the print_error() function, that is not helpful

instrumentation/afl-compiler-rt.o.c
hexcoder-: Neither do I, so just _add_ that?

vanhauser-thc: this write error specifies a specific location. so it is unique. if you prefer a different message that is unique too - go for it. this one tells me where it goes wrong.

instrumentation/afl-compiler-rt.o.c
hexcoder-: Think macro instead of function like this:
```
--- a/instrumentation/afl-compiler-rt.o.c
+++ b/instrumentation/afl-compiler-rt.o.c
@@ -630,7 +630,9 @@ static void __afl_unmap_shm(void) {
 
 }
 
-void write_error(char *text) {
+#define write_error(text) write_error_with_location(text, __FILE__, __LINE__)
+
+void write_error_with_location(char *text, char* filename, int linenumber) {
 
   u8 *  o = getenv("__AFL_OUT_DIR");
   char *e = strerror(errno);
@@ -643,14 +645,14 @@ void write_error(char *text) {
 
     if (f) {
 
-      fprintf(f, "Error(%s): %s\n", text, e);
+      fprintf(f, "File %s, line %d: Error(%s): %s\n", filename, linenumber, text, e);
       fclose(f);
 
     }
 
   }
 
-  fprintf(stderr, "Error(%s): %s\n", text, e);
+  fprintf(stderr, "File %s, line %d: Error(%s): %s\n", filename, linenumber, text, e);
 
 }
 
@@ -2079,3 +2081,5 @@ void __afl_coverage_interesting(u8 val, u32 id) {
 
 }
 
+#undef write_error

```

hexcoder-: Just changed it so.

instrumentation/afl-compiler-rt.o.c
hexcoder-: ok, fine with me.

PR #964, opened 2021-06-07
split-comparison llvm pass refactor for smaller compilation times (and a small bug fix)

f0rki:
I have a large target that I compile with LTO and I noticed that the comparison splitting was very slow. So painfully slow that I decided to investigate. I believe the reason was that the pass unnecessarily loops over the whole llvm module multiple times, which takes quite some time when using full LTO on a somewhat larger project. So I started to refactor the code a bit s.t., one pass over the module is enough to simplify and split all the integer comparisons, which resulted in acceptable compilation times for me. I did not touch the floating point code. 

Also the `AFL_LLVM_LAF_SPLIT_COMPARES_BITW` was not working as I expected. Previously, due to the fallthrough in switch-case, all comparisons were split to 8-bit, which might not be desirable e.g., 16 or 32 bit might be enough. With this PR now all comparison are split until they are smaller or equal to the `target_bitwidth` member of the Pass C++ class, which is controlled through the `AFL_LLVM_LAF_SPLIT_COMPARES_BITW` environment variable.

Small Q: did the clang-format style change or do I have a configuration error on my side? do I need to reformat the code?

### DIFF
--- a/instrumentation/split-compares-pass.so.cc
+++ b/instrumentation/split-compares-pass.so.cc
@@ -137,8 +137,8 @@ class SplitComparesTransform : public ModulePass {
 
 char SplitComparesTransform::ID = 0;
 
-/* This function splits FCMP instructions with xGE or xLE predicates into two
- * FCMP instructions with predicate xGT or xLT and EQ */
+/// This function splits FCMP instructions with xGE or xLE predicates into two
+/// FCMP instructions with predicate xGT or xLT and EQ
 bool SplitComparesTransform::simplifyFPCompares(Module &M) {
   LLVMContext &              C = M.getContext();
   std::vector<Instruction *> fcomps;
@@ -254,22 +254,24 @@ bool SplitComparesTransform::simplifyFPCompares(Module &M) {
 
   return true;
 }
-/* This function splits ICMP instructions with xGE or xLE predicates into two
- * ICMP instructions with predicate xGT or xLT and EQ */
+
+/// This function splits ICMP instructions with xGE or xLE predicates into two
+/// ICMP instructions with predicate xGT or xLT and EQ
 bool SplitComparesTransform::simplifyOrEqualsCompare(
     CmpInst *IcmpInst, Module &M, std::vector<CmpInst *> &worklist) {
   LLVMContext &C = M.getContext();
   IntegerType *Int1Ty = IntegerType::getInt1Ty(C);
 
+  /* find out what the new predicate is going to be */
+  auto cmp_inst = dyn_cast<CmpInst>(IcmpInst);
+  if (!cmp_inst) { return false; }
+  
   BasicBlock *bb = IcmpInst->getParent();
 
   auto op0 = IcmpInst->getOperand(0);
   auto op1 = IcmpInst->getOperand(1);
 
-  /* find out what the new predicate is going to be */
-  auto cmp_inst = dyn_cast<CmpInst>(IcmpInst);
-  if (!cmp_inst) { return false; }
-  auto               pred = cmp_inst->getPredicate();
+  CmpInst::Predicate pred = cmp_inst->getPredicate();
   CmpInst::Predicate new_pred;
 
   switch (pred) {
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
Hi, thank you for the patch. looks good to me, but will have a closer look the next days.
The formatting is likely that you have a different clang format version running, but we can fix this after merging.

hexcoder-:
Thanks from me too!
These are a lot of changes (a bit more complicated due to reformatting).
I will need some time for a review because I want to write some test cases, in order to compare the instrumentation with the previous state.

hexcoder-:
@f0rki: I have found an error in your PR with this test case:

```
  volatile short a, b;
  a = -1;
  b = 1;
  assert((a < b));
  assert((a <= b));
  assert(!(a > b));
  assert(!(a >= b));
  assert((a != b));
  assert(!(a == b));
```

hexcoder-:
Here is my test. Compile with -DINT_TYPE="short" to reproduce.
[test-int_cases.zip](https://github.com/AFLplusplus/AFLplusplus/files/6618531/test-int_cases.zip)

hexcoder- (changes requested):
Test cases should work.

f0rki:
@hexcoder- nice testcase. the int testcase works now (for every combination of int type and target BITW)

[EDIT] does it make sense to integrate that integer testcase into `make test` ? (if yes I can add it to this PR if you want)

hexcoder-:
@f0rki:  Nice work, now my test cases all succeed (see new files `test/test-{u}int_cases.c`).
These tests are not exhaustive, of course. But I would not mind having them added them to `make tests` (PR welcome). For a practical test, we could compile and run ffmpeg, I propose.

But before waving this PR through, I would like to understand why the number of instrumentations from the test cases decreased (compared to the old splitting). Do you know why?

Type | Old | New
------|------|-------
Char | 960 | 864
Short | 1164 | 1068
Int | 1810 | 1510
Long | 3646 | 2394
UChar | 26 | 26
UShort| 200 | 200
UInt| 652 | 544
ULong| 1924 | 1252

hexcoder-:
@f0rki more analysis done...
In this program (test-compare.c)
```
int main() {

  volatile unsigned int a, b;
  a = 1234;
  b = 4321;
  if (a <= b) {
    return 1;
  } else {
    return 2;
  }
}
```
your code does not break all comparisons down to 8 bit size.

Here is the LLVM IR (generated with `AFL_LLVM_LAF_ALL=1 afl-clang-fast -S -emit-llvm test-compare.c`) with the old splitting, reduced to broken down comparisons:
```
...
%23 = icmp ugt i8 %20, %22
...
%34 = icmp ult i8 %20, %22
...
%55 = icmp ugt i8 %53, %54
...
%72 = icmp ult i8 %69, %71
...
%83 = icmp ugt i8 %69, %71
...
%104 = icmp ult i8 %102, %103
...
%123 = icmp ugt i8 %120, %122
...
%134 = icmp ult i8 %120, %122
...
%155 = icmp ugt i8 %153, %154

```
and here is the list of icmps from your code:
```
%23 = icmp ugt i8 %20, %22
...
%34 = icmp ult i8 %20, %22
...
%55 = icmp ugt i8 %53, %54
...
%68 = icmp ult i16 %16, %18 <---- 16 bit compare
...
%85 = icmp ugt i8 %82, %84
...
%96 = icmp ult i8 %82, %84
...
%96 = icmp ult i8 %82, %84
...
%117 = icmp ugt i8 %115, %116
```
So, the refactoring introduced another error. Please fix (and test/compare).

f0rki:
OK. I rebased, updated the `test-llvm.sh` script to test for miscompilations due to the splitting pass and fixed the issue discovered by @hexcoder-: I simply overlooked that for the ULT/UGT case the splitting code generates three new comparisons instead of two new comparisons in the EQ/NE case. Maybe this was also the reason for the mismatch in number of instrumentations @hexcoder- ?

hexcoder- (approved):
LGTM, the number of instrumentations match with the older state.

hexcoder-:
@vanhauser-thc Do you want to take a look before merging?

vanhauser-thc:
LGTM and great work!
and I hope you got to appreciate how much effort it is to write effective llvm modules :)

f0rki:
thanks for maintaining such a great project! :)

PR #965, opened 2021-06-07
FRIDA AARCH64 support

WorksButNotTested:


vanhauser-thc:
can this be merged?

WorksButNotTested:
I still need to look at the persistent ret functionality for aarch64 and try to understand why the performance with the hook is slower. 

There shouldn’t be any regressions, so it’s an improvement on the current version, but happy if you’d rather wait until the loose ends are tidied up?

vanhauser-thc:
ah yes, better to have it all in this PR IMHO

WorksButNotTested:
Issues with `AFL_FRIDA_PERSISTENT_RET` fixed. Also fixed weird performance of `AFL_FRIDA_PERSISTENT_HOOK` being slower.

Now ready to merge.

vanhauser-thc:
> Also fixed weird performance of `AFL_FRIDA_PERSISTENT_HOOK` being slower.

what was the issue?

WorksButNotTested:
> > Also fixed weird performance of `AFL_FRIDA_PERSISTENT_HOOK` being slower.
> 
> what was the issue?

I had commented out this line...

https://github.com/AFLplusplus/AFLplusplus/blob/48c878a76ddec2c133fd5708b185b2ac27740084/frida_mode/src/persistent/persistent_arm64.c#L431

Basically it meant that the persistent count was effectively hard coded to 1. Performance for both persistent and persistent hook mode were severely impacted. Oddly the latter more so.

PR #966, opened 2021-06-09
Instrumentation Suppression

WorksButNotTested:
**DO NOT MERGE INTO DEV**

This PR includes changes for x64 which support the suppression of the generation of coverage information when either:

- There is a `call`, `jmp` `jrcxz` etc where the target is given as an immediate
- There is a conditional branch `jae`, `ja`, `jbe`, `jb` .... where the branch is not taken.

The theory is that in the case of the former, the branch was always going to happen, and always going to the same location. Hence given that execution reached the block of the caller, the generated coverage information was inevitable and hence superfluous.

In the case of the latter, whilst the conditional branch is considered the end of a basic block, the branch was not actually taken and hence the absence of the coverage information from the taken branch is sufficiently distinguishing.

The intention of this approach is to remove unnecessary coverage information from polluting the coverage map and potentially creating collisions. However, this comes at the expense of a performance overhead (albeit mostly incurred during block compilation rather than execution).

The code which updates the coverage map is in the destination block (since calculating the target address at the source block is likely expensive and complicated). However, at the destination block we don't know how the block was reached (e.g. whether by a `call imm` or `call reg`, each block could be reached by multiple different instructions) and hence whether to suppress the coverage generation.

Therefore, we instrument the instruction which causes the block transition to write a cookie into the stack (far beyond the stack pointer, since stalker internally is at liberty to use the stack beyond the red-zone). This cookie is then detected in the coverage generation code.

### DIFF
--- a/frida_mode/include/instrument.h
+++ b/frida_mode/include/instrument.h
@@ -8,6 +8,7 @@
 extern __thread uint64_t previous_pc;
 extern uint8_t *         __afl_area_ptr;
 extern uint32_t          __afl_map_size;
+extern gboolean          instrument_suppression;
 
 void instrument_init(void);
 
@@ -17,6 +18,9 @@ GumStalkerTransformer *instrument_get_transformer(void);
 gboolean instrument_is_coverage_optimize_supported(void);
 
 void instrument_coverage_optimize(const cs_insn *   instr,
+                                  GumStalkerOutput *output);
+
+void instrument_coverage_suppress(const cs_insn *   instr,
                                   GumStalkerOutput *output);
 
 void     instrument_debug_init(void);
--- a/frida_mode/src/instrument/instrument_x64.c
+++ b/frida_mode/src/instrument/instrument_x64.c
@@ -1,8 +1,10 @@
 #include "frida-gum.h"
 
 #include "config.h"
+#include "debug.h"
 
 #include "instrument.h"
+#include "ranges.h"
 
 #if defined(__x86_64__)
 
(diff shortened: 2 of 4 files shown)


vanhauser-thc:
> This PR includes changes for x64 which support the suppression of the generation of coverage information when either:
>     * There is a `call`, `jmp` `jrcxz` etc where the target is given as an immediate

yes this does not need coverage

>     * There is a conditional branch `jae`, `ja`, `jbe`, `jb` .... where the branch is not taken.

this however does. e.g. lets say you have:

```
       A
    /    \
  B        C
  |       /  \
   \    /     E
     D
```
and C->D would be the jump not taken, then it would look the same to afl-fuzz if A -> B -> D or A -> C -> D was taken.

WorksButNotTested:
> > This PR includes changes for x64 which support the suppression of the generation of coverage information when either:
> > * There is a `call`, `jmp` `jrcxz` etc where the target is given as an immediate
> 
> yes this does not need coverage
> 
> > ```
> > * There is a conditional branch `jae`, `ja`, `jbe`, `jb` .... where the branch is not taken.
> > ```
> 
> this however does. e.g. lets say you have:
> 
> ```
>        A
>     /    \
>   B        C
>   |       /  \
>    \    /     E
>      D
> ```
> 
> and C->D would be the jump not taken, then it would look the same to afl-fuzz if A -> B -> D or A -> C -> D was taken.

Wouldn't the first have the edges A/B and B/D. Whereas the second would have A/C? Or have I misunderstood?

vanhauser-thc:
B->D has a simple jmp (no condition) so it is not instrumented, so only A->D is visible. and C wont be instrumented if it goes to D as default - that is how I understood how you did thing. but it doesnt matter. if you not instrument conditional jumps (if default is taken or not) then you fuck up the coverage :)

WorksButNotTested:
> B->D has a simple jmp (no condition) so it is not instrumented, so only A->D is visible. and C wont be instrumented if it goes to D as default - that is how I understood how you did thing. but it doesnt matter. if you not instrument conditional jumps (if default is taken or not) then you fuck up the coverage :)

Given you can branch from A->B or A->C, then either:

1. The instruction at the end of A is a jmp without an immediate (e.g. `jmp rax`)
2. The instruction at the end of A is a conditional jump (with either B or C the block immediately following it).

In the case of the case of 1, then instrumentation is emitted on both edges, in case 2, then instrumentation is emitted on only one edge. If B->D is a `jmp imm` then no instrumentation occurs on the edge.

Let's assume that we have the harder (second) case.
If we assume the convention that the branch to the left is the non-conditional one (e.g. branch not taken), and that all edges with only one choice are immediate branches. 

Then for the path A->B->D we therefore emit nothing. For the path A->C->D we emit A->C and for the path A->C->E we emit A->C->E.

If we assume that the branch to the right is the non-conditional one.
A->B->D emits A->B, A->C->D emits C->D and A->C->E emits nothing.

Does this work ok?

vanhauser-thc:
no. you cannot omit instrumentation on a block that has 2+ outgoing edges or 2+ incoming edges.

andreafioraldi:
Yes the second option is not going to work

andreafioraldi:
Simply consider when the not taken block is a leaf in the CFG with just one parent. Simply you will never notice when you reach the not taken block, that is bad. But the approach does not work in general, otherwise we could have implemented it for LLVM detecting leafs at compile time but there are other counter-examples that I'm too lazy to think about now.

andreafioraldi:
You are not the first to think about that btw, does not work trust me.

vanhauser-thc:
> You are not the first to think about that btw, does not work trust me.

exactly :)

WorksButNotTested:
Certainly not saying you are wrong. I just don't quite get it yet, sorry!

In the example of a leaf block in the CFG with a single parent, will you not know that you have reached the not taken branch as a result of detecting you have reached the parent and the subsequent lack of any other edges being emitted? e.g considering the following...

```
parent:
 <some instructions>
 jcc some_other_block
 leaf:
<some more instructions>

If you reach the parent, and then don't take the branch, then you *MUST* run the code in the leaf. Effectively the code after the not-taken branch is considered part of the parent block. I guess in the case that `<some instructions>` caused a crash, you'd be unable to distinguish whether it happened before or after the branch?

vanhauser-thc:
@WorksButNotTested 
you will be able to come up with an example where what you are doing might work, however we will always come up with examples where it wont work :)

> you cannot omit instrumentation on a block that has 2+ outgoing edges or 2+ incoming edges.
this holds (for you doing classic instrumentation. for source instrumentation there are different solutions).

WorksButNotTested:
> @WorksButNotTested
> you will be able to come up with an example where what you are doing might work, however we will always come up with examples where it wont work :)
> 
> > you cannot omit instrumentation on a block that has 2+ outgoing edges or 2+ incoming edges.
> > this holds (for you doing classic instrumentation. for source instrumentation there are different solutions).

Obviously it needs to work all of the time, not just partially. I'm not saying I'm not wrong. Just trying to understand why I'm wrong. If you've got an example, then it might help me figure it out.

vanhauser-thc:
make an example (pseudo C code or graph) with shows what is instrumented and what not so I actually understand where you want to put the instrumentation and where not. then I try to come up with an example where this will not work

vanhauser-thc:
@WorksButNotTested this can be closed, right?

WorksButNotTested:
I want to come back to the discussion when I’ve wrapped up the rest of what I’m working on, so I fully understand what’s possible.

But that doesn’t need the PR to be open to look back. I’ll keep my local branch in case there’s fragments worth saving, and I can make another PR when we’ve discussed and I’ve fixed things.

PR #969, opened 2021-06-09
Add debug output to alert user to calibration progress/issues

virtuald:


src/afl-fuzz-run.c
```diff
@@ -402,6 +404,8 @@ u8 calibrate_case(afl_state_t *afl, struct queue_entry *q, u8 *use_mem,
 
         }
 
+        if (!var_detected) WARNF("instability detected during calibration\n");
```
virtuald: This is particularly useful if you expect a testcase to be 100% deterministic and it isn't.

vanhauser-thc: you need to check too if he user us NOT using the UI. because otherwise this will f*ck up the UI view.

src/afl-fuzz-run.c
```diff
@@ -355,6 +355,8 @@ u8 calibrate_case(afl_state_t *afl, struct queue_entry *q, u8 *use_mem,
 
   for (afl->stage_cur = 0; afl->stage_cur < afl->stage_max; ++afl->stage_cur) {
 
+    if (afl->debug) DEBUGF("calibration stage %d/%d\n", afl->stage_cur+1, afl->stage_max);
```
vanhauser-thc: please put `if (unlikely(afl->debug)) ...`

src/afl-fuzz-run.c
virtuald: It appears that `from_queue` is only true when not in the UI. Is there a different way to detect that?

vanhauser-thc: `if (afl->afl_env.afl_no_ui && !var_detected)  { WARNF() ; }`

src/afl-fuzz-run.c
```diff
@@ -402,6 +404,14 @@ u8 calibrate_case(afl_state_t *afl, struct queue_entry *q, u8 *use_mem,
 
         }
 
+        if (unlikely(!var_detected)) {
+          if (from_queue) {
```
vanhauser-thc: this still needs afl->afl_env.afl_no_ui instead so the PR can be merged :)
thanks

virtuald: Your original statement:

> you need to check too if he user us NOT using the UI.

I don't think `afl_no_ui` will tell me if the UI is currently active or not, it's only set if the `AFL_NO_UI` environment variable is set?

virtuald: And for example, in `afl-fuzz-init.c` shortly after `calibrate_case` is called, it calls WARNF if a timeout occurs. Why doesn't that check for afl_no_ui if that's the correct solution?

vanhauser-thc: I dont understand what you mean by "if the UI is active or not". there are two output puts. text lines, or the graphical UI. in graphical UI you cant print these messages. this will fuck up the screen and would only be visible for 1/2 a second. in text mode output you can print this message. and that is what this is checking for.

vanhauser-thc: > And for example, in `afl-fuzz-init.c` shortly after `calibrate_case` is called, it calls WARNF if a timeout occurs. Why doesn't that check for afl_no_ui if that's the correct solution?

because there the UI is not up yet.

virtuald: I agree with you that this message shouldn't be printed when the UI is currently showing. However, afl_no_ui is only set when you *never* want the UI to be shown, and is only set via environment variable.

My current solution of relying on `from_queue` to be true works in the sense that it is only true when called from afl-fuzz-init. I agree this is brittle, but I don't see any other variable that AFL sets when it transitions from initialization to UI.

vanhauser-thc: ah now I understand why you did it like that.
but also then, `if (from_queue || afl_no_ui) ...` is even more what you want?

virtuald: I don't think there's a need to check afl_no_ui here. `AFL_NO_UI` seems to be about disabling the UI, not to disable output completely. As stated in env_variables.md:

>   - Setting `AFL_NO_UI` inhibits the UI altogether, and just periodically prints
    some basic stats. This behavior is also automatically triggered when the
    output from afl-fuzz is redirected to a file or to a pipe.

`WARNF` calls `SAYF` which selectively disables colors via `colorfilter` which checks the `AFL_NO_UI` variable directly.

vanhauser-thc: I think you dont get what I mean. if the UI is disabled, then this message can be safely written *even during a run, not only during the initial calibration*

vanhauser-thc: if you are only interested in this information during the init process then its fine, although if you are interested in this, why not also during a run? (if no UI is used)

virtuald: Oh, I see! Got it.

PR #970, opened 2021-06-09
FRIDA - Remove need for AFL_FRIDA_PERSISTENT_RETADDR_OFFSET

WorksButNotTested:
This change does away with the need for setting `AFL_FRIDA_PERSISTENT_RETADDR_OFFSET`. Just before we branch to the target persistent loop function, we stash the return address into a global. When we subsequently reach the persistent epilogue at `AFL_FRIDA_PERSISTENT_RETADDR` we retrieve the return address from there rather than from the stack. (Note since we call the target persistent loop function, if we don't set `AFL_FRIDA_PERSISTENT_RETADDR`, we end up back at the same place anyway when the target function returns).

From here, the persistent loop continue running using the stack as it was left by the target application (in fact we store any necessary data beyond any redzone) until all of the registers are restored from the `saved_regs` global.


vanhauser-thc:
ty!

PR #971, opened 2021-06-10
Frida complog fix

WorksButNotTested:


frida_mode/src/cmplog/cmplog.c
```diff
@@ -27,6 +33,39 @@ static gint cmplog_sort(gconstpointer a, gconstpointer b) {
...
+
+  if (unlink(fname) < 0) {
+
+    FATAL("Failed to unlink temp file: %s (%d), errno: %d", fname, fd, errno);
+
+  }
+
+  if (ftruncate(fd, 0) < 0) {
+
+    FATAL("Failed to unlink temp file: %s (%d), errno: %d", fname, fd, errno);
```
vanhauser-thc: that would be truncate not unlink :-)

frida_mode/src/cmplog/cmplog.c
```diff
@@ -81,6 +130,27 @@ gboolean cmplog_is_readable(guint64 addr, size_t size) {
...
+  /*
+   * Our address map can change (e.g. stack growth), use write as a fallback to
+   * validate our address.
+   */
+  ssize_t written = syscall(__NR_write, fd_null, (void *)addr, size);
+
+  /*
+   * Seek back to the start, we don't want our temp file to continue growing!
+   */
+  if (ftruncate(fd_null, 0) < 0) {
```
vanhauser-thc: you want to truncate for every write?? rather do that once per fork() or persistent run

frida_mode/src/cmplog/cmplog.c
WorksButNotTested: Fixed.

frida_mode/src/cmplog/cmplog.c
WorksButNotTested: I went with tracking the file size and truncating when it reached a given limit. It kept the interface between complog and the rest of the code more minimal and should still have the desired effect.

vanhauser-thc:
LGTM
a few ideas:
  * increase the allowed size of the temp file. 1MB should be fine IMHO
  * use likely()/unlikely() in your if() for a bit more performance (everywhere in code that runs during fuzzing)
  * compile with perf and perform an analysis what is called very often, then use attribute(hot) for the most used functions, also gives quite a speed gain.

PR #972, opened 2021-06-10
Unstable

WorksButNotTested:
This PR moves the `var_bytes` array used by `afl++` to track unstable edges into a shared memory segment so that it can be accessed by the target application (This uses the same mechanics as used for in-memory test-cases). This should then be accessible to other in-process binary fuzzers such as QEMU as well as FRIDA.

Once this data is in memory, when passed the flag `AFL_FRIDA_INST_UNSTABLE`, the FRIDA instrumentation will automatically enable `AFL_FRIDA_INST_NO_OPTIMIZE` (which allows the instrumentation to be carried out in C code rather the inline generated assembly). This code then checks the given edge index in the bitmap in shared memory to determine whether the edge is unstable as well as storing the actual previous RIP when generating instrumentation. If it is marked as unstable, then a log message is generated with the current RIP as well as the previous RIP to allow the user to diagnose the unstable edge.

A similar shared memory region of the same size is also created which is used to mark unstable edges which have already been reported. This prevents us from spamming the logs.


src/afl-sharedmem.c
```diff
@@ -239,6 +264,37 @@ u8 *afl_shm_init(sharedmem_t *shm, size_t map_size,
 
   }
 
+  snprintf(shm->var_bytes_g_shm_file_path, L_tmpnam, "/afl_var_bytes_%d_%ld",
+           getpid(), random());
+
+  /* create the shared memory segment as if it was a file */
+  shm->var_bytes_g_shm_fd =
+      shm_open(shm->var_bytes_g_shm_file_path, O_CREAT | O_RDWR | O_EXCL,
```
domenukk: Just a nitpick: the original idea behind the `afl_shm` functions was to create one `sharedmem_t` per shared map.
That way we don't have to call `shm_open`(or whatever the names are on other operating systems), but it "just works".

WorksButNotTested:
Following discussion with @vanhauser-thc, I have removed the changes to the fuzzer itself and instead focussed on improving the logging provided by the already supported trace functionality to include edge IDs. This can then be cross matched by the user using the `fuzzer_stats` to identify instability.

frida_mode/src/instrument/instrument.c
```diff
@@ -164,18 +189,28 @@ void instrument_init(void) {
...
 
   OKF("Instrumentation - optimize [%c]", optimize ? 'X' : ' ');
   OKF("Instrumentation - tracing [%c]", tracing ? 'X' : ' ');
+  OKF("Instrumentation - unique [%c]", unique ? 'X' : ' ');
 
   if (tracing && optimize) {
 
-    FATAL("AFL_FRIDA_INST_OPTIMIZE and AFL_FRIDA_INST_TRACE are incompatible");
+    FATAL("AFL_FRIDA_INST_TRACE requires AFL_FRIDA_INST_NO_OPTIMIZE");
+
```
vanhauser-thc, line 203: why dont you just hardcode to not optimize if trace/unique are set?

PR #975, opened 2021-06-11
Dynamic map2

WorksButNotTested:
**DO NOT MERGE**

This PR is an experiment to test whether dynamically changing the map size based upon the contents of the buckets can help a target application reduce the number of hash collisions. The function `reconsider_map_size` contains the logic for deciding when to increase the map size (which may or may not be very good).

- All shared memory maps are now 256Mb in size (an adjustable constant which might not be amazing on 32-bit platforms), but are marked as `MAP_NORESERVE` or similar.
- The shared memory code has been refactored to greatly reduce duplication and complexity.
- The use of large non-reserved maps means that they don't have to be re-allocated if the size is changed, so this code has been removed.
- The cmplog structure was organised so that there were `n` headers, followed by `n` operands. This obviously won't allow us to change the value of `n` without changing the structure definition. We therefore re-organise this data so that each header is followed immediately by its operands.
- The code for mapping the shared memory regions in afl-compiler-rt.o.c has similarly been refactored to reduce duplication.
- An addiitonal shared memory segment has been created which contains only the size of the mapping (this size dictates the number of buckets for edges as well as for cmplog and the various supporting structures and temporary values which they use during their calculations). Accodingly the cmplog structure is modified to use a zero length array and this map size is used to calculate its actual size instead of `sizeof`.
- The code in FRIDA which is updating the cmplog map and generating edges, similarly must read the dynamic size from the shared memory region rather than using a constant. This means that the which the shift operations of the edge id calculationcan still be completed at compile (instrumentation) time, due to prefetching the masking of this value must be carried out at runtime. Or otherwise we must discard our prefetched blocks (perhaps the map size will stabilize as we stop finding new paths, so the performance trade-off will be worthwhile), but wholesale invalidation of prefetched blocks isn't supported in FRIDA for now.
- Lastly some logic has been added to adjust the size of the map every `n` runs (where `n` is 10 for testing and probably should be much larger) and update the necessary variables given some (possibly shoddy) reasoning about the contents of the map.
- Note that on the tests I have run, the conditional logic doesn't result in the size of the map being increased, (maybe 64kb is enough, who knew), but when hard coding the resize, it seems to work successfully.

And that's all there is to it. The size of the map can be changed dynamically based upon how densely populated it is.

Note that this branch is called `dynamic-map`, the original branch `dynamic-map` attempted to add the size of the map to the beginning of the shared memory itself. But this proved to be a horror show with code changes required everywhere, and was duly adandoned.

### DIFF
--- a/frida_mode/GNUmakefile
+++ b/frida_mode/GNUmakefile
@@ -20,6 +20,7 @@ RT_CFLAGS:=-Wno-unused-parameter \
 		   -Wno-unused-function \
 		   -Wno-unused-result \
 		   -Wno-int-to-pointer-cast \
+		   -Wno-pointer-sign \
 
 LDFLAGS+=-shared \
 		 -lpthread \
--- a/frida_mode/include/instrument.h
+++ b/frida_mode/include/instrument.h
@@ -4,10 +4,11 @@
 #include "frida-gum.h"
 
 #include "config.h"
+#include "sharedmem.h"
 
 extern __thread uint64_t previous_pc;
 extern uint8_t *         __afl_area_ptr;
-extern uint32_t          __afl_map_size;
+extern struct map_size * __afl_map_size_addr;
 
 void instrument_init(void);
 
--- a/frida_mode/src/cmplog/cmplog_arm64.c
+++ b/frida_mode/src/cmplog/cmplog_arm64.c
@@ -114,6 +114,7 @@ static void cmplog_call_callout(GumCpuContext *context, gpointer user_data) {
   uintptr_t k = address;
 
   k = (k >> 4) ^ (k << 8);
+  // TODO: DONT HARD CODE THE MAP SIZE
   k &= CMP_MAP_W - 1;
 
   __afl_cmp_map->headers[k].type = CMP_TYPE_RTN;
@@ -191,6 +192,7 @@ static void cmplog_handle_cmp_sub(GumCpuContext *context, gsize operand1,
   register uintptr_t k = (uintptr_t)address;
 
   k = (k >> 4) ^ (k << 8);
+  // TODO: DONT HARD CODE THE MAP SIZE
   k &= CMP_MAP_W - 1;
 
   __afl_cmp_map->headers[k].type = CMP_TYPE_INS;
--- a/frida_mode/src/cmplog/cmplog_x64.c
+++ b/frida_mode/src/cmplog/cmplog_x64.c
@@ -5,6 +5,7 @@
 
 #include "ctx.h"
 #include "frida_cmplog.h"
+#include "instrument.h"
 #include "util.h"
 
 #if defined(__x86_64__)
@@ -109,7 +110,7 @@ static void cmplog_call_callout(GumCpuContext *context, gpointer user_data) {
   uintptr_t k = address;
 
   k = (k >> 4) ^ (k << 8);
-  k &= CMP_MAP_W - 1;
+  k &= __afl_map_size_addr->size - 1;
 
   __afl_cmp_map->entries[k].type = CMP_TYPE_RTN;
 
@@ -175,7 +176,7 @@ static void cmplog_handle_cmp_sub(GumCpuContext *context, gsize operand1,
   register uintptr_t k = (uintptr_t)address;
 
   k = (k >> 4) ^ (k << 8);
-  k &= CMP_MAP_W - 1;
+  k &= __afl_map_size_addr->size - 1;
 
   __afl_cmp_map->entries[k].type = CMP_TYPE_INS;
 
--- a/frida_mode/src/cmplog/cmplog_x64.c
+++ b/frida_mode/src/cmplog/cmplog_x64.c
@@ -111,18 +111,16 @@ static void cmplog_call_callout(GumCpuContext *context, gpointer user_data) {
   k = (k >> 4) ^ (k << 8);
   k &= CMP_MAP_W - 1;
 
-  __afl_cmp_map->headers[k].type = CMP_TYPE_RTN;
+  __afl_cmp_map->entries[k].type = CMP_TYPE_RTN;
 
-  u32 hits = __afl_cmp_map->headers[k].hits;
-  __afl_cmp_map->headers[k].hits = hits + 1;
+  u32 hits = __afl_cmp_map->entries[k].hits;
+  __afl_cmp_map->entries[k].hits = hits + 1;
 
-  __afl_cmp_map->headers[k].shape = 31;
+  __afl_cmp_map->entries[k].shape = 31;
 
   hits &= CMP_MAP_RTN_H - 1;
-  gum_memcpy(((struct cmpfn_operands *)__afl_cmp_map->log[k])[hits].v0, ptr1,
-             32);
-  gum_memcpy(((struct cmpfn_operands *)__afl_cmp_map->log[k])[hits].v1, ptr2,
-             32);
+  gum_memcpy(__afl_cmp_map->entries[k].log[hits].fn.v0, ptr1, 32);
+  gum_memcpy(__afl_cmp_map->entries[k].log[hits].fn.v1, ptr2, 32);
 
 }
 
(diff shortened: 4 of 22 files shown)


vanhauser-thc:
I have never seen a map larger than 9MB and that was for open office excel. so 256 is absolutely overkill.
also a better approach would be to have a tool that tries to identify the edges, plus the libs and their edges, and report that back, and using that plus a percentage as the size for the map. basically a script for Ghidra doing that.

WorksButNotTested:
I did just pick a number out of the air. I guess if you do static analysis, then you’ll end up with a theoretical maximum (assuming you reach every possible block of code in every binary).

Would it be worthwhile having a dynamic tool like afl-showmap that you run against the corpus and it can tell you how many edges it encounters? Then you can perform an initial fuzzing run with a set map size, then use that to set ‘AFL_MAP_SIZE’ for the next run?

Does having a large map size have a significant impact on performance? 10Mb memory isn’t a lot, so can you just run with a large map size to start with?

vanhauser-thc:
> I did just pick a number out of the air. I guess if you do static analysis, then you’ll end up with a theoretical maximum (assuming you reach every possible block of code in every binary).

but you need the theoretical maximum because every edge has to have a unique id

> Would it be worthwhile having a dynamic tool like afl-showmap that you run against the corpus and it can tell you how many edges it encounters? Then you can perform an initial fuzzing run with a set map size, then use that to set ‘AFL_MAP_SIZE’ for the next run?

afl-showmap already does that. 
but what you propose that would be not nice user experience.
constantly rerunning to have a "good" map size 

> Does having a large map size have a significant impact on performance? 10Mb memory isn’t a lot, so can you just run with a large map size to start with?

yes it has a huge impact. but the maximum edge ID could be written to the first bytes of the map

WorksButNotTested:
You will want the theoretical maximum of the the number of edges reachable from the entry-point though, and not the maximum for the whole binary right? Also isn't tracing cross-references from static analysis particularly difficult?

I guess how good the user-experience is depends on how frequently you would have to measure and then re-adjust the map-size manually?

Presumably the performance overhead is from the back-end processing of the map to interpret the coverage rather than the overhead from forking with a larger amount of pages?

I'm still leaning toward a dynamically re-sized map. The optimal map size at any given point is based on the number of edges encountered so far. e.g. if you have discovered `n` edges, you want to discover the `n+1`th edge. In order to do that, you need to ensure (probabilistically) that a single new edge won't collide with another edge which you have already found (it doesn't matter if it collides with another edge you are yet to uncover). Equally, even if you will in due course find many more edges, you want to keep the map size as small as possible until you have found enough of them to saturate the current map. So the number of buckets you need when you first start out may be quite different to the number at the end.

By sampling the number of vacant buckets, you should be able to calculate the probability of a collision (much like the resizing of a hash table) and then adjust the size of the map accordingly. Obviously this performing this calculation will incur some overhead and so again, the period for re-calculating could be tuned empirically and/or left for the user to configure.

WorksButNotTested:
Oh BTW don't run on anything other than x64 (e.g. x86 or aarch64) without `AFL_FRIDA_INST_NO_OPTIMIZE` as they both need a small tweak to make them work.

vanhauser-thc:
> You will want the theoretical maximum of the the number of edges reachable from the entry-point though, and not the maximum for the whole binary right? Also isn't tracing cross-references from static analysis particularly difficult?

yes, however I doubt you could cut that out of the calculation without a lot of extra effort.
what do you mean by cross references?
the safe approach would be to just count all the edges of the target + all libs and report that number back with a tool (Ghidra script).
Then set this value for AFL_FRIDA_UNIQUE_COVERAGE=xxx
the edge id set by frida will be incremental and start with 4, when the ID is increased write that new value into the first 4 bytes of the coverage map.

IMHO that is the easiest way to implement.

> I guess how good the user-experience is depends on how frequently you would have to measure and then re-adjust the map-size manually?

yes. more than 0 is too much, because you would not be able to fuzz unattended.

> Presumably the performance overhead is from the back-end processing of the map to interpret the coverage rather than the overhead from forking with a larger amount of pages?

yes, the memset and (more) the map analysis

> In order to do that, you need to ensure (probabilistically) that a single new edge won't collide with another edge which you have already found

that is not possible at run time, and even at compile/analysis time it is impossible. it would take hours. I tried it :)
The only way to go is edge_id++, no prev_loc, no xoring (this also makes it faster). this will result in block coverage though, not edge coverage. but because of non-colliding this will be better.

> I'm still leaning toward a dynamically re-sized map. The optimal map size at any given point is based on the number of edges encountered so far. e.g. if you have discovered `n` edges, you want to discover the `n+1`th edge. 

the two plus sides of this approach:
  * less ram usage of the shared map
  * user does not need execute a ghidra script

on the neg side:
  * a lot of complexity that would need to be implemented in afl-fuzz and frida that can fail

andreafioraldi:
> The cmplog structure was organised so that there were n headers, followed by n operands. This obviously won't allow us to change the value of n without changing the structure definition. We therefore re-organise this data so that each header is followed immediately by its operands.

This will hurt a lot the performance of cmplog map reset

vanhauser-thc:
> > The cmplog structure was organised so that there were n headers, followed by n operands. This obviously won't allow us to change the value of n without changing the structure definition. We therefore re-organise this data so that each header is followed immediately by its operands.
> 
> This will hurt a lot the performance of cmplog map reset

which is a rare occasion so rather neglectable IMHO. 
but I dont see the reason why we would need to change this.
cmplog is not collision free even for collision-free coverage, but that is not a problem unless there are too many hits.

WorksButNotTested:
By cross references I meant edges. Block numbering should be easy with FRIDA as with prefetch, each is only instrumented once. Just need a monotonic counter. Or perhaps wrap around based on the maps size. Thus with each block numbered avoiding collisions should be easy if you’re just using those for the bitmap. You just need the map to have capacity for enough bits of index to represent the block.

you can still use edges though, just need twice the number of bits. The index is the concatenation of the current and the previous. So the target can effectively feed back the map size required.

alternatively if there are collisions, you can detect easily if a bigger map reduces collisions by simply trying it and comparing the number of populated buckets when a larger size is used.

does the complog map need to adjust its map size too? Does it need to match? Or are collisions in that not as important?

WorksButNotTested:
Another option could be to use a different formula for calculating the bucket ID. So given formula ‘f’ if edges a & b collide, you swap to a formula ‘f2’ where different edges will collide.

vanhauser-thc:
> you can still use edges though, just need twice the number of bits. 

no because of switch()es that wont be correct




> alternatively if there are collisions, you can detect easily if a bigger map reduces collisions by simply trying it and comparing the number of populated buckets when a larger size is used.



> does the complog map need to adjust its map size too? Does it need to match? Or are collisions in that not as important?

as I wrote that map is colliding  (so fixed size) and that is not a problem

> Another option could be to use a different formula for calculating the bucket ID. So given formula ‘f’ if edges a & b collide, you swap to a formula ‘f2’ where different edges will collide.

how do you want to do that at runtime without overhead? is will slow down too much. even if it is only done once per edge, the math and lookups gets more and more complicates the more edges you have. and that time lost is never recovered. and takes way too long.

WorksButNotTested:
Not sure I follow your point about switches? Can you explain? Each edge is a transition from one block to another, each block has a unique incremental number. Thus each edge can be uniquely represented using the ID of the block it came from and the ID of the one it’s going to?

fixed map for complog sounds good and keeps things simple.

with the idea of swapping formula, the idea was you just do it blind. You don’t compute which edges collide, or which don’t, you just alternate through different ways of combining the block ids at each end of the edge. So in f1 you may rotate the source block ID by some number of bits, and in f2, a different number. It’s not perfect, buts computationally quite inexpensive.

but that probably isn’t as effective as knowing how many blocks you’ve encountered and this how many edges and scaling the map accordingly.

vanhauser-thc:
you were saying "you can still use edges though, just need twice the number of bits." and this it not true for
```
switch(foo) { case 0: break; case 1: break ... }
```
which are usually instrumented as "jmp [rax + rbx*8]" and then you have not two edges for one basic block

WorksButNotTested:
Twice as many bits of index, rather than twice the size. Sorry if I wasn’t very clear. Assume our map is a 2d array. The index of the source block is the first dimension, the index of the target block the other. If there are n edges. Then the array needs to be ‘n x n’ is size. (If we just did blocks, it would be a 1d array). This array would have no collisions.

Because multiplication is slow though, you’d probably want the size of the arrays to be rounded up to a power of 2. Then you can just shift the the source block ID left some places and add the destination block ID.

PR #978, opened 2021-06-15
Perf regression

WorksButNotTested:


frida_mode/src/cmplog/cmplog.c
```diff
@@ -35,63 +49,76 @@ static gint cmplog_sort(gconstpointer a, gconstpointer b) {
...
+
+  GArray *cmplog_ranges =
+      g_array_sized_new(false, false, sizeof(GumMemoryRange), 100);
+  gum_process_enumerate_ranges(GUM_PAGE_READ, cmplog_range, cmplog_ranges);
+  g_array_sort(cmplog_ranges, cmplog_sort);
+
+  if (cmplog_ranges->len > RANGE_DATA_MAX) {
 
-    FATAL("Failed to unlink temp file: %s (%d), errno: %d", fname, fd, errno);
+    FATAL("Too many ranges: %u > %lu", cmplog_ranges->len, RANGE_DATA_MAX);
```
vanhauser-thc, line 70: why do you FATAL? is this necessary? you make a target unfuzzable with that. if possible just work with the number of ranges you can handle and live with that that you will miss stuff.

vanhauser-thc:
if I understand this correctly then mappings of the process are read every 60 seconds and cmplog rtn ranges compared to this.
IMHO this is not a good solution.
If a process does a malloc, writes to the memory region and then uses it for a function call it is highly unlikely that just in these microseconds the map update would happen.
you are already importing afl-compiler-rt.o - why dont you just use the `area_is_valid(void *mem, size_t len)` function?

PR #979, opened 2021-06-15
Perf regression4

WorksButNotTested:


frida_mode/src/cmplog/cmplog.c
```diff
@@ -119,12 +145,15 @@ gboolean cmplog_is_readable(guint64 addr, size_t size) {
   /* Check our addres/length don't wrap around */
   if (SIZE_MAX - addr < size) { return false; }
 
+  if (size > MAX_MEMFD_SIZE) { return false; }
```
vanhauser-thc: wouldnt it make more sense to call lseek rather than to return false?

WorksButNotTested: This is if a single request is larger than 64k (rather than we’ve accidentally let the file get too big). It should never happen (since we only query up to 32 bytes), but seemed like a good idea to have some sort of validation.

vanhauser-thc: if it cant happen then dont do it because it costs time? if it is called anywhere accidently with higher values IMHO it is better to crash so this bug can be fixed

WorksButNotTested: Fixed.

WorksButNotTested:
Thanks as always!

PR #986, opened 2021-06-22
Misc

WorksButNotTested:
Lots of little odds and ends per the commit log.

### DIFF
--- a/frida_mode/GNUmakefile
+++ b/frida_mode/GNUmakefile
@@ -25,7 +25,7 @@ RT_CFLAGS:=-Wno-unused-parameter \
 LDFLAGS+=-shared \
 		 -lpthread \
 		 -lresolv \
-		 -ldl \
+		 -ldl
 
 ifdef DEBUG
 CFLAGS+=-Werror \
--- a/frida_mode/GNUmakefile
+++ b/frida_mode/GNUmakefile
@@ -25,7 +25,7 @@ RT_CFLAGS:=-Wno-unused-parameter \
 LDFLAGS+=-shared \
 		 -lpthread \
 		 -lresolv \
-		 -ldl
+		 -ldl \
 
 ifdef DEBUG
 CFLAGS+=-Werror \
@@ -59,6 +59,7 @@ else
 ifdef DEBUG
  RT_CFLAGS:=$(RT_CFLAGS) -Wno-prio-ctor-dtor
 endif
+LDFLAGS+=-z noexecstack
 endif
 
 ifeq "$(shell uname)" "Linux"
--- a/frida_mode/src/cmplog/cmplog.c
+++ b/frida_mode/src/cmplog/cmplog.c
@@ -1,7 +1,8 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
-#include <syscall.h>
+#include <sys/syscall.h>
+#include <unistd.h>
 
 #include "frida-gum.h"
 
@@ -16,8 +17,8 @@ extern struct cmp_map *__afl_cmp_map;
 static GArray *        cmplog_ranges = NULL;
 static GHashTable *    hash = NULL;
 
-static int    memfd = -1;
-static size_t memfd_size = 0;
+static int    tmpfd = -1;
+static size_t tmpfd_size = 0;
 
 static gboolean cmplog_range(const GumRangeDetails *details,
                              gpointer               user_data) {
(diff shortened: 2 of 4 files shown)


frida_mode/src/cmplog/cmplog_x64.c
```diff
@@ -256,6 +254,9 @@ static void cmplog_instrument_cmp_sub(const cs_insn *     instr,
   if (operand1->type == X86_OP_INVALID) return;
   if (operand2->type == X86_OP_INVALID) return;
 
+  if (operand1->size != operand2->size) FATAL("Operand size mismatch");
```
vanhauser-thc: why fatal?

frida_mode/GNUmakefile
```diff
@@ -25,7 +25,7 @@ RT_CFLAGS:=-Wno-unused-parameter \
 LDFLAGS+=-shared \
 		 -lpthread \
 		 -lresolv \
-		 -ldl
+		 -ldl \
```
vanhauser-thc: that backslash is unnecessary and can lead to problems

frida_mode/src/cmplog/cmplog.c
```diff
@@ -94,38 +94,41 @@ static gboolean cmplog_contains(GumAddress inner_base, GumAddress inner_limit,
...
-  ssize_t written = syscall(SYS_write, memfd, (void *)addr, size);
-  if (written < 0 && errno != EFAULT && errno != 0) {
-    FATAL("CMPLOG - Failed __NR_write, errno: %d", errno);
-  }
-  /*
-   * If the write succeeds, then the buffer must be valid otherwise it would
-   * return EFAULT
-   */
-  if (written > 0) { memfd_size += written; }
+  if (msync(page_addr, page_offset + size, MS_ASYNC) < 0) {
```
vanhauser-thc, line 110: does detection of invalid memory work with ASYNC?

frida_mode/src/cmplog/cmplog_x64.c
WorksButNotTested: It should never happen. It's more a sanity check. Let me remove it.

frida_mode/src/cmplog/cmplog.c
WorksButNotTested: Seems to be ok for me. I copied it from here...
https://github.com/AFLplusplus/qemuafl/blob/d73b0336b451fd034e5f469089fb7ee96c80adf2/accel/tcg/tcg-runtime.c#L195

frida_mode/src/cmplog/cmplog_x64.c
vanhauser-thc: yes. sanity checks are only a waste of time in fuzzing :)

frida_mode/src/cmplog/cmplog_x64.c
```diff
@@ -247,13 +254,8 @@ static void cmplog_instrument_cmp_sub(const cs_insn *     instr,
...
 
-  if ((operand1->type == X86_OP_MEM) &&
-      (operand1->mem.segment != X86_REG_INVALID))
-    return;
-
-  if ((operand2->type == X86_OP_MEM) &&
-      (operand2->mem.segment != X86_REG_INVALID))
-    return;
+  /* Both operands are the same size */
+  if (operand1->size == 1) { return; }
```
vanhauser-thc, line 258: can that even happen? I mean there is a cmp. so why ignore it?

frida_mode/src/cmplog/cmplog_x64.c
WorksButNotTested: Done.

frida_mode/src/cmplog/cmplog_x64.c
WorksButNotTested: Single byte compares/subtracts are a thing. But QEMU mode ignores them...
https://github.com/AFLplusplus/qemuafl/blob/d73b0336b451fd034e5f469089fb7ee96c80adf2/qemuafl/cpu-translate.h#L88
Presumably, this is because the overhead of cmplog is not worthwhile compared to just letting the fuzzer guess the byte the normal way.

frida_mode/src/cmplog/cmplog_x64.c
domenukk: leaving sanity checks in place for debug builds would/could make sense?

frida_mode/src/cmplog/cmplog_x64.c
vanhauser-thc: you are right

PR #987, opened 2021-06-22
push to stable

vanhauser-thc:
accurate map density calculation, cmplog fix for qemu + frida, afl-showmap -i subdir descending (also for afl-cmin)


afl-cmin
```diff
@@ -296,13 +296,13 @@ BEGIN {
...
   }
 
-  if (0 == system( "test -d "in_dir"/default" )) {
-    in_dir = in_dir "/default"
-  }
-
-  if (0 == system( "test -d "in_dir"/queue" )) {
-    in_dir = in_dir "/queue"
-  }
+  #if (0 == system( "test -d "in_dir"/default" )) {
```
domenukk, line 299: why comment and not remove?

src/afl-fuzz-init.c
```diff
@@ -710,7 +710,10 @@ void read_testcases(afl_state_t *afl, u8 *directory) {
 
   }
 
-  for (i = 0; i < (u32)nl_cnt; ++i) {
+  i = nl_cnt;
+  do {
```
domenukk: why not keep this as (more readable) for loop? What happens if `nl_cnt == 0`?

domenukk (approved):

afl-cmin
vanhauser-thc: I like to keep it in case I want to revert the behaviour. at a later time I usually remove it - when I am sure it is not needed anymore.

src/afl-fuzz-init.c
vanhauser-thc: value 0 is already handled above. I found reversing the for loop less readable and more error prone than moving to while.

afl-cmin
domenukk: It's still in git though?

src/afl-fuzz-init.c
domenukk: I don't think it's handled? If you'd use a for loop here we wouldn't have this discussion ;)

domenukk: At least make it a while loop instead of do while?

vanhauser-thc: > I don't think it's handled? If you'd use a for loop here we wouldn't have this discussion ;)

above, not visible in the diff is `if (nl_cnt == 0) { .... return;}`.



> At least make it a while loop instead of do while?

I find it best like this, for the same reason I did not use for() is why I did not use while(), but do {} while instead.
change it if you want. but dont introduce a bug :)

afl-cmin
vanhauser-thc: of course. I talk about a timeframe like in a few months, I will at some point remove it.

src/afl-fuzz-init.c
domenukk: Where? All I see is`nl_cnt = scandir(dir, &nl, NULL, alphasort);`.
It's unlikely that it is `0` on linux as it should(?) return `.` and `..`,  so I guess it usually works(?), just not intuitive to look at :)

vanhauser-thc: uh right! I looked at foreign_sync. but it is impossible that it is empty, at least on unix, as all unix directories have "." and "..", so nl_cnt is always < 0 on error or at minimum 2.

vanhauser-thc: but for windows that is likely different so I just added a check. just not to run into this in the future.

PR #992, opened 2021-06-23
Added JS support

WorksButNotTested:
Initial JavaScript scripting support to allow use of FRIDA APIs in assisting configuration.


domenukk:
Ooh this is cool!

WorksButNotTested:
> Ooh this is cool!

The plan is to add support for writing the persistent hook in JavaScript too and maybe even Fridas CModule support (which uses tinycc to dynamically  compile C code).

Also, you should be able to use (with this PR) all the normal Frida goodness to patch up the target under test. E.g. making any checks pass, or patching out a bug you already found so that boring crashes don’t cause the target to crash and hence stop persistent mode running until the end of the loop.

vanhauser-thc:
can you please add the feature to the README of frida?

WorksButNotTested:
I was going to update the docs when I’ve finished all the JS stuff, but would you prefer I do it as I go?

vanhauser-thc:
if you add a feature in a PR the PR should also have the update to the documentation :)

WorksButNotTested:
> if you add a feature in a PR the PR should also have the update to the documentation :)

Fair enough. Hold tight for now and I'll get some docs sorted out for tonight.

vanhauser-thc:
thank you :)

PR #1014, opened 2021-07-09
remove redundant check

yuawn:
The `ck_realloc()` function already checked the pointer returned.


vanhauser-thc:
true, thank you!

PR #1017, opened 2021-07-13
Support AFL_NO_FORKSRV env-var in afl-tmin and afl-cmin

jhertz:
Does what it says on the can


domenukk:
Great, thx! Do you feel like pasting these three lines into `afl-showmap.c`, too? Else I will :)

jhertz:
done :)

domenukk:
Guess we'll need to copy&paste the info about this env variable to the help screens in showmap, cmin, cmin.bash and tmin still but not urgent

PR #1021, opened 2021-07-14
Improved block and edge numbering to reduce collisions

WorksButNotTested:
See [here](https://github.com/WorksButNotTested/AFLplusplus/blob/map-density/frida_mode/MapDensity.md) for the details.


domenukk:
The idea of a different "seed" for each instance is great! Basically could allow us to have smaller maps, but distributed over many cores

vanhauser-thc:
> The idea of a different "seed" for each instance is great! Basically could allow us to have smaller maps, but distributed over many cores

what is does is making the edge IDs as random as they are in afl-cc CLASSIC mode. (we need this in qemu too)
the random seeding is like recompiling the target with afl-cc CLASSIC for every afl-fuzz instance (which was originally a recommendation of mine for large programs).
I would not make the maps smaller because of that :)

PR #1035, opened 2021-07-20
add optimin corpus minimizer

adrianherrera:
Hi AFL++ devs!

Per our Twitter chat, here is a PR for integrating our MaxSAT-based corpus minimizer (from our paper, [Seed Selection for Successful Fuzzing](https://dl.acm.org/doi/10.1145/3460319.3464795)) as an AFL++ "util".

I'm making the PR against the dev branch, because that is what was suggested in the "contributing guidelines". However, it makes no modifications to the rest of the AFL++ code base. I've tried to make the code as self-contained as possible, however it does rely on [EvalMaxSAT](https://github.com/FlorentAvellaneda/EvalMaxSAT) as the underlying MaxSAT engine and the LLVM support API to simplify the code/avoid duplicating common functionality.

Let me know what you think!


utils/optimin/src/OptiMin.cpp
```diff
@@ -0,0 +1,443 @@
...
+  if (TargetArgsHasAtAt)
+    AFLShowmapArgs.append({"-A", Seed});
+  else
+    Redirects[/* stdin */ 0] = Seed;
+
+  AFLShowmapArgs.append({"--", TargetProg});
+  AFLShowmapArgs.append(TargetArgs.begin(), TargetArgs.end());
+
+  // Run afl-showmap
+  const int RC = sys::ExecuteAndWait(AFLShowmapPath, AFLShowmapArgs,
```
andreafioraldi, line 176: Here problably this should be replaced with the code in forkserver.h

andreafioraldi:
Nice! The only thing here that can be a problem is the use of cmake, but as it is in the utils/ folder I have mixed feelings maybe we can just use cmake for it

vanhauser-thc:
> Nice! The only thing here that can be a problem is the use of cmake, but as it is in the utils/ folder I have mixed feelings maybe we can just use cmake for it

same for me. I would prefer not having cmake in this, but it is in a utils/ folder ... and EvalMaxSAT needs cmake too, so I don't see an easy way to get rid of it.

fix the conflict please I just created with a different push :)

domenukk:
Another question about `utils`, should we make this work without `git`? I don't think anybody would build this from a tarball, but @vanhauser-thc may disagree?
At least for non-utils stuff, we use a wrapper script that either updates the submodule, or wgets a tar, see
https://github.com/AFLplusplus/AFLplusplus/blob/stable/unicorn_mode/update_uc_ref.sh

vanhauser-thc:
@domenukk yes good point.
@adrianherrera just take the custom_mutators/gramatron/build_gramatron_mutator.sh as a template

vanhauser-thc:
btw the CI failures are issues in github's CI docker, failing to get essential packages. nothing to do with this

utils/optimin/src/OptiMin.cpp
adrianherrera: I did consider this, but given that I use the LLVM API so much already, I figured I may as well use it exclusively. Let me know if you want me to change this, however!

adrianherrera:
Hey @vanhauser-thc and @domenukk,

I've added a build script and fixed the conflict.

vanhauser-thc:
LGTM
thank you!

PR #1038, opened 2021-07-21
Fix links in FAQ.md

yuawn:
Fix relative path in docs/FAQ.md.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -58,8 +58,8 @@ Solution: `git pull ; make clean install` of afl++
 
 ## How to improve the fuzzing speed?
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
you are right that the links are outdated, however you fix is incomplete :)
the referenced docs are in instrumentation/... but as the FAQ.md is in docs/ it would need to be ../instrumentation/foobar.md
you are using /instrumentation in most placed instead, but that does not work well either.

yuawn:
I tested it, it works in web.
But still change it to ../instrumentation.

yuawn:
@vanhauser-thc done!

vanhauser-thc:
yes it would work on github, but not once it is checked out anymore. now its fine, thank you.

PR #1042, opened 2021-07-23
Remove redundant if statement in afl-cc

yuawn:
When `instrument_mode == INSTRUMENT_LTO`, control always passes to the first if statement, and it seems correct too.


vanhauser-thc:
thank you

PR #1050, opened 2021-07-30
improved error message

fbeqv:
took me way to long to figure out the problem here and only afterwards learned about afl-system-config.


vanhauser-thc:
sure why not :) 
for the future: note that we only accept PRs to dev.

PR #1055, opened 2021-08-07
`afl-plot-ui` window starts maximized now

DMaroo:
* Window starts maximized now (can be resized again by user if necessary)
* Updated the util specific `README` with the build and install instructions


vanhauser-thc:
thank you!

PR #1061, opened 2021-08-09
Removed loading of "paths_total" property from stats file

mqf20:
When running `afl-fuzz` with `AFL_AUTORESUME=1` and/or if the input directory is specified as `-`, the property `paths_total` should *not* be read from the `fuzzer_stats` file by `load_stats_file()`. 

This is because:

- Prior to calling `load_stats_file()`, `afl-fuzz` has already set the value of `afl->queued_paths` by parsing the `queue/` folder and loaded the queued seeds into `afl->queue_buf`. 
- It is redundant to read the value of `afl->queued_paths` again from the `fuzzer_stats` file (and overwrite its value). 
- Consider the following scenario:
  - The user runs a fuzzing experiment and stops it after some time.
  - The user modifies the `queue/` folder by culling/minimizing the queued seeds.
  - The user does not update the `fuzzer_stats` file.
  - The user continues the fuzzing experiment by setting `AFL_AUTORESUME=1`.
  - As a result, the read value of `afl->queued_paths` and the *actual number of queued seeds* will be inconsistent.
  - Since the read value of `afl->queued_paths` is larger than the *actual number of queued seeds*, an out-of-bounds access on the array `afl->queue_buf` will take place.

### DIFF
--- a/src/afl-fuzz-stats.c
+++ b/src/afl-fuzz-stats.c
@@ -146,10 +146,6 @@ void load_stats_file(afl_state_t *afl) {
           if (!strcmp(keystring, "execs_done        "))
             afl->fsrv.total_execs = strtoull(lptr, &nptr, 10);
           break;
-        case 10:
-          if (!strcmp(keystring, "paths_total       "))
-            afl->queued_paths = strtoul(lptr, &nptr, 10);
-          break;
         case 12:
           if (!strcmp(keystring, "paths_found       "))
             afl->queued_discovered = strtoul(lptr, &nptr, 10);


vanhauser-thc:
I have to think about this.

A user should not tamper with the contents of the sync folder.

Why would you do it?
In my experience with minimizing queues it degrades the performance rather than increasing it.
Our changed algorithm also tries to be clever which will results in ignored existing queue entries on restarts if they are less interesting than others.
And if the minimizing really is important, why is it then also so important to have the old stats included?

that being said your point that it is not needed to be read in is true, however I am thinking rather about getting the value anyway and either print a fat warning (and use the real number) or abort.

mqf20:
@vanhauser-thc  thanks for reviewing this PR.

> I have to think about this.
> 
> A user should not tamper with the contents of the sync folder.
> 
> Why would you do it?

In my case, I stopped the fuzzer so that I can apply a custom minimization strategy on the queued seeds.

> In my experience with minimizing queues it degrades the performance rather than increasing it.
> Our changed algorithm also tries to be clever which will results in ignored existing queue entries on restarts if they are less interesting than others.
> And if the minimizing really is important, why is it then also so important to have the old stats included?

I agree that there are better ways to minimize queues and restart the experiment. However, I assumed (perhaps incorrectly?) that this is the purpose of `AFL_AUTORESUME` -- for the user to adjust the fuzzing parameters (including the queued seeds), then resume the experiment gracefully.

> that being said your point that it is not needed to be read in is true, however I am thinking rather about getting the value anyway and either print a fat warning (and use the real number) or abort.

Would you like me to amend the PR?

vanhauser-thc:
> However, I assumed (perhaps incorrectly?) that this is the purpose of `AFL_AUTORESUME` -- for the user to adjust the fuzzing parameters (including the queued seeds), then resume the experiment gracefully.

The resume feature is mostly for accidental aborts (laptop ran out of power, system reboot, accidently pressing Control-C), and for switching around command line options (like setting `-p explore -L0` and restarting).
as it is not that well documented how the queue works etc. I would not have thought that people could get the idea that this is the purpose of the resume option :)

I think the best way forward is to change the PR to compare the fuzzer_stats value to the real value and print a WARNF("...") that the queue was modified, things might not work and they on their own :) (and use the real value, not the fuzzer_stats value)

mqf20:
Resubmitted PR as https://github.com/AFLplusplus/AFLplusplus/pull/1064

PR #1063, opened 2021-08-10
Fix "havoc_mutations" not working in Python module

ASBOSNIA:
The Python module of custom mutator was loaded without initializing the variables stacked_custom and stacked_custom_prob. This caused the havoc_mutation and havoc_mutation_probability APIs not to work. So I wrote a fix referring to "afl-fuzz-mutators.c".


vanhauser-thc:
thank you!
for the future please note that we only accept PRs to the dev branch, a fixed that for you.
waiting for the CI to finish then merging.

PR #1068, opened 2021-08-16
Edit FAQ

llzmb:
Changes:
- Edit FAQ.
- Add history.
- Add best practices.

One of the links does not work yet (links_examples_writeups.md). The missing file will be added with another PR where some of the content from the README.md will be moved to docs/.


vanhauser-thc:
The file name links_examples_writeups.md is weird. how about just "tutorials.md" ?

also the history_afl++.md file is what I meant yesterday with 100+ md files with small content is not helping. I think this content could easily be added to another file as an own chapter.

llzmb:
The file name was an attempt to suggest that the file contains only links. There is also a file with links to tools that help with fuzzing. Both files would be grouped if sorted alphabetically. I will try to come up with another idea.

docs/FAQ.md
```diff
@@ -1,116 +1,57 @@
...
-  * [Why is my stability below 100%?](#why-is-my-stability-below-100)
-  * [How can I improve the stability value?](#how-can-i-improve-the-stability-value)
+# Frequently asked questions (FAQ)
 
 If you find an interesting or important question missing, submit it via
-[https://github.com/AFLplusplus/AFLplusplus/issues](https://github.com/AFLplusplus/AFLplusplus/issues)
+[https://github.com/AFLplusplus/AFLplusplus/discussions](https://github.com/AFLplusplus/AFLplusplus/discussions).
 
-## What is the difference between AFL and AFL++?
+### General
```
vanhauser-thc: general, targets and performance headers have 3 # .... shouldnt it be 2 # ?

docs/FAQ.md
```diff
@@ -144,100 +86,70 @@ code example above):
...
-  2. Second step: Identify which edge ID numbers are unstable
+A value above 90% is usually fine and a value above 80% is also still ok, and even a value above 20% can still result in successful finds of bugs.
+However, it is recommended that for values below 90% or 80% you should take countermeasures to improve stability.
 
-     run the target with `export AFL_DEBUG=1` for a few minutes then terminate.
-     The out/fuzzer_stats file will then show the edge IDs that were identified
-     as unstable in the `var_bytes` entry. You can match these numbers
-     directly to the data you created in the first step.
-     Now you know which functions are responsible for the instability
+### How can I improve the stability value?
```
vanhauser-thc: I think "how can I improve the stability value" should be moved to best practices" and the previous chapter ending with a link there.

vanhauser-thc:
left 2 comments, otherwise looks good to me :)

docs/FAQ.md
llzmb: To fix that and to avoid the redundant questions, I went with another approach:
https://github.com/llzmb/AFLplusplus/blob/docs_edit_faq/docs/FAQ.md

What do you think?

vanhauser-thc: I think it is very readable in MD, and still readable with a text editor so that is a good solution. what is not possible anymore though is to have a link to a specific entry from a different document, or is that still somehow possible?

vanhauser-thc:
IMHO this can be merged after the last item with links is resolved one way or another.

docs/FAQ.md
llzmb: Good point! I can add an ID to each summary tag. It makes the MD a little less human-readable, but it should work for internal and external linking:
```html
<details>
  <summary id="where-can-i-find-tutorials">Where can I find tutorials?</summary><p>

  We compiled a list of tutorials and exercises, see [tutorials.md](tutorials.md).
</p></details>
```

```markdown
[internal link](#where-can-i-find-tutorials)
[external link](FAQ.md#where-can-i-find-tutorials)
```

What do you think about this solution?

vanhauser-thc: I think this is good. @domenukk / @andreafioraldi  ?

llzmb:
The last commit already adds the IDs: https://github.com/llzmb/AFLplusplus/blob/docs_edit_faq/docs/FAQ.md

If there is nothing else to change -- can you please merge?

vanhauser-thc:
@llzmb I would propose we merge faq + readme together. can both be merged now?

PR #1074, opened 2021-08-19
push to stable

vanhauser-thc:
before we get a larger readme change from Jana maybe we should push to stable.


domenukk (approved):

frida_mode/DEBUGGING.md
```diff
@@ -0,0 +1,160 @@
...
+
+It should be noted that attempting to debug code using gdb which has been
+instrumented in FRIDA is unlikely to be successful since the debugger will be
+inserting breakpoints by patching the code in memory. FRIDA works by reading
+this code and generating an instrumented copy to execute. In any case, unless
+you are very familiar with the implementation of Stalker, the instrumented code
+generated by FRIDA is likely to be very difficult to follow. For this reason,
+the following debugging strategies are outlined below.
+
+Byte convention below all files should be provided with their path (they are
```
hexcoder-: "By convention" instead of "Byte convention"?

instrumentation/README.persistent_mode.md
```diff
@@ -2,13 +2,16 @@
 
 ## 1) Introduction
 
-The most effective way is to fuzz in persistent mode, as the speed can easily
+In persistent mode, AFL++ fuzzes a target multiple times
+in a single process, instead of forking a new process for each fuzz execution.
```
hexcoder-: This is technically unprecise, I think, as the forkserver already avoided fork()ing and used clone() and COW. 
The main points here (for me) is: 
 - the life time of the target is extended over multiple fuzzing runs.
 - The target implements a loop and runs like a nonterminating server waiting for inputs.
 - AFL supplies these inputs.

instrumentation/SanitizerCoverageLTO.so.cc
```diff
@@ -626,12 +626,41 @@ bool ModuleSanitizerCoverage::instrumentModule(
...
+                             !FuncName.compare("Curl_safe_strcasecompare") ||
+                             !FuncName.compare("cmsstrcasecmp"));
+            isStrncasecmp &= (!FuncName.compare("strncasecmp") ||
+                              !FuncName.compare("strnicmp") ||
+                              !FuncName.compare("ap_cstr_casecmpn") ||
+                              !FuncName.compare("OPENSSL_strncasecmp") ||
+                              !FuncName.compare("xmlStrncasecmp") ||
+                              !FuncName.compare("g_ascii_strncasecmp") ||
+                              !FuncName.compare("Curl_strncasecompare") ||
+                              !FuncName.compare("g_strncasecmp"));
```
hexcoder-, line 662: If these lists keep on growing, we could soon have namespace issues (clashes with user functions named the same).

hexcoder- (approved):
LGTM, thanks.

andreafioraldi (approved):

frida_mode/DEBUGGING.md
vanhauser-thc: @WorksButNotTested

instrumentation/SanitizerCoverageLTO.so.cc
vanhauser-thc: that is not an issue IMHO. it is not only matched on function name but also parameter types and return type. and even if these all match and still it does something completely different the impact is minimal

instrumentation/README.persistent_mode.md
vanhauser-thc: I tried to clarify it a little bit better. I am not sure that your main points make it easier to understand either, could work for some and not for others. I think this only works with seeing example code, whatever the writing.

frida_mode/DEBUGGING.md
WorksButNotTested: Yep. Typo sorry.

hexcoder-: fixed, Thanks.

instrumentation/README.persistent_mode.md
hexcoder-: Yes, I agree; lets delegate this to an example.

instrumentation/SanitizerCoverageLTO.so.cc
hexcoder-: ok with me. Lets see, if someone experiences this conflict at all.

PR #1093, opened 2021-09-20
fix a bug in frida_mode/test/libpcap/GNUMakefile

Mart11n:
A bug in frida_mode/test/libpcap/GNUMakefile

line 62 should be : `ifeq "$(ARCH)" "arm64"`


vanhauser-thc:
thank you!

PR #1095, opened 2021-09-27
Fix null ptr de-reference of unresolved symbols on early init

kotee4ko:
When libcompcov preloaded to the environment of complicated programs - it crash other libs.

изображение
As you can see from crash back-trace `memcmp()` pointed to by zero.
изображение

So, whole process crashes even before `_start`.

This patch should fix that.


andreafioraldi:
Ty!

PR #1096, opened 2021-10-03
frida mode macOs warning fix, syscall being deprecated, using pthread…

devnexen:
…_threadid_np available since snow leopard


vanhauser-thc:
this breaks our CI test for frida, are you sure this works for Linux too?

devnexen:
ah right no I tought this was macos only. hold on

PR #1101, opened 2021-10-08
Dev

vanhauser-thc:


hexcoder- (approved):

src/afl-common.c
```diff
@@ -51,6 +55,66 @@ u8  last_intr = 0;
   #define AFL_PATH "/usr/local/lib/afl/"
 #endif
 
+u32 check_binary_signatures(u8 *fn) {
```
domenukk, line 58: I (still) don't understand why we need this, the compiled target knows if it is using a deferred forkserver or not...
There should be a better way than to have the parent scan the binary, imho...

domenukk (approved):

src/afl-common.c
vanhauser-thc: yeah maybe. we can put that on the todo list and do something about this at some point. eg. add a change to a global var in afl-compiler-rt.o.c in the __AFL_INIT macro etc.

TODO.md
```diff
@@ -1,30 +1,27 @@
 # TODO list for AFL++
 
-## Roadmap 3.00+
+## TODO
```
domenukk: TODO is already in the title (minor nitpick)

src/afl-fuzz-queue.c
```diff
@@ -315,7 +315,96 @@ void mark_as_redundant(afl_state_t *afl, struct queue_entry *q, u8 state) {
...
+/* check if pointer is ascii or UTF-8 */
+
+u8 check_if_text_buf(u8 *buf, u32 len) {
+
+  u32 offset = 0, ascii = 0, utf8 = 0;
+
+  while (offset < len) {
+
+    // ASCII: <= 0x7F to allow ASCII control characters
+    if ((buf[offset + 0] == 0x09 || buf[offset + 0] == 0x0A ||
```
domenukk, line 327: why `+ 0` here?

domenukk (approved):

src/afl-fuzz-queue.c
vanhauser-thc: just for copy-paste-ability + readability

PR #1138, opened 2021-11-03
cmplog enhancement

vanhauser-thc:


instrumentation/afl-compiler-rt.o.c
```diff
@@ -154,6 +158,12 @@ static void at_exit(int signal) {
 
 }
 
+#ifdef WORD_SIZE_64
+  #define default_hash(a, b) XXH64(a, b, HASH_CONST)
+#else
+  #define default_hash(a, b) XXH64(a, b, HASH_CONST)
```
devnexen: did you mean something else ?

vanhauser-thc: no, why do you ask? afl-compiler-rt has `#include "xxhash.h"`

devnexen: I meant `default_hash` is the same for both cases unless I missed something.

vanhauser-thc: :D ah yes you are right. I thought we have a different one for 32 bit, but we have not. ty!

vanhauser-thc:
the other variant will be merged.

PR #1139, opened 2021-11-03
Fix request size & remove redundant code

yuawn:
Fix request size.
`ck_alloc` clears the array with `memset`, adding null terminator is no need.

### DIFF
--- a/src/afl-common.c
+++ b/src/afl-common.c
@@ -217,7 +217,7 @@ char **get_qemu_argv(u8 *own_loc, u8 **target_path_p, int argc, char **argv) {
 
   }
 
-  char **new_argv = ck_alloc(sizeof(char *) * (argc + 4));
+  char **new_argv = ck_alloc(sizeof(char *) * (argc + 3));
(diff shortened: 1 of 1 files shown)


domenukk:
I think argv may need to be null-terminated at certain places, if you remove the last element, it'll break.
Also, I'm not a fan of leaving out the explicit initialization, explicit is better than implicit, and this is not a hot code path at all.

domenukk:
Ah the main issue is that the terminator was off-by-one?

yuawn:
It is null-terminated, `ck_alloc` assigned null to all elements, just no need to assign again.

yuawn:
Hi @domenukk 
It request exceeded size originally, there are two null terminator in the array.

I see, so keep the unnecessary null terminator initialization? 
IMHO `ck_alloc` already do the explicit initialization that includes clearing the returned buffer.

Thanks!

domenukk:
Guess it's fine this way, too. Thx for spotting it

PR #1148, opened 2021-11-07
Fix index of cmp_map->log[key].

tl455047:
In my understanding, the index of log[key] should be j at src/afl-fuzz-redqueen.c:1672,  otherwise the condition at line 1672 will always be true.


vanhauser-thc:
yes this looks wrong. thank you!

PR #1149, opened 2021-11-07
Use FIFOs instead of hacky sleep code to pass messages in afl-plot

DMaroo:
FIFOs (named pipes) are more reliable and appropriate for the task of passing messages. We can also remove the sleep period and rely on the fact that `cat` will only end once the name pipe's stream is closed.

### DIFF
--- a/afl-plot
+++ b/afl-plot
@@ -196,14 +196,15 @@ exit 1
 fi
 
 mkdir -p "$outputdir/tmp"
+mkfifo "$outputdir/tmp/win_ids"
 afl-plot-ui > "$outputdir/tmp/win_ids" &
 
-sleep 0.5
+W_IDS=$(cat $outputdir/tmp/win_ids)
 
-W_ID1=$(cat $outputdir/tmp/win_ids | head -1)
-W_ID2=$(cat $outputdir/tmp/win_ids | head -2 | tail -1)
-W_ID3=$(cat $outputdir/tmp/win_ids | head -3 | tail -1)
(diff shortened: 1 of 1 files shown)


afl-plot
```diff
@@ -196,14 +196,15 @@ exit 1
 fi
 
 mkdir -p "$outputdir/tmp"
+mkfifo "$outputdir/tmp/win_ids"
```
vanhauser-thc: can you please add a `rm -f "$outputdir/tmp/win_ids"` prior?
plus an `|| exit 1` on the mkfifo command?

afl-plot
```diff
@@ -196,14 +196,15 @@ exit 1
...
 afl-plot-ui > "$outputdir/tmp/win_ids" &
 
-sleep 0.5
+W_IDS=$(cat $outputdir/tmp/win_ids)
```
vanhauser-thc, line 209: isnt an `rm -f "$outputdir/tmp/win_ids"` needed here?

afl-plot
```diff
@@ -196,14 +196,15 @@ exit 1
...
```
vanhauser-thc: this is missing the double quotes on the parameter

vanhauser-thc:
sure why not, thank you!
please fix the nits I identified then I will merge it.

btw if you know enough about gnuplot to add a feature to afl-plot to combine multiple afl instances in the plot at once - that would be awesome.

vanhauser-thc:
thank you

PR #1153, opened 2021-11-09
Frida

WorksButNotTested:


vanhauser-thc:
you should only filter out OK, but of course not FATAL, and mostly not WARNings :)

WorksButNotTested:
Changed to not filter FWARNF and FFATAL.

vanhauser-thc:
thank you!

PR #1156, opened 2021-11-11
Add CoreSight mode support

retrage:
This PR proposes adding CoreSight mode to AFL++.
Please read [our blog post](https://ricercasecurity.blogspot.com/2021/11/armored-coresight-towards-efficient.html) for detailed background. Thanks to the AFL++'s high extensibility, we can achieve this mode with only ~400 LoC to AFL++.

It mainly just adds `-A` option to switch CoreSight mode, but I have an issue regarding the option. As I noted in the comment, The `-A` option is already used in `afl-showmap` for the undocumented feature "Another afl-cmin specific feature." Because I saw this option exists on the original AFL code, I'm not sure what change should I make. I temporarily use `-P` for `afl-showmap`, but it's inconsistent. Please give me advice on this.

NOTE: I submit this PR from my personal account repository because the [original CoreSight mode repository](https://github.com/RICSecLab/AFLplusplus-cs/tree/retrage/cs-mode-support) is not recognized by GitHub as a fork of this repository, and we cannot submit a PR from the original repository.

### DIFF
--- a/src/afl-analyze.c
+++ b/src/afl-analyze.c
@@ -77,6 +77,7 @@ static volatile u8 stop_soon;          /* Ctrl-C pressed?                   */
 static u8 *target_path;
 static u8  frida_mode;
 static u8  qemu_mode;
+static u8  cs_mode;
 static u32 map_size = MAP_SIZE;
 
 static afl_forkserver_t fsrv = {0};   /* The forkserver                     */
@@ -790,6 +791,8 @@ static void set_up_environment(char **argv) {
 
     } else {
 
+      /* CoreSight mode uses the default behavior. */
+
       setenv("LD_PRELOAD", getenv("AFL_PRELOAD"), 1);
       setenv("DYLD_INSERT_LIBRARIES", getenv("AFL_PRELOAD"), 1);
 
@@ -845,6 +848,7 @@ static void usage(u8 *argv0) {
       "  -f file       - input file read by the tested program (stdin)\n"
       "  -t msec       - timeout for each run (%u ms)\n"
       "  -m megs       - memory limit for child process (%u MB)\n"
+      "  -A            - use binary-only instrumentation (CoreSight mode)\n"
       "  -O            - use binary-only instrumentation (FRIDA mode)\n"
       "  -Q            - use binary-only instrumentation (QEMU mode)\n"
       "  -U            - use unicorn-based instrumentation (Unicorn mode)\n"
@@ -890,7 +894,7 @@ int main(int argc, char **argv_orig, char **envp) {
 
   afl_fsrv_init(&fsrv);
 
-  while ((opt = getopt(argc, argv, "+i:f:m:t:eOQUWh")) > 0) {
+  while ((opt = getopt(argc, argv, "+i:f:m:t:eAOQUWh")) > 0) {
 
     switch (opt) {
 
@@ -989,6 +993,14 @@ int main(int argc, char **argv_orig, char **envp) {
 
         break;
 
+      case 'A':                                           /* CoreSight mode */
+
+        if (cs_mode) { FATAL("Multiple -A options not supported"); }
+
+        cs_mode = 1;
+        fsrv.cs_mode = cs_mode;
+        break;
+
       case 'O':                                               /* FRIDA mode */
 
         if (frida_mode) { FATAL("Multiple -O options not supported"); }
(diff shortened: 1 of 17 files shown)


vanhauser-thc:
Thanks for the PR!

you are right, I forgot about the hidden -A parameter in afl-showmap.
But that is easy to fix as it is undocumented, so I can switch it to something else afterwards, update afl-cmin*, and then rename -P to -A.

One wish from my side: could you please update the help for -A to state "ARM Coresight" instead of just "Coresight"? This makes it more obvious for people what it is for.

And a question because I am curious - does it scale to run multiple times in parallel? (e.g. 20 x -S instances)

vanhauser-thc:
and an idea - wouldnt it be easier to just use an LD_PRELOAD lib with a late constructor to establish the forkserver? that would eliminate the glibc need.

src/afl-analyze.c
```diff
@@ -989,6 +993,14 @@ int main(int argc, char **argv_orig, char **envp) {
 
         break;
 
+      case 'A':                                           /* CoreSight mode */
+
+        if (cs_mode) { FATAL("Multiple -A options not supported"); }
```
domenukk, line 1002: should the flag be available on non aarch64 compile targets? What about non-linux?

retrage: I added the platform checks by c2feee4.

retrage:
> And a question because I am curious - does it scale to run multiple times in parallel? (e.g. 20 x -S instances)

Unfortunately, the answer is no. It is theoretically possible to support fuzzing in parallel, but the current implementation cannot handle tracing multiple processes. So the afl-fuzz needs to check if `-S` is not supplied on CoreSight mode.

retrage:
> and an idea - wouldnt it be easier to just use an LD_PRELOAD lib with a late constructor to establish the forkserver? that would eliminate the glibc need.

I previously tried to use `LD_PRELOAD`, but could not. I would like to try again this time.

vanhauser-thc:
> > And a question because I am curious - does it scale to run multiple times in parallel? (e.g. 20 x -S instances)
> 
> Unfortunately, the answer is no. It is theoretically possible to support fuzzing in parallel, but the current implementation cannot handle tracing multiple processes. So the afl-fuzz needs to check if `-S` is not supplied on CoreSight mode.

that check it not there yet? 
so basically it should only work if -S and -M are not supplied (so -S default only)?
that limits this a lot.

src/afl-analyze.c
```diff
@@ -989,6 +993,20 @@ int main(int argc, char **argv_orig, char **envp) {
...
+
+        if (!(__aarch64__ && __linux__)) {
```
vanhauser-thc: this has to be an #if

retrage: Oops, sorry, commit feff819 should fix this.

retrage:
> > > And a question because I am curious - does it scale to run multiple times in parallel? (e.g. 20 x -S instances)
> > 
> > 
> > Unfortunately, the answer is no. It is theoretically possible to support fuzzing in parallel, but the current implementation cannot handle tracing multiple processes. So the afl-fuzz needs to check if `-S` is not supplied on CoreSight mode.
> 
> that check it not there yet? so basically it should only work if -S and -M are not supplied (so -S default only)? that limits this a lot.

I added the check as a commit d4a0fd4.
I agree. Most of the high-end ARM-based SoCs are multi-core, so we would like to support parallel fuzzing on this mode in the future.

vanhauser-thc:
IMHO it is fine for merging. is there anything else on your todo list before that @retrage ?

retrage:
That's all so far. I pushed minor fixes 2d4b18f and TODO list 3f864fa.

vanhauser-thc:
moved afl-showmap -A to -H and -P to -A and modified the afl-cmin* tools for the change.
Also modified the help output so that only the options available are visible.

hexcoder-:
@retrage : While compiling on aarm64 I am having problems cloning coresight-trace and its submodules.
This is because I get lines like these:
`Submodule 'coresight_mode/coresight-trace' (git@github.com:RICSecLab/coresight-trace.git) registered for path 'coresight-trace'`
The url **git@github.com:RICSecLab/coresight-trace.git** does not work with non-repo owners like me. I get this
```
Submodule path 'coresight-trace': checked out 'ec0fd6104720ac0b59967616363dc18209adc02e'
Cloning into '/home/heiko/AFLplusplus/coresight_mode/coresight-trace/CSAL'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

The form https://github.com/RICSecLab/coresight-trace would work. But there are clones of other recursive submodules with the same problem.

Could you fix this, please? Thanks a lot!

PR #1164, opened 2021-11-17
Fix

WorksButNotTested:
**NOTE** the following PR contains changes to `afl-compiler-rt` besides just changes to `afl-frida-trace.so`.

1. When sorting address ranges for inclusion/exclusion, a numeric overflow caused them to be incorrectly sorted on occasion.
2. The ASAN DSO should be excluded from coverage so that it doesn't end up generating instrumentation for itself.
3. When generating coverage data, prefetching of adjacent blocks should be disabled as the coverage data is based on blocks compiled, not when they are executed.
4. **afl-compiler-rt** `afl-compilter-rt`** sets the `is_persistent` variable depending on the `__AFL_PERSISTENT` environment variable in a constructor here...
https://github.com/AFLplusplus/AFLplusplus/blob/773baf9391ff5f1793deb7968366819e7fa07adc/instrumentation/afl-compiler-rt.o.c#L1261

The corresponding environment variable is set by `afl-fuzz` depending on the configuration of `AFL_FRIDA_PERSISTENT_ADDR` here...
https://github.com/AFLplusplus/AFLplusplus/blob/773baf9391ff5f1793deb7968366819e7fa07adc/src/afl-fuzz-init.c#L2791

With FRIDA mode, one of the options is to configure `afl-frida-trace.so` using JS bindings. This includes the configuration of 
persistent mode. But obviously this code runs within `afl-frida-trace.so` and so after the constructor has run. This patch defers
the reading of this environment variable until it is actually needed.


instrumentation/afl-compiler-rt.o.c
```diff
@@ -150,6 +146,34 @@ u32 __afl_already_initialized_second;
 
 static int __afl_dummy_fd[2] = {2, 2};
 
+/* Running in persistent mode? */
+
+static inline u8 __afl_is_persistent(void) {
+
+  static u8 is_initialized = 0;
+  static u8 is_persistent = 0;
```
vanhauser-thc: please keep is_persistent as a global static, not a local static, just to be sure this does not destroying other uses.

instrumentation/afl-compiler-rt.o.c
```diff
@@ -150,6 +146,34 @@ u32 __afl_already_initialized_second;
...
+
+  if (is_initialized) { return is_persistent; }
```
vanhauser-thc: this needs to be `if (likely(is_initialized)) ...`

vanhauser-thc:
how do the changes in afl-compiler-rt.o.c change the assembly?
can you please paste it for x86_x64?

and of course - why is that change needed?
EDIT - rephrasing that - why do you not do a __afl_set_persistent_mode(u8 mode)... and set it via frida to 0 or 1 as you like?

instrumentation/afl-compiler-rt.o.c
WorksButNotTested: I’m going to change the approach like you suggest below, so this shouldn’t be relevant any more.

instrumentation/afl-compiler-rt.o.c
WorksButNotTested: I’m going to change the approach like you suggest below, so this shouldn’t be relevant any more.

WorksButNotTested:
> how do the changes in afl-compiler-rt.o.c change the assembly? can you please paste it for x86_x64?
> 
> and of course - why is that change needed? EDIT - rephrasing that - why do you not do a __afl_set_persistent_mode(u8 mode)... and set it via frida to 0 or 1 as you like?

Good idea. This keeps the change much more minimal.

vanhauser-thc:
can this be merged or are you still working on it?

WorksButNotTested:
All good to merge if you’re happy with it.

WorksButNotTested:
Just added another little commit on the top. Just a FRIDA version bump with a few very minor tweaks.

PR #1168, opened 2021-11-19
fork server: restore SIGCHLD handler for child process

kcwu:


vanhauser-thc:
uh, right, it was already too late yesterday ... thanks ...

PR #1169, opened 2021-11-19
Add llvm 14 compat

martinetd:
add/hasAttribute(s) were renamed add/hasAttribute(s)AtIndex in llvm 14

Note I didn't actually get to test this yet and just fixed build error for now, but the comment string in the header is the same so these are most likely drop in replacements.
In particular I don't see `hasAttribute(-1, ...)` as -1 being a valid index meaning anywhere? So it might make sense to convert this one to hasAttrSomewhere if that's what was intended, but I honestly have no idea what any of this does so I'll leave that to you.

### DIFF
--- a/instrumentation/afl-llvm-lto-instrumentation.so.cc
+++ b/instrumentation/afl-llvm-lto-instrumentation.so.cc
@@ -242,7 +242,11 @@ bool AFLLTOPass::runOnModule(Module &M) {
 
     // the instrument file list check
     AttributeList Attrs = F.getAttributes();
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
this is fixed in dev for a few weeks now. as there are bigger changes in dev which still need some riping this will take some more time until we push to stable. :)
EDIT: but thank you for your effort. in the future check dev for the current state of afl++

martinetd:
oh, why didn't I see this. Thanks, I'll give the dev branch a try instead!

PR #1185, opened 2021-11-28
Include cmake in the Dockerfile

eriktews:
The README contains an examples how afl can be used with cmake, but
cmake is not included in the Dockerfile. Thus, when running afl using
docker, one cannot invoke cmake from within the container.

This commit adds cmake to the list of packages in the Dockerfile that
are installed when the docker image is generated.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,8 @@ env NO_ARCH_OPT 1
 RUN apt-get update && \
     apt-get -y install --no-install-suggests --no-install-recommends \
     automake \
+    cmake \
+    meson \
     ninja-build \
     bison flex \
     build-essential \


vanhauser-thc:
that is true. however then we should also add meson and ninja. can you add them please?

eriktews:
> that is true. however then we should also add meson and ninja. can you add them please?

Of course, I will update the branch in a few minutes.

eriktews:
> that is true. however then we should also add meson and ninja. can you add them please?

I have added meson as well, ninja was already included in the Dockerfile.

vanhauser-thc:
ah yes, because ninja is used by qemu. thanks!

eriktews:
Thanks, do you also build and push the official dev branch images somewhere?

vanhauser-thc:
the stable branch is auto-generated and pushed to docker.io.
we push every 2-10 weeks from dev to stable.

PR #1186, opened 2021-11-29
[fix] Fix custom mutator trim bug

eternalsakura:
Custom mutator does not implement all three trim APIs, standard trimming will be used, but now the actual implementation does not match the description, fix this problem.


vanhauser-thc:
right, thank you!

PR #1191, opened 2021-12-02
Docs content - quality assurance

llzmb:


vanhauser-thc:
looks great so far!

llzmb:
Quick question: Is "never zero" still a thing? We deleted the README.neverzero.md and I cannot find anything in the docs.

> AFL++ performs "never zero" counting in its bitmap. You can read more about this
> here:
> * [instrumentation/README.neverzero.md](../instrumentation/README.neverzero.md)

Source: https://github.com/AFLplusplus/AFLplusplus/blob/3506077fd6f250f3c080b58ea3bae117c3b122da/docs/fuzzing_in_depth.md?plain=1#L159-L161

vanhauser-thc:
yes it is the default for most instrumentations. there are environment variables to disable it. 
Maybe it should be preserved... just add the original document as an extra content to README.llvm.md

llzmb:
It's a wrap! :)

PR #1192, opened 2021-12-04
Fix memory leak in libprotobuf-mutator-example

Kiprey:
Since `afl_custom_fuzz` do not free the memory of `out_buf` , there maybe a memory leak in example code.


vanhauser-thc:
thank you!

vanhauser-thc:
for the future: please send PRs to the dev branch only

PR #1205, opened 2021-12-11
rename active_paths

yuawn:
Rename `active_paths` to `active_items` for better readability.


yuawn:
`xxhash.h` still cannot pass code-format after code-format.
It is broken after twice code-format.

vanhauser-thc:
yes I am aware of the issue. just restore xxhash.h (eg. copy from our afl++ repo)

yuawn:
I resolved it in this way https://github.com/AFLplusplus/AFLplusplus/pull/1205/commits/602eafc223d76987e447b431fc75903147a40c38. It can now pass twice code format.
What do you think?

vanhauser-thc:
if it still works as before I am fine with it! :)

PR #1206, opened 2021-12-11
afl-cc: fix incorrect CLANGPP_BIN

Mindavi:
A minor bug slipped in on a probably not-often used path.

This caused CMake to complain that the c compiler was set to C++ mode,
causing the CMake configuration step to fail for all c targets.

aflplusplus was built with

```shell
make source-only -j8 LLVM_BINDIR="" AFL_REAL_LD=`which ld.lld`
```

(I'm setting LLVM_BINDIR to the empty string since it's making assumptions that don't hold on nixpkgs, and getting the binaries from PATH seems to work better).


vanhauser-thc:
true, thank you!

PR #1224, opened 2021-12-20
Update grammar_ref.sh

CityOfLight77:
now can fetch from dev branch


domenukk:
You can already just use the commit hash, no need to specify a branch

CityOfLight77:
> You can already just use the commit hash, no need to specify a branch

Should I close this PR?

vanhauser-thc:
the correct fix is to get the dev state to stable of the Grammar-Mutator, just did that. :) (and updated the ref normally)

PR #1227, opened 2021-12-23
Frida

WorksButNotTested:
AARCH64 coverage instrumentation optimization.

### DIFF
--- a/frida_mode/src/instrument/instrument_arm64.c
+++ b/frida_mode/src/instrument/instrument_arm64.c
@@ -1,50 +1,121 @@
+#include <stddef.h>
+
 #include "frida-gumjs.h"
 
 #include "config.h"
 
 #include "instrument.h"
+#include "ranges.h"
+#include "stalker.h"
+#include "util.h"
+
+#define G_MININT33 ((gssize)0xffffffff00000000)
+#define G_MAXINT33 ((gssize)0x00000000ffffffff)
+
+#define PAGE_MASK (~(GUM_ADDRESS(0xfff)))
+#define PAGE_ALIGNED(x) ((GUM_ADDRESS(x) & PAGE_MASK) == GUM_ADDRESS(x))
 
 #if defined(__aarch64__)
 
-static GumAddress current_log_impl = GUM_ADDRESS(0);
+__attribute__((aligned(0x1000))) static guint8 area_ptr_dummy[MAP_SIZE];
 
-static const guint8 afl_log_code[] = {
+  #pragma pack(push, 1)
+typedef struct {
 
-    // __afl_area_ptr[current_pc ^ previous_pc]++;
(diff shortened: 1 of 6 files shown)


frida_mode/src/instrument/instrument_arm64.c
```diff
@@ -1,101 +1,221 @@
...
 
-static GumAddress current_log_impl = GUM_ADDRESS(0);
+__attribute__((aligned(0x1000))) static guint8 area_ptr_dummy[MAP_SIZE];
+
+  #pragma pack(push, 1)
+typedef struct {
+
+  // cur_location = (block_address >> 4) ^ (block_address << 8);
+  // shared_mem[cur_location ^ prev_location]++;
+  // prev_location = cur_location >> 1;
```
flyingmutant, line 27: @WorksButNotTested this looks to revert the `ROR` logic (`// previous_pc = current_pc ROR 1;`) from 94999782f1a3742e3e755a66f5d76e84573ae6ef that is described in the [MapDensity.md](https://github.com/WorksButNotTested/AFLplusplus/blob/376d1736a8f9de28e76c739c448ada2fb4fcb5a7/frida_mode/MapDensity.md), was this intended?

frida_mode/src/instrument/instrument_arm64.c
```diff
@@ -1,101 +1,221 @@
...
+  code.code.mov_x0_curr_loc |= area_offset << 5;
 
-  }
+  instrument_patch_ardp(
+      &code.code.adrp_x1_area_ptr,
+      code_addr + offsetof(afl_log_code, code.adrp_x1_area_ptr),
+      GUM_ADDRESS(__afl_area_ptr));
+
+  map_size_pow2 = util_log2(__afl_map_size);
+  area_offset_ror = util_rotate(area_offset, 1, map_size_pow2);
```
WorksButNotTested, line 189: @flyingmutant Instead of the `ROR` operation being carried out within the inline assembly, the already rotated value is encoded into the instruction `mov_x1_curr_loc_shr_1`. This rotation can therefore be carried out at instrumentation time rather than execution time.

frida_mode/src/instrument/instrument_arm64.c
WorksButNotTested: It's a long time since I made this change, but I think the calculation remains the same. Please see the comment below.

frida_mode/src/instrument/instrument_arm64.c
flyingmutant: Thanks for the explanation! I've got confused by the code comments and instruction variable names.

WorksButNotTested: That's no problem. It's quite a complicated piece of code. If you have time to PR some additional comments or renaming then that would always be welcome.

PR #1229, opened 2021-12-27
Fix big-little support on Linux

WorksButNotTested:


vanhauser-thc:
can you please explain?

WorksButNotTested:
> can you please explain?

Does the comment in the PR suffice?

vanhauser-thc:
would I ask if that would be the case?
the PR is wrong to me, but I do not know what you try to achieve.

WorksButNotTested:
> would I ask if that would be the case?
> the PR is wrong to me, but I do not know what you try to achieve.

Sorry. Didn’t know if you’d seen it. I’m hoping that in the event you run a single instance of afl++ it will be bound to one of the big cores rather than a little core. As that is likely what the user would want?

vanhauser-thc:
replaced with 7bd2899

PR #1237, opened 2021-12-30
Fix LeakSanitizer Usage.

MegaManSec:
Previously, __lsan_do_leak_check() was run when using __AFL_LEAK_CHECK,
however this was the incorrect function to use. According to the
documentation: "Subsequent calls to this function will have no effect
and end-of-process leak check will not run".
This meant that if the memory did not leak on the first usage of
__AFL_LEAK_CHECK, subsquent calls to this macro would never do anything.

Likewise, it is not possible to use an LSAN suppression list with
symbolize=0, so instead __lsan_disable and __lsan_enable are used to
'ignore' certain memory allocations where needed.

### DIFF
--- a/docs/env_variables.md
+++ b/docs/env_variables.md
@@ -105,7 +105,8 @@ fairly broad use of environment variables instead:
       within your program at a certain point (such as at the end of an
       `__AFL_LOOP()`), you can run the macro  `__AFL_LEAK_CHECK();` which will
       cause an abort if any memory is leaked (you can combine this with the
-      `LSAN_OPTIONS=...` suppression option to suppress some known leaks).
+      `__AFL_LSAN_OFF();` and `__AFL_LSAN_ON();` macros to avoid checking for
+      memory leaks from memory allocated between these two calls.
     - `AFL_USE_MSAN=1` - activates the memory sanitizer (uninitialized memory)
     - `AFL_USE_TSAN=1` - activates the thread sanitizer to find thread race
       conditions
--- a/docs/fuzzing_in_depth.md
+++ b/docs/fuzzing_in_depth.md
@@ -201,7 +201,10 @@ The following sanitizers have built-in support in AFL++:
   security issue, but for developers this can be very valuable. Note that unlike
   the other sanitizers above this needs `__AFL_LEAK_CHECK();` added to all areas
   of the target source code where you find a leak check necessary! Enabled with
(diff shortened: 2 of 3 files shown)


MegaManSec:
Example code:
```
#include <stdlib.h>
#include <unistd.h>
#include <stdio.h>
__AFL_FUZZ_INIT();

char *do_some_config() {
	return malloc(100); //some random memory allocation which will never be de-allocated for some reason (i.e. deliberately leaked.)
}


void leak_mem(char *buf) {
	__AFL_LEAK_CHECK();
	printf("passed 1\n");
	free(buf);

	buf = do_some_config();
	__AFL_LEAK_CHECK();
	printf("passed 2\n");

	buf = do_some_config(); //This leaks memory. If LSAN_OFF/LSAN_ON have not been used, this should exit the program with errorcode 23.
	__AFL_LEAK_CHECK();

	printf("passed 3\n");
	free(buf);
}

// Due to __AFL_LSAN_OFF/__AFL_LSAN_ON being used, memory allocated by do_some_config in this case will not be checked for leaks.
void do_no(unsigned char *fbuf){

	__AFL_LSAN_OFF();
	char *buf = do_some_config();
	 __AFL_LSAN_ON();
	leak_mem(buf);
}

void do_yes(unsigned char *fbuf){
	char *buf = do_some_config();
	leak_mem(buf);
}


int main() {
	__AFL_INIT();
	unsigned char *buf = __AFL_FUZZ_TESTCASE_BUF;
	do_no(buf);
	do_yes(buf);
	return 0;
}
```
outputs the following:
```
passed 1
passed 2
passed 3

=================================================================
==11382==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 100 byte(s) in 1 object(s) allocated from:
    #0 0x408c35 in malloc (/root/a.out+0x408c35)
    #1 0x4280dc in do_some_config /root/test.c:7:9
    #2 0x428157 in leak_mem /root/test.c:16:8
    #3 0x428261 in do_no /root/test.c:33:2
    #4 0x428351 in main /root/test.c:45:2
    #5 0x7ffff6e22bf6 in __libc_start_main /build/glibc-S9d2JN/glibc-2.27/csu/../csu/libc-start.c:310

SUMMARY: LeakSanitizer: 100 byte(s) leaked in 1 allocation(s).
```

vanhauser-thc:
thank you :)

PR #1240, opened 2022-01-04
optimin: -A -> -H

adrianherrera:
Inline with afl-showmap change


vanhauser-thc:
uh yes, forgot about that, sorry.
is the fixing of optimin still on your todo list?

PR #1255, opened 2022-01-12
Edit SVG files

llzmb:
@domenukk and @stevenagy, thank you for the hint about the missing background!

@vanhauser-thc, thank you for the feedback on the content!


vanhauser-thc:
thank you!

PR #1256, opened 2022-01-13
Lpm patches

dmell:
While developing a custom mutator based on the example in the repo, I found out some minor issues in how it dynamically allocates/frees memory. This PR contains my proposed fixes.


vanhauser-thc:
honestly, this was copied from somewhere and at least me I never used it :)
if you say this is better, I think it is fine :)

PR #1259, opened 2022-01-15
Adjustments to unicorn docs and speedtest sample

dmell:
With this PR, I am proposing some adjustments to the docs about AFL++ in Unicorn mode.

While checking out the speedtest sample, I realized some of the commands provided were wrong, but also that the second part of the README seems to refer to a different example. I think the purpose of this one is to compare the speed of different harnesses against the same target, which is target.c in the folder - not simple_target. Before fixing this though, I wanted to check with the original author (@domenukk probably) if my understanding is indeed correct.

### DIFF
--- a/unicorn_mode/samples/speedtest/rust/Cargo.toml
+++ b/unicorn_mode/samples/speedtest/rust/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "unicornafl_harness"
+name = "unicornafl"
 version = "0.1.0"
 authors = ["Dominik Maier <domenukk@gmail.com>"]
 edition = "2018"
--- a/unicorn_mode/samples/speedtest/rust/Cargo.toml
+++ b/unicorn_mode/samples/speedtest/rust/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "unicornafl"
+name = "harness"
 version = "0.1.0"
 authors = ["Dominik Maier <domenukk@gmail.com>"]
 edition = "2018"
--- a/unicorn_mode/README.md
+++ b/unicorn_mode/README.md
@@ -9,7 +9,7 @@ The CompareCoverage and NeverZero counters features are by Andrea Fioraldi <andr
 ## 1) Introduction
 
 The code in ./unicorn_mode allows you to build the
-(Unicorn Engine)[https://github.com/unicorn-engine/unicorn] with AFL++ support.
+[Unicorn Engine](https://github.com/unicorn-engine/unicorn) with AFL++ support.
 This means, you can run anything that can be emulated in unicorn and obtain instrumentation
 output for black-box, closed-source binary code snippets. This mechanism
 can be then used by afl-fuzz to stress-test targets that couldn't be built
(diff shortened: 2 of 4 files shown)


vanhauser-thc:
@domenukk ?

unicorn_mode/samples/speedtest/rust/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "unicornafl_harness"
+name = "unicornafl"
```
domenukk: Why rename the package, here? I prefer the current name or just `harness`

dmell: Oops, my bad. Sure I agree with you, I wanted to rename it to `harness` like the other ones, but I edited the string wrong. Just fixed.

domenukk:
To answer your question @dmell it looks like the readme refers to an older version, or simply was a copy&paste fail, yes

dmell:
> To answer your question @dmell it looks like the readme refers to an older version, or simply was a copy&paste fail, yes

Alright, if you want I can clean it up. In the same file, there is also a TODO in the results: I tried to run the three harnesses for a few minutes, and the results are actually pretty surprising. It seems that the Python harness goes 2x faster than the Rust/C ones, which perform similarly. I would have expected the opposite, do you have any idea of why this is happening?

domenukk:
Yes feel free to edit the docs, thanks! :)

WRT python speeds, it sounds _very_ obscure.
When I measured it, python was slower so I spent some time optimising the bindings to keep up with the others, but it's very hard to believe it should be that much faster than C now!
How did you measure it? Did you bind the process to the cores correctly and all that?

In any case I think we should redo the speed tests on UC2 at some point cc @wtdcode

dmell:
I just repeated the experiment. I used the same commands as in the README, only adding `-b 0`. After around 10 minutes, the results are 11.2k exec/sec for C, 10.5k exec/sec for Rust and 22.1k exec/sec for Python. I have just committed the updated docs, leaving the TODO. If you can give it a try, let me know if you get different results :)

llzmb:
Hi @dmell,

thank you for your contribution to the AFL++ docs! You won one of the very rare AFL++ t-shirts (see https://twitter.com/aflplusplus/status/1480836318175408138)!

Please e-mail @vanhauser-thc (mh@mh-sec.de) your address, size, and style (straight or fitted) until this upcoming Friday, January 21st, 2022.

Congrats! 🥳

vanhauser-thc:
is this mergable? or still things to do?

dmell:
I fixed the docs related to the speedtest sample, now a user can try it out following the provided README without any tweak needed. There is still a TODO in the results section, I was waiting for a comment from @domenukk on those I get on my end, since they seem counterintuitive.

vanhauser-thc:
yes @domenukk please comment :)
@dmell can it be converted from a draft to a full PR? or is it 100% sure that there is still stuff to be done? background - we want to release monday/tuesday.

domenukk:
First of all, we just switched to unicorn 2, so numbers might be a bit higher overall (thanks @wtdcode )

For ideal tests, you probably want to add a fixed seed (i.e. `-s 123`), and redo the experiments 3 times - but it shouldn't change speed _that_ much.
If these are really the results, by all means feel free to add them (if you can do them with uc2)

It's for sure something we need to investigate into further as the compiled versions should be quicker, or, at least, roughly equal.
Maybe some optimisation or linker flags? Or shared mem fuzzing broke, or ...

domenukk:
@dmell We merged this for now to get the docs ready for release, if you still find the time to do the Speedtest, feel free to add a new PR thx :)

PR #1298, opened 2022-01-19
Changes to support variations in the mapsize

WorksButNotTested:


frida_mode/GNUmakefile
```diff
@@ -259,7 +259,7 @@ endif
 
 
 ############################## AFL #############################################
-$(AFL_COMPILER_RT_OBJ): $(AFL_COMPILER_RT_SRC)
+$(AFL_COMPILER_RT_OBJ): $(AFL_COMPILER_RT_SRC) $(ROOT)include/config.h
```
WorksButNotTested, line 262: Add dependency on header containing the MAP_SIZE macro

frida_mode/src/instrument/instrument_x64.c
```diff
@@ -216,6 +216,8 @@ static gboolean instrument_coverage_find_low(const GumRangeDetails *details,
   static GumAddress last_limit = (64ULL << 10);
   gpointer *        address = (gpointer *)user_data;
 
+  last_limit = GUM_ALIGN_SIZE (last_limit, __afl_map_size);
```
WorksButNotTested, line 219: Fix the alignment. The map should be aligned based on it's size.

frida_mode/src/instrument/instrument_x64.c
```diff
@@ -216,6 +216,8 @@ static gboolean instrument_coverage_find_low(const GumRangeDetails *details,
   static GumAddress last_limit = (64ULL << 10);
```
WorksButNotTested, line 216: We start looking from 64kb as that is a conservative value for the size of the zero page.

frida_mode/src/instrument/instrument_x64.c
```diff
@@ -326,7 +328,7 @@ void instrument_coverage_optimize_init(void) {
   FVERBOSE("Low address: %p", low_address);
 
   if (low_address == 0 ||
-      GPOINTER_TO_SIZE(low_address) > ((2UL << 20) - __afl_map_size)) {
+      GPOINTER_TO_SIZE(low_address) > ((2UL << 30) - __afl_map_size)) {
```
WorksButNotTested, line 331: We actually want our memory to be in the low 2GB of the address space. This was broken before.

frida_mode/src/instrument/instrument.c
```diff
@@ -318,12 +318,6 @@ void instrument_init(void) {
 
   if (instrument_unique) { instrument_tracing = TRUE; }
 
-  if (__afl_map_size != 0x10000) {
```
WorksButNotTested, line 321: Changing the map size isn't that scary after all!

PR #1315, opened 2022-01-29
afl fuzz enable getpgid on other platforms.

devnexen:


vanhauser-thc:
the macos CI failure is something different.
so it seems to work everywhere? nice! thank you!

PR #1318, opened 2022-01-30
Build docker images for arm64 as well as x86.

moyix:
This involves:

* A small change to the Dockerfile, so that it checks the target architecture and only installs gcc-multilib on x86_64.
* Updates to the GitHub Actions workflow build_aflplusplus_docker.yaml to build for both linux/amd64 and linux/arm64.

These changes should allow Mac M1 users to get a working version of AFL++ by doing `docker pull aflplusplus/aflplusplus`.

The only downside is that the Docker build now takes longer (1h 50m) because it runs the arm64 build inside QEMU.


vanhauser-thc:
I changed the base to dev because I realized this also updates Dockerfile, and the is also relevant for dev. so will push to stable tomorrow or so.
Thank you!

PR #1323, opened 2022-02-06
Add GCC hot attribute to hot functions

yuawn:
Add GCC hot attribute to `afl_fsrv_run_target()` and `afl_fsrv_write_to_testcase()` which be called frequently in fuzzing loop to optimize performance.


vanhauser-thc:
that is useful, thanks!

PR #1334, opened 2022-02-22
Fix compilation against LLVM 14

Teemperor:
This is just compilation fixes for LLVM 14 which comes out in less than a month. Also some people such as me are already using it since a while and currently can't compile afl++ because of the API breakages, so this PR should rectify that.

Note: This isn't doing any pass porting the New PM. IIUC LLVM 14 is the last release with the old PM.

Some notes:

* `isGEPWithNoNotionalOverIndexing` has been removed without replacement in LLVM commit ed0cdb29397ecd7d03579b846360906081d80aea . From my basic reading of the source our intention was just to check if we have a GEP? I'm anyway not sure if our logic there is super sound, so I might be wrong here.
* `-flegacy-pass-manager` is an alias since LLVM 10. `-fno-experimental-new-pass-manager` was removed in LLVM 14, but `-flegacy-pass-manager` is still around for this release.

### DIFF
--- a/instrumentation/SanitizerCoverageLTO.so.cc
+++ b/instrumentation/SanitizerCoverageLTO.so.cc
@@ -17,6 +17,7 @@
 #include "llvm/Transforms/Instrumentation/SanitizerCoverage.h"
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/Triple.h"
 #include "llvm/Analysis/EHPersonalities.h"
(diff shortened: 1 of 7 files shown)


Teemperor:
I think I'm out of the loop about the status of the pass porting to the NewPM, but given that the respective branch seems stale I assume this stop-gap fix here is fine.

vanhauser-thc:
thank you!
I hope your patch didnt break older llvm versions...
the newpm2 branch is current, a 1/3 of the passes have been ported.

PR #1337, opened 2022-02-24
Fix build for the LLVM pass for LLVM >= 13

devnexen:
Due to change of inheritance for more modern LLVM versions, the overridable
members are not the same.


vanhauser-thc:
thank you. did the tshirt arrive?

devnexen:
> thank you. did the tshirt arrive?

Yes it fits fine ;-) thx.

Issue #1328, opened 2022-02-13
AFL_TARGET_ENV does not play nice with QEMU mode


adamdoupe:
**Describe the bug**
When using `AFL_TARGET_ENV` to set the environment for the target binary, when using `afl-fuzz -Q` in QEMU mode, the `AFL_TARGET_ENV` environment applies to QEMU and the target binary, not just the target binary.

This is important when fuzzing a target binary in QEMU mode using specific versions of libraries that will cause QEMU to crash.

**To Reproduce**
1. Fuzz a binary with local library that is different from the system, something like: `AFL_TARGET_ENV="LD_LIBRARY_PATH=./lib" afl-fuzz -Q -- ./binary`
2. Note that QEMU will pick up the libraries in `./lib` and crash if the libraries are not compatible with QEMU.

**Expected behavior**
`afl-fuzz` should either translate the `AFL_TARGET_ENV` into `-E` options to QEMU or set the `QEMU_SET_ENV` environment variable. One complication is that the format of these are different, `QEMU_SET_ENV` and `-E` expect a comma separated list while `AFL_TARGET_ENV` is a space separated list.

Another option is to update the documents in [env_variables.md](https://github.com/AFLplusplus/AFLplusplus/blob/stable/docs/env_variables.md) to make it clear that `AFL_TARGET_ENV` will apply to `QEMU` and the target binary, and that in this case users should use `QEMU_SET_ENV`. Although this will require more work on the part of users.

**Proposed solution**
I am happy to create a PR for either of the two options presented previously, if the AFL++ team can tell me which is the preferred approach.

vanhauser-thc:
That is not a bug but as things ought to be.
Note that qemu also has env vars to set for a target so in your case use these.

vanhauser-thc:
so to clarify, if you want to send a PR for env.md to better explain the limitations, that would be nice

domenukk:
This seems to be done

PR #1338, opened 2022-02-26
Clarify in docs that AFL_TARGET_ENV will apply to QEMU in QEMU mode

adamdoupe:
As discussed in #1328, it can be confusing to someone using QEMU mode to have `AFL_TARGET_ENV` apply to QEMU and not the binary itself.

This commit updates the docs to clarify that when using QEMU mode, the user might want to use `QEMU_SET_ENV`.


adamdoupe:
Sorry, I realized that I made this against the wrong branch, going to close and resubmit.

Issue #1341, opened 2022-03-01
Documentation: libdesock to fuzz network servers


martinclauss:
In https://github.com/AFLplusplus/AFLplusplus/blob/stable/docs/fuzzing_in_depth.md#k-known-limitations--areas-for-improvement you mention preeny to support the fuzzing of network servers. Check out https://github.com/fkie-cad/libdesock and maybe add it as an additional reference to support network fuzzing with AFL++. Thanks! :)

vanhauser-thc:
feel free to send a PR :)

PR #1342, opened 2022-03-01
Update fuzzing_in_depth.md

martinclauss:
add libdesock as an additional way to ease the fuzzing for network servers (#1341)


vanhauser-thc:
thank you!

Issue #944, opened 2021-05-28
Writing a custom post processor


r00tus3r:
Hi, I also have a question regarding the custom post processor. 

In the example https://github.com/AFLplusplus/AFLplusplus/blob/stable/custom_mutators/examples/post_library_gif.so.c#L48 it is mentioned that if the post process function returns 0 the input is not sent to the target program and fuzzing is continued. When I tried to do that same the fuzzer throws the error: `Custom_post_process failed (ret: 0)`

I guess that example is outdated and the latest one is https://github.com/AFLplusplus/AFLplusplus/blob/stable/custom_mutators/examples/example.c.

Also, I found that https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-fuzz-run.c#L189 expects the return value to be 0 only when there is an error.

Is there any other way I could make the post processor skip an unwanted test case and continue fuzzing?

vanhauser-thc:
yes the post process example is outdated. 0 is an illegal return.
I see the point that it can make sense to instead of "post processing" just to do "input verification" and then skip on the skip if unworthy.
If you want you can send a PR that skips the fuzz attempt instead of erroring.

ShubhamPalriwala:
@vanhauser-thc Can you elaborate more on this please? I'm new to the project and would like to try AFL++ out

vanhauser-thc:
@ShubhamPalriwala there are examples in custom_mutators/examples
what do you want to do`

1984gobrr:
Hi, I would like to work on this issue :smiley:

domenukk:
https://github.com/AFLplusplus/AFLplusplus/blob/1d4f1e48797c064ee71441ba555b29fc3f467983/src/afl-fuzz-run.c#L208

Instead of exiting here for `== 0`, we could instead (somehow) skip this testcase completely.

vanhauser-thc:
@1984gobrr just send a PR :)

VARUN-009:
I wanna work on this issue. Please assign this to me and since I'm new to this repo so I request you to please guide me through

vanhauser-thc:
implemented.

PR #1351, opened 2022-03-16
Fix post processor skip test case if return 0

1984gobrr:
https://github.com/AFLplusplus/AFLplusplus/issues/944

### DIFF
--- a/src/afl-fuzz-run.c
+++ b/src/afl-fuzz-run.c
@@ -97,8 +97,8 @@ write_to_testcase(afl_state_t *afl, void *mem, u32 len, u32 fix) {
   if (unlikely(afl->custom_mutators_count)) {
 
     ssize_t new_size = len;
-    u8 *    new_mem = mem;
-    u8 *    new_buf = NULL;
+    u8     *new_mem = mem;
+    u8     *new_buf = NULL;
 
     LIST_FOREACH(&afl->custom_mutator_list, struct custom_mutator, {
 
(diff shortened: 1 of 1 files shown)


src/afl-fuzz-run.c
```diff
@@ -205,14 +205,17 @@ static void write_with_gap(afl_state_t *afl, u8 *mem, u32 len, u32 skip_at,
         new_size =
             el->afl_custom_post_process(el->data, new_mem, new_size, &new_buf);
 
-        if (unlikely(!new_buf || new_size <= 0)) {
+        if (unlikely(!new_buf)) { FATAL("Custom_post_process failed"); }
```
vanhauser-thc, line 208: I would keep this as is and just check for "new_size < 0" instead of "<=".

src/afl-fuzz-run.c
```diff
@@ -205,14 +205,17 @@ static void write_with_gap(afl_state_t *afl, u8 *mem, u32 len, u32 skip_at,
...
 
-          FATAL("Custom_post_process failed (ret: %lu)",
-                (long unsigned)new_size);
+        if (unlikely(new_size <= 0)) {
 
-        }
+          LIST_REMOVE_CURRENT_EL_IN_FOREACH()
```
vanhauser-thc, line 212: why remove? and why remove if the size is 0 (which is a legal case, that is what this PR should be about)

vanhauser-thc:
beside the two comments I left another thing that is missing is that afl-fuzz has to be instructed to skip the testcase if the size is 0. this is what this PR has to be about.
thank you for the effort so far.

1984gobrr:
Hummm crap, I misunderstood the issue.
I will make an other PR to try to fix it, sorry for that :face_in_clouds:

PR #1355, opened 2022-03-18
docs: Document linker scripts change required in some projects

rwmjones:
AFL instrumentation modifies the binary to export extra symbols which
must be visible if the main binary dlopen()s an instrumented DLL.
This change was required by nbdkit since it uses a linker script to
limit the symbols exported from the main binary:

https://gitlab.com/nbdkit/nbdkit/-/commit/355cacb812126f5e37a2468f6939821b15017373

Signed-off-by: Richard W.M. Jones <rjones@redhat.com>


vanhauser-thc:
I think a better place for this would be the FAQ, but I will merge it.
btw we only accept PRs to the dev branch, switched it for you.

PR #1362, opened 2022-03-23
push to stable

vanhauser-thc:
getting the new pass manager out there and fixing unicorn issues


docs/Changelog.md
```diff
@@ -12,6 +12,9 @@ sending a mail to <afl-users+subscribe@googlegroups.com>.
   - fix */build_...sh scripts to work outside of git
   - new custom_mutator: libafl with token fuzzing :)
   - afl-fuzz:
+    - when you just want to compile once and set CMPLOG, then just
```
domenukk, line 15: I feel like that could be default, and the binary (and other instrumentations) could tell afl-fuzz it supports inline cmplog on handshake?
Different binaries to get more perf is an advanced use-case

docs/fuzzing_in_depth.md
```diff
@@ -817,9 +835,9 @@ Here are some of the most important caveats for AFL++:
 
 - There is no direct support for fuzzing network services, background daemons,
   or interactive apps that require UI interaction to work. You may need to make
-  simple code changes to make them behave in a more traditional way. Preeny may
+  simple code changes to make them behave in a more traditional way. Preeny or libdesock may
   offer a relatively simple option, too - see:
-  [https://github.com/zardus/preeny](https://github.com/zardus/preeny)
+  [https://github.com/zardus/preeny](https://github.com/zardus/preeny) or [https://github.com/fkie-cad/libdesock](https://github.com/fkie-cad/libdesock)
```
domenukk, line 840: We could port over FitM QEMU ideas/network emu at some point in time or hint at it here (https://github.com/fgsect/FitM)

instrumentation/SanitizerCoveragePCGUARD.so.cc
```diff
@@ -992,6 +1011,15 @@ bool ModuleSanitizerCoverage::InjectCoverage(Function &             F,
                   FixedVectorType::get(Int32PtrTy, elements);
               Value *x, *y;
 
+              if (!FunctionGuardArray) {
```
domenukk, line 1014: @vanhauser-thc `scan-build` complained that `FunctionGuardArray` may be set to `null` here, maybe either fixing the root cause or exiting the build here could be a better solution (and at two other locations)

domenukk (approved):

docs/Changelog.md
vanhauser-thc: hmm that could be an interesting option indeed. and how to tell afl-fuzz not to enable cmplog then (if the target has cmplog compiled in)?

docs/fuzzing_in_depth.md
vanhauser-thc: that would be cool and could be then also applied to libafl right?

instrumentation/SanitizerCoveragePCGUARD.so.cc
vanhauser-thc: these are false positives in scan-build. it is impossible that FunctionGuardArray is NULL, otherwise the function would not be called ... I think better than adding that if (!...) rather put a scan-build comment that this is a false positive? (sure there is a way to flag these)

domenukk: I prefer the if, the flow looks sane I think so there is a chance this might happen, and if not, it doesn't hurt

docs/fuzzing_in_depth.md
domenukk: LibAFL would be a perfect fit for this, yes, but it needs time and someone to do it

docs/Changelog.md
domenukk: F0r example `AFL_AUTO_CMPLOG=0` or `-c -` or something similar?

vanhauser-thc: yeah maybe ... lets put that on the TODO list

docs/fuzzing_in_depth.md
vanhauser-thc: I mean the patch to qemu could be used for both (or maybe the fitm qemu changes are already generic and dont need any changes?)

domenukk: For LibAFl it'd be cleaner to reimplement the FitM changes on top of the LibAFL's syscall API imho

PR #1363, opened 2022-03-23
fix installation packages commands for debian 10

clchrlls:
Hi !
I tried to install afl++ on a fresh install of debian 10 and the command listed in INSTALL.md to install gcc failed. I think it might be due to a change in the way gcc report its version number. So I just added a small section for Debian 10 to INSTALL.md so the instructions work. 
I know the documentation cannot reflect all the small differences accross all distributions so feel free to close the pull request if you think this fix is easy enough for users to figure out by themselves and not needing an edit to the installation instructions.
Thanks for reading and for maintaining afl++ !

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -1,4 +1,4 @@
-# Building and installing AFL++
+
 
 ## Linux on x86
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
the only thing that needed changing was the `gcc --version|head -n1|sed 's/\..*//'|sed 's/.* //'` line right?
that line works for me too, so you can simply just change the line where this is used twice.

clchrlls:
Yeah, only this line changed. 
I just didn't have another version of linux to try it on so I didn't want just to edit the line which could break the command on every other distribution and version. Also, I didn't know how to make it clear in the code block directly only one of the two version of the command should be executed depending on the system (provided they each fail on some systems). 
But I'll just change the line in the code block then.

vanhauser-thc:
thank you!

PR #1376, opened 2022-04-05
[wontfix] make error @SIMPLE_FILES

0x39f4:
Hi there!
I tried to compile with SIMPLE_FILES, but got  an error message.
Please have a look through it.
Sincerely.

```
src/afl-fuzz-bitmap.c: In function ‘save_if_interesting’:
src/afl-fuzz-bitmap.c:706:39: error: ‘afl_state_t’ {aka ‘struct afl_state’} has no member named ‘last_kill_signal’
  706 |                afl->saved_crashes, afl->last_kill_signal);
      |                                       ^~
make: *** [GNUmakefile:437: afl-fuzz] Error 1
```


vanhauser-thc:
ah good find, that is a feature that is rarely used by anyone so things like this easily happen ...

PR #1379, opened 2022-04-05
network client mem leak fix.

devnexen:


vanhauser-thc:
IMHO these are not memleaks as the program exit cleans that up and that is the only thing between those free and the exit ... but cant hurt :)

PR #1381, opened 2022-04-06
add make flag to avoid building nyx mode

f0rki:
I don't use nyx, but whenever I do `make source-only` it takes ages to build nyx, which is kind of annoying, so I updated the Makefile s.t. I can do `make source-only NO_NYX=1`. Maybe this is useful for others?

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -625,7 +625,9 @@ ifeq "$(ARCH)" "aarch64"
 	-$(MAKE) -C coresight_mode
 endif
 ifeq "$(SYS)" "Linux"
+ifndef NO_NYX
 	-cd nyx_mode && ./build_nyx_support.sh
+endif
 endif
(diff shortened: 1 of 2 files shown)


domenukk:
@vanhauser-thc should nyx be part of `source-only` anyway? It can _kinda_ be used for source-based fuzzing, but I doubt it's a common usecase

vanhauser-thc:
@domenukk for larger targets it is beneficial, and it is for source-only targets too so it has to be built there too. otherwise "source-only" would need to be renamed to "most-source-only" :)

I am all for this PR! can you please add the option to the GNUmakefile help output and also add it to docs/INSTALL.md? thank you!

f0rki:
@vanhauser-thc done.

vanhauser-thc:
thank you!

PR #1382, opened 2022-04-07
Fix GCC plugin crash when using deny/allow list

dkm:
The provided function declaration F may not have valid location
information. Return an empty string in this case as the two callers are
already using this convention to filter out functions from being
instrumented when deny/allow list are used.

### DIFF
--- a/instrumentation/afl-gcc-pass.so.cc
+++ b/instrumentation/afl-gcc-pass.so.cc
@@ -714,9 +714,11 @@ struct afl_pass : gimple_opt_pass {
 
   }
 
+  /* Returns the source file name attached to the function declaration F. If
+     there is no source location information, returns an empty string.  */
   std::string getSourceName(function *F) {
 
-    return DECL_SOURCE_FILE(F->decl);
+    return DECL_SOURCE_FILE(F->decl) ? DECL_SOURCE_FILE(F->decl) : "";
 
   }
 


instrumentation/afl-gcc-pass.so.cc
```diff
@@ -714,9 +714,11 @@ struct afl_pass : gimple_opt_pass {
 
   }
 
+  /* Returns the source file name attached to the function declaration F. If
+     there is no source location information, returns an empty string.  */
   std::string getSourceName(function *F) {
 
-    return DECL_SOURCE_FILE(F->decl);
+    return DECL_SOURCE_FILE(F->decl) ? "" : DECL_SOURCE_FILE(F->decl);
```
domenukk: I didn't look into it deeply but isn't this flipped? If there is a file, it returns empty string?

hexcoder- (approved):
Thanks very much!

instrumentation/afl-gcc-pass.so.cc
dkm: You are correct, let me fix that and check why this has not been spotted before!

vanhauser-thc:
on a different topic, how is the cmplog/redqueen implementation coming along?

vanhauser-thc:
looks better :) I am curious though, if the fix actually did not fix it but make it worse, how was did not found?
@domenukk well spotted!

dkm:
> looks better :) I am curious though, if the fix actually did not fix it but make it worse, how was did not found? @domenukk well spotted!

It was found during some investigation that this feature is crashing the plugin, but this has not been used (yet?). Clearly, it is lacking some testing, which I hope to be able to do later.

As for the cmplog/redqueen, I'll ask my colleague about the status :)

vanhauser-thc:
please give a heads up after this has been tested then we will merge. thank you!

dkm:
Thanks for merging this. I am still planning on testing this correctly, but have been busy with other tasks. As for the cmplog/redqueen, there's not much to be shared at the moment but we'll keep you updated when we have something.

PR #1389, opened 2022-04-14
update llvm version in INSTALL.md

yuawn:


vanhauser-thc:
depends on ubuntu/debian version, but yeah this should encourage to update :)

PR #1390, opened 2022-04-15
libdislocator, new option to name an anonymous address range.

devnexen:
For performance matter tough, disabled by default.

### DIFF
--- a/utils/libdislocator/Makefile
+++ b/utils/libdislocator/Makefile
@@ -19,9 +19,10 @@ HELPER_PATH  = $(PREFIX)/lib/afl
 VERSION     = $(shell grep '^\#define VERSION ' ../../config.h | cut -d '"' -f2)
 
 CFLAGS      ?= -O3 -funroll-loops -D_FORTIFY_SOURCE=2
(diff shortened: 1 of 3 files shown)


utils/libdislocator/Makefile
```diff
@@ -19,9 +19,10 @@ HELPER_PATH  = $(PREFIX)/lib/afl
 VERSION     = $(shell grep '^\#define VERSION ' ../../config.h | cut -d '"' -f2)
 
 CFLAGS      ?= -O3 -funroll-loops -D_FORTIFY_SOURCE=2
-override CFLAGS += -I ../../include/ -Wall -g -Wno-pointer-sign
+CFLAGS += -I ../../include/ -Wall -g -Wno-pointer-sign
```
vanhauser-thc, line 22: why removing override? if CFLAGS are given then compilation will fail, right? isntt config.h etc. used?

vanhauser-thc: ah sorry you are right its not needed, good catch.

vanhauser-thc:
looks useful, thank you

PR #1393, opened 2022-04-20
instrumentation fixes for wasm

jschwartzentruber:
Something was missed across merges such that `afl-clang-fast` broke wasm. Use the passthrough flag instead so we really don't touch wasm.

Also fix `-Wunused-but-set-variable`. Attribute `used` is so `_B` isn't optimized out. `unused` is to avoid the warning.


vanhauser-thc:
thank you!

PR #1395, opened 2022-04-28
Properly null CmpLog map in deinit 

che30122:
When deinit cmplog shared memory.
shm->cmp_map should be NULL?


domenukk:
Thanks!

PR #1401, opened 2022-05-06
large shared segment optimisation (in TEST_MMAP mode)

devnexen:
for FreeBSD supporting such feature.

Grabbing the first kind which fit the needed size otherwise
falling back to the classic shared segment allocation.


vanhauser-thc:
thank you!

PR #1408, opened 2022-05-12
update nyx submodules

schumilo:
This PR basically bumps up all Nyx submodules to the latest stable commits. With this PR we also fix a bug in Nyx that can occur in vanilla KVM mode.


domenukk:
Sweet

PR #1412, opened 2022-05-16
Fix missing dependency

HexRabbit:
Add `instrumentation/afl-llvm-common.o` to `SanitizerCoverageLTO.so`'s dependency list


vanhauser-thc:
thank you! please note for the future to direct the PRs to dev not stable.

HexRabbit:
Oh, I didn't notice that, thanks!

PR #1415, opened 2022-05-17
afl->cmplog_binary memory leak problem

che30122:
afl->cmplog_binary memory should be free when afl deinit to avoid memory leak.


vanhauser-thc:
thank you. please push to the dev  branch in the future, fixed that for you.

PR #1425, opened 2022-05-24
Add AFL_SYNC_TIME variable for synchronization time tuning

apach301:
Hi!

I'm trying to combine AFL++ with symbolic engine for the sake of hybrid fuzzing. One of the metrics I track is imported path - number of generated inputs by symbolic engine that have proven to be useful for fuzzer. I found out that time of synchronization between fuzzing instances is set in `include/config.h` to 30 minutes by default. The value can be tuned only by re-compiling AFL++, so I decided it would be cool to make this parameter tweakable via environment variable.

### DIFF
--- a/docs/env_variables.md
+++ b/docs/env_variables.md
@@ -517,6 +517,10 @@ checks or alter some of the more exotic semantics of the tool:
     (empty/non present) will add no tags to the metrics. For more information,
     see [rpc_statsd.md](rpc_statsd.md).
 
+  - `AFL_SYNC_TIME` allows you to specify a different minimal time (in minutes)
+    between fuzzing instances synchronization. Default sync time is 30 minutes,
+    note that time is halfed for -M main nodes.
+
   - Setting `AFL_TARGET_ENV` causes AFL++ to set extra environment variables for
     the target binary. Example: `AFL_TARGET_ENV="VAR1=1 VAR2='a b c'" afl-fuzz
     ... `. This exists mostly for things like `LD_LIBRARY_PATH` but it would
--- a/src/afl-fuzz.c
+++ b/src/afl-fuzz.c
@@ -295,6 +295,7 @@ static void usage(u8 *argv0, int more_help) {
       "AFL_STATSD_TAGS_FLAVOR: set statsd tags format (default: disable tags)\n"
       "                        Supported formats are: 'dogstatsd', 'librato',\n"
       "                        'signalfx' and 'influxdb'\n"
+      "AFL_SYNC_TIME: sync time between fuzzing instances (in minutes)\n"
       "AFL_TESTCACHE_SIZE: use a cache for testcases, improves performance (in MB)\n"
       "AFL_TMPDIR: directory to use for input file generation (ramdisk recommended)\n"
       "AFL_EARLY_FORKSERVER: force an early forkserver in an afl-clang-fast/\n"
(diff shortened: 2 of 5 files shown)


vanhauser-thc:
looks useful, thanks.
can you please also make an entry to docs/envs.md?

docs/env_variables.md
```diff
@@ -517,6 +517,10 @@ checks or alter some of the more exotic semantics of the tool:
     (empty/non present) will add no tags to the metrics. For more information,
     see [rpc_statsd.md](rpc_statsd.md).
 
+  - `AFL_SYNC_TIME` allows you to specify a different minimal time (in minutes)
+    between fuzzing instances synchronization. Default sync time is 30 minutes,
+    note that time is halfed for -M main nodes.
```
domenukk, line 522: Doesn't the extra halving make it more complex than needed? Maybe only do it (once) if the time is not set?

apach301: As far as I understand, the fuzzer logic is that main node should perform syncing more often than secondary nodes. There is no extra halving when specifying time via environment variable - the sync_time value is halved directly in afl-fuzz: https://github.com/AFLplusplus/AFLplusplus/blob/ba3c7bfe40f9b17a691958e3525828385127ad25/src/afl-fuzz.c#L2514

domenukk: Yes but is that really what a user wants if they specify the time manually?

apach301: I thought it should have same behaviour like when sync_time is set in config.h. So you think that when user specify sync_time manually then we should omit implicit logic with halving value and use same times? In that case, I suggest to replace the environment variable with command line option. Then it could be possible to manually set proper sync times for both kinds of instances. The default algorithm with halving time remains if there is no option provided.

domenukk: You can set env variables for each node as well. I don't think it's important enough to justify a parameter.
In any case I'm not fighting to change the current behavior if you are happy with it, just pointing out it may be surprising

apach301: Oh, right, environment is unique for each instance:) So I figured out how to change code to save its original behavior when no user sync_time specified and not perform implicit halving of time. It surely opens more possibilities for testing (user can explicitly define exact syncing time for every fuzzing instance).

src/afl-fuzz.c
```diff
@@ -2512,7 +2512,8 @@ int main(int argc, char **argv_orig, char **envp) {
         if (unlikely(afl->is_main_node)) {
 
           if (unlikely(get_cur_time() >
-                       (afl->sync_time >> 1) + afl->last_sync_time)) {
+                       (afl->sync_time ? afl->sync_time : SYNC_TIME >> 1) +
```
domenukk: If we always set the `sync_time` on startup, after the initial env is read, we'd get rid of this extra branch, what do you think?

docs/env_variables.md
domenukk: Nice

src/afl-fuzz.c
apach301: If I set `afl->sync_time` on startup then I should make SYNC_TIME/2 for main node (if no AFL_SYNC_TIME set). But there are situations when secondary node could become main during fuzzing process. In that case original behavior will be lost.

vanhauser-thc:
The halfing comes from me and is important :)

domenukk:
Sorry @apach301 Marc thinks your original pr was better, as in, the halving in the main node is a better default if people only skim over the documentation.
Want to roll back to the initial version? 🙈🙈🙈

apach301:
Sure, the simpler code the better :)

apach301:
Last changes are reverted

domenukk:
Thank you and sorry for the extra work 🙈

PR #1429, opened 2022-05-27
Fix wrong memchr size in android

cyanpencil:
Need to fix this otherwise ASAN will always complain about heap buffer overflows and refuse to run.


domenukk:
Thanks! Please open PRs against `dev` in the future :)

PR #1437, opened 2022-06-10
Fix Byte Decrement Havoc Mutation

Scepticz:
While looking at the source code of the havoc mutations I realized that there seems to be a typo / copy+paste error with the SUBBYTE_ mutation. It is currently incrementing, instead of decrementing the value.

Alternative Fix: Change the documentation to "/* Decrease byte by minus 1. */" to make it work as documented :-P


domenukk:
Thanks! :)
Please direkt PRs towards dev

vanhauser-thc:
thank you!

PR #1440, opened 2022-06-17
require value in env variables

rhtenhove:
This is a partial proposal for a change. Let me know what you think, and I'll add the changes listed below.

I came across this when trying to disable affinity in the container image, in which during the build `AFL_TRY_AFFINITY` is set to `1`. Setting that value to `0` or anything else doesn't do anything, as the code just checks if the variable exists (even when empty it'll return true).

The only way to unset it is using `unset AFL_TRY_AFFINITY` before `afl-fuzz`. A small thing to do, but not obvious, especially as the verbosity when starting is unclear, as it mentions the value but that value with booleans doesn't mean anything.  `[+] Loaded environment variable AFL_TRY_AFFINITY with value 0` still enables that function.

### TODO

- [ ] Have every boolean `afl->afl_env.*` check to be set to `1` else they should be false

### DIFF
--- a/src/afl-common.c
+++ b/src/afl-common.c
@@ -1249,3 +1249,4 @@ s32 create_file(u8 *fn) {
   return fd;
 
 }
+
--- a/src/afl-common.c
+++ b/src/afl-common.c
@@ -715,17 +715,23 @@ char *get_afl_env(char *env) {
 
   char *val;
 
-  if ((val = getenv(env)) != NULL) {
+  if ((val = getenv(env))) {
 
-    if (!be_quiet) {
+    if (*val) {
+      
+      if (!be_quiet) {
 
-      OKF("Loaded environment variable %s with value %s", env, val);
+        OKF("Loaded environment variable %s with value %s", env, val);
+
+      }
+
+      return val;
 
     }
 
   }
 
-  return val;
+  return NULL;
 
 }
 
@@ -1243,4 +1249,3 @@ s32 create_file(u8 *fn) {
   return fd;
 
 }
-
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
I would propose the change differently. if the variable is set it is "true", and if it is unset or set with an empty value then it is false. basically the change would be to ensure that the length of the value is larger than zero.
WDYT?

vanhauser-thc:
reason being that the proposed change would break scripts as the behaviour changes fundamentaly. but just ensure that the env var is also set is something most people would expect - like you.

rhtenhove:
> I would propose the change differently. if the variable is set it is "true", and if it is unset or set with an empty value then it is false. basically the change would be to ensure that the length of the value is larger than zero. WDYT?

That indeed sounds safer, and would work fine for me.

The environment variable output would still be confusing, as `[+] Loaded environment variable AFL_TRY_AFFINITY with value 0` would still mean enabled. I don't see a way to get the booleans separated from the other env vars to make that more clear. Any ideas?

Either way, I've added what you mentioned before. It's ready to merge regardless of getting a pretty output :smile:

vanhauser-thc:
I agree that setting a value with 0 is not what a new user would expect, but this is the original behaviour of vanilla afl, and would not want to change this as the chance of breaking some scripts somewhere is pretty high.

The PR is IMHO the best possible and would want to merge it unless you think a slightly different way would be better and want to discuss this.

rhtenhove:
Slept on it, and would like to propose the latest commit just made. Feels much clearer, and would provoke a new user to investigate what "enabled with 0" actually means, if any unexpected behavior is seen after seemingly disabling a variable with 0.

vanhauser-thc:
yep, that looks good :)

PR #1441, opened 2022-06-17
Allow to skip README.txt creation in crashes/

rhtenhove:
When automating processing of crash results, the README.txt is always an oddity to skip (for example, a similar file doesn't appear in any other directory which has significant meaning, such as `hangs` or `queue`).

This env variable allows the user to skip the creation of that file if so desired.


vanhauser-thc:
I totally agree on that one, also didnt understand the point of it :)

PR #1442, opened 2022-06-17
Push both stable and dev

rhtenhove:
Preferring Docker to try out AFL++, I'd like to also be able to run the dev branch easily.

It also makes more sense to have `dev` as the `latest`, and have a `stable` tag for those wanting, well... stable.
I'd understand if you'd disagree, and I would be fine with that too.

Furthermore, having no experience with github actions, please verify if these jobs are set correctly :sweat_smile:

### DIFF
--- a/.github/workflows/build_aflplusplus_docker.yaml
+++ b/.github/workflows/build_aflplusplus_docker.yaml
@@ -2,7 +2,9 @@ name: Publish Docker Images
 
 on:
   push:
-    branches: [ stable ]
+    branches:
+    - stable
+    - dev
 #    paths:
 #    - Dockerfile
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
I am not sure if the changes are correct or not ;)
if the dev one is called `latest` I guess by default this one would be grabbed by people automatically? that would be too risky IMHO. if the default if nothing is given can be set to the stable one somehow then this is fine, otherwise I would propose keeping stable as "latest" and putting dev to "dev" instead?

rhtenhove:
That should do it :smile: 

`latest` is indeed the same as pulling without a tag.

vanhauser-thc:
looks good, thank you!

vanhauser-thc:
that is weird I get the error

```
Invalid workflow file: .github/workflows/build_aflplusplus_docker.yaml#L23
You have an error in your yaml syntax on line 23
```
can you have a look? I have no time for this until monday or so :(

PR #1447, opened 2022-06-18
fixes github actions

rhtenhove:
This fixes the problem introduced by #1442 . Also updated the workflow versions.

I've also made the `:stable` tag explicit everywhere. That should make it more clear for every user what they're pulling. Also follows the recommendation that you should always use a tagged image. `:latest` still points to stable, so nothing changes for existing users.

So to keep with that idea, I've also added a push for tagged releases, so things like `aflplusplus/aflplusplus:4.00d` will work in the future.

Finally, as now everyone is hosting a registry, and some container daemons decided to not enforce `docker.io` as the default anymore, I've made them all fully qualified, i.e.: `docker.io/aflplusplus/aflplusplus:stable`.

WDYT ?

### DIFF
--- a/.github/workflows/build_aflplusplus_docker.yaml
+++ b/.github/workflows/build_aflplusplus_docker.yaml
@@ -5,6 +5,8 @@ on:
     branches:
       - stable
       - dev
+    tags:
+      - '*'
 
 jobs:
   push_to_registry:
@@ -37,3 +39,11 @@ jobs:
           push: true
           tags: aflplusplus/aflplusplus:${{ github.ref_name }},aflplusplus/aflplusplus:latest
         if: ${{ github.ref_name == 'stable' }}
+      - name: Publish tagged release to docker.io registry
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: aflplusplus/aflplusplus:${{ github.ref_name }}
+        if: ${{ github.ref_type == 'tag' }}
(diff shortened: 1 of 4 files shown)


vanhauser-thc:
lets try if it works :)

rhtenhove:
It works! :)

Currently fuzzing with `:dev`

vanhauser-thc:
hmm there is still something wrong, the dev one is not being built anymore:

  /usr/bin/docker pull tonistiigi/binfmt:latest
  Error response from daemon: Head "https://registry-1.docker.io/v2/tonistiigi/binfmt/manifests/latest": Get "https://auth.docker.io/token?account=githubactions&scope=repository%3Atonistiigi%2Fbinfmt%3Apull&service=registry.docker.io": EOF

maybe the tonistiigi docker registry is not a good one to use?

rhtenhove:
> hmm there is still something wrong, the dev one is not being built anymore:
> 
> /usr/bin/docker pull tonistiigi/binfmt:latest Error response from daemon: Head "https://registry-1.docker.io/v2/tonistiigi/binfmt/manifests/latest": Get "https://auth.docker.io/token?account=githubactions&scope=repository%3Atonistiigi%2Fbinfmt%3Apull&service=registry.docker.io": EOF
> 
> maybe the tonistiigi docker registry is not a good one to use?

I also see this in the Docker login action of the latest (cpubind) merge:

```
Error: Error response from daemon: Get "https://registry-1.docker.io/v2/": received unexpected HTTP status: 503 Service Unavailable
```

It seems Dockerhub had some downtime. Can you rerun the workflow?

The `tonistiigi` is the default image used by the QEMU action to emulate arm64. Best not to change that.

PR #1450, opened 2022-06-19
Update workflows to ubuntu 22.04

jrozner:
22.04 is the most recent LTS release and the official docker container
is running on it. It probably makes sense to run the unit tests on that
as well.


vanhauser-thc:
thank you!

PR #1453, opened 2022-06-21
improve CI and container image build

rhtenhove:
Attempting to get extra features working in the container, I made a few improvements to the build process:

1. GHA now also tests the resulting docker image before pushing it.
2. Pin gcc and llvm using an ARG (and remove gcc testing ppa, as it looks unmaintained)
3. ~make no longer silently fails (removed the `-` prefix), which uncovered some problems~ Dockerfile now removes `-` prefixes before build, only uncovering errors during CI container image build
4. Added Rust for nyx, but also disabled nyx support for now as that fails to build in the container (needs gcc 12.1 for the fix or  workarounds with unknown effects)
5. Disable gtk support for nyx if gtk3-devel is missing. Removes quite a few optional dependencies.
6. Various little shell quirks refactored. Let me know if any are unclear, I'll explain.

More additions:

7. Fixes but disables coresight build in container (it was being `.dockerignore`d, but has a build failure due to some errors in the source).
8. Fixes unicorn but disables it only for arm64 as the build is currently broken
9. Allows the container tests to fail, as various things currently fail, but does give direct insight in what's failing from the logs.
10. Building now takes longer, because more things successfully build. Especially the emulated arm64 build is slow (1+ hour...).

Last additions

11. ~Use concurrency to build container image. Uses ghcr.io registry for caching.~ Doesn't work with forks and their pull requests as there's no common registry usable.
12. Fix CodeQL job, which now CodeQL's way more, but also takes longer. Concurrency helps here.
13. Container now builds even more, so is larger, but is also made smaller again by having fewer and more efficient layers
14. Container jobs now run on pull requests to dev, so any additions are tested before going into dev
15. `dockerignore` even more, saving quite a lot of unused stuff (such as `.git`) 
16. Fix `clang` command availability
17. Add TEST_BUILD arg to Dockerfile so tests actually work as they don't get cleaned in CI testing


vanhauser-thc:
Thank you!
This looks all fine to me except:

>     3. make no longer silently fails (removed the `-` prefix), which uncovered some problems

this is bad. it expects that llvm, gcc plugin etc. are all present. afl++ is meant to build what it can in whatever environment it is, with clang or without, with internet connectivity or not etc.
your changes will fail if the system setup is not perfect.
that would be OK for docker, but not for anything else ...

rhtenhove:
That sounds like a valid reason, but the method seems to be mostly hiding problems during development.

I just gave it an offline try, and with all of the `-`s removed, but invoking make with `-i` (`--ignore-errors`), it behaves the same, i.e. missing clang but being fine with it, skipping over qemu, nyx, etc because it couldn't pull the git repos, etc.

Would that be a good alternative? It allows for the best of both worlds.

In an ideal world, offline systems would install a container daemon and run the container image :sunglasses:

vanhauser-thc:
> That sounds like a valid reason, but the method seems to be mostly hiding problems during development.

not for us, because the should CI catch most of that.

> I just gave it an offline try, and with all of the `-`s removed, but invoking make with `-i` (`--ignore-errors`), it behaves the same, i.e. missing clang but being fine with it, skipping over qemu, nyx, etc because it couldn't pull the git repos, etc.

but that would a user need to specify. for makefiles being called from the main GNUmakefile that would be ok for me, but the user should not need to specify this by hand.

rhtenhove:
The current CI job fails to build QEMU-Nyx (search for `(ignored)` in the [logs](https://github.com/AFLplusplus/AFLplusplus/runs/6978680929?check_suite_focus=true)) which does not show in the build result.

This succeeds in this new pipeline, as the requirement of gtk3 has been made optional (to which one of their developers tends to [agree](https://github.com/nyx-fuzz/QEMU-Nyx/issues/23)). It will now fail the pipeline next time if that build no longer succeeds for any reason, before it gets pushed to stable or dev.

I understand and agree with the requirement to make it as easy as possible for every user, and I would expect that user to copy-paste the instructions in the README, where it is trivial to add a `-i` here. This would remove the requirement of _forty_ `-`s which would need to be added to each new make instruction, and which cause for example the build failure of QEMU-Nyx to go unnoticed in CI (or even locally if not paying attention).

It also allows more advanced users to figure out what's built and what isn't, helping them fix their environment if they do want to have all the goodies, but need extra steps to get the right packages.

WDYT?

vanhauser-thc:
> The current CI job fails to build QEMU-Nyx (search for `(ignored)` in the [logs](https://github.com/AFLplusplus/AFLplusplus/runs/6978680929?check_suite_focus=true)) which does not show in the build result.

yes. not all is cought. but it makes not much sense - or needs a special setup to be able to use nyx from a docker container.

> I understand and agree with the requirement to make it as easy as possible for every user, and I would expect that user to copy-paste the instructions in the README

sadly this is not the case. people do not read the readme before building. I know, blame is on them, but we get the issues posted, discord questions etc. etc.

a more advanced user will check if anything failed, so I can turn that argument around :)
what we could do though is at the end of the build process in GNUmakefile to display a list of what is compiled and there and what is not. that could help.

vanhauser-thc:
I see in your last commit more of the -i and removal of the dash in front of make commands.

As I said the gmake -i needed to be used by a user gets a hard no by me. This just results in tons of issues being created and questions asked on the discord with adding any real value to users or us - people rarely read docs :(
 I already outlined better alternatives (better ci checks, makefile reports what could and could not be build) - but I don’t expect you to provide them. 

The j4 to nproc changes are obvious but we didn’t do them for a reason. Zero impact. Nothing is compiled in parallel although it should be. :-( we have never figured out why …

vanhauser-thc:
FYI we need to release a new version tomorrow, if this should be in that it should be done latest tonight.

rhtenhove:
Thanks for the heads up! I've been away for a few days, and still need to work out a few kinks in the PR. If it's not done before the release, it can wait until the next. Would rather have it working well next month than cause problems tomorrow.

About your remarks, I fully understand not wanting to have a barrage of the same question because of what 'should' be. I've reverted it and added a `sed` to do it in the Dockerfile. I believe everybody wins now.

I can't find anything summarizing a `(g)make` run, and making something custom here would need maintenance, so perhaps that's indeed best left as a manual task to the advanced user. They could use the same `sed` added to the `Dockerfile` before running `make`.

The container workflow should now build and test everything that can build and test on bare metal, and the time it now takes and the size of the resulting image it seems to do just that. Several things failing as well, but that's just revealing existing problems.


Reason for my interest is research into integrating fuzzing into CI. It's a world of its own, where tiny fuzzing campaigns (i.e. minutes) on limited/shared hardware using containers are of interest and seemingly useful indeed. For that, I like to have all the cool features of AFL++ in the container image, and this was an opportunity to make that possible and get to know (the limitations of) GitHub Actions in the meantime.

rhtenhove:
That should do it! Does need to be tested in `dev` with a push event, but no problems expected there. If there are, I'm available to fix.

vanhauser-thc:
> Thanks for the heads up! I've been away for a few days, and still need to work out a few kinks in the PR. If it's not done before the release, it can wait until the next. Would rather have it working well next month than cause problems tomorrow.

great to hear, then lets have that for the next. it is also more for container building and this is release independent.

> About your remarks, I fully understand not wanting to have a barrage of the same question because of what 'should' be. I've reverted it and added a `sed` to do it in the Dockerfile. I believe everybody wins now.

yes that sounds like a good idea

> I can't find anything summarizing a `(g)make` run, and making something custom here would need maintenance, so perhaps that's indeed best left as a manual task to the advanced user. They could use the same `sed` added to the `Dockerfile` before running `make`.

yeah maybe I will do that. i put it on my todo

> The container workflow should now build and test everything that can build and test on bare metal, and the time it now takes and the size of the resulting image it seems to do just that. Several things failing as well, but that's just revealing existing problems.

great!

> Reason for my interest is research into integrating fuzzing into CI. It's a world of its own, where tiny fuzzing campaigns (i.e. minutes) on limited/shared hardware using containers are of interest and seemingly useful indeed. For that, I like to have all the cool features of AFL++ in the container image, and this was an opportunity to make that possible and get to know (the limitations of) GitHub Actions in the meantime.

sounds like a win-win :)

rhtenhove:
The thing with fixed pipelines is that the [cracks start appearing](https://github.com/AFLplusplus/AFLplusplus/pull/1453/checks?check_run_id=7081556059) :stuck_out_tongue: 

@vanhauser-thc how would you want to approach these? I can make the job allow to fail, but that'll just hide the problems behind a green checkmark. You can also allow merging PRs with failing pipelines (I can't see if you allow that anyway).

vanhauser-thc:
I set those which do not belong to us to ignore and will then look at the rest. can you please do a dummy commit to restart it?

rhtenhove:
> I set those which do not belong to us to ignore and will then look at the rest. can you please do a dummy commit to restart it?

Done

vanhauser-thc:
can you please do another dummy commit? thanks

rhtenhove:
A not so dummy commit has been made.

It adds a formatting checker using the `make code-format` command as in the README. Saves having to do this afterwards all the time.

It does show a small but frequent change, as it now uses the clang version used to build the image (14), instead of hardcoding to 11. Shall I push the changes it proposes?

.custom-format.py
```diff
@@ -19,13 +19,18 @@
...
 # string_re = re.compile('(\\"(\\\\.|[^"\\\\])*\\")') # future use
 
 with open(".clang-format") as f:
     fmt = f.read()
 
 CLANG_FORMAT_BIN = os.getenv("CLANG_FORMAT_BIN")
+
+if CLANG_FORMAT_BIN is None:
+    CLANG_FORMAT_BIN = shutil.which("clang-format")
+
```
vanhauser-thc: the problem here is that clang-format stuff changes with versions. hence if I use it on my system the result will be different to what the container has for testing.
it should be clang-13 IMHO

rhtenhove: That'll probably make it harder for contributors to know what to do here, and that opinion might change over time (13 was released just half a year ago). The README also mostly mentions 12 (and 13 on one occurence in the FAQ).

Perhaps correlate it with whatever is the compiler du jour for compiling AFL++? The stable Dockerfile is boasting 14, which is a clear setpoint of what it should be. The README could be generalized with a variable before the install command as in the Dockerfile in this PR.

Then 14 could be the thing now, and it'll be 15 when AFL++ is ready for that version. 

In the end it's just formatting, a goal here is to reduce maintenance, and I believe consistency is better than preference.

I've made a commit suggesting a way to approach this:
1. If CLANG_FORMAT_BIN is set _and_ exists, use that
2. else check if clang-format-$LLVM_VERSION exists (as in the container) and use that
3. else check if clang-format-CURRENT_LLVM exists (hardcoded in the .py) and use that
4. else complain that we need clang-format-$LLVM_VERSION (or hardcoded CURRENT_LLVM if that env doesn't exist as with most users))

WDYT?

domenukk: Probably it also makes sense to have the current `$LLVM_VERSION` somewhere in the repo in a single file, instead of multiple

vanhauser-thc: hmm I think a format checker is not a good idea. every submitter would need to use the same llvm and update when we command everyone does. I prefer that we do it ourselves from time to time.
so please disable this..

vanhauser-thc:
can you please disable the format checker and fix the dockerfile conflict? 
then from my side this can be merged.
thank you for your effort!

rhtenhove:
That should do it! Changed to gcc 11, added some packages, disabled the formatter, and enabled building nyx (assuming that works now? we'll see soon enough)

vanhauser-thc:
hmm ok what is problematic is that codeql is looking at **all** code, so nyx, unicornafl, aflqemu. all include qemu, so 3x qemu which has tons of findings.
this makes this CI check impossible and will make things fail even for stuff that a PR is not touching.

I only see two ways to solve this:
  1) disable the codeql check
  2) the codeql check only work on the github checkout, not including the submodules

rhtenhove:
> hmm ok what is problematic is that codeql is looking at **all** code, so nyx, unicornafl, aflqemu. all include qemu, so 3x qemu which has tons of findings.
> 
> this makes this CI check impossible and will make things fail even for stuff that a PR is not touching.
> 
> 
> 
> I only see two ways to solve this:
> 
>   1) disable the codeql check
> 
>   2) the codeql check only work on the github checkout, not including the submodules

I think that can be achieved by not having it build any of the submodules. Which `make _something_` do you recommend?

vanhauser-thc:
`make all` builds all except sub projects *_mode

vanhauser-thc:
looks good, thank you!

Issue #1451, opened 2022-06-20
Confusion about State of Frida_mode for Android


GanbaruTobi:
Hello, I am little confused. 
I try to build frida_mode for android. The release log from 4.00c states "frida_mode: ...    David Carlier added Android support :)". 
When I am able to build the afl-frida-trace.so (after lots of changes in GNUMake) I get an error

 'cannot locate symbol "__libc_start_main" referenced by afl-frida-trace.so'

But on Android it should be __libc_init, like here : https://github.com/Vector35/binaryninja-api/issues/558. Could somebody help me understand? Was it maybe OSX support instead?

vanhauser-thc:
@devnexen do you know?
also @WorksButNotTested

devnexen:
I may look into a bit later unless someone beats me to it :-) not sure yet how straightforward that is.

WorksButNotTested:
Looks like there is a ‘ifdef’ EMBEDDED in there which may allow for systems which have a ‘main’ but no ‘__libc_start_main’?

GanbaruTobi:
So I tried building with EMBEDDED, got errors for seccomp. For now i excluded all references to the seccomp code.
Next error was a cmplog error, which at least could be bypassed with -c 0 option.

Now i rewrote the __libc_init and also tried with EMBEDDED and got the following error:

[-] PROGRAM ABORT : shmctl (IPC_RMID) < 0 - errno: 22
 Location : prefetch_init(), /home/gbaru/android/lineage/AFLplusplus/frida_mode/src/prefetch.c:304


![grafik](https://user-images.githubusercontent.com/15239745/174830173-07b51d8d-a97d-4006-ba3d-4a13d3608486.png)

This is probably over my head, but is this the android ashmem problem?

WorksButNotTested:
This section of code is generating a section of memory which can be used for IPC between the fork server and the child (running the target code). 

FRIDA generates instrumented blocks of code on the fly as the target runs. The shared memory is used so that the child can pass the address of executed blocks of code so that the parent can also generate an instrumented copy of executed blocks so that they can be inhered by the child.

There isn’t anything too particular about this shared memory except that the two processes see the same data rather than being copy on write.

Running ‘afl-system-config’ does something to the shared memory configuration I believe, so that might help you?

GanbaruTobi:
Indeed that helped:
[...]

PR #1456, opened 2022-06-22
** DO NOT MERGE ** frida android changes

GanbaruTobi:
This is the PR to my changes for frida-mode to have it working with Android.
They should not getting merged as is. 
Some problems to solve:

1. The seccomp code has to be excluded for android (or get reworked to work, I did not look into it) in GNUmakefile
2. The GNUmakefile pulled the wrong gum_devkit (pulled linux_x86_64 bc of the host system)
3. Merge of old GNUmakefile and overwriten parts for Android to have both working (here I changed flags to work for android)
4. Some compiler warnings have been ignored by me for now
5. Note in Readme to use Android >= 26 to have prefetch working
6. Integration of frida-mode into Android.bp

This was the issue:  https://github.com/AFLplusplus/AFLplusplus/issues/1451


frida_mode/src/prefetch.c
```diff
@@ -299,12 +300,15 @@ void prefetch_init(void) {
   /*
    * Configure the shared memory region to be removed once the process dies.
    */
+  #ifdef __ANDROID__
```
GanbaruTobi: Ah! I forgot about this part. Its not solved. I just verified it will still cause crash:
grafik

frida_mode/GNUmakefile
```diff
@@ -27,10 +33,16 @@ AFL_CFLAGS:=-Wno-unused-parameter \
 		   -Wno-unused-result \
 		   -Wno-int-to-pointer-cast \
 		   -Wno-pointer-sign
-
+ifneq "$(findstring android, $(shell $(TARGET_CC) --version 2>/dev/null))" ""
```
WorksButNotTested: Might be clearer to read if you do something like...
```
IS_ANDROID:=$(findstring android, $(shell $(TARGET_CC) --version 2>/dev/null))
#ifdef IS_ANDROID
...
```

Repeat for similar expressions below.

frida_mode/GNUmakefile
```diff
@@ -116,7 +125,8 @@ ifndef OS
  $(error "Operating system unsupported")
 endif
 
-GUM_DEVKIT_VERSION=15.1.22
+
+GUM_DEVKIT_VERSION=15.1.25
```
WorksButNotTested: Why the version bump? No reason not to use the latest, but are there any new features/fixes in FRIDA which you need? Think there is a newer version still? I assume Ole has been busy as he's a bit behind on the release notes.

frida_mode/GNUmakefile
```diff
@@ -218,12 +228,22 @@ TRACE_LDFLAGS+=$(FRIDA_DIR)build/frida-$(OS)-$(ARCH)/lib/libfrida-gum-1.0.a \
 			   $(FRIDA_DIR)build/sdk-$(OS)-$(ARCH)/lib/libgmodule-2.0.a \
 
 else ifeq "$(ARCH)" "arm64"
-
+ifeq "$(OS)" "android"
```
WorksButNotTested: Might be simpler to have...
```
CFLAGS+=-I $(FRIDA_DIR)build/frida_thin-$(OS)-$(ARCH)/include/frida-1.0 \
	    -I $(FRIDA_DIR)build/frida_thin-sdk-$(OS)-$(ARCH)/include/glib-2.0/ \
		-I $(FRIDA_DIR)build/frida_thin-sdk-$(OS)-$(ARCH)/lib/glib-2.0/include/ \	
		-I $(FRIDA_DIR)build/frida_thin-sdk-$(OS)-$(ARCH)/include/capstone/ \
		-I $(FRIDA_DIR)build/frida_thin-sdk-$(OS)-$(ARCH)/include/json-glib-1.0/

ifeq "$(OS)" "android"
  CFLAGS += -static-libstdc++
endif
```
```

File: frida_mode/GNUmakefile [hunk 5]
```diff
@@ -364,6 +394,7 @@ $(AFLPP_QEMU_DRIVER_HOOK_OBJ): $(AFLPP_QEMU_DRIVER_HOOK_SRC) | $(BUILD_DIR)
 hook: $(AFLPP_FRIDA_DRIVER_HOOK_OBJ) $(AFLPP_QEMU_DRIVER_HOOK_OBJ)
 
 ############################# ADDR #############################################
+ifneq "$(OS)" "android"
```
WorksButNotTested, line 396: All these individual options should probably be split into a separate env variable definition.
Then you can just have...
```
ifneq "$(OS)" "android"
  MY_ENV+=-lrt
endif
````

File: frida_mode/src/entry.c [hunk 6]
```diff
@@ -10,7 +11,7 @@
 #include "instrument.h"
 #include "persistent.h"
 #include "ranges.h"
-#include "seccomp.h"
+
```
WorksButNotTested, line 14: I'd try copying the approach for __APPLE__ for seccomp support. That way only the feature is changed depending on the OS, rather than the various points where it is integrated with the wider instrumentation engine.

frida_mode/src/persistent/persistent_arm64.c
```diff
@@ -105,6 +105,7 @@ static void instrument_persitent_save_regs(GumArm64Writer *  cw,
                                           offsetof(persistent_ctx_t, rflags));
 
   /* Q */
+  #if defined(__arm64__)
```
WorksButNotTested: Why is this needed? The file has a guard for `__aarch64__`, how does `__arm64__` differ? A comment in the code might be helpful here too.

frida_mode/src/prefetch.c
WorksButNotTested: Looks like AFL++ itself uses this header to fix Android issues with shared memory. Should be able to add a similar header to FRIDA mode. Perhaps shared memory should be abstracted to a BSP layer rather than using the Linux APIs directly. Check [here](https://grep.app/search?q=shmctl&filter[repo][0]=AFLplusplus/AFLplusplus&filter[path][0]=frida_mode/) to see where it is used.

frida_mode/ts/lib/afl.ts
```diff
@@ -266,11 +266,12 @@ class Afl {
    * See `AFL_FRIDA_SECCOMP_FILE`. This function takes a single `string` as
    * an argument.
    */
+  /* dk how to exlcude
```
WorksButNotTested: Should be able to re-instate this if the `__APPLE__` model for excluding seccomp is usable.

WorksButNotTested:
> 5\. Note in Readme to use Android >= 26 to have prefetch working

It would be good if we could have the Makefile detect this too and error out.

WorksButNotTested:
I've chucked in a few comments. Hope they are helpful. Totally appreciate that this was just a rough draft. It doesn't look like it should be too much effort to tidy and merge.

frida_mode/GNUmakefile
GanbaruTobi: True, they come fast. I did my tests with .25 and I only saw internal used lib updates. So nothing I assumed having breaking changes. 15.1.27 seems to be the same in that regard. I guess i will revert this for now, since it has nothing to do with this PR

frida_mode/src/persistent/persistent_arm64.c
GanbaruTobi: My bad, its not needed. I had compile errors coming from that in my first setup. Now it only breaks by frida  >= 15.1.25.

GanbaruTobi: grafik

frida_mode/GNUmakefile
WorksButNotTested: Happy with whatever version you want to go with. Just wondered if there was any problems with 15.1.27 which means that 15.1.25 was preferabe.

frida_mode/src/persistent/persistent_arm64.c
WorksButNotTested: I think Ole added more support for NEON/SIMD registers recently which maybe have changed the structure slightly....
https://github.com/frida/frida-gum/blob/eb4faac39fa2db6d221d329ee50ca2c8059543f1/gum/gumdefs.h#L276

GanbaruTobi: Okey, so I might go against the latest version 15.1.27 then.
The variable naming in the x86-backend also changed. After I changed that I could run some tests on my x64 successfully.

My problem is just that I dont have figured out how to use persistent mode on android yet. The addr binary is not working. I will look into the test examples and addr source and try to figure out my alternative.

frida_mode/src/persistent/persistent_arm64.c
```diff
@@ -107,16 +107,16 @@ static void instrument_persitent_save_regs(GumArm64Writer *  cw,
   /* Q */
   gum_arm64_writer_put_stp_reg_reg_reg_offset(
```
WorksButNotTested: Looks like ARM64 has 32 of these Q registers... https://github.com/capstone-engine/capstone/blob/fee83fcc1ad096c22d4f2066ccb58ad1a76a9886/include/capstone/arm64.h#L453

Gum meanwhile provides 16 `v` registers (each 16 bytes in size)...
https://github.com/frida/frida-gum/blob/eb4faac39fa2db6d221d329ee50ca2c8059543f1/gum/gumdefs.h#L259

Thus I think the index into the `v` array should increment by `1` for each pair of registers rather than `4`, and we are missing pushing the remainder of the Q registers. Given the increased number of registers, this series of similar function calls should probably be replaced by a loop.

GanbaruTobi: I don't know whats a beautiful way to tackle this problem. I just came up with:


// https://d-capstone.dpldocs.info/capstone.arm64.arm64_reg.html
int reg_Q = 104;

for (int i = 0; i < 16; i++) {

    gum_arm64_writer_put_stp_reg_reg_reg_offset(
             cw, (arm64_reg)reg_Q, (arm64_reg)(reg_Q + 1), ARM64_REG_X0,
             offsetof(GumCpuContext, v[i]), GUM_INDEX_SIGNED_OFFSET);

     reg_Q = reg_Q + 2;

}

frida_mode/src/persistent/persistent_arm64.c
WorksButNotTested: Yeah. Bumping to the latest certainly shouldn't do any harm. The `addr` binary is just used to work out the default load address for PIE executables when ASLR is disabled.

Rather than relying on an...
```
echo 0 > /proc/sys/kernel/randomize_va_space
```
FRIDA mode instead uses the `personality` API with `ADDR_NO_RANDOMIZE`. This takes effect from the next `execve`, so the process then `execve`'s itself to cause it to be loaded at a fixed address (without requiring any system-wide configuration). Hence the need to hook `main` or similar to access the arguments and environment so that they can be passed to the `execve` function to ensure that the new process keeps all of it's options.

WorksButNotTested: Yeah. Bumping to the latest certainly shouldn't do any harm. The `addr` binary is just used to work out the default load address for PIE executables when ASLR is disabled.

Rather than relying on an...
```
echo 0 > /proc/sys/kernel/randomize_va_space
```
FRIDA mode instead uses the `personality` API with `ADDR_NO_RANDOMIZE`. This takes effect from the next `execve`, so the process then `execve`'s itself to cause it to be loaded at a fixed address (without requiring any system-wide configuration). Hence the need to hook `main` or similar to access the arguments and environment so that they can be passed to the `execve` function to ensure that the new process keeps all of it's options.

`addr` is used by the makefile for various tests to help in calculating the address of symbols in the target application. `addr` provides the base address, and [get_symbol_addr.sh](https://github.com/AFLplusplus/AFLplusplus/blob/stable/frida_mode/util/get_symbol_addr.sh) provides the offset (by means of grepping the output of `nm`). This allows variables such as `AFL_FRIDA_PERSISTENT_ADDR` to be set appropriately when testing with persistent mode.

Of course, using the [scripting](https://github.com/AFLplusplus/AFLplusplus/blob/stable/frida_mode/Scripting.md) mode avoids all this since you can use FRIDA's symbol engine to find symbols of interest and use the API exposed by AFL++ FRIDA mode to configure these settings.

frida_mode/src/persistent/persistent_arm64.c
WorksButNotTested: Yeah. I don't think a beautiful way exists, but that's along the lines I was thinking...

```
for (int i = 0; i < 16; i++) {

    gum_arm64_writer_put_stp_reg_reg_reg_offset(
         cw, ARM64_REG_Q0 + (i*2), ARM64_REG_Q0 + (i*2) + 1, ARM64_REG_X0,
         offsetof(GumCpuContext, v[i]), GUM_INDEX_SIGNED_OFFSET);

}
```

frida_mode/src/main.c
```diff
@@ -220,7 +220,7 @@ __attribute__((visibility("default"))) void afl_frida_start(void) {
   persistent_config();
   prefetch_config();
   ranges_config();
-  #if !defined(__ANDROID__)
+  #if defined(__linux__) && !defined(__ANDROID__)
   seccomp_config();
```
WorksButNotTested: I don't think we should need any conditional compilation in the rest of AFL++ FRIDA mode, I think it should be possible to isolate all the conditional compilation to the `seccomp` portion like [this](https://github.com/AFLplusplus/AFLplusplus/blob/ba3c7bfe40f9b17a691958e3525828385127ad25/frida_mode/src/seccomp/seccomp.c#L12). e.g. the user can still attempt to use `seccomp`, but it will provide an error at runtime saying it's unsupported rather than being compiled out.

vanhauser-thc:
FYI we need to release a new version tomorrow, if this should be in that it should be done latest tonight.

GanbaruTobi:
Hey @[vanhauser-thc](https://github.com/vanhauser-thc), I will push my latest changes but there is no fix for the shm problem for prefetch for now. I started to look into it but I am not sure how to do this. I will not manage to look into it the next some days.

WorksButNotTested:
> Hey @[vanhauser-thc](https://github.com/vanhauser-thc), I will push my latest changes but there is no fix for the shm problem for prefetch for now. I started to look into it but I am not sure how to do this. I will not manage to look into it the next some days.

I think the shmem issues should be resolved by following the same approach as AFL++ itself. @domenukk pointed us at [this](https://github.com/AFLplusplus/AFLplusplus/blob/ba3c7bfe40f9b17a691958e3525828385127ad25/include/android-ashmem.h#L29) in the [issue](https://github.com/AFLplusplus/AFLplusplus/issues/1451) which I think should fix it.

I'd be tempted to abstract the shared memory to a separate source file and conditionally include the above android fixes as the shared memory stuff is used in a few [places](https://grep.app/search?q=IPC_RMID&filter[repo][0]=AFLplusplus/AFLplusplus&filter[path][0]=frida_mode/) in FRIDA mode.

@vanhauser-thc thanks for the heads up. @GanbaruTobi if you are able to make the deadline, then great, but if not it will be picked up in the next release (it doesn't tend to be too long between releases), obviously it's better to wait than introduce a regression trying to rush things.

GanbaruTobi:
So to follow up on the state and all comments:
1.  The seccomp code has to be excluded for android (or get reworked to work, I did not look into it) in GNUmakefile
       <-> seccomp on __ANDROID__ is handled like __APPLE__ now
2. The GNUmakefile pulled the wrong gum_devkit (pulled linux_x86_64 bc of the host system)
       <-> This should be fixed, i tested for linux_x86_64, android and mac
3. Merge of old GNUmakefile and overwriten parts for Android to have both working (here I changed flags to work for android)
       <-> Now both is integrated
4. Some compiler warnings have been ignored by me for now 
      <-> There are still some warnings
5. Note in Readme to use Android >= 26 to have prefetch working
     <-> This is not needed. Android shm changed in their headers for Android >=26 but the shm functionality somewhat didn't.  
           Thats the reason the compile errors disappeared, but the error for prefetch stayed.
6. Integration of frida-mode into Android.bp
    <-> I didn't work on that yet

For GNUmakefile I tried to work in all comments from @WorksButNotTested besides [env var seperation](https://github.com/AFLplusplus/AFLplusplus/pull/1456#discussion_r904731110). 

For the Frida changes in 15.1.25 and above i did the loops as talked about.

WorksButNotTested:
A few comments. Should be easy fixes. Then it should at least allow things to run on Android. With the potential limitation of leaked shared memory regions. But since this doesn't make anything worse and indeed offers useful improvement, it seems sensible to merge in it's current form. I'll take a look at abstracting the shared memory stuff to a separate module and incorporating that Android header later in the week, but I will likely need your help @GanbaruTobi to check it works. If that's ok?

domenukk:
FWIW, https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-sharedmem.c was originally abstracted and would "just work", although a few dirty cmplog-specific patches seem to now be in place. In any case, it might be very easy to use/adapt

GanbaruTobi:
@WorksButNotTested happy to help

WorksButNotTested:
> FWIW, https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-sharedmem.c was originally abstracted and would "just work", although a few dirty cmplog-specific patches seem to now be in place. In any case, it might be very easy to use/adapt

Cool, thanks. I'll use it as a reference.

WorksButNotTested:
@GanbaruTobi - I have made the few tweaks above, re-jigged the commits a bit and abstracted out the shared memory management in this PR (https://github.com/AFLplusplus/AFLplusplus/pull/1467). Can you test it out for me and check it works OK for you?

vanhauser-thc:
this now has conflicts, also what is the status @GanbaruTobi ?

WorksButNotTested:
I think this has been superseded by this one now. https://github.com/AFLplusplus/AFLplusplus/pull/1467. But I will need a bit of feedback as to whether it fixes this.

GanbaruTobi:
Yes, its superseded with https://github.com/AFLplusplus/AFLplusplus/pull/1467

PR #1457, opened 2022-06-23
status_screen.md added

Alireza-Razavi:
AFL++ status screen explanations based on https://aflplus.plus/docs/status_screen with some changes in newer versions.

### DIFF
--- a/docs/status_screen.md
+++ b/docs/status_screen.md
@@ -1,0 +1,265 @@
+# Understanding the status screen
+This document provides an overview of the status screen - plus tips for troubleshooting any warnings and red text shown in the UI.
+
+## A note about colors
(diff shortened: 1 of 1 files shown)


domenukk:
It would probably make sense to link this from somewhere in the main docs, right?

vanhauser-thc:
That information is just in a different file (during GSoD)

Alireza-Razavi:
@domenukk @vanhauser-thc 
All docs and `md` files about the AFLplus are integrated in `docs/` and i think it makes a good sense when we can find all information about the status_screen shown by AFLplus in that directory.

vanhauser-thc:
look into docs/afl-fuzz_approach.md - as I said it was merged into another document.
if you think that text there needs improvement - I will be happy with a PR fo rthat.

Issue #1458, opened 2022-06-24
python custom mutator empty string and single byte error causes SEGFAULT


cielonet:
**IMPORTANT**
1. You have verified that the issue to be present in the current `dev` branch. Yes
2. Please supply the command line options and relevant environment variables,
# compile command
./afl-clang-fast custom_mutators/libprotobuf-mutator-example/vuln.c -o vuln_prog

# environment variables:
AFL_DEBUG=1 
PYTHONPATH=`dirname ~/AFLplusplus_dev/custom_mutators/examples/example.py` 
AFL_PYTHON_MODULE=example 
AFL_CUSTOM_MUTATOR_ONLY=1 
./afl-fuzz -i in -o out -- ./vuln_prog 
# command line:
'./afl-fuzz' '-i' 'in' '-o' 'out' '--' './vuln_prog'

**Describe the bug**
1. fuzz() in the python custom-mutator freezes afl-fuzz when the return value is an empty string (i.e. return '')

2. fuzz() in the python custom-mutator SEGFAULTS afl-fuzz when the return value is a "byte" (i.e. return b'\x41\x41')

**To Reproduce**
1. change example.py fuzz() => return ''
2. change example.py fuzz() => return b'\x41\x41'

**Expected behavior**
Expected error msg and/or no segfault

vanhauser-thc:
@domenukk can you have a look? you are a python guy :)

domenukk:
Thank you for the detailed report! It was super easy to reproduce.
The issue is handled more gracefully in #1464.
Please not that you have to return a `bytearray` (starting with this PR, `bytes` will also work) from `fuzz()`, not a utf8 String.

PR #1464, opened 2022-06-28
Python mutators: Gracious error handling for illegal return type

domenukk:
This fixed bug 2 in #1458 - when instead of a `bytearray`, the returned null pointer would be subsequently dereferenced.
Insetad, `afl-fuzz` will now crash with a meaningful error. 
Additonally, this PR allows `bytes`, on top of `bytearray`. utf8-strings, as used in  #1458 are still not supported (with good reason)


vanhauser-thc:
@domenukk thank you! fixed the target branch of the PR.

domenukk:
Bug 1 should also be fixed with this.

PR #1466, opened 2022-06-30
feat: show bound cpu

rhtenhove:
When using variables setting the CPU binding, it's not always trivial to display that value, and this seems like the right place to debug this flag.


vanhauser-thc:
thank you!

PR #1468, opened 2022-07-07
add missing dependencies for "make source-only"

waugustus:
When I build this project with `make source-only`, I got the following errors,

```
$ make source-only
...
[*] Checking libnyx ...
./build_nyx_support.sh: line 58: cargo: command not found
make[1]: Leaving directory '/root/fuzzer/afl++/nyx_mode/QEMU-Nyx/libxdc'
...
ERROR: User requested feature gtk
       configure was not able to find it.
       Install gtk3-devel

GNUmakefile:661: recipe for target 'source-only' failed
make: [source-only] Error 1 (ignored)
```

I have added the instruction to install the missing dependencies for `make source-only` in docs/INSTALL.md.

```
sudo apt install -y cargo libgtk-3-dev
```


domenukk:
@vanhauser-thc I don't think nyx should be part of source-only, right?

vanhauser-thc:
@waugustus 
can you please move the apt call above `sudo apt-get install -y ninja-build # for QEMU mode` please?

@domenukk no that is correct, nyx is also for source-only

domenukk:
I don't think people that are fuzzing with source will want nyx 99% of the time, it's a special use case with a lot of dependencies

waugustus:
@vanhauser-thc OK, done.

vanhauser-thc:
nyx is faster due to snapshot so if people dont want it ... it is a bad decision :)

vanhauser-thc:
thank you!

PR #1469, opened 2022-07-07
push to stable

vanhauser-thc:


instrumentation/split-compares-pass.so.cc
```diff
@@ -566,8 +566,17 @@ bool SplitComparesTransform::splitCompare(CmpInst *cmp_inst, Module &M,
     case CmpInst::ICMP_NE:
     case CmpInst::ICMP_UGT:
     case CmpInst::ICMP_ULT:
+    case CmpInst::ICMP_UGE:
+    case CmpInst::ICMP_ULE:
+    case CmpInst::ICMP_SGT:
+    case CmpInst::ICMP_SLT:
+    case CmpInst::ICMP_SGE:
+    case CmpInst::ICMP_SLE:
```
vanhauser-thc, line 574: @hexcoder- this was needed as otherwise several compares were not split. not sure if my other changes below are ok, maybe you can have a look?

hexcoder-: @vanhauser-thc: You said "this was needed as otherwise several compares were not split."
Are you really sure? All these predicates had been replaced by their simplifications (see `simplifyOrEqualsCompare()` and `simplifySignedCompare()`) and should not be present here anymore.
So this is a nop (does not change anything). 
If you have an example that is not working, I would like to see that, thanks.

vanhauser-thc: either this or the other one. I noticed that during the compilation of a target errors were printed and so I added them and then they were correctly split and no error message anymore. cant tell you though which of the two locations was the necessary one.

PR #1480, opened 2022-07-22
Rename and cleanup

sjmluo:


sjmluo:
renamed variable names and cleaned up files

PR #1496, opened 2022-08-13
libtokencap annotate overriden calls to pyt them together in a special

devnexen:
elf (sub)section and improve their locality.


vanhauser-thc:
thank you!

PR #1499, opened 2022-08-19
push to stable

vanhauser-thc:


src/afl-cc.c
```diff
@@ -310,6 +310,69 @@ static u8 *find_object(u8 *obj, u8 *argv0) {
 
 }
 
+void parse_fsanitize(char *string) {
+
```
domenukk, line 314: I'd probably double-check that `strlen(string) > strlen("-fsanitize=")`, here. Just because it's C...

src/afl-cc.c
```diff
@@ -310,6 +310,69 @@ static u8 *find_object(u8 *obj, u8 *argv0) {
...
+  char *p, *ptr = string + strlen("-fsanitize=");
+  char *new = malloc(strlen(string) + 1);
+  char *tmp = malloc(strlen(ptr));
```
domenukk, line 317: ... and check that `new && tmp`

src/afl-cc.c
vanhauser-thc: nah that is a given. at a minimum it is 1 byte longer than strlen("-fsanitize="), the branch which calls the function ensures that.

src/afl-cc.c
vanhauser-thc: yes you are right :)

src/afl-cc.c
domenukk: Famous last words of a C developer 😛

PR #1501, opened 2022-08-24
Add android documentation

TheCoryBarker:


vanhauser-thc:
Hi,
thank you for your PR.
I don't really understand what you are trying to do though.
Why is there no CMPLOG possible in your opinion?
and for what are you setting the env FUZZ_FRAMEWORK seems not to be used?

TheCoryBarker:
Maybe this isn't the right place to check in this doc, I've seen some people asking questions about fuzzing Android source code before so I thought it might be good to add this documentation here for visibility.

I'm working to integrate AFL++ into Androids build system directly. Support is now available to build AFL++ fuzzers to fuzz Android source code from the check ins below:

https://android-review.googlesource.com/c/platform/build/soong/+/2151702
https://android-review.googlesource.com/c/platform/build/soong/+/2178196

for example, (randomly grabbing a fuzz target checked in to AOSP) 
this file: (https://android.googlesource.com/platform/system/core/+/refs/heads/master/libutils/CallStack_fuzz.cpp) 
with cc_fuzz definied here: https://android.googlesource.com/platform/system/core/+/refs/heads/master/libutils/Android.bp
can now be built as AFL++ fuzzer by instead of building with
> mma libutils_fuzz_callstack -j8  

build with
> FUZZ_FRAMEWORK=AFL mma libutils_fuzz_callstack -j8

This work is supposed to make it very very easy to fuzz Android source code with AFL++. CMPLOG is not yet supported in AOSP directly yet, but I'm hoping to get it in later this year.

vanhauser-thc:
> I'm working to integrate AFL++ into Androids build system directly. Support is now available to build AFL++ fuzzers to fuzz Android source code from the check ins below:
> 
> https://android-review.googlesource.com/c/platform/build/soong/+/2151702 https://android-review.googlesource.com/c/platform/build/soong/+/2178196

ah this is important context you did not provide ;-)
and would belong in this documentation.

I am bit torn. it is good to have this here, but expectations would be different for readers - some fuzz with frida_mode, others want to instrument their own targets, and those two groups would find nothing in this doc.

TheCoryBarker:
I see, so maybe this will just cause confusion. Maybe this should only be documented in Android then. Should we abandon this check in?

vanhauser-thc:
unless you also want to document how to use frida and qemu mode on android and how to compile afl++ on android - yes this should better be closed :)

PR #1509, opened 2022-09-07
analyze_run_target return dataconversion

CodeLogicError:
the function analyze_run_target create u64 data and return u32, then in function analyze xor_ff == orig_cksum will always be 0. also like 
   ```
 xff_orig = (xor_ff == orig_cksum);
    x01_orig = (xor_01 == orig_cksum);
    s10_orig = (sub_10 == orig_cksum);
    a10_orig = (add_10 == orig_cksum);
```
then the b_data[i] = RESP_NONE; always false . 

So I change the u32 to u64 to fix the bug

### DIFF
--- a/src/afl-analyze.c
+++ b/src/afl-analyze.c
@@ -203,7 +203,7 @@ static void read_initial_file(void) {
 /* Execute target application. Returns exec checksum, or 0 if program
    times out. */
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thank you!

PR #1516, opened 2022-09-13
Fix LLVM 15 build

nikic:
Fix LLVM 15 build by removing the call to `initializeModuleSanitizerCoverageLegacyPassPass()`. It looks like things got confused here -- `initializeModuleSanitizerCoverageLegacyPassPass()` is the initialization function for LLVMs own sanitizer coverage pass (for which the LegacyPM version no longer exists), which is independent of AFL's sanitizer coverage pass. The initialization function for AFL's pass is commented out at the end of the file, so I went with just dropping the call entirely.

### DIFF
--- a/instrumentation/SanitizerCoveragePCGUARD.so.cc
+++ b/instrumentation/SanitizerCoveragePCGUARD.so.cc
@@ -209,57 +209,6 @@ class ModuleSanitizerCoverageAFL
 
 };
 
-class ModuleSanitizerCoverageLegacyPass : public ModulePass {
-
- public:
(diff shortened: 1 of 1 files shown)


nikic:
Just confirmed that with this change `make test` passes against an LLVM 15 build.

nikic:
After looking into this a bit more, I went for a more complete fix: Just delete ModuleSanitizerCoverageLegacyPass entirely. The use of the legacy PM variant of this pass was removed at some point in the past, so there is no point in declaring it anymore.

vanhauser-thc:
closing in favor of #1518

PR #1519, opened 2022-09-15
push to stable

vanhauser-thc:


Dockerfile
```diff
@@ -9,18 +9,31 @@ FROM ubuntu:22.04 AS aflplusplus
 LABEL "maintainer"="afl++ team <afl@aflplus.plus>"
 LABEL "about"="AFLplusplus container image"
 
+### Comment out to enable these features
+# Only available on specific ARM64 boards
+ENV NO_CORESIGHT=1
+# Possbile but unlikely in a docker container
+ENV NO_NYX=1
+# Unicorn issues on ARM64
+ENV NO_UNICORN_ARM64=1
```
domenukk: Unicorn on aarch should be fixed

Dockerfile
```diff
@@ -9,18 +9,31 @@ FROM ubuntu:22.04 AS aflplusplus
 LABEL "maintainer"="afl++ team <afl@aflplus.plus>"
 LABEL "about"="AFLplusplus container image"
 
+### Comment out to enable these features
+# Only available on specific ARM64 boards
+ENV NO_CORESIGHT=1
```
domenukk, line 14: Shouldn't we disable Coresight by default, and have a flag that's positive?

src/afl-showmap.c
```diff
@@ -1241,11 +1241,15 @@ int main(int argc, char **argv_orig, char **envp) {
...
-        fsrv->map_size = map_size;
-    }
-    else {
-        fsrv->map_size = 4194304; // dummy temporary value
+
+      fsrv->map_size = map_size;
+
+    } else {
+
+      fsrv->map_size = 4194304;  // dummy temporary value
```
domenukk, line 1249: Should the default value be in config.h?

domenukk (approved):

Dockerfile
```diff
@@ -9,18 +9,29 @@ FROM ubuntu:22.04 AS aflplusplus
...
+# Possbile but unlikely in a docker container
+ENV NO_NYX=1
+
+### Only change these if you know what you are doing:
+# LLVM 15 does not look good so we stay at 14 to still have LTO
```
hexcoder-, line 19: I thought, we have now LTO with LLVM 15, correct?

src/afl-showmap.c
vanhauser-thc: nah it is more of a dummy value

hexcoder- (approved):
Nice work, LTO with LLVM especially!

Dockerfile
vanhauser-thc: true we could change this, but for safety I would leave it at 14 for another 1-2 months

PR #1529, opened 2022-09-21
Potential fix for instrumentation corrupting heap 

### LINKED ISSUE #1528: Instrumentation overflows the heap when using a instrumented shared library
**Describe the bug**
This is the problem I was talking on discord, and I'll upload the minimum PoC here
In short, If I compiles the harness into two parts, one to a shared lib, the other to the main executable. The instrumented code overflows into another heap chunk.

cc. @octaviogalland found this issue.

**To Reproduce**
main.c:
```
int main(int argc) {

  int v = 0;
  char *p = calloc(1, 0x10);

  scanf("%d", &v);
  printf("Before p[0]: %d\n", p[0]);

  if (v == 0) {
    printf("In main.c: 0\n");
  } else if (v == 1) {
    printf("In main.c: 1\n");
  } else if (v == 2) {
    printf("In main.c: 2\n");
  } else if (v == 3) {
    printf("In main.c: 3\n");
  } else if (v == 4) {
    printf("In main.c: 4\n");
  } else if (v == 5) {
    printf("In main.c: 5\n");
  } else if (v == 6) {
    printf("In main.c: 6\n");
  } else if (v == 7) {
    printf("In main.c: 7\n");
  } else if (v == 8) {
    printf("In main.c: 8\n");
  } else if (v == 9) {
    printf("In main.c: 9\n");
  } else if (v == 10) {
    printf("In main.c: 10\n");
  } else if (v == 11) {
    printf("In main.c: 11\n");
  } else if (v == 12) {
    printf("In main.c: 12\n");
  } else if (v == 13) {
    printf("In main.c: 13\n");
  } else if (v == 14) {
    printf("In main.c: 14\n");
  } else if (v == 15) {
    printf("In main.c: 15\n");
  } else if (v == 16) {
    printf("In main.c: 16\n");
  } else if (v == 17) {
    printf("In main.c: 17\n");
  } else if (v == 18) {
    printf("In main.c: 18\n");
  } else if (v == 19) {
    printf("In main.c: 19\n");
  } else if (v == 20) {
    printf("In main.c: 20\n");
  } else if (v == 21) {
    printf("In main.c: 21\n");
  } else if (v == 22) {
    printf("In main.c: 22\n");
  } else if (v == 23) {
    printf("In main.c: 23\n");
  } else if (v == 24) {
    printf("In main.c: 24\n");
  } else if (v == 25) {
    printf("In main.c: 25\n");
  }
  printf("After p[0]: %d\n", p[0]);
  a(3);
}
```
a.c:
```
void a(int v) {
  if (v == 0x0) {
    printf("In a.c: 0\n");
  } else if (v == 0x1) {
    printf("In a.c: 1\n");
  } else if (v == 0x2) {
    printf("In a.c: 2\n");
  } else {
    printf("In a.c: 3\n");
  }
}
```

Compile:
```
~/AFLplusplus/afl-clang-fast -O0 -shared -fPIC a.c -o liba.so
~/AFLplusplus/afl-clang-fast -O0 main.c -L. -la -Wl,-rpath=. -o main
```

Execution: (type 24 for the scanf)
```
24
Before p[0]: 0
In main.c: 24
After p[0]: 1
In a.c: 3
```

In my code, I calloc-ed 0x10 buffer for p, and makes no change to it, but the instrumented code changes it from 0 to 1
**Expected behavior**
The overflow should not happen

Source code here: [test.zip](https://github.com/AFLplusplus/AFLplusplus/files/9613809/test.zip)

**Additional context**
The reason for this is that the pcguard array starts from 0x7 to 0x23
but __afl_area_ptr has only 0x20 bytes allocated
```
pwndbg> x/4gx __afl_area_ptr - 0x10
0x60b290:	0x0000000000000000	0x0000000000000021
0x60b2a0:	0x0100000000000000	0x0000000000000000
```
```
pwndbg> x/20gx 0x40a1a0
0x40a1a0 <__sanitizer_cov_trace_pc_guard_init.offset>:	0x0000000200000004	0x0000000700000002
0x40a1b0:	0x0000000900000008	0x0000000b0000000a
0x40a1c0:	0x0000000d0000000c	0x0000000f0000000e
0x40a1d0:	0x0000001100000010	0x0000001300000012
0x40a1e0:	0x0000001500000014	0x0000001700000016
0x40a1f0:	0x0000001900000018	0x0000001b0000001a
0x40a200:	0x0000001d0000001c	0x0000001f0000001e
0x40a210:	0x0000002100000020	0x0000002300000022
0x40a220 <completed.0>:	0x0000000000000000	0x0000000000000000
0x40a230 <__afl_area_initial>:	0x0000000000000000	0x0000000000000000
```

This problem happens after https://github.com/AFLplusplus/AFLplusplus/commit/35d49c7c5c398c6de5d3091fdda298e45726ae1b this change.
after this change, __afl_area_ptr & __afl_area_ptr_dummy on line 511, 512 does not get updated in the else if branch anymore, so it fails(?)

tokatoka:
potential fix for #1528 

But I'm not confident about the conditions for the if branches..
cc @OctavioGalland

### DIFF
--- a/instrumentation/afl-compiler-rt.o.c
+++ b/instrumentation/afl-compiler-rt.o.c
@@ -363,7 +363,7 @@ static void __afl_map_shm(void) {
 
       }
 
-    } else {
+    } else if (!__afl_final_loc){
 
       __afl_map_size = MAP_INITIAL_SIZE;
 
@@ -516,11 +516,22 @@ static void __afl_map_shm(void) {
 
     }
 
-  } else if (__afl_final_loc > __afl_map_size) {
+  } else if ((_is_sancov && __afl_area_ptr && __afl_area_ptr != __afl_area_initial) || (__afl_final_loc > __afl_map_size)) {
 
-    if (__afl_area_initial != __afl_area_ptr_dummy) {
+    // Reallocate __afl_area_ptr if
+    // 1. __afl_area_ptr was allocated in __afl_auto_first or __afl_auto_second before, or
+    // 2. __afl_map_size from the env is smaller than __afl_final_loc
+
+    if (__afl_area_initial != __afl_area_ptr_dummy && __afl_final_loc > __afl_map_size) {
 
       free(__afl_area_ptr_dummy);
+
+    }
+    if (__afl_area_initial != __afl_area_ptr && __afl_area_ptr && __afl_area_ptr_dummy != __afl_area_ptr) {
+      // In case this __afl_area_ptr got malloc-ed in __afl_auto_first or __afl_auto_second
+      // but then __afl_final_loc is larger now than before, so we must re-request the memory chunk with larger size
+
+      free(__afl_area_ptr);
 
     }
 


instrumentation/afl-compiler-rt.o.c
```diff
@@ -516,12 +516,23 @@ static void __afl_map_shm(void) {
 
     }
```
vanhauser-thc, line 517: No further else? Is that as you wanted it?

tokatoka: yes I don't think we should set 
__afl_map_size to MAP_INITIAL_SIZE;
if __afl_final_loc is already equivalent to the map size, and we should just keep it as it is

https://github.com/AFLplusplus/AFLplusplus/issues/1528#issuecomment-1253318827
here in your comment it says `__afl_map_size 2097152` (MAP_INITIAL_SIZE) but it is wrong isn't it?
this `else if` here intends to fix this wrong __afl_map_size

tokatoka:
It angers LeakSanitizer so this looks wrong..

PR #1541, opened 2022-10-03
auto shmem fuzzing

vanhauser-thc:
a harness must be able to disable shared memory fuzzing: if it is a multi harnesses implementation where one or more instances can't use it for what ever reason.

on the other hand, the only occasion shared memory fuzzing can erroneously be disabled is the aflpp driver.

so in this PR I remove the disable shared memory coding in the aflpp driver, but to ensure that shared memory is never used when afl-fuzz is not present, I moved this check into afl-compiler-rt.

@domenukk

### DIFF
--- a/src/afl-fuzz.c
+++ b/src/afl-fuzz.c
@@ -2132,6 +2132,16 @@ int main(int argc, char **argv_orig, char **envp) {
 
   }
 
+  if (afl->fsrv.out_file && afl->fsrv.use_shmem_fuzz) {
+
+    afl->fsrv.out_file = NULL;
+    afl->fsrv.use_stdin = 0;
+    WARNF(
+        "You specified -f or @@ on the command line but the target harness "
+        "specified fuzz cases via shmem, switching to shmem!");
+
+  }
+
   deunicode_extras(afl);
   dedup_extras(afl);
   if (afl->extras_cnt) { OKF("Loaded a total of %u extras.", afl->extras_cnt); }
--- a/src/afl-fuzz.c
+++ b/src/afl-fuzz.c
@@ -2136,9 +2136,13 @@ int main(int argc, char **argv_orig, char **envp) {
 
     afl->fsrv.out_file = NULL;
     afl->fsrv.use_stdin = 0;
-    WARNF(
-        "You specified -f or @@ on the command line but the target harness "
-        "specified fuzz cases via shmem, switching to shmem!");
+    if (!afl->unicorn_mode && !afl->fsrv.use_stdin) {
+
+      WARNF(
+          "You specified -f or @@ on the command line but the target harness "
+          "specified fuzz cases via shmem, switching to shmem!");
+
+    }
 
   }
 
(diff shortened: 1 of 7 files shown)


domenukk:
This doesn't fix __AFL_LOOP-style harnesses using shared memory. What would such a multi-harness look like?

vanhauser-thc:
> This doesn't fix __AFL_LOOP-style harnesses using shared memory. 

why doesnt it? shared memory fuzzing is only available it if you request it in the harness. (__AFL_FUZZ_INIT() macro + ..._BUF + ..._LEN )

> What would such a multi-harness look like?
```
switch (argv[1]) {
  case 0: harness_fuzz_header(argc, argv); break;
  case 1: harness_fuzz_body(argc, argv); break;
  case 2: harness_fuzz_data(argc, argv); break;
}
```
and harness_fuzz_header could require reading from a file, harness_fuzz_body from stdin and harness_fuzz_data from shared memory.

not saying that this is effective, but I have seen multiharness files.

domenukk:
You can still use __afl_fuzz_ptr as indicator if shmem fuzzing should be used (set to NULL to disable)
The extra variable isn't needed imho

vanhauser-thc:
> You can still use __afl_fuzz_ptr as indicator if shmem fuzzing should be used (set to NULL to disable) The extra variable isn't needed imho

that variable is only set when the forkserver is reached. with manual_init that is pretty late in the harness obviously for performance reason.

the __afl_sharedmem_fuzzing is set to 1 in main if it is actually running within afl-fuzz and 0 if not, so during option parsing you already have that information which should be before the forkserver as otherwise the forkserver is in the wrong location.

domenukk:
No, I mean the harness can set the variable to NULL (after delayed init) if it doesn't want to use shmem

domenukk:
@vanhauser-thc Sharedmem fuzzing is now only used when the driver is used, or `__AFL_FUZZ_TESTCASE_BUF` is executed, but then it should always be used. Please double/triplecheck :)

src/afl-fuzz.c
```diff
@@ -2132,6 +2132,16 @@ int main(int argc, char **argv_orig, char **envp) {
 
   }
 
+  if (afl->fsrv.out_file && afl->fsrv.use_shmem_fuzz) {
+
+    afl->fsrv.out_file = NULL;
+    afl->fsrv.use_stdin = 0;
+    WARNF(
+        "You specified -f or @@ on the command line but the target harness "
```
domenukk: This warning is very confusing for unicorn, where specifying @@ is the common way. I'd remove it again (it doesn't make a difference anyway?)
Else, hide for -U

utils/aflpp_driver/aflpp_driver.c
```diff
@@ -290,6 +291,13 @@ int LLVMFuzzerRunDriver(int *argcp, char ***argvp,
 
   }
 
+  bool in_afl = (!getenv(SHM_ENV_VAR) || fcntl(FORKSRV_FD, F_GETFD) == -1 || fcntl(FORKSRV_FD + 1, F_GETFD) == -1);
```
domenukk: Not sure if that isn't redundant now

src/afl-fuzz.c
vanhauser-thc: Hiding for unicorn would be best yes

vanhauser-thc:
so, now everything should work ...

utils/aflpp_driver/aflpp_driver.c
vanhauser-thc: no it isnt, we cannot use the shmem_fuzz var until the forkserver, but we do the cmdline parsing before that where this is needed

domenukk:
Seems good to me now, thx! :)

domenukk (approved):

PR #1542, opened 2022-10-03
fix Rust speedtest example dependency

jma-qb:
Hi, the Rust capstone dependency was updated on the unicornafl side but not on the AFL++-rust-speedtest side preventing it from compiling.


domenukk:
Sweet, thank you!

PR #1543, opened 2022-10-05
Support building LTO on Gentoo

quinox:
Running `make all` on Gentoo with LLVM, ld.ldd and LLVMgold installed through the package manager doesn't result in an LTO version. The makefile expects `ld.ldd` to be inside `$LLVM_BINDIR` which isn't the case for me. 

I don't know if my solution is appropriate but it does make everything work on my system so hence this PR. I'm happy to make further changes if needed.

Before this patch:
```
make[1]: Entering directory '/home/quinox/.cache/tmp/AFLplusplus'
[+] llvm_mode detected llvm 10+, enabling neverZero implementation and c++14
[+] llvm_mode detected llvm 11+, enabling afl-lto LTO implementation
GNUmakefile.llvm:218: ld.lld not found, cannot enable LTO mode
[+] shmat seems to be working.
...
Build Summary:
[+] afl-fuzz and supporting tools successfully built
[+] LLVM basic mode successfully built
[+] LLVM mode successfully built
[-] LLVM LTO mode could not be build, it is optional, if you want it, please install LLVM 11-14. More information at instrumentation/README.lto.md on how to build it
[+] gcc_mode successfully built
```

After this patch:
```
make[1]: Entering directory '/home/quinox/.cache/tmp/AFLplusplus'
[+] llvm_mode detected llvm 10+, enabling neverZero implementation and c++14
[+] llvm_mode detected llvm 11+, enabling afl-lto LTO implementation
GNUmakefile.llvm:219: ld.ldd found in a weird location (/usr/bin/ld.lld), you better know what you're doing
[+] shmat seems to be working.
...
Build Summary:
[+] afl-fuzz and supporting tools successfully built
[+] LLVM basic mode successfully built
[+] LLVM mode successfully built
[+] LLVM LTO mode successfully built
[+] gcc_mode successfully built
```

---

Information from my package manger:

ld.lld:
```
quinox@gofu ~> emerge --search '^lld$'

[ Results for search key : ^lld$ ]
Searching...

*  sys-devel/lld
      Latest version available: 14.0.6
      Latest version installed: 14.0.6
      Size of files: 103.143 KiB
      Homepage:      https://llvm.org/
      Description:   The LLVM linker (link editor)
      License:       Apache-2.0-with-LLVM-exceptions UoI-NCSA

[ Applications found : 1 ]

quinox@gofu ~> equery files lld | grep 'ld\.lld'
/usr/bin/ld.lld

quinox@gofu ~> /usr/bin/ld.lld --version
LLD 14.0.6 (compatible with GNU linkers)
```

LLVM:
```
quinox@gofu ~> emerge --search '^llvm$'

[ Results for search key : ^llvm$ ]
Searching...

*  sys-devel/llvm
      Latest version available: 14.0.6-r2
      Latest version installed: 14.0.6-r2
      Size of files: 103.372 KiB
      Homepage:      https://llvm.org/
      Description:   Low Level Virtual Machine
      License:       Apache-2.0-with-LLVM-exceptions UoI-NCSA BSD public-domain rc

[ Applications found : 1 ]

quinox@gofu ~> equery files llvm | grep llvm-config
/usr/lib/llvm/14/bin/llvm-config
/usr/lib/llvm/14/bin/x86_64-pc-linux-gnu-llvm-config
/usr/lib/llvm/14/include/llvm/Config/llvm-config.h
/usr/lib/llvm/14/include/x86_64-pc-linux-gnu/llvm/Config/llvm-config.h
/usr/lib/llvm/14/share/man/man1/llvm-config.1.bz2

quinox@gofu ~> /usr/lib/llvm/14/bin/llvm-config --bindir
/usr/lib/llvm/14/bin

quinox@gofu ~> ls -al /usr/lib/llvm/14/bin/ld.lld
ls: cannot access '/usr/lib/llvm/14/bin/ld.lld': No such file or directory
```

### DIFF
--- a/GNUmakefile.llvm
+++ b/GNUmakefile.llvm
@@ -214,6 +214,17 @@ ifeq "$(LLVM_LTO)" "1"
     ifeq "$(AFL_REAL_LD)" ""
       ifneq "$(shell readlink $(LLVM_BINDIR)/ld.lld 2>&1)" ""
         AFL_REAL_LD = $(LLVM_BINDIR)/ld.lld
+      else ifneq "$(shell command -v ld.lld 2>/dev/null)" ""
+        AFL_REAL_LD = $(shell command -v ld.lld)
+        TMP_LDLDD_VERSION = $(shell $(AFL_REAL_LD) --version | awk '{ print $$2 }')
+        ifeq "$(LLVMVER)" "$(TMP_LDLDD_VERSION)"
+          $(warning ld.lld found in a weird location ($(AFL_REAL_LD)), but its the same version as LLVM so we will allow it)
+        else
+          $(warning ld.lld found in a weird location ($(AFL_REAL_LD)) and its of a different version than LLMV ($(TMP_LDLDD_VERSION) vs. $(LLVMVER)) - cannot enable LTO mode)
+          AFL_REAL_LD=
+          LLVM_LTO = 0
+        endif
+        undefine TMP_LDLDD_VERSION
       else
         $(warning ld.lld not found, cannot enable LTO mode)
         LLVM_LTO = 0
@@ -229,7 +240,7 @@ AFL_CLANG_FUSELD=
 ifeq "$(LLVM_LTO)" "1"
   ifeq "$(shell echo 'int main() {return 0; }' | $(CLANG_BIN) -x c - -fuse-ld=`command -v ld` -o .test 2>/dev/null && echo 1 || echo 0 ; rm -f .test )" "1"
     AFL_CLANG_FUSELD=1
-    ifeq "$(shell echo 'int main() {return 0; }' | $(CLANG_BIN) -x c - -fuse-ld=ld.lld --ld-path=$(LLVM_BINDIR)/ld.lld -o .test 2>/dev/null && echo 1 || echo 0 ; rm -f .test )" "1"
+    ifeq "$(shell echo 'int main() {return 0; }' | $(CLANG_BIN) -x c - -fuse-ld=ld.lld --ld-path=$(AFL_REAL_LD) -o .test 2>/dev/null && echo 1 || echo 0 ; rm -f .test )" "1"
       AFL_CLANG_LDPATH=1
     endif
   else


vanhauser-thc:
uh interesting. honestly gentoo fucked it up.
you can have multiple llvm versions installed in parallel, but with gentoo this would not work with ld.lld because it would be overwritten.

the issue with your PR is that we do not know if ld.lld is actually the correct version, so it could also be 13 or 15 instead of the correct 14 (in your system).

I see 3 solutions:
1. open a ticket in gentoo to get this fixed
2. add code to only do this if the Linux distribution is gentoo
3. add code to check the version of ld.lld matching the one of LLVM_CONFIG

quinox:
Thanks for the feedback, I've amended the PR with a version check:
```
[+] llvm_mode detected llvm 11+, enabling afl-lto LTO implementation
GNUmakefile.llvm:221: ld.lld found in a weird location (/usr/bin/ld.lld), but its the same version as LLVM so we will allow it
[+] shmat seems to be working.
```

```
[+] llvm_mode detected llvm 11+, enabling afl-lto LTO implementation
GNUmakefile.llvm:223: ld.lld found in a weird location (/usr/bin/ld.lld) and its of a different version than LLMV (13.0.1 vs. 14.0.6) - cannot enable LTO mode
[+] shmat seems to be working.
```

It checks the version including patch level, which is identical for my system. It uses an `awk` call so it's not super reliable but it won't make it any worse than without this patch.

---

About Gentoo: Yes, it seems peculiar. The LLVM+clang packages uses slots so I can install both 13 and 14 at the same time without issues. `lld` does not support slots so you can only have 1 of them.

I've done some digging on Gentoo's bugtracker, there are a few reports about failing to build stuff like Firefox / Rust due to  incompatible lld versions. ([#838121](https://bugs.gentoo.org/838121#c17)). They also echo the sentiment that Gentoo's current setup is odd. Saillant quotes:

> Comment 16: Therefore, it would seem that in reality, multiple versions of LLVM/clang CAN NOT be installed simultaneously as long as lld is dictated by the earliest version of LLVM. At least from my observations with Firefox.

> Comment 17: I know this is a question that arises often, but from a quick search, only found 
[#691900](https://bugs.gentoo.org/691900). AFAIK this subject has been raised even in our gentoo-dev mailing list. Don't remember the outcome, but clearly slotted lld wasn't it.

I think [#691900](https://bugs.gentoo.org/691900) isn't actually a solution to my current predicament: I want clang to bring its own ld.ldd along, which feels like a separate issues from slotting `/usr/bin/ld.lld`. But if the solution was simple I'm sure they would have done that by now.

Anyway, right now I don't have the energy to deal with them. I'm happy to further amend this PR should you desire so, otherwise feel free to close it without merging.

vanhauser-thc:
looks good, thank you!

PR #1555, opened 2022-10-13
Fix exit_on_time.

kobrineli:
Hi! We've found the bug, that if no coverage was found during all the fuzzing process, exit_on_time is ignored and fuzzing can become endless. We fixed it with checking whether `run_time`, which is equal to `prev_run_time + cur_ms - start_time`, can be greater than `exit_on_time`.


domenukk:
Thank you! please always direct PRs to the `dev` branch

PR #1563, opened 2022-10-20
Fix child reaping on fuzzer termination

nbars:
This pull request contains multiple changes to address issues that cause zombie processes to be left behind after AFL++'s termination (brought to my attention by @j4ck1000):
- Call `waitpid()` on the child and the fork server when terminating the fuzzer; thus, we do not end up with zombies.
- Rename `fsrv.kill_signal` to `fsrv.child_kill_signal`, since the documentation states that the signal is used to terminate the **child**.
- Use SIGTERM instead of `fsrv.(child)_kill_signal`; thus the fork server can always reap the child.

During implementation, I realized that if `afl_states_stop()` is called during handling of SIGTERM the function races with the `afl_fsrv_deinit(&afl->fsrv)` call in `src/afl-fuzz.c:2658` due to the `stop_soon` flag
```
  LIST_FOREACH(&afl_states, afl_state_t, {

    el->stop_soon = 1;

  });
```
that is set in `afl_states_stop` before killing the child and fork server.

In other words, two independent program executions are doing (partially) the same:
1. SIGTERM -> ... -> `afl_states_stop` -> set `stop_soon` flag -> kill child -> kill fork server
2. `main` -> ... -> some fuzzing loop stuff reads `stop_soon` -> `afl_fsrv_deinit` -> `afl_fsrv_kill` -> kill child -> kill fork server

Since I am unfamiliar with the code base, I am unsure why this is the case. Are there any reasons for not stripping the clean-up part from `afl_states_stop` and only setting the `stop_soon` flag there?

Best,
Nils


vanhauser-thc:
please note that we only accept PRs to the dev branch. fixed that for you.

> Use SIGTERM instead of fsrv.(child)_kill_signal; thus the fork server can always reap the child.`

not sure this works. there are childs that should not be killed, e.g. when afl++ i used on a target that is running in a kvm (which would kill the kvm). you should honor what the kill signal is.

src/afl-forkserver.c
```diff
@@ -1242,10 +1242,10 @@ void afl_fsrv_start(afl_forkserver_t *fsrv, char **argv,
 
 void afl_fsrv_kill(afl_forkserver_t *fsrv) {
 
-  if (fsrv->child_pid > 0) { kill(fsrv->child_pid, fsrv->kill_signal); }
+  if (fsrv->child_pid > 0) { kill(fsrv->child_pid, fsrv->child_kill_signal); }
   if (fsrv->fsrv_pid > 0) {
 
-    kill(fsrv->fsrv_pid, fsrv->kill_signal);
+    kill(fsrv->fsrv_pid, SIGTERM);
```
vanhauser-thc: this should not be changed

src/afl-fuzz-state.c
```diff
@@ -653,9 +654,14 @@ void afl_states_stop(void) {
...
 
   LIST_FOREACH(&afl_states, afl_state_t, {
-
-    if (el->fsrv.child_pid > 0) kill(el->fsrv.child_pid, el->fsrv.kill_signal);
-    if (el->fsrv.fsrv_pid > 0) kill(el->fsrv.fsrv_pid, el->fsrv.kill_signal);
+    /* NOTE: We need to make sure that the parent (the forkserver) reap the child (see below). */
+    if (el->fsrv.child_pid > 0) kill(el->fsrv.child_pid, el->fsrv.child_kill_signal);
+    if (el->fsrv.fsrv_pid > 0) {
+      /* This must be SIGTERM, to allow the forkserver to reap the child before exiting. */
+      kill(el->fsrv.fsrv_pid, SIGTERM);
```
vanhauser-thc: same here, the kill signal selected should be honored

vanhauser-thc:
besides that - looks good.

nbars:
Thanks for the response. What about the race condition and the doubled clean-up code, any thoughts?

src/afl-forkserver.c
nbars: > 

I would argue that then, there should be a new environment variable to set the kill signal for the fork server. Since SIGKILL is fine (for the default setup AFL++ is used for) to kill the child, it is not fine to be used to terminate the fork server because it then has no chance to reap the child.

nbars: There should be a `fsrv.fsrv_kill_signalI` defaulting to `SIGTERM` that can be configured via an environment variable for cases as yours mentioned previously.

vanhauser-thc:
> Thanks for the response. What about the race condition and the doubled clean-up code, any thoughts?

that looks good :)

src/afl-forkserver.c
vanhauser-thc: maybe like this: have another variable for the forkserver, defaulting to SIGTERM. if the KILL_SIGNAL env is set, this value is set for the child and the forkserver.

The need for the KILL_SIGNAL env this is only with kvm-like setups, and there the signal is changed to a dummy signal. so and extra env to control the forkserver signal would not be needed.

nbars:
> > Thanks for the response. What about the race condition and the doubled clean-up code, any thoughts?
> 
> that looks good :)

I guess we have a misunderstanding here. What I am trying to say, is that `afl_fsrv_kill` and `afl_states_stop` are doing the cleanup twice. In the worst case, this might cause processes to be killed that are not descendants of AFL++'s main process. At least, the pids should be set to `-1`, I guess?

nbars:
> maybe like this: have another variable for the forkserver, defaulting to SIGTERM. if the KILL_SIGNAL env is set, this value is set for the child and the forkserver.
> 
> The need for the KILL_SIGNAL env this is only with kvm-like setups, and there the signal is changed to a dummy signal. so and extra env to control the forkserver signal would not be needed.

I added an `AFL_FORK_SERVER_KILL_SIGNAL` environment variable (defaulting to `SIGTERM`). I am not sure about setting both signals to the same value if `AFL_KILL_SIGNAL` is set. Do you want this behavior for backward compatibility, or what is the reason?

vanhauser-thc:
> > maybe like this: have another variable for the forkserver, defaulting to SIGTERM. if the KILL_SIGNAL env is set, this value is set for the child and the forkserver.
> > The need for the KILL_SIGNAL env this is only with kvm-like setups, and there the signal is changed to a dummy signal. so and extra env to control the forkserver signal would not be needed.
> 
> I added an `AFL_FORK_SERVER_KILL_SIGNAL` environment variable (defaulting to `SIGTERM`). I am not sure about setting both signals to the same value if `AFL_KILL_SIGNAL` is set. Do you want this behavior for backward compatibility, or what is the reason?

yes for compatability. otherwise you are killing kafl and other tools that are using afl++.

nbars:
> > > maybe like this: have another variable for the forkserver, defaulting to SIGTERM. if the KILL_SIGNAL env is set, this value is set for the child and the forkserver.
> > > The need for the KILL_SIGNAL env this is only with kvm-like setups, and there the signal is changed to a dummy signal. so and extra env to control the forkserver signal would not be needed.
> > 
> > 
> > I added an `AFL_FORK_SERVER_KILL_SIGNAL` environment variable (defaulting to `SIGTERM`). I am not sure about setting both signals to the same value if `AFL_KILL_SIGNAL` is set. Do you want this behavior for backward compatibility, or what is the reason?
> 
> yes for compatability. otherwise you are killing kafl and other tools that are using afl++.

Alright. Now, if `AFL_KILL_SIGNAL` is set, `AFL_FORK_SERVER_KILL_SIGNAL` will be set to the same value. However, if `AFL_FORK_SERVER_KILL_SIGNAL` itself is set, it will take precedence.

vanhauser-thc:
wow including doc update - looks great!
I will wait for the CI to check if everything is fine and then merge.

PR #1567, opened 2022-10-25
Parallel afl-cmin

sei-eschwartz:
This PR is not quite ready and is more to generate discussion.  It seems pretty easy to collect the traces in parallel, so is there a reason this isn't being done already?

Asking for a friend who is running afl-cmin on 200,000+ files :)

### DIFF
--- a/afl-cmin.bash
+++ b/afl-cmin.bash
@@ -358,14 +358,17 @@ echo "[*] Obtaining traces for input files in '$IN_DIR'..."
 
   if [ "$STDIN_FILE" = "" ]; then
 
-    STR=$(echo "$SHOWMAP" -m "$MEM_LIMIT" -t "$TIMEOUT" -o "$TRACE_DIR/fn" -Z $EXTRA_PAR -- "$@" "<$IN_DIR/fn")
+    STR=$(echo "$SHOWMAP" -m "$MEM_LIMIT" -t "$TIMEOUT" -o "$TRACE_DIR/fn" -Z $EXTRA_PAR -- "$@" "<$IN_DIR/fn" || exit 255)
 
   else
 
-    STR=$(echo cp "$IN_DIR/fn" "$STDIN_FILE" && "$SHOWMAP" -m "$MEM_LIMIT" -t "$TIMEOUT" -o "$TRACE_DIR/fn" -Z $EXTRA_PAR -H "$STDIN_FILE" -- "$@" "</dev/null")
+    STR=$(echo cp "$IN_DIR/fn" "$STDIN_FILE" && "$SHOWMAP" -m "$MEM_LIMIT" -t "$TIMEOUT" -o "$TRACE_DIR/fn" -Z $EXTRA_PAR -H "$STDIN_FILE" -- "$@" "</dev/null" || exit 255)
   fi
 
-  ls "$IN_DIR" | xargs -P$NPROC -Ifn bash -c "$STR"
+  if !(ls "$IN_DIR" | xargs -P$NPROC -Ifn bash -c "$STR"); then
+    echo "[-] Error: Bad exit code received from afl-showmap (this means a crashing or timeout input is likely present), terminating..."
+    exit 1
+  fi
 
 )
 
--- a/afl-cmin.bash
+++ b/afl-cmin.bash
@@ -47,13 +47,14 @@ echo
 
 MEM_LIMIT=none
 TIMEOUT=none
+NPROC=1
 
 unset IN_DIR OUT_DIR STDIN_FILE EXTRA_PAR MEM_LIMIT_GIVEN \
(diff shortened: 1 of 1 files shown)


domenukk:
In any case you probably don't want to use the bash version, it's lacking a forkserver and hence is pretty slow

sei-eschwartz:
Ah, it looks like the awk script basically moved the iteration into `afl-showmap` itself.

sei-eschwartz:
The only easy way I can think of to do this would be to modify `afl-showmap` and add an option that is similar to `-i` but allows a list of input files to be specified, maybe by passing a file/pipe containing their names.  Then it would be possible to use `xargs` with a suitably large `-n` value to get a speed-up from the fork server, while also being able to execute in parallel.

vanhauser-thc:
I think it is a good idea. I too had the issue of needing to use afl-cmin on 20.000 files but only reaching 0.7 execs/s ... took days.

but as dominik said, afl-cmin.bash is outdated and only there if something in the awk implementation is not working.
so implement it in afl-cmin instead :)

EDIT: and of course the changes need to be in afl-showmap too. if you want to add then this feature to afl-cmin.bash this is totally optional.

domenukk:
> I think it is a good idea. I too had the issue of needing to use afl-cmin on 20.000 files but only reaching 0.7 execs/s ... took days.
> 
> but as dominik said, afl-cmin.bash is outdated and only there if something in the awk implementation is not working. so implement it in afl-cmin instead :)
> 
> EDIT: and of course the changes need to be in afl-showmap too. if you want to add then this feature to afl-cmin.bash this is totally optional.

The way to do it right now is to manually split the folders, run cmin on each in parallel, then do a final cmin on the remainder.
Still extra overhead, of course.
My approach would have been to do spawn threads/fork inside showmap directly for `n` cores, but xargs may work. One thing to look out for is that right now showmap supports recursive folders as well (or was that cmin directly?)

andreafioraldi:
> then do a final cmin on the remainder

In the first round you can temporary save the bitmaps and do the other rounds without re-executing the testcases

andreafioraldi:
You can also implement some parallel version of minset cover, something like (from google) https://www.researchgate.net/publication/358460080_A_parallel_algorithm_for_minimum_weight_set_cover_with_small_neighborhood_property

vanhauser-thc:
@sei-eschwartz so do you still want to tackle this problem? ;)

sei-eschwartz:
For the moment, I'm testing the PR to see if it solves my problem, and see how much overhead from not using the forkserver.  If this bash script solves my problem, then my motivation to parallelize the awk script is significantly less (at least during work hours...).

edmcman:
Weird.  It seems like one of the test cases is hanging, and now that the timeout is working as expected the whole CI fails

vanhauser-thc:
a forkserver will bring a huge speedup (x10 on average I think, depends on target complexity).

vanhauser-thc:
@sei-eschwartz so what do you think about parallelizing? :)

edmcman:
I think for now it probably makes sense to merge this commit so that afl-cmin.bash at least has parallelization.  If I find some free time I might add support to the awk version, but I can't commit to doing it.

vanhauser-thc:
I dont think it makes much sense to merge.

first you broke `@@` support with your change.

then afl-cmin (awk) runs approx 6-8x faster than afl-cmin.bash due to using the forkserver. so that a user would have an advantage with this change, it would require either using `-n 16`, or that the target is very slow.

finally afl-cmin.bash is deprecated, we do not want to really touch it besides that it still works with afl-showmap changes.

edmcman:
> first you broke `@@` support with your change.

Oops, you are right.  I was not using @@ so I didn't notice that.  I assumed the tests would cover both cases.

> then afl-cmin (awk) runs approx 6-8x faster than afl-cmin.bash due to using the forkserver. so that a user would have an advantage with this change, it would require either using `-n 16`, or that the target is very slow.
>
> finally afl-cmin.bash is deprecated, we do not want to really touch it besides that it still works with afl-showmap changes.

FTR, I don't think anyone besides the AFL++ developers are aware of the difference between afl-cmin and afl-cmin.bash.  If this is a concern, deleting afl-cmin.bash immediately is probably a good idea.  You probably have people running the bash version because they don't know there is a difference.

vanhauser-thc:
> > finally afl-cmin.bash is deprecated, we do not want to really touch it besides that it still works with afl-showmap changes.
> 
> FTR, I don't think anyone besides the AFL++ developers are aware of the difference between afl-cmin and afl-cmin.bash. If this is a concern, deleting afl-cmin.bash immediately is probably a good idea. You probably have people running the bash version because they don't know there is a difference.

the reason we keep afl-cmin.bash is that we encountered old awk implementations that dont work (on some BSD machines). so we keep it around although it dont reference it in the documentation. all documenation point to afl-cmin, which is the awk script.

PR #1570, opened 2022-10-27
Fix -t none in afl-showmap

edmcman:


src/afl-showmap.c
```diff
@@ -1015,7 +1015,7 @@ int main(int argc, char **argv_orig, char **envp) {
 
           }
 
-        }
+        } else fsrv->exec_tmout = 0;
```
vanhauser-thc: can you please put `{}` around it? thank you

vanhauser-thc:
thank you, I wasnt even aware that there is a -t none variant :)

edmcman:
Me neither.  But cmin uses that as the default unless you override it.

edmcman:
My change doesn't work for the fork server.  I am looking into it.  I think it has to do with read_s32_timed.

vanhauser-thc:
I think -t none should rather translate to -t 120000 (120 seconds). it doesnt make sense to allow infinity.

edmcman:
Yeah, I was leaning in that direction, or just changing the default in afl-cmin to be 1000 (or 120000).

vanhauser-thc:
are you still on this? :)

edmcman:
This should be ready to go once the CI finishes.  It treats `-t none` as `-t 120000`

vanhauser-thc:
thanks! I will update the help though.

Issue #1565, opened 2022-10-21
Reconsider Persistent Mode in the Compiler Runtime


domenukk:
Right now, persistent mode is enabled the following way:

- afl-fuzz scans the complete binary and checks if `PERSIST_SIG` was inserted (which is automatically done by afl-cc if __AFL_LOOP is used) (and of course this will break for shared objects or wrapper scripts/libraries)
- afl-fuzz sets the `PERSIST_SIG` env variable before launching the target
- When the target starts, it checks the value of this env variable.
- _For every execution_, the target cheks if persistent mode is used. If so, it loops around, else, it exits.

Instead, I propose to change the following macro:
https://github.com/AFLplusplus/AFLplusplus/blob/cbfa5207ba2853e249ffb256d99880368ee224e0/src/afl-cc.c#L1145
to instead set the `persistent_mode` variable to 1 at compile time.

This way, we can potentially (in the future, we don't want a regression for older targets) remove the target scanning, and the persistent mode check can be optimised at compile time.

For new implementations of the forkserver, like in LibAFL, we can also reduce complexity from the get-go.

I haven't yet looked into deferred forkserver, but I assume it should be possible there as well, somehow

vanhauser-thc:
this would break multiharness files if different techniques are used there. it is a rare thing sure, but breaking something that currently works ....

how would you want to set a value in the client at compile time? I dont see a way how this could work. the target forkserver must know if it is persistent mode, but the AFL_LOOP comes later so you cannot set a global var with the AFL_LOOP macro, that would be too late.

domenukk:
> this would break multiharness files if different techniques are used there. it is a rare thing sure, but breaking something that currently works ....

How so? If anything, this can *fix* multiharness files. Right now, it will always default to persistent mode, if one of them is persistent.
Setting the variable to 1 in `__AFL_LOOP` is early enough, the target doesn't need to know it before it either exits, or it doesn't. AFL++ itself doesn't need to know if it's persistent mode or not (we can keep the binary signature around if we really want to, for this case, but have it not used)

vanhauser-thc:
> Setting the variable to 1 in `__AFL_LOOP` is early enough, the target doesn't need to know it before it either exits, or it doesn't. 

 `forkserver -> persistent_loop`. the forkserver must know if there is a persistent loop. look in the code (for the waitpid).
[...]

PR #1576, opened 2022-11-09
Persistent mode always

domenukk:
This PR always enables persistent mode, as discussed in #1565

### DIFF
--- a/instrumentation/afl-compiler-rt.o.c
+++ b/instrumentation/afl-compiler-rt.o.c
@@ -105,6 +105,9 @@ u32 __afl_dictionary_len;
 u64 __afl_map_addr;
 u32 __afl_first_final_loc;
 
+/* 1 if we are running in afl, and the forkserver was stared, else 0 */
+u32 __afl_connected = 0;
+
 // for the __AFL_COVERAGE_ON/__AFL_COVERAGE_OFF features to work:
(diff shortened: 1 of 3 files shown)


domenukk:
The logic is simple: if it's running inside afl, it's persistent (loop count >1), else it's not and runs but once.

instrumentation/afl-compiler-rt.o.c
```diff
@@ -105,6 +105,9 @@ u32 __afl_dictionary_len;
 u64 __afl_map_addr;
 u32 __afl_first_final_loc;
 
+/* 1 if we are running in afl, and the forkserver was stared, else 0 */
```
vanhauser-thc, line 108: typo

vanhauser-thc:
the persistent signature must still be preserved as otherwise this will break oss-fuzz in this coming PR https://github.com/google/oss-fuzz/pull/7343/files

domenukk:
It is preserved

vanhauser-thc:
can you please fix the typo?
also, what is the benefit this creates? as it is this this is just a different way and does not improve anything, or am I missing something?

domenukk:
It'll enable persistent mode in LibAFL for relevant targets, without the need to scan the binary or set an env

Issue #1546, opened 2022-10-07
cmplog colorization type_replace often not random


guyf2010:
The type_replace function used to in cmplog's colorization step often does not mutate randomly depending on the data in the current byte.

The switch statement at the following location contains many data values, but a large number of destinations in the switch do not apply any sort of randomness to their replacements. https://github.com/AFLplusplus/AFLplusplus/blob/7e0171006a32bf6b510e08df76ca6d8555272140/src/afl-fuzz-redqueen.c#L182-L254

In particular, any byte containing a value in the range from 128 to 254 will simply be XORed with 0x7F. I assume the intention was to XOR with a random value from 0 to 0x7F.

Additionally, the original RedQueen paper uses an EXT4 filesystem as a motivating example for colorization, where more than 60,000 instances of the value 0 were present, and colorization can greatly reduce the number of candidate replacements. However, the current cmplog implementation can only replace a value of 0 with 1, making it far less effective in this use case. 

In my testing the input-to-state stage for a single seed can run for extremely long periods of time (Several minutes to hours in some cases). Discovering seeds prone to this seems particularly common from the clone mutation with a constant value. https://github.com/AFLplusplus/AFLplusplus/blob/7e0171006a32bf6b510e08df76ca6d8555272140/src/afl-fuzz-one.c#L2476-L2517

I'm unsure whether this is a bug or an intentional trade-off made for certain test cases. If it is an intentional decision, perhaps some comments should be added to those switch cases explaining why colorization does not randomize those byte values. The type_replace does seem very focused on ASCII, maybe a config option can be added for cmplog to not use these types when a non-ASCII based program is being fuzzed.

Thanks,
Guy

vanhauser-thc:
Everything in that function is intentional, also the xor with a fixed value in specific cases.

I am on my phone so can’t write a thesis long rational now, but as with most in fuzzing it is guesswork and experience gathering with targets (fuzzbench).

if you want then make a pr that you think would be better and I do a fuzzbench comparison run.

Theories never help to know what is better in fuzzing, only real world testing :)

vanhauser-thc:
> if you want then make a pr that you think would be better and I do a fuzzbench comparison run.

@guyf2010 any thinking about this? would you want to try?

guyf2010:
> > if you want then make a pr that you think would be better and I do a fuzzbench comparison run.
> 
> @guyf2010 any thinking about this? would you want to try?

[...]

PR #1580, opened 2022-11-13
Add option for random cmplog colorization

guyf2010:
Added option to do random colorization for cmplog, rather than type based. Addresses issue #1546

### DIFF
--- a/src/afl-fuzz-redqueen.c
+++ b/src/afl-fuzz-redqueen.c
@@ -167,6 +167,13 @@ static u8 get_exec_checksum(afl_state_t *afl, u8 *buf, u32 len, u64 *cksum) {
 
 }
 
+/* replace everything with different values */
+static void random_replace(afl_state_t *afl, u8 *buf, u32 len){
+  for(u32 i=0; i < len; i++){
+    buf[i] = rand_below(afl, 256);
+  }
+}
+
 /* replace everything with different values but stay in the same type */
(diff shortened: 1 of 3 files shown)


src/afl-fuzz-redqueen.c
```diff
@@ -167,6 +167,17 @@ static u8 get_exec_checksum(afl_state_t *afl, u8 *buf, u32 len, u64 *cksum) {
 
 }
 
+/* replace everything with different values */
+static void random_replace(afl_state_t *afl, u8 *buf, u32 len) {
+
+  for (u32 i = 0; i < len; i++) {
+
+    buf[i] = rand_below(afl, 256);
```
vanhauser-thc: would it not be better to do something like:
```
for (...) {
  u8 c;
  do { c = rand_below(afl, 256);  } while (c == buf[i]);
  buf[i] = c;
}
```

vanhauser-thc:
one comment, otherwise this looks fine :)

src/afl-fuzz-redqueen.c
guyf2010: I think it depends on how successful replacement is going to be. If many bytes are identical and cannot be colorized, forcing those that were colorizable to be changed would be beneficial. If every byte is colorizable, then there's slightly more entropy in allowing 256 possible values rather than 255.

I suspect you're right in the majority of cases, I'll add the changes.

vanhauser-thc:
thank you, waiting for CI then I will merge.

PR #1583, opened 2022-11-16
Fix argv-fuzz.

kobrineli:
Hi! I tried to use `argv-fuzz-inl.h` to build target to fuzz argv, I executed binary and faced the problem that all the input from `stdin` is placed to `argv[0]`. And if I use `AFL_INIT_SET0`, the `argv[0]` is rewritten by passed program name, so nothing from `stdin` data stays in the `argv`.

I fix this problem with using `strtok` from `string.h` for command line arguments parsing.

Please, tell me, if I just didn't understand how to use this, I'll close the PR. Moreover, I didn't understand, why do we need to check `0x02` values.

### DIFF
--- a/utils/argv_fuzzing/argv-fuzz-inl.h
+++ b/utils/argv_fuzzing/argv-fuzz-inl.h
@@ -34,6 +34,7 @@
 #ifndef _HAVE_ARGV_FUZZ_INL
 #define _HAVE_ARGV_FUZZ_INL
 
+#include <string.h>
 #include <unistd.h>
 
 #define AFL_INIT_ARGV()          \
@@ -63,22 +64,22 @@ static char **afl_init_argv(int *argc) {
   char *ptr = in_buf;
   int   rc = 0;
 
-  if (read(0, in_buf, MAX_CMDLINE_LEN - 2) < 0) {}
+  ssize_t num = 0;
+  if ((num = read(0, in_buf, MAX_CMDLINE_LEN - 2)) < 0) {}
+  if (in_buf[num - 1] == '\n') {
+      in_buf[num - 1] = 0;
+  }
 
-  while (*ptr && rc < MAX_CMDLINE_PAR) {
-
-    ret[rc] = ptr;
+  char delim = ' ';
+  char *curarg = strtok(ptr, &delim);
+  while (curarg && rc < MAX_CMDLINE_PAR) {
+    ret[rc] = curarg;
     if (ret[rc][0] == 0x02 && !ret[rc][1]) ret[rc]++;
     rc++;
-
-    while (*ptr)
-      ptr++;
-    ptr++;
-
+    curarg = strtok(NULL, &delim);
   }
 
   *argc = rc;
-
   return ret;
 
 }
@@ -87,4 +88,3 @@ static char **afl_init_argv(int *argc) {
 #undef MAX_CMDLINE_PAR
 
 #endif                                              /* !_HAVE_ARGV_FUZZ_INL */
-
--- a/utils/argv_fuzzing/argv-fuzz-inl.h
+++ b/utils/argv_fuzzing/argv-fuzz-inl.h
@@ -34,7 +34,7 @@
 #ifndef _HAVE_ARGV_FUZZ_INL
 #define _HAVE_ARGV_FUZZ_INL
 
-#include <string.h>
+#include <stdlib.h>
 #include <unistd.h>
 
 #define AFL_INIT_ARGV()          \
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
@afflux that is your original code, can you have a look?

for the original code:
```
if (read(0, in_buf, MAX_CMDLINE_LEN - 2) < 0) {}
```
this is weird because now length is obtained which results in previous fuzz data being present making crashes potentially unreplayable.

from the new code:
```
 if ((num = read(0, in_buf, MAX_CMDLINE_LEN - 2)) < 0) {}
 if (in_buf[num - 1] == '\n') {
```
if num is 0 or -1 the in_buf access is out of bounds.
what is also problematic in the new code that it does not handle 0 bytes in the fuzz input

both codes suffer from not returning at once with a dummy argv if `read() <= 0`.

kobrineli:
Alright
I've added `read() <= 0` handling by setting `argc` to 0 and returning `ret`. I think that is correct, because in case of `AFL_INIT_SET0` the first arg would be set to the program name and `argc` would be set to 1.

afflux:
> @afflux that is your original code, can you have a look?

not mine, but lcamtuf's. I've just built argvfuzz.so to make it work on binaries with LD_PRELOAD.

That being said, I don't see why the original code is not working for you. The current code splits on null-bytes, as documented in lines 24-27.

The original code is indeed suffering from not properly handling empty inputs, so my suggestion would be something along these lines:

```
num = read(0, in_buf, MAX_CMDLINE_LEN - 2);
if (num < 0) {
  perror("I think fuzzing shouldn't cotinue on on read error...?");
  abort();
}
in_buf[num] = '\0';
in_buf[num+1] = '\0';
```

(side note: I think `strtok` wants a null-terminated `const char*` for `delim`, so " " instead of &' ')

Yes, `AFL_INIT_SET0` discards the first argument, but you should still see fuzzed data on the following `argc`s

kobrineli:
@afflux
The problem is that there are no following `args` in `argv`.


    while (*ptr && rc < MAX_CMDLINE_PAR) {

        ret[rc] = ptr;
        if (ret[rc][0] == 0x02 && !ret[rc][1]) ret[rc]++;
        rc++;

        while (*ptr)
            ptr++;
        ptr++;

    }

Here `ptr`  points to the same data as `in_buf`, so there is stored all the input data. We assign `ret[rc]` equal to `ptr`, where `ret[rc]` corresponds to `argv[0]` ('cause `rc` equals 0 at this time). And then in the nested `while`  `ptr` is increased until `*ptr` equals 0. Then `ptr` is incremented, and `*ptr` equals 0 as well. Right after it we check the outer loop condition, discover that `*ptr` equals 0 and leave the loop, assigning `*argc` to 0 and returning `ret` with only the first element filled.

afflux:
> Then ptr is incremented, and *ptr equals 0 as well. 

After the the [increment](https://github.com/AFLplusplus/AFLplusplus/blob/8cdc48f73a17ddd557897f2098937a8ba3bfe184/utils/argv_fuzzing/argv-fuzz-inl.h#L76), *ptr points to the byte *following* the 0 byte.
The idea is to separate this fuzzing input:

```
const char *in_buf = "AAAA\0BBBB\0CCCC\0\0";
```

Into this argv:
```
argv[0] = &in_buf[0]; /* "AAAA\0..." */
argv[1] = &in_buf[5]; /* "BBBB\0..." */
argv[2] = &in_buf[10]; /* "CCCC\0..." */
argc = 3;
```

kobrineli:
Oh, I just didn't understand the input data format. Now that works, thank you very much! I think that it makes sense to fix `read()` error handling then.

utils/argv_fuzzing/argv-fuzz-inl.h
```diff
@@ -63,7 +64,12 @@ static char **afl_init_argv(int *argc) {
   char *ptr = in_buf;
   int   rc = 0;
 
-  if (read(0, in_buf, MAX_CMDLINE_LEN - 2) < 0) {}
+  ssize_t num = read(0, in_buf, MAX_CMDLINE_LEN - 2);
+  if (num < 0) {
+      abort();
```
vanhauser-thc: this should be `exit` not abort.

utils/argv_fuzzing/argv-fuzz-inl.h
```diff
@@ -63,7 +64,12 @@ static char **afl_init_argv(int *argc) {
...
+  }
+  in_buf[num] = '\0';
```
vanhauser-thc, line 71: this is missing a `if (num == 0) ...` returning argc of 0.

afflux: I don't think so. With empty input, we have `in_buf = "\0\0"`. The while condition is false, hence you already get argc = 0

utils/argv_fuzzing/argv-fuzz-inl.h
afflux: agree, makes way more sense. Sorry!

kobrineli: fixed

utils/argv_fuzzing/argv-fuzz-inl.h
kobrineli: Yeah, agree with @afflux

vanhauser-thc:
thank you guys!

PR #1606, opened 2022-12-23
afl-untracer freebsd 13.1 and above update.

devnexen:
with the new kern.elf*.allow_wx setting, we try
 to make sure we still can make the maps w/x,
fixing few build warnings while at it.

### DIFF
--- a/utils/afl_untracer/afl-untracer.c
+++ b/utils/afl_untracer/afl-untracer.c
@@ -156,7 +156,7 @@ void read_library_information(void) {
         *e = 0;
         if (n[strlen(n) - 1] == '\n') n[strlen(n) - 1] = 0;
 
-        liblist[liblist_cnt].name = strdup(n);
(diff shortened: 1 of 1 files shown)


utils/afl_untracer/afl-untracer.c
```diff
@@ -156,7 +156,7 @@ void read_library_information(void) {
         *e = 0;
         if (n[strlen(n) - 1] == '\n') n[strlen(n) - 1] = 0;
 
-        liblist[liblist_cnt].name = strdup(n);
+        liblist[liblist_cnt].name = strdup((char *)n);
```
domenukk: Should the result of strdup be casted to `u8*` then?

devnexen: yes it s fair.

vanhauser-thc:
thank you!

PR #1608, opened 2022-12-28
Support for clang-format from pip in the .custom-format.py

ahpaleus:
Hello,
This pull request adds support for the clang-format installed via pip. It makes contributing more convenient.


vanhauser-thc:
thank you!

PR #1609, opened 2022-12-30
Fix kill signals configuring.

kobrineli:
Hi! I've found the bug related with configuring afl kill signals for QEMU and unicorn modes in `afl-showmap.c`.
The `configure_afl_kill_signals` function is called only under condition `!fsrv->cs_mode && !fsrv->qemu_mode && !unicorn_mode`, and as a result no signals for QEMU and unicorn modes are set, and `afl-showmap` (and `afl-cmin`) hangs. Fixed the error by moving configuring out of this scope.

### DIFF
--- a/src/afl-showmap.c
+++ b/src/afl-showmap.c
@@ -1305,10 +1305,6 @@ int main(int argc, char **argv_orig, char **envp) {
                                  : 0);
     be_quiet = save_be_quiet;
 
-    configure_afl_kill_signals(
-        fsrv, NULL, NULL,
-        (fsrv->qemu_mode || unicorn_mode) ? SIGKILL : SIGTERM);
-
     if (new_map_size) {
 
       // only reinitialize when it makes sense
@@ -1332,6 +1328,10 @@ int main(int argc, char **argv_orig, char **envp) {
     fsrv->map_size = map_size;
 
   }
+
+  configure_afl_kill_signals(
+      fsrv, NULL, NULL,
+      (fsrv->qemu_mode || unicorn_mode) ? SIGKILL : SIGTERM);
 
   if (in_dir) {
 


vanhauser-thc:
you cant move the location, the forkserver is then already started.
but what is the issue? you usually do not want to kill the main qemu if there is a target on a timeout

kobrineli:
@vanhauser-thc 
I dont understand the point of locating kill signals configuring for qemu and unicorn modes under condition `!fsrv->cs_mode && !fsrv->qemu_mode && !unicorn_mode`, which doesn't allow to set kill signals for these modes.
The issue is that the corpus minimization for the target with no timeout  inputs just cannot be ended, because the forkserver tries to be killed with the zero signal (because no signal was set) and the program hangs on `wait`.

kobrineli:
Here is the example of running `afl-showmap` without changes from these PR.

    (gdb) b src/afl-forkserver.c:1249
    Breakpoint 1 at 0x409daa: file src/afl-forkserver.c, line 1249.
    (gdb) run
    Starting program: /usr/local/bin/afl-showmap -t 10000 -o out -i /corpus/ -Q -- /load_sydr @@
    warning: Error disabling address space randomization: Operation not permitted
    [Thread debugging using libthread_db enabled]
    Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
    afl-showmap++4.05a by Michal Zalewski
    [*] Executing '/load_sydr'...
    [*] Reading from directory '/corpus/'...
    [*] Spinning up the fork server...
    [Detaching after fork from child process 478]
    [+] All right - fork server is up.
    [*] Target map size: 65536
    [*] Scanning '/corpus/'...
    -- Program output begins --
    -- Program output ends --
    ...
    -- Program output begins --
    -- Program output ends --
    [!] WARNING: Test case '/corpus//large.xlsx' is too big (1.84 MB, limit is 1.00 MB), partial reading
    [!] WARNING: Input file '/corpus//large.xlsx' is too large, only reading 1048576 bytes.
    -- Program output begins --
    -- Program output ends --
    -- Program output begins --
    -- Program output ends --
    [!] WARNING: Test case '/corpus//very_large.xlsx' is too big (4.13 MB, limit is 1.00 MB), partial reading
    [!] WARNING: Input file '/corpus//very_large.xlsx' is too large, only reading 1048576 bytes.
    -- Program output begins --
    -- Program output ends --
    [+] Processed 28 input files.
    [+] Captured 9867 tuples (map size 65536, highest value 8, total values 792761) in 'out'.
    
    Breakpoint 1, afl_fsrv_kill (fsrv=0x7ffc6d46cad0) at src/afl-forkserver.c:1249
    1249	    waitpid(fsrv->fsrv_pid, NULL, 0);
    (gdb) p fsrv->fsrv_kill_signal
    $1 = 0 '\000'
    (gdb) c
    Continuing.
    ^C
    Program received signal SIGINT, Interrupt.
    0x00007fcc05b2bc3a in wait4 () from /lib/x86_64-linux-gnu/libc.so.6
    (gdb) bt
    #0  0x00007fcc05b2bc3a in wait4 () from /lib/x86_64-linux-gnu/libc.so.6
    #1  0x0000000000409db6 in afl_fsrv_kill (fsrv=0x7ffc6d46cad0) at src/afl-forkserver.c:1249
    #2  0x0000000000409e09 in afl_fsrv_deinit (fsrv=0x7ffc6d46cad0) at src/afl-forkserver.c:1687
    #3  0x0000000000406421 in main (argc=11, argv_orig=0x0, envp=<optimized out>) at src/afl-showmap.c:1482
    (gdb)

Here I run `afl-showmap` under the gdb. As you can see, all the inputs were processed and there were no timeouts. But in the end the program hanged, because `afl_fsrv_kill` failed to kill forkserver (because `fsrv->fsrv_kill_signal` is 0) and the program hanged on `wait`. With changes from this PR the signal is set to 9 and the forkserver is successfully killed in the end of `afl-showmap` executing.

vanhauser-thc:
I tried a fix in the dev branch, can you test it please?

kobrineli:
Now that works and doesn't hang, thanks! But what is the principal difference between this PR changes and your fix? Does it matter where to locate `configure_afl_kill_signals`: before `if` scope or after?

kobrineli:
Oh, ok, I've understood. Should I close this PR then or can I move configuring before scope to merge this PR?

vanhauser-thc:
the fix is in and this needs to be closed.
the signals need to be set before the forkserver is started.
the bug was that the kill signals were only set when not in qemu etc. mode, so it was in the wrong scope.
thanks for reporting!

PR #1621, opened 2023-01-20
LLVM plugin 16+ support proposal.

devnexen:
- Lifting the standard to C++17.
- Beyond the cosmetic changes, it boils down to BasicBlock::getInstList being no longer available (and reading the header it is no accident).

### DIFF
--- a/instrumentation/split-switches-pass.so.cc
+++ b/instrumentation/split-switches-pass.so.cc
@@ -225,12 +225,20 @@ BasicBlock *SplitSwitchesTransform::switchConvert(
   BasicBlock  *NewNode = BasicBlock::Create(Val->getContext(), "NodeBlock", F);
   Shift = BinaryOperator::Create(Instruction::LShr, Val,
                                  ConstantInt::get(ValType, smallestIndex * 8));
+#if LLVM_VERSION_MAJOR >= 16
+  Shift->insertInto(NewNode, NewNode->end());
+#else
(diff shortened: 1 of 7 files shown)


vanhauser-thc:
oh thank you so much for this!
I will have a look on Monday :)

vanhauser-thc:
LGTM. is anything missing? (PR is a draft)

devnexen:
No the todo can be seen later finally

instrumentation/split-switches-pass.so.cc
```diff
@@ -452,7 +468,11 @@ bool SplitSwitchesTransform::splitSwitches(Module &M) {
     BranchInst::Create(SwitchBlock, OrigBlock);
 
     /* We are now done with the switch instruction, delete it. */
+#if LLVM_VERSION_MAJOR >= 16
+    // TODO to erase range of instructions instead ?
```
vanhauser-thc: I think that is the right thing to do.
to be sure you could check with a switch example for llvm < 16 and llvm >= 16 and see if the same instructions are generated?
I did not code this, so I do not know that well what happens here exactly and why ...

vanhauser-thc:
did you test that it does the same? :)

devnexen:
I did not notice differences in my test with clang 15 and 16

vanhauser-thc:
thank you a lot! those llvm changes always drive me crazy as they usually also give zero support how fix the changes in code.

PR #1623, opened 2023-01-25
Fix FRIDA mode for OSX x64

WorksButNotTested:


vanhauser-thc:
is that all that is needed? 
and it doesnt break old macos?

WorksButNotTested:
It needed a tweak for OSX on arm64. 10.9 is Maverick, but that's pretty old anyway right?

PR #1634, opened 2023-02-03
Explicitly print error code if sched_setaffinity fails

nataraj-hates-MS-for-stealing-github:
I found `afl-gotcpu` does not work properly in lxc containers (May be it should not, this is not the issue now).

While figuring out what's wrong with it, I've make simple addition to `afl-gotcpu.c`, may be you will find it useful in main code...


vanhauser-thc:
not sure but why not :)

vanhauser-thc:
(note that we only accept PRs to dev, not stable, changed this for you)

Issue #1625, opened 2023-01-26
Bad performance when fuzzing Python native extensions in Qemu mode


stulle123:
Hi there!

I'm trying to fuzz Python native extensions in Qemu mode. In normal LLVM fork-based mode everything works fine, but when I try to fuzz the native extension in binary-only Qemu mode I'm facing terrible performance issues (only 16.73/sec exec speed):

`LD_PRELOAD=~/AFLplusplus/qemu_mode/fastexit/fastexit64.so ~/AFLplusplus/afl-fuzz -i in/ -o out/ -Q -- ./afl_fuzz.a`

I think I'm facing a similar issue as described [here](https://github.com/AFLplusplus/AFLplusplus/issues/1547), however the `LD_PRELOAD` "speed hack" doesn't give me any performance boost.

**My setup**

I'm on Ubuntu 20.04 x86_64 running AFL++ on the current `dev` branch.

**fuzz_target.c**

```
#define PY_SSIZE_T_CLEAN
#include <Python.h>

#pragma clang optimize off

static PyObject *corruption(PyObject* self, PyObject* args) {
    char arr[3];
    Py_buffer name;

    if (!PyArg_ParseTuple(args, "y*", &name))
        return NULL;

    if (name.buf != NULL) {
        if (strcmp(name.buf, "FUZZ") == 0) {
            arr[0] = 'B';
            arr[1] = 'O';
            arr[2] = 'O';
            arr[3] = 'M';
        }
    }

    PyBuffer_Release(&name);
    Py_RETURN_NONE;
}

static PyMethodDef MemoryMethods[] = {
    {"corruption", corruption, METH_VARARGS, "BOOM!"},
    {NULL, NULL, 0, NULL}
};

static struct PyModuleDef memory_module = {
    PyModuleDef_HEAD_INIT,
    "memory",
    "BOOM!",
    -1,
    MemoryMethods
};

PyMODINIT_FUNC PyInit_memory(void) {
    return PyModule_Create(&memory_module);
}
```

**setup.py**

```
from distutils.core import setup, Extension

module = Extension("memory", sources=["fuzz_target.c"])

setup(
    name="memory",
    version="1.0",
    description='A simple "BOOM!" extension',
    ext_modules=[module],
)
```

**fuzz_harness.c**

```
#include <Python.h>

#ifndef __AFL_FUZZ_TESTCASE_LEN

ssize_t       fuzz_len;
unsigned char fuzz_buf[1024000];

[...]

PR #1635, opened 2023-02-06
Add Qemu Deferred Initialization Example for Binary-Only Targets

stulle123:
Hi there,

This PR adds documentation that explains how to use LLVM mode with deferred initialization for binary-only targets.

See issue https://github.com/AFLplusplus/AFLplusplus/issues/1625 for some background information.

Hope this is helpful!

Cheers


vanhauser-thc:
thank this is helpful!

One request though: please change  `fuzz_harness.a` to `fuzz_harness`.
A `.a` file extension is for an archive of object files (a library compiled for static compilation), so that naming is asking for trouble.

stulle123:
Thanks for the quick response! Good spot! Just renamed all occurrences to `fuzz_harness`.

vanhauser-thc:
thank you!

PR #1639, opened 2023-02-07
Increase fuzz_level for mopt_common_fuzzing

### LINKED ISSUE #1633: Linear power schedule
AFL++ hangs in endless loop without new corpus finds if "-L 0 -p lin" combination specified. It also make afl-fuzz produce very huge logs (2Tb for 12h).

Is it a bug or linear power schedule is incompatible with MOpt? Does it worth using in parallel fuzzing? There are no `quad` and `lin` schedules in clusterfuzz launcher (https://github.com/google/clusterfuzz/blob/fe38978a76d81701f2c144933ad94c4be203e68c/src/clusterfuzz/_internal/bot/fuzzers/afl/launcher.py#L324).

apach301:
Trying to fix #1633 

One of the possible fix is to prevent factorizing perf_score for unfuzzed seeds like it is done for other power schedules. These changes fixes hanging in a fuzzing loop without any mutations performed.

Also I discovered that fuzz_level of testcases doesn't increased when MOpt mode is enabled (-L 0 options, `mopt_common_fuzzing` function). This change itself also fixes the original issue (even without changing factorization strategy for lin and quad),

As I understand, the bugfixes in 03e6d33a4044115c44afeb6c1ae735c0310018af and 6596284cc41484ec5062ca53109ec5bd7899e56f not needed anymore and could be reverted

### DIFF
--- a/src/afl-fuzz-queue.c
+++ b/src/afl-fuzz-queue.c
@@ -1007,10 +1007,16 @@ u32 calculate_score(afl_state_t *afl, struct queue_entry *q) {
       break;
 
     case LIN:
+      // Don't modify perf_score for unfuzzed seeds
+      if (!q->fuzz_level) break;
+
       factor = q->fuzz_level / (afl->n_fuzz[q->n_fuzz_entry] + 1);
       break;
 
     case QUAD:
+      // Don't modify perf_score for unfuzzed seeds
+      if (!q->fuzz_level) break;
+
       factor =
           q->fuzz_level * q->fuzz_level / (afl->n_fuzz[q->n_fuzz_entry] + 1);
       break;
--- a/src/afl-fuzz-one.c
+++ b/src/afl-fuzz-one.c
@@ -5683,6 +5683,7 @@ pacemaker_fuzzing:
 
   }                                                                /* block */
 
+  ++afl->queue_cur->fuzz_level;
   return ret_val;
 
 }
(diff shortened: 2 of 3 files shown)


src/afl-fuzz-queue.c
```diff
@@ -1007,10 +1007,16 @@ u32 calculate_score(afl_state_t *afl, struct queue_entry *q) {
...
+      // Don't modify perf_score for unfuzzed seeds
+      if (!q->fuzz_level) break;
+
       factor = q->fuzz_level / (afl->n_fuzz[q->n_fuzz_entry] + 1);
       break;
 
     case QUAD:
+      // Don't modify perf_score for unfuzzed seeds
+      if (!q->fuzz_level) break;
+
```
vanhauser-thc, line 1019: why only for for lin and quad, why not for all?

apach301: Factorization is only performed for exploit, fast, coe, lin, and quad schedules. And fuzz_level checked in all of them except lin and quad.

apach301: I don't know how other power schedules should handle seeds with fuzz_level=0, but for lin and quad perf_score will be turned to a null in this case. And when combined with MOpt score would not be bounded to a 1 at https://github.com/AFLplusplus/AFLplusplus/blob/f01bf77604c7d5b2ee12a44cbb3dbd615ad9565e/src/afl-fuzz-queue.c#L1058

vanhauser-thc:
thank you!

PR #1643, opened 2023-02-10
Cmplog should search for unprefixed target in PATH, not PWD

YSaxon:
This fixes a bug in cmplog, where, if afl-fuzz is called with an unprefixed-path target (eg `-- targetapp` instead of `-- /path/to/targetapp`), while afl-fuzz itself will correctly look for `target` in `PATH`, cmplog instead looks for it in `PWD`.

If the target does not happen to be in PWD, then it will fail with the confusing error message: 
```
[-] PROGRAM ABORT : Unable to execute target application ('targetapp')
         Location : afl_fsrv_start(), src/afl-forkserver.c:1077
```

This bug is present in standard instrumented mode and Frida mode, but not QEMU mode.

Inasmuch as it violates the general principle and expectation that unprefixed executables should be executed by PATH rather than PWD, this bug being fixed also constitutes a security vulnerability, which could theoretically be exploited by placing a malicious executable of the same name into the user's working directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,7 @@ Thank you! (For people sending pull requests - please add yourself to this list
     Thomas Rooijakkers                    David Carlier
     Ruben ten Hove                        Joey Jiao
     fuzzah                                @intrigus-lgtm
+    Yaakov Saxon
   ```
 
 </details>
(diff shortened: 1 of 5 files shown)


vanhauser-thc:
thanks for the PR, letting the CI check this ... :)

PR #1645, opened 2023-02-13
LLVM cmplog factoring custom Instruction iterator with added restrict…

devnexen:
…ion.


vanhauser-thc:
Thanks! Are the changes for qemu and unicorn on purpose?

devnexen:
ah not at all I often forget those sub git references

PR #1653, opened 2023-02-23
push to stable

vanhauser-thc:


custom_mutators/autotokens/README
```diff
@@ -0,0 +1,33 @@
+# autotokens
```
domenukk: Capital A?

docs/custom_mutators.md
```diff
@@ -132,13 +136,21 @@ def deinit():  # optional for Python
     for a specific queue entry, use this function. This function is most useful
     if `AFL_CUSTOM_MUTATOR_ONLY` is **not** used.
 
+- `splice_optout` (optional):
```
domenukk, line 139: How can this possibly work in python? Wouldn't the fn always be available?

docs/custom_mutators.md
```diff
@@ -48,6 +48,7 @@ C/C++:
 ```c
 void *afl_custom_init(afl_state_t *afl, unsigned int seed);
 unsigned int afl_custom_fuzz_count(void *data, const unsigned char *buf, size_t buf_size);
+void afl_custom_splice_optout(void *data);
```
domenukk, line 51: Rust implementation?

src/afl-gotcpu.c
```diff
@@ -214,7 +214,13 @@ int main(int argc, char **argv) {
   #if defined(__linux__)
       if (sched_setaffinity(0, sizeof(c), &c)) {
 
-        PFATAL("sched_setaffinity failed for cpu %d", i);
+        const char *error_code = "Unkown error code";
+        if (errno == EFAULT) error_code = "EFAULT";
```
domenukk, line 218: switch case for legibility?

test/test-llvm.sh
```diff
@@ -257,12 +257,13 @@ test -e ../afl-clang-fast -a -e ../split-switches-pass.so && {
   }
   rm -f test-compcov test.out instrumentlist.txt
   AFL_LLVM_CMPLOG=1 ../afl-clang-fast -o test-cmplog test-cmplog.c > /dev/null 2>&1
+  ../afl-clang-fast -o test-c test-cmplog.c > /dev/null 2>&1
```
domenukk, line 260: Maybe add test-c to `.gitignore`?

domenukk (approved):
Not sure if the python custom mutator still works as intended, else looks good.
Rust bindings need support for the new custom mutator functions eventually

docs/custom_mutators.md
vanhauser-thc: why would the function be available when it is not in the python script?  I dont know python really, but logically it does not make sense to me :)

docs/custom_mutators.md
vanhauser-thc: uh can you do that? :) not sure what has to be done. pretty sure afl_custom_send is missing too

src/afl-gotcpu.c
vanhauser-thc: dunno what you mean :)

test/test-llvm.sh
vanhauser-thc: ok

src/afl-gotcpu.c
domenukk: Like
```c
switch (errno) {
    case EFAULT:
        error_code = "EFAULT";
        break;
    case ...
```

docs/custom_mutators.md
domenukk: Do you check somewhere in the code if it's set in python?

src/afl-gotcpu.c
vanhauser-thc: was a PR, nobody is using the tool really ... yes that would be nicer, but I like to invest my time better :)

docs/custom_mutators.md
vanhauser-thc: fixed

PR #1654, opened 2023-02-24
Redqueen typo?

tokatoka:
probably this is a typo?
Should be `o_diff` not `diff`


vanhauser-thc:
you are right, thanks!
please note that we only accept PRs to dev, I changed that for you.

PR #1662, opened 2023-03-02
Fix exit on time

kobrineli:
Hi! I've realized that my fix for exit on time #1555 was only for the pizza mode and, moreover, incorrect. Here I fix exit on time again and in both modes, and now that works.

By the way, @domenukk didn't notice as well that fix #1555 was only for the pizza mode. @vanhauser-thc, does AFL++ still need supporting pizza mode with the exact copy of normal mode code? It can be really misleading.

### DIFF
--- a/src/afl-fuzz-stats.c
+++ b/src/afl-fuzz-stats.c
@@ -669,9 +669,15 @@ void show_stats_normal(afl_state_t *afl) {
 
   /* AFL_EXIT_ON_TIME. */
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
I am busy with a lot of stuff - are you sure the exit on time conditions are correct now? because I cannot check :)

SweetVishnya (approved):
LGTM

kobrineli:
@vanhauser-thc, yes, they are :) I've checked on the project with the little exit on time, where new coverage cannot be found from the very beginning.

PR #1667, opened 2023-03-05
Added env variable to allow custom interval update of fuzzer_stats file

amitelka:
I wanted to allow custom interval for dumping stats into fuzzer_stats file (as I'd like to use this in my own fuzzing automations instead of parsing stdout, etc.)


vanhauser-thc:
PRs need to be for the dev branch, I changed this for you.

i am not sure it is that useful as the updates will not be exact and with slow targets it can take seconds until there is a slice for the time test, but well that is a user problem :)

I am fine with the idea, what I do not like is sacrificing a command line parameter for this. Please change this to use an env var instead.

amitelka:
> PRs need to be for the dev branch, I changed this for you.
> 
> i am not sure it is that useful as the updates will not be exact and with slow targets it can take seconds until there is a slice for the time test, but well that is a user problem :)
> 
> I am fine with the idea, what I do not like is sacrificing a command line parameter for this. Please change this to use an env var instead.

I assumed there'd be a timing error but I see the usage of this even with the non-exact flushing.

I moved the CLI flag to be an env variable with sufficient checks excluding overflow due to sec to msec conversion (which might happen if someone uh, never wants the file to be flushed..)

src/afl-fuzz-state.c
```diff
@@ -565,6 +567,26 @@ void read_afl_environment(afl_state_t *afl, char **envp) {
 
             }
 
+          } else if (!strncmp(env, "AFL_FUZZER_STATS_UPDATE_INTERVAL",
+
+                              afl_environment_variable_len)) {
+
+            u64 stats_update_freq_sec =
+                strtoull(get_afl_env(afl_environment_variables[i]), NULL, 0);
+            if (ULLONG_MAX == stats_update_freq_sec ||
```
vanhauser-thc: if this is ULLONG_MAX - 1 and is then multiplied by 1000 then we have an overflow right? better `if >= UINT_MAX`?

vanhauser-thc:
one minor issue, otherwise it is ready to merge :)

src/afl-fuzz-state.c
amitelka: Yeah, correct. Fixed.

vanhauser-thc:
thank you!

PR #1680, opened 2023-03-25
add check for n_fuzz update

yuawn:
Check for AFLFast before update `afl->n_fuzz`, or fuzzer will crash when `afl->schedule >= FAST && afl->schedule <= RARE` is not true (e.g. `afl-fuzz -p exploit`) because `afl->n_fuzz` is not initialized:
https://github.com/AFLplusplus/AFLplusplus/blob/24503fba5fd2580559223ec3c6ee408dfa15e080/src/afl-fuzz.c#L1555-L1559


vanhauser-thc:
Thank you! Can you please add a likely() to the if cksum

vanhauser-thc:
Thank you :)

PR #1683, opened 2023-03-27
Clarify confusing version message

cdown:
When running, the following gets printed in quick succession on startup:

    afl-fuzz++4.00c based on afl by Michal Zalewski and a large online community
    [...]
    [+] NOTE: This is v3.x which changes defaults and behaviours - see README.md

Don't assert that this is v3, just that v3+ changes defaults and behaviours.


vanhauser-thc:
thank you :)

PR #1686, opened 2023-03-28
Minor fixes

tuliom:
I found a few issues while building AFLplusplus against Fedora Rawhide with LLVM 16.0.0 and Python 3.11.


vanhauser-thc:
thank you!

Issue #1688, opened 2023-03-31
when use lto mode, ld.lld Assertion failed.


nj00001:
**IMPORTANT**
4.05c and dev  branch can reproduce
ubuntu 22.04
llvm 14 and clang 14

**To Reproduce**
target: php-8.2.4 tag

build-command:
apt install re2c libsqlite3-dev
./buildconf --force
CC=afl-clang-lto CXX=afl-clang-lto++ RANLIB=llvm-ranlib-14 AR=llvm-ar-14 ld=afl-ld-lto ./configure
make -j6

**Expected behavior**
when ld.lld link the target object, SanitizerCoverageLTO.so plugin crashed. No error should occur under normal circumstances

**Screen output/Screenshots**
```
clang: error: unable to execute command: Aborted (core dumped)
clang: error: linker command failed due to signal (use -v to see invocation)
make: *** [Makefile:300：sapi/phpdbg/phpdbg] 错误 254
afl-llvm-lto++4.06a by Marc "vanHauser" Heuse <mh@mh-sec.de>
ld.lld: /usr/lib/llvm-14/include/llvm/IR/User.h:170: llvm::Value *llvm::User::getOperand(unsigned int) const: Assertion `i < NumUserOperands && "getOperand() out of range!"' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
[...]

PR #1692, opened 2023-04-03
fix getOperand out of range

nj00001:
This commit fixes a sloppy function call that should normally check the number of insn's Operand before calling insn's getOperand method. The fix is that if it is 0 it should continue the loop. I solved problem https://github.com/AFLplusplus/AFLplusplus/issues/1688#issue-1648543691 using this modification

### DIFF
--- a/instrumentation/afl-llvm-common.cc
+++ b/instrumentation/afl-llvm-common.cc
@@ -289,6 +289,9 @@ void scanForDangerousFunctions(llvm::Module *M) {
 
     StringRef ifunc_name = IF.getName();
     Constant *r = IF.getResolver();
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
I have never seen that this occurs, but apparently you have and it makes sense to have this check. thank you!

nj00001:
> I have never seen that this occurs, but apparently you have and it makes sense to have this check. thank you!

Yes, it's strange, I also use lto mode a lot, but it's the first time I've encountered this, I've also tried it on other ubuntu servers, but they all have this problem, so I think it might be caused by some optimization code used by the php target

PR #1693, opened 2023-04-08
display instance name

hardik05:
change to display instance name like master,slave1 etc... with afl-whatsup


vanhauser-thc:
looks good, thanks.
for the future note that all PRs must be for the dev branch.

PR #1694, opened 2023-04-08
afl-cc: Avoid casts of string literals to char*, in definition of __AFL_INIT() etc.

neuschaefer:
With the right -W options, compilers may complain about the cast of string literals (for PERSIST_SIG and DEFER_SIG) to (char*), and they're right to do so, because string literals are constant. Since some projects enable -Werror, this can lead to a broken build with afl-cc.

Let's simply cast to (const char *), which preserves the constness of the string literal.


vanhauser-thc:
cannot hurt for sure, thank you

PR #1697, opened 2023-04-10
LLVM instrumentation disable build warning.

devnexen:
Since clang 16 is the version for Ubuntu 23 04/Fedora 38 and is easy enough to fix..


vanhauser-thc:
Thank you!

PR #1702, opened 2023-04-14
add Nyx support in afl-showmap, afl-tmin, afl-cmin and afl-analyze

schumilo:
This PR is still work in progress, but everything seems to work so far. 
However, the current patches lack a proper implementation of a helper function to remove a temporarily created Nyx work directory safely. I also plan to switch to a more recent version of libnyx to add some other useful features (like the option to re-use an already created Nyx fuzzing snapshots; to avoid the bootstrapping procedure of the target and its environment in the VM and to improve reproducibility in some specific cases).


vanhauser-thc:
looks good let's try this :)

PR #1703, opened 2023-04-14
Fixing typo: & (background) to && (and)

YSaxon:
Presently the actual outcome of the test is being ignored by using & (running the test in background) instead of && (conditioning response on outcome of test). This is clearly a typo


vanhauser-thc:
Thank you!

PR #1710, opened 2023-04-21
frida mode: add dynamic loaded code exclusion

### LINKED ISSUE #1708: Simply exclude runtime loaded code from instrumentation with Frida mode

I am currently experiencing a problem with the Frida mode in Android. My harness calls the Android runtime to setup an environment that allows targeting a JNI function. The most convenient way is to use dlopen to load the library so I don't have to link it at compile time. The Android runtime is something quite large with many dependencies, so I don't want to instrument this part. However, currently it doesn't seem possible to simply prevent instrumentation of the loaded code at runtime (and potentially its dependencies).

It might be handy with Frida mode to be able to exclude from instrumentation all code that is loaded at runtime so that instrumentation is only kept on what is specified in the Fuzzer configuration.

I think the most convenient way to do this would be to keep the current default behavior, but offer the user to explicitly exclude code loaded at runtime with an environment variable.

The variable could be : `AFL_FRIDA_INST_NO_DYNLOAD` and the associated JS function `setInstrumentNoDynamicLoad()`.

What do you think about it? I could do the PR with pleasure.

Ha0ris:
Add the AFL_FRIDA_INST_NO_DYNAMIC_LOAD environment variable and its associated JS function setInstrumentNoDynamicLoad to prevent the instrumentation of late dynamic loaded code.

Resolve #1708


vanhauser-thc:
thanks!
handing over to @WorksButNotTested  :)

WorksButNotTested:
The PR looks great to me. But the CI seems to be failing (I think unrealated)?

vanhauser-thc:
yes the ci failures is do a different change and an old gcc ... fixed now but I cannot convince the ci to try against the new dev branch state ... weird.

PR #1714, opened 2023-04-25
 Add env var to ignore coverage from dynamically loaded code after forkserver

choller:
When using TRACEPC instrumentation, loading code dynamically (e.g. through dlopen()) it can be useful to completely ignore the loaded code, esp. when it cannot be preloaded and is not the target to be tested. This patch allows setting AFL_LLVM_IGNORE_PROBLEMS_COVERAGE=1 to do so.


vanhauser-thc:
thank you!

PR #1718, opened 2023-04-27
afl-showmap: Start a only a single fork server

nick-potenski:
A forkserver is started by afl_fsrv_get_mapsize() when dynamically finding the map size.  When an input directory option is specified a second fork server was also started.  This commit re-arranges the inits for several forkserver struct members so that we can re-use the server started by the get_mapsize() call when not in coresight/qemu/unicorn modes and just start the server otherwise.

The duplicate forkservers lead to two copies of the program-under-test running simultaneously.  In my case, I have an application  that tries to reserve a specific port before the deferred init (__AFL_INIT()).  The second attempt to reserve the port will fail.  It is too troublesome to try to remove the port access or otherwise make it fault tolerant, and it didn't seem like afl-showmap really needed two fork servers anyway.

### DIFF
--- a/src/afl-showmap.c
+++ b/src/afl-showmap.c
@@ -1421,6 +1421,14 @@ int main(int argc, char **argv_orig, char **envp) {
     // If @@ are in the target args, replace them and also set use_stdin=false.
     detect_file_args(argv + optind, stdin_file, &fsrv->use_stdin);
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
looks ok, but I hardly ever look into afl-showmap, lets hope this doesnt break anything :)

PR #1720, opened 2023-04-28
Source Code Coverage support for Nyx (Part 1)

choller:
This commit adds core code in the AFL++ runtime and compiler to measure source code coverage in Nyx mode.

This PR does not yet include:

* Changes required to the Nyx preload code (in particular allocating the pcmap buffer)
* Example code how to dump the pcmap, the module info and the coverage in the Nyx preload code
* External tooling to post-process the resulting data


vanhauser-thc:
looks good, thank you!
I will document the new option build option in a few places where it is mising.

Issue #1721, opened 2023-04-30
Uninitalized values of effector map is used


wtdcode:
This block of code:

https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-fuzz-one.c#L843-L852

```C
eff_map = afl_realloc(AFL_BUF_PARAM(eff), EFF_ALEN(len));
```

`eff_map` is allocated from `afl_realloc` which is backed by `realloc`. However, without any initialization, it's accessed by:

```C
(EFF_APOS(len - 1) != 0) 
```

and notably many conditions check afterwards like

```C
if (!eff_map[EFF_APOS(afl->stage_cur)])
```

Valgrind also complains a lot about jumping depending on uninitialized values.

My fix is adding one line to `afl_realloc` but probably is not what you wish:

```c
static inline void *afl_realloc(void **buf, size_t size_needed) {

  struct afl_alloc_buf *new_buf = NULL;

  size_t current_size = 0;
  size_t next_size = 0;

  if (likely(*buf)) {

    /* the size is always stored at buf - 1*size_t */
    new_buf = (struct afl_alloc_buf *)afl_alloc_bufptr(*buf);
    current_size = new_buf->complete_size;

  }

  size_needed += AFL_ALLOC_SIZE_OFFSET;

  /* No need to realloc */
  if (likely(current_size >= size_needed)) { return *buf; }

  /* No initial size was set */
  if (size_needed < INITIAL_GROWTH_SIZE) {

    next_size = INITIAL_GROWTH_SIZE;

  } else {

    /* grow exponentially */
    next_size = next_pow2(size_needed);

    /* handle overflow: fall back to the original size_needed */
    if (unlikely(!next_size)) { next_size = size_needed; }

  }

  /* alloc */
  struct afl_alloc_buf *newer_buf =
      (struct afl_alloc_buf *)realloc(new_buf, next_size);
  if (unlikely(!newer_buf)) {

    free(new_buf);  // avoid a leak
    *buf = NULL;
    return NULL;

  } else {

    new_buf = newer_buf;

  }

  /* Initialize the new content */
  memset( ((u8*)new_buf) + current_size, 0, next_size - current_size ); // <------ this line

  new_buf->complete_size = next_size;
  *buf = (void *)(new_buf->buf);
  return *buf;

}
```

vanhauser-thc:
Uh you are right, as I read in the man pages: `if the new size is larger than the old size, the added memory will not be initialized.`
[...]

PR #1722, opened 2023-05-01
Should memset EFF_ALEN(len) of eff_map

wtdcode:
9065d4ba86ecdafeade50e5235ee1e99f4179692 seems not incorrect on the lenght of the `eff_map`, I fix it here.

Regarding #1721, I carefully checked all usages of `afl_realloc` and the `eff_map` seems the only place that forgets to do the initialization so the original fix is no longer needed once this PR is accepted.


vanhauser-thc:
yes, right.
and please send the PR for memset in afl_realloc

wtdcode:
> yes, right. and please send the PR for memset in afl_realloc

I think it's not needed anymore because `eff_map` already memset correctly and other cases using `afl_realloc` also does initialization correctly too. Apply `memset` in `afl_realloc` will introduce additional performance penalty.

vanhauser-thc:
ok thanks!

PR #1725, opened 2023-05-04
afl-fuzz.c: Document -i - in --help

moshekaplan:
afl-fuzz.c: Document `-i -` in `--help`, to write that `-i` can be passed '-' to resume the prior fuzzing job. Also reference AFL_AUTORESUME so users know they can set that parameter to sidestep the issue entirely.


vanhauser-thc:
thank you!

PR #1728, opened 2023-05-05
GNUmakefile: Update LLVM instructions

moshekaplan:
Update LLVM instructions, because versions higher than 14 are supported and to be explicit that LLD is also required


vanhauser-thc:
thank you

PR #1738, opened 2023-05-22
Fix wrong option name of AFL_LLVM_INSTRUMENT

wtdcode:
According to:

https://github.com/AFLplusplus/AFLplusplus/blob/c4b1566ba35c697cda7822bd0cf30e2e3eeee0c7/src/afl-cc.c#L1840

the docs here seem outdated.


vanhauser-thc:
good point! I fixed it however to also allow `native` as a keyword, I think that is the better solution.
thanks for reporting!

PR #1750, opened 2023-05-31
Support for instrumentation more than GB away from data structures

WorksButNotTested:
Fix https://github.com/AFLplusplus/AFLplusplus/issues/1743


vanhauser-thc:
Thanks!

PR #1757, opened 2023-06-05
Fix llvm 17 pcguard compile error

cocochpie:
For building with llvm 17,
f567a89dae29afb2e421d649f0e750e77913f08c was good, and then suddenly broke in 9324f3f6289c62451e2add1f7553a7eda0d7d642.


vanhauser-thc:
thanks for this.
It worked for my llvm 17 install but that is not a current version.

can you please change the ‘#if’ to >= 17 instead of < 17, so switching things around? Thanks!

vanhauser-thc:
thank you!

PR #1758, opened 2023-06-06
build: adjust LLVM development version check

fanquake:
Adjust version check to only warn for LLVM 17.x and newer, which are the development versions. Otherwise we'll get:
```bash
make LLVM_CONFIG=llvm-config-15 CC=clang-15 CXX=clang++-15
<snip>
GNUmakefile.llvm:69: you are using an in-development llvm version - this might break llvm_mode!
```

for versions that are supported, and not in development.


vanhauser-thc:
yeah true, thanks!

PR #1771, opened 2023-06-14
Fixes #1770: afl-cmin in -T mode doesn't correctly divide inputs amon…

### LINKED ISSUE #1770: afl-cmin in -T mode doesn't correctly divide inputs among threads
**Describe the bug**
afl-cmin (the awk script) when run in -T mode, does not correctly divide inputs among threads due to a float divide result.

**To Reproduce**
For example, say there are 188 inputs and run with -T 16:
```
inputsperfile = in_count / threads # 11.75
```
If there is a remainder (there is) this is increased by 1:
```
if (in_count % threads) {
  inputsperfile++; # 12.75
}
```
Therefore in the loop creating inputs for each thread, there will be 13 inputs for threads 1-15, and the remainder goes into thread 16:
```
    for (instance = 1; instance < threads; instance++) { # For threads 1-15
      for (i = 0; i < inputsperfile; i++) { # Insert 13 inputs, not the expected 12.
        print in_dir"/"infilesSmallToBigFull[cnt] >> tmpfile"."instance
        cnt++
      }
    }
    for (; cnt < in_count; cnt++) { # This loop will never enter as cnt is now 195 and in_count is 188.
      print in_dir"/"infilesSmallToBigFull[cnt] >> tmpfile"."threads
    }
```

The result is that there are 7 spurious lines with no input case for thread 15, and no input file at all for thread 16. (Produces error "could not open .filelist.16".)

**Expected behavior**
Each thread in 1..n-1 should have int(in_count/threads), and remainder in thread n.

forky2:
…g threads

### DIFF
--- a/afl-cmin
+++ b/afl-cmin
@@ -488,7 +488,7 @@ BEGIN {
 
   if (threads) {
 
-    inputsperfile = in_count / threads
+    inputsperfile = int(in_count / threads)
     if (in_count % threads) {
       inputsperfile++;
     }


vanhauser-thc:
Thanks! Is the same issue present in afl-cmin.bash?

forky2:
> Thanks! Is the same issue present in afl-cmin.bash?

The bash version is unafflicted and has truncation down to an integer.

PR #1772, opened 2023-06-14
Fixing rtn_fuzz bug?

tokatoka:


tokatoka:
from my last discord message @vanhauser-thc 

this code is like

```
for (int idx = 0, idx < its_len; idx++) {
 if (buf[0..idx] matches with pattern[0..idx]) {
  its_fuzz();
 }
}
```

so basically it is running its_fuzz for its_len times. 
we are substituting 

```
buf[0] to pattern[0]
buf[0..1] to pattern[0..1]
buf[0..2] to pattern[0..2]
...
...
buf[0..(its_fuzz - 1)] to pattern[0..(its_fuzz-1)]
```

and these are its_len trials.
I think we only need substituting buf[0..(its_fuzz - 1)] to pattern[0..(its_fuzz-1)] and not the others

then we can run its_fuzz() for the largest match

PR #1773, opened 2023-06-16
fix bug

cuanduo:
when using network proxy fuzzing，out_file will close after one fuzz. Howerver fsrv->out_file is zero, out_file wiil not be reopen, write testcase will fail.


vanhauser-thc:
Thanks!

PR #1779, opened 2023-06-21
updated llvm requirements

mmisc:
Updated the LLVM requirements printed in case the build of SanitizerCoveragePCGUARD.so fails.  They are now consistent with 9324f3f6289c62451e2add1f7553a7eda0d7d642.


vanhauser-thc:
thanks!

PR #1782, opened 2023-06-22
Support for long form instrumentation on x64

WorksButNotTested:
Fix https://github.com/AFLplusplus/AFLplusplus/issues/1743#issuecomment-1601176434 on x64


WorksButNotTested:
Thanks for the merge!

PR #1784, opened 2023-06-24
delete duplicate branches

tokatoka:
`# #if LLVM_VERSION_MAJOR >= 16` and `else` have the same branch

### DIFF
--- a/instrumentation/SanitizerCoverageLTO.so.cc
+++ b/instrumentation/SanitizerCoverageLTO.so.cc
@@ -331,7 +331,7 @@ llvmGetPassPluginInfo() {
 #if LLVM_VERSION_MAJOR <= 13
             using OptimizationLevel = typename PassBuilder::OptimizationLevel;
 #endif
-#if LLVM_VERSION_MAJOR >= 16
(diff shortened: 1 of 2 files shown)


tokatoka:
btw does this mean we already dropped the support for the old pass manager & llvm non-compatible with new manager

instrumentation/SanitizerCoverageLTO.so.cc
```diff
@@ -331,7 +331,7 @@ llvmGetPassPluginInfo() {
 #if LLVM_VERSION_MAJOR <= 13
             using OptimizationLevel = typename PassBuilder::OptimizationLevel;
 #endif
-#if LLVM_VERSION_MAJOR >= 16
+#if LLVM_VERSION_MAJOR >= 15
```
tokatoka, line 334: also put my clang-15/lto fix here.
i didn't check if it works on 14 or earlier

vanhauser-thc:
yes. AFL++ PCGUARD now requires llvm13 and that means we do not need old pass manager support.
thanks!

PR #1786, opened 2023-06-26
Increase the number of afl-cc supported params

### LINKED ISSUE #1785: afl-cc.c: error ‘Too many command line parameters’
## Description
When I compiled qemu-system-aarch64 using afl, I got an error in AFL-CC.
```
......
[2882/2882] Linking target qemu-system-aarch64
FAILED: qemu-system-aarch64 
afl-clang-fast++ -m64 -mcx16 @qemu-system-aarch64.rsp
)B
[-] PROGRAM ABORT : Too many command line parameters
         Location : process_params(), src/afl-cc.c:388
```

## Reproducer
```
#git clone qemu & cd
CC=afl-clang-fast CXX=afl-clang-fast++ ./configure --target-list=aarch64-softmmu --enable-sanitizers
make -j 8
```

## Analysis
This [commit](https://github.com/AFLplusplus/AFLplusplus/commit/7f636dbfc247fbe75910fa8fb681ea55d230ba79) introduced the problem. The maximum number of params is 1024. However, it seems easy to exceed this number when programs are linked.  I changed 1024 to 2048 and compiled successfully. I wonder if AFL++ can be more flexible to solve this problem. And the previous version used argc + 128 to determine the number of params, which does not seem to cause this problem.

coc-cyqh:
Fix #1785

### DIFF
--- a/src/afl-cc.c
+++ b/src/afl-cc.c
@@ -383,9 +383,11 @@ static u8 fortify_set = 0, asan_set = 0, x_set = 0, bit_mode = 0,
           have_o = 0, have_pic = 0, have_c = 0, partial_linking = 0,
           non_dash = 0;
 
+#define MAX_PARAMS_NUM 2048
+
 static void process_params(u32 argc, char **argv) {
 
-  if (cc_par_cnt + argc >= 1024) { FATAL("Too many command line parameters"); }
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
Thanks!

PR #1792, opened 2023-06-30
*** WIP *** DO NOT MERGE - Deferred map update

WorksButNotTested:


WorksButNotTested:
I'd welcome any thoughts on this potential performance improvement...

Instead of updating the map at the start of each block, we instead log the block ID to a global array.
```c
typedef struct {

  void           *bitmap;
  fnPurgeBlockLog fn_purge;
  uint32_t        remain;
  uint32_t        log[LOG_SIZE];

} block_log_t;
```

Our inline assembly code at the start of each block is therefore shorter. Since we use `loop`, we don't clobber flags and we still only need spill 2 registers.
```asm
mov qword ptr [rsp-0x88], rax
mov qword ptr [rsp-0x90], rcx
mov rax, <structure base>
mov ecx, dword ptr [rax]
loop 1f
lea rsp, qword ptr [rsp - 0x98]
call qword ptr [rax - <purge>]
lea rsp, qword ptr [rsp + 0x98]
1:
mov dword ptr [rax], ecx
mov dword ptr [rax + rcx * 4], 0x<block_id>
mov rcx, qword ptr [rsp-0x90]
mov rax, qword ptr [rsp-0x88]
```

The intention then is that the purge function can process data in bulk using AVX (GPU offload isn't worthwhile due to the latency of sending the data down the memory bus).

sajjadali8779:
WorksButNotTested:frida-perf

PR #1796, opened 2023-07-03
Fix MAX_PARAMS_NUM define.

kobrineli:
Hi! I've fixed `MAX_PARAMS_NUM` definition and added `#ifndef`. Without it custom value of the variable is shadowed by the `define`.


vanhauser-thc:
Can’t hurt :)

PR #1798, opened 2023-07-05
push to stable

vanhauser-thc:


frida_mode/src/instrument/instrument_arm64.c
```diff
@@ -22,7 +22,7 @@ gboolean           instrument_cache_enabled = FALSE;
 gsize              instrument_cache_size = 0;
 static GHashTable *coverage_blocks = NULL;
 
-__attribute__((aligned(0x1000))) static guint8 area_ptr_dummy[MAP_SIZE];
+__attribute__((aligned(0x1000))) static guint8 area_ptr_dummy[2UL << 20];
```
hexcoder-: Now we have two size definitions '2UL << 20' in instrument_arm32.cpp and ...64.cpp (before we had one definition MAP_SIZE). I prefer the one definition rule.

hexcoder- (commented):
Apart from two small nits, I approve it.

- The typo was fixed by myself.
- The MAP_SIZE copy-and-paste change I leave to you.

hexcoder- (approved):

PR #1802, opened 2023-07-07
*** WIP *** DO NOT MERGE - Changes to select only 1/n blocks to add coverage

WorksButNotTested:


WorksButNotTested:
Discontinued. Although this can lead to early gains in comparison to normal coverage, after a couple of hours, it quickly falls away.

PR #1815, opened 2023-07-28
fix the file descriptor without determining null

mark0-cn:


mark0-cn:
fix the file descriptor without determining null

vanhauser-thc:
Can’t hurt :)

PR #1818, opened 2023-07-29
Fix bug: compilation cannot pass when DEBUG macro is enabled

mark0-cn:
image

### DIFF
--- a/include/afl-mutations.h
+++ b/include/afl-mutations.h
@@ -2456,7 +2456,7 @@ inline u32 afl_mutate(afl_state_t *afl, u8 *buf, u32 len, u32 steps,
         }
 
         char buf[20];
-        snprintf(buf, sizeof(buf), "%ld", val);
+        snprintf(buf, sizeof(buf), "%lld", val);
         u32 old_len = off2 - off;
         u32 new_len = strlen(buf);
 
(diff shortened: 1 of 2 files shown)


domenukk:
Is this fine on Linux, MacOS and gcc and clang? IIRC the platform independent / correct way would be to use `, "%" PRId64,` here (but it's super ugly so I'd prefer `%lld` if it works fine..)

mark0-cn:
> Is this fine on Linux, MacOS and gcc and clang? IIRC the platform independent / correct way would be to use `, "%" PRId64,` here (but it's super ugly so I'd prefer `%lld` if it works fine..)

I find this issue on MacOS and have not tried it on other platforms. But I feel like there's no problem on other platforms either. Perhaps it would be better to change to '%' PRId64?

domenukk:
Probably safer, yes, IIRC on linux sometimes I needed `%ld` instead of `%lld` specifically, but I don't have a machine to test right now

domenukk:
Tried it, seems fine, thanks! :)

domenukk:
Oops actually now it warns on linux, will fix it manually

domenukk:
Fixed in d0782a7

PR #1819, opened 2023-07-31
C++ compilation should use CPPFLAGS

jschwartzentruber:
`CPPFLAGS` should be used for C++ compilation too (ref: [implicit GNU Make rule](https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html#index-C_002b_002b_002c-rule-to-compile)).

This makes it easier to use `--sysroot` when required, which needs to be used with all flag variations: `CFLAGS`, `CFLAGS_SAFE`, and `CXXFLAGS`.


vanhauser-thc:
I am afk - is the gcc plugin makefile affected too?

jschwartzentruber:
> is the gcc plugin makefile affected too?

Yes, you're right. Updated.

vanhauser-thc:
Thank you!

PR #1821, opened 2023-08-02
Implement an option for treating crashing seeds as new crash

junwha:
- the feature discussed in #1810 
- added new environment variable `AFL_CRASHING_SEEDS_AS_NEW_CRASH`
- if the env variable set, classify and save the corpus to crash folder with specifying the origin seed and "dry_run" stage


include/afl-fuzz.h
```diff
@@ -539,7 +539,8 @@ typedef struct afl_state {
       expand_havoc,                /* perform expensive havoc after no find */
       cycle_schedules,                  /* cycle power schedules?           */
       old_seed_selection,               /* use vanilla afl seed selection   */
-      reinit_table;                     /* reinit the queue weight table    */
+      reinit_table,                     /* reinit the queue weight table    */
+      crashing_seeds_as_new_crash; /* treat crashing seeds as normal corpus */
```
vanhauser-thc: this should be in afl_env_vars

src/afl-fuzz-init.c
```diff
@@ -1056,6 +1056,13 @@ void perform_dry_run(afl_state_t *afl) {
               "skipping",
               fn, (int)(s8)afl->fsrv.crash_exitcode);
 
+        } else if (afl->crashing_seeds_as_new_crash) {
+          
+          WARNF(
+              "Test case '%s' results in a crash,"
+              "as AFL_CRASHING_SEEDS_AS_NEW_CRASH is set, "
+              "saving as a crash", fn);
```
vanhauser-thc: maybe `saving as a new crash`?

src/afl-fuzz-init.c
```diff
@@ -1078,38 +1085,94 @@ void perform_dry_run(afl_state_t *afl) {
 
         }
 
-        q->disabled = 1;
-        q->perf_score = 0;
+        /* Crashing corpus will regrad as normal, and categorized as new crash at fuzzing  */
```
vanhauser-thc: typos: `Crashing seeds will be regarded as new crashes on startup` or something like that

src/afl-fuzz-init.c
```diff
@@ -1056,6 +1056,13 @@ void perform_dry_run(afl_state_t *afl) {
               "skipping",
               fn, (int)(s8)afl->fsrv.crash_exitcode);
 
+        } else if (afl->crashing_seeds_as_new_crash) {
```
vanhauser-thc: you need to take care of a crash with uses_crash_exitcode as well (above)

vanhauser-thc: basically, just reverse the if and else parts

src/afl-fuzz-init.c
```diff
@@ -1078,38 +1085,94 @@ void perform_dry_run(afl_state_t *afl) {
...
+          ++afl->saved_crashes;
 
-          afl->queue = afl->queue_buf[i];
+          fd = open(crash_fn, O_WRONLY | O_CREAT | O_EXCL, DEFAULT_PERMISSION);
+          if (unlikely(fd < 0)) { PFATAL("Unable to create '%s'", crash_fn); }
+          ck_write(fd, use_mem, read_len, crash_fn);
+          close(fd);
+
+          afl->last_crash_time = get_cur_time();
+          afl->last_crash_execs = afl->fsrv.total_execs;
```
vanhauser-thc, line 1140: you still need to set
```
q->disabled = 1;
q->perf_score = 0;
```
so the else should not be an else but mandatory code.

vanhauser-thc:
fyi - in your code change you saved the crash away but do not disable the crashing seed for fuzzing

include/afl-fuzz.h
junwha: moved it as env variable!

src/afl-fuzz-init.c
junwha0511: fixed

src/afl-fuzz-init.c
junwha: moved it to the inside of else statement

src/afl-fuzz-init.c
junwha: moved it to the ouside of else

junwha:
Thank you for your kindness review!:) is there anything else to fix?

vanhauser-thc:
finger crossed AFL_CRASHING_SEEDS_AS_NEW_CRASH will behave as it should, that startup code is quite complex with the queue setup :)

PR #1823, opened 2023-08-03
Nyx mode updates

schumilo:
This PR contains the following changes:

- update all Nyx submodules
- fix for https://github.com/nyx-fuzz/packer/issues/25
- configurable Nyx aux buffers size (useful if crash reports are truncated due to very long ASAN reports; see: https://github.com/nyx-fuzz/QEMU-Nyx/pull/52)
- `NYX_VM=1` is set by default as env var
- certain information on the snapshot and the virtual machine state are now also written to a yaml file, which can be later processed by the fuzzing frontend (see: https://github.com/nyx-fuzz/QEMU-Nyx/pull/53) 
- other minor changes and fixes


vanhauser-thc:
@schumilo great!!

@choller

PR #1825, opened 2023-08-05
disable exceptions on LLVM/GCC plugins, decreasing further the librar…

devnexen:
…ies's binaries size in the process.


vanhauser-thc:
Thank you!

PR #1826, opened 2023-08-05
Don't corrupt instruction if map offset is too large

WorksButNotTested:


vanhauser-thc:
why is that only for arm64? and what if a user wants to use a map size of 128kb - does this still work then?

WorksButNotTested:
I’m not certain if the other architectures are affected, but leave the issue open for me and I’ll check.

arm64 has two different ways of encoding the instrumentation at the top of each block. One uses relative offsets (which can only cope with the map offset being 64k) and the other uses absolute offsets stored alongside the code which can cope with any size. It selects which encoding to use on a block by block basis.

vanhauser-thc:
I will leave this open for now until someone can do proper testing

WorksButNotTested:
I've confirmed that things work ok with large map sizes on x86, x64, arm and aarch64 now. I have also added a fix to `util/frida_get_symbol_addr.sh` as PIEs on aarch64 are identified as `shared object` rather than `pie` by the file utility. So this should be ready to merge now.

vanhauser-thc:
Great, thanks!

PR #1830, opened 2023-08-09
Adding Coverage Estimation using Species Richness Estimators

Jonas-Eppard:
In a project together with [@MartinRingCR](https://github.com/MartinRingCR) we included species Richness Estimators to AFL++. This tracks the path coverage and estimates the amount of unseen code paths.
This is an extension of the approach in [this](https://mboehme.github.io/paper/ICSE23.Effectiveness.pdf) paper, where Basic Blocks were counted.

In [docs/coverage_estimation.md](https://github.com/Jonas-Eppard/AFLplusplus-coverage-estimation/blob/coverage_estimation/docs/coverage_estimation.md) the usage is described. Merging this would enable Users/Researchers to gather more inside into the fuzzing campaign. Furthermore, the data gathered can be exposed in files, to enable further development of estimators (maybe even optimized for fuzzing).

### DIFF
--- a/docs/coverage_estimation.md
+++ b/docs/coverage_estimation.md
@@ -1,0 +1,39 @@
+# Coverage Estimation in AFL++
+
+This file describes the Coverage Estimation of AFL++. or general information about AFL++, see
+[README.md](../README.md).
+
+## Table of Content
+* [Introduction](#1-introduction)
+* [Setup](#2-setup)
+* [Status Screen extension](#3-status-screen-extension)
+
+## 1 Introduction
+The Coverage Estimation inside AFL++ is based on Path Coverage. It used STADS (Security Testing As Discovery of Species) to use Species Richness estimators for Coverage estimation.
+The estimated coverage should help developers when to stop a Fuzzing campaign.
+The coverage estimation can only be estimated over fuzzable/reachable paths.
+
+Coverage estimation is not tested on multiple Fuzzing instances (-M/-S Options). It's also not tested on resuming a fuzz run (AFL_AUTORESUME, -i -).
+
+## 2 Setup
+To use coverage estimation you don't have to change your workflow, just add following environment variables:
+ * Set `AFL_CODE_COVERAGE` to enable Coverage Estimation.
+ * Consider Setting `AFL_N_FUZZ_SIZE` to something bigger then (1 << 21)(default) to mitigate (Re-)Hash collisions
+ * Consider the use of `AFL_CRASH_ON_HASH_COLLISION` if (slightly) incorrect coverage estimation is worse then a abort
+ * If the Coverage estimation should update more often change `COVERAGE_INTERVAL` in [config.h](../config.h) (This requires rebuilding of AFL++)
+
+More information's about these environment variables in [env_variables.md](./env_variables.md).
+
+## 3 Status Screen extension
+The status screen will be extended with following box:
+```
+ +- code coverage information ------------------------+
+ |              coverage : 57.12% - 63.21%            |
+ | collision probability : 1.02%                      |
+ +----------------------------------------------------+
+```
+ * coverage - This is the estimated path coverage. The first number is a lower bound estimate.
+ The second number is a upper bound estimate. It's only possible to estimate the fuzzable/reachable paths.
+ If the coverage is very fast very high you either fuzzing a simple target or don't have a good corpus.
+ * collision propability - This is a estimate for the probability of Hash Collisions. If this number gets high you should increase `AFL_N_FUZZ_SIZE`. Hash collisions will cause errors in coverage estimation.
+ If `AFL_CRASH_ON_HASH_COLLISION` is set afl-fuzz will abort on a detected Hash collision.
--- a/docs/coverage_estimation.md
+++ b/docs/coverage_estimation.md
@@ -38,3 +38,5 @@ The status screen will be extended with following box:
  Unfortunately there is not enough research done to prove accuracy and precision of these estimators, but they hold in asymptotic behavior and the accuracy and precision of the estimators, go down with more hash collisions (see collision probability).
  * collision propability - This is a estimate for the probability of hash collisions. If this number gets higher than 25% you should consider increasing `AFL_N_FUZZ_SIZE`. The coverage estimation will lose accuracy (overestimates) when hash collisions occur. You can also chose to ignore this, but then the coverage estimation will suffer.
  If `AFL_CRASH_ON_HASH_COLLISION` is set afl-fuzz will abort on a detected Hash collision.
+
+This box takes the place of the fuzzing strategy yields, since there are not used when non-deterministic mode is used. Therefore the coverage estimation is not compatible with the deterministic mode (-D).
--- a/docs/coverage_estimation.md
+++ b/docs/coverage_estimation.md
@@ -1,6 +1,6 @@
-# Coverage Estimation in AFL++
+# Coverage estimation in AFL++
 
-This file describes the Coverage Estimation of AFL++. or general information about AFL++, see
+This file describes the coverage estimation of AFL++. For general information about AFL++, see
 [README.md](../README.md).
 
 ## Table of Content
@@ -9,18 +9,18 @@ This file describes the Coverage Estimation of AFL++. or general information abo
 * [Status Screen extension](#3-status-screen-extension)
 
 ## 1 Introduction
-The Coverage Estimation inside AFL++ is based on Path Coverage. It used STADS (Security Testing As Discovery of Species) to use Species Richness estimators for Coverage estimation.
-The estimated coverage should help developers when to stop a Fuzzing campaign.
+The coverage estimation inside AFL++ is based on path coverage. It used STADS (Security Testing As Discovery of Species) to use species richness estimators for coverage estimation.
+The estimated coverage should help developers when to stop a fuzzing campaign.
 The coverage estimation can only be estimated over fuzzable/reachable paths.
 
-Coverage estimation is not tested on multiple Fuzzing instances (-M/-S Options). It's also not tested on resuming a fuzz run (AFL_AUTORESUME, -i -).
+Coverage estimation is not tested on multiple fuzzing instances (-M/-S Options). It's also not tested on resuming a fuzz run (AFL_AUTORESUME, -i -).
 
 ## 2 Setup
 To use coverage estimation you don't have to change your workflow, just add following environment variables:
- * Set `AFL_CODE_COVERAGE` to enable Coverage Estimation.
+ * Set `AFL_CODE_COVERAGE` during fuzzing (afl-fuzz) to enable coverage estimation.
  * Consider Setting `AFL_N_FUZZ_SIZE` to something bigger then (1 << 21)(default) to mitigate (Re-)Hash collisions
  * Consider the use of `AFL_CRASH_ON_HASH_COLLISION` if (slightly) incorrect coverage estimation is worse then a abort
- * If the Coverage estimation should update more often change `COVERAGE_INTERVAL` in [config.h](../config.h) (This requires rebuilding of AFL++)
+ * If the coverage estimation should update more often change `COVERAGE_INTERVAL` in [config.h](../config.h) (This requires rebuilding of AFL++)
 
 More information's about these environment variables in [env_variables.md](./env_variables.md).
 
@@ -35,5 +35,6 @@ The status screen will be extended with following box:
  * coverage - This is the estimated path coverage. The first number is a lower bound estimate.
  The second number is a upper bound estimate. It's only possible to estimate the fuzzable/reachable paths.
  If the coverage is very fast very high you either fuzzing a simple target or don't have a good corpus.
- * collision propability - This is a estimate for the probability of Hash Collisions. If this number gets high you should increase `AFL_N_FUZZ_SIZE`. Hash collisions will cause errors in coverage estimation.
+ Unfortunately there is not enough research done to prove accuracy and precision of these estimators, but they hold in asymptotic behavior and the accuracy and precision of the estimators, go down with more hash collisions (see collision probability).
+ * collision propability - This is a estimate for the probability of hash collisions. If this number gets higher than 25% you should consider increasing `AFL_N_FUZZ_SIZE`. The coverage estimation will lose accuracy (overestimates) when hash collisions occur. You can also chose to ignore this, but then the coverage estimation will suffer.
  If `AFL_CRASH_ON_HASH_COLLISION` is set afl-fuzz will abort on a detected Hash collision.
--- a/docs/env_variables.md
+++ b/docs/env_variables.md
@@ -314,6 +314,10 @@ mode.
 The main fuzzer binary accepts several options that disable a couple of sanity
 checks or alter some of the more exotic semantics of the tool:
 
+  - Coverage estimation only: `AFL_ABUNDANT_CUT_OFF` describes the cut-off for
+    species richness estimators. Default is 10. The Value should only be changed
+    for research.
+
   - Setting `AFL_AUTORESUME` will resume a fuzz run (same as providing `-i -`)
     for an existing out folder, even if a different `-i` was provided. Without
     this setting, afl-fuzz will refuse execution for a long-fuzzed out dir.
@@ -326,6 +330,11 @@ checks or alter some of the more exotic semantics of the tool:
     newly found test cases and not for test cases that are loaded on startup
     (`-i in`). This is an important feature to set when resuming a fuzzing
     session.
+
+  - `AFL_CODE_COVERAGE` will enable code coverage estimation. See also 
+    [coverage_estimation.md](./coverage_estimation.md). Related enviroment
+    variables: `AFL_ABUNDANT_CUT_OFF`, `AFL_CRASH_ON_HASH_COLLISION` and
+    `AFL_N_FUZZ_SIZE`
 
   - Setting `AFL_CRASH_EXITCODE` sets the exit code AFL++ treats as crash. For
     example, if `AFL_CRASH_EXITCODE='-1'` is set, each input resulting in a `-1`
@@ -407,6 +416,11 @@ checks or alter some of the more exotic semantics of the tool:
     Others need not apply, unless they also want to disable the
     `/proc/sys/kernel/core_pattern` check.
 
+  - Coverage estimation only: `AFL_CRASH_ON_HASH_COLLISION` will crash on detected
+    (Re)Hash collisions. If this is not set only a warning is displayed instead of
+    Aborting the Fuzzing campaign. You could also increase `AFL_N_FUZZ_SIZE` to
+    mitigate the chance of a (Re)Hash collision.
+
   - If afl-fuzz encounters an incorrect fuzzing setup during a fuzzing session
     (not at startup), it will terminate. If you do not want this, then you can
     set `AFL_IGNORE_PROBLEMS`. If you additionally want to also ignore coverage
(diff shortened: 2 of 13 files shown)


vanhauser-thc:
can you please resolve the conflict (sorry for that, other PR that was earlier than yours) plus the two codeql items?

docs/coverage_estimation.md
```diff
@@ -0,0 +1,39 @@
+# Coverage Estimation in AFL++
+
+This file describes the Coverage Estimation of AFL++. or general information about AFL++, see
```
vanhauser-thc: typo: `. For general`

docs/coverage_estimation.md
```diff
@@ -0,0 +1,39 @@
...
+[README.md](../README.md).
+
+## Table of Content
+* [Introduction](#1-introduction)
+* [Setup](#2-setup)
+* [Status Screen extension](#3-status-screen-extension)
+
+## 1 Introduction
+The Coverage Estimation inside AFL++ is based on Path Coverage. It used STADS (Security Testing As Discovery of Species) to use Species Richness estimators for Coverage estimation.
+The estimated coverage should help developers when to stop a Fuzzing campaign.
```
vanhauser-thc: typos: please lower case for "path coverage" and "fuzzing". also in the rest of the readme :)

docs/coverage_estimation.md
```diff
@@ -0,0 +1,39 @@
...
+The status screen will be extended with following box:
+```
+ +- code coverage information ------------------------+
+ |              coverage : 57.12% - 63.21%            |
+ | collision probability : 1.02%                      |
+ +----------------------------------------------------+
+```
+ * coverage - This is the estimated path coverage. The first number is a lower bound estimate.
+ The second number is a upper bound estimate. It's only possible to estimate the fuzzable/reachable paths.
+ If the coverage is very fast very high you either fuzzing a simple target or don't have a good corpus.
```
vanhauser-thc, line 37: please add a line describing how likely that estimate is.

docs/coverage_estimation.md
```diff
@@ -0,0 +1,39 @@
...
+To use coverage estimation you don't have to change your workflow, just add following environment variables:
+ * Set `AFL_CODE_COVERAGE` to enable Coverage Estimation.
+ * Consider Setting `AFL_N_FUZZ_SIZE` to something bigger then (1 << 21)(default) to mitigate (Re-)Hash collisions
+ * Consider the use of `AFL_CRASH_ON_HASH_COLLISION` if (slightly) incorrect coverage estimation is worse then a abort
+ * If the Coverage estimation should update more often change `COVERAGE_INTERVAL` in [config.h](../config.h) (This requires rebuilding of AFL++)
+
+More information's about these environment variables in [env_variables.md](./env_variables.md).
+
+## 3 Status Screen extension
+The status screen will be extended with following box:
```
vanhauser-thc, line 28: I have not run this yet. where is that added to the status screen? on the bottom so the window would need more lines?
If so I would propose to rather reuse the lines for the yields for determinstic fuzzing (which is disabled by default. you can enforce that -D does not work with AFL_CODE_COVERAGE if you want)

docs/coverage_estimation.md
```diff
@@ -0,0 +1,39 @@
...
+## 1 Introduction
+The Coverage Estimation inside AFL++ is based on Path Coverage. It used STADS (Security Testing As Discovery of Species) to use Species Richness estimators for Coverage estimation.
+The estimated coverage should help developers when to stop a Fuzzing campaign.
+The coverage estimation can only be estimated over fuzzable/reachable paths.
+
+Coverage estimation is not tested on multiple Fuzzing instances (-M/-S Options). It's also not tested on resuming a fuzz run (AFL_AUTORESUME, -i -).
+
+## 2 Setup
+To use coverage estimation you don't have to change your workflow, just add following environment variables:
+ * Set `AFL_CODE_COVERAGE` to enable Coverage Estimation.
```
vanhauser-thc: when running afl-fuzz or when instrumenting the target? I think you mean for afl-fuzz, however for a user this is not clear

docs/coverage_estimation.md
```diff
@@ -0,0 +1,39 @@
...
+```
+ +- code coverage information ------------------------+
+ |              coverage : 57.12% - 63.21%            |
+ | collision probability : 1.02%                      |
+ +----------------------------------------------------+
+```
+ * coverage - This is the estimated path coverage. The first number is a lower bound estimate.
+ The second number is a upper bound estimate. It's only possible to estimate the fuzzable/reachable paths.
+ If the coverage is very fast very high you either fuzzing a simple target or don't have a good corpus.
+ * collision propability - This is a estimate for the probability of Hash Collisions. If this number gets high you should increase `AFL_N_FUZZ_SIZE`. Hash collisions will cause errors in coverage estimation.
```
vanhauser-thc: "if this number gets too high" ... maybe reword "if this number gets larger than 15% ..." or what a good value would be.

docs/env_variables.md
```diff
@@ -327,6 +331,11 @@ checks or alter some of the more exotic semantics of the tool:
     (`-i in`). This is an important feature to set when resuming a fuzzing
     session.
 
+  - `AFL_CODE_COVERAGE` will enable code coverage estimation. See also 
+    [coverage_estimation.md](./coverage_estimation.md). Related enviroment
+    variables: `AFL_ABUNDANT_CUT_OFF`, `AFL_CRASH_ON_HASH_COLLISION` and
+    `AFL_N_FUZZ_SIZE`
+
```
vanhauser-thc, line 334: hmm I would propose to move AFL_ABUNDANT_CUT_OFF and AFL_CRASH_ON_HASH_COLLISION just below here.

src/afl-fuzz-bitmap.c
```diff
@@ -474,7 +474,8 @@ save_if_interesting(afl_state_t *afl, void *mem, u32 len, u8 fault) {
 
   /* Generating a hash on every input is super expensive. Bad idea and should
      only be used for special schedules */
-  if (unlikely(afl->schedule >= FAST && afl->schedule <= RARE)) {
+  if (unlikely((afl->schedule >= FAST && afl->schedule <= RARE) ||
+               afl->coverage_estimation)) {
```
vanhauser-thc: can you please change this to:

`if (likely(afl->schedule >= FAST && afl->schedule <= RARE) || unlikely(afl->coverage_estimation))`

src/afl-fuzz-bitmap.c
```diff
@@ -483,8 +484,91 @@ save_if_interesting(afl_state_t *afl, void *mem, u32 len, u8 fault) {
...
 
     /* Saturated increment */
-    if (likely(afl->n_fuzz[cksum % N_FUZZ_SIZE] < 0xFFFFFFFF))
-      afl->n_fuzz[cksum % N_FUZZ_SIZE]++;
+    if (LARGE_INDEX(afl->n_fuzz, cksum % afl->n_fuzz_size, MAX_ALLOC,
+                    sizeof(u32)) < 0xFFFFFFFF)
+      ++LARGE_INDEX(afl->n_fuzz, cksum % afl->n_fuzz_size, MAX_ALLOC,
+                    sizeof(u32));
+
+    if (afl->coverage_estimation) {
```
vanhauser-thc: please insert `unlikely(...)`

src/afl-fuzz-bitmap.c
```diff
@@ -594,11 +678,43 @@ save_if_interesting(afl_state_t *afl, void *mem, u32 len, u8 fault) {
     /* For AFLFast schedules we update the new queue entry */
     if (likely(cksum)) {
 
-      afl->queue_top->n_fuzz_entry = cksum % N_FUZZ_SIZE;
-      afl->n_fuzz[afl->queue_top->n_fuzz_entry] = 1;
+      afl->queue_top->n_fuzz_entry = cksum % afl->n_fuzz_size;
+      if (afl->coverage_estimation) {
```
vanhauser-thc: again, please add `unlikely(...)`

src/afl-fuzz-init.c
```diff
@@ -1770,6 +1770,12 @@ static void handle_existing_out_dir(afl_state_t *afl) {
   if (delete_files(fn, CASE_PREFIX)) { goto dir_cleanup_failed; }
   ck_free(fn);
 
+#if defined COVERAGE_ESTIMATION_LOGGING && COVERAGE_ESTIMATION_LOGGING
+  fn = alloc_printf("%s/path_data", afl->out_dir);
```
vanhauser-thc: doesnt this need `if (unlikely(afl->coverage_estimation)) {' ?

src/afl-fuzz-init.c
```diff
@@ -1903,6 +1909,10 @@ static void handle_existing_out_dir(afl_state_t *afl) {
   if (unlink(fn) && errno != ENOENT) { goto dir_cleanup_failed; }
   ck_free(fn);
 
+  fn = alloc_printf("%s/coverage_estimation", afl->out_dir);
```
vanhauser-thc: same?

src/afl-fuzz-init.c
```diff
@@ -2111,6 +2121,33 @@ void setup_dirs_fds(afl_state_t *afl) {
         "pending_total, pending_favs, map_size, saved_crashes, "
         "saved_hangs, max_depth, execs_per_sec, total_execs, edges_found\n");
 
+#if defined COVERAGE_ESTIMATION_LOGGING && COVERAGE_ESTIMATION_LOGGING
+    tmp = alloc_printf("%s/path_data", afl->out_dir);
```
vanhauser-thc: same?

src/afl-fuzz-stats.c
```diff
@@ -332,7 +331,8 @@ void write_stats_file(afl_state_t *afl, u32 t_bytes, double bitmap_cvg,
       "afl_version       : " VERSION
       "\n"
       "target_mode       : %s%s%s%s%s%s%s%s%s%s\n"
-      "command_line      : %s\n",
+      "command_line      : %s\n"
+      "hash_collisions   : %lu",
```
vanhauser-thc: please move this just after `testcache_evict`

src/afl-fuzz-stats.c
```diff
@@ -511,7 +597,15 @@ static void check_term_size(afl_state_t *afl) {
   if (ioctl(1, TIOCGWINSZ, &ws)) { return; }
 
   if (ws.ws_row == 0 || ws.ws_col == 0) { return; }
-  if (ws.ws_row < 24 || ws.ws_col < 79) { afl->term_too_small = 1; }
+  if (afl->coverage_estimation) {
```
vanhauser-thc: `unlikely(...)` please

src/afl-fuzz-stats.c
```diff
@@ -520,6 +614,123 @@ static void check_term_size(afl_state_t *afl) {
 
 void show_stats(afl_state_t *afl) {
 
+  if (afl->coverage_estimation) {
```
vanhauser-thc: unlikely pls.

vanhauser-thc: in all other locations too please

vanhauser-thc:
and one further question - what is the impact (e.g. 5% less exec/s) when this feature is activated? have you measured it?
maybe it is not a lot and could be on by default if it is helpful.

docs/coverage_estimation.md
Jonas-Eppard: Yes, I currently allocate two extra lines, but only if the code coverage is enabled. Should I still change it? I don't like the approach of overwriting the deterministic fields, to ensure compatibility with -D.

src/afl-fuzz-bitmap.c
Jonas-Eppard: Are you sure this would change the effect? Since the unlikely was already there. I can change it to likely, but then I would change the meaning of the line. I haven't changed the unlikely.
Do you mean:
`if (unlikely(afl->schedule >= FAST && afl->schedule <= RARE) || unlikely(afl->coverage_estimation))`
I could change that, but I don't know if this would differ to the current state.

Jonas-Eppard:
I have measured the impact of the runtime, but the results weren't clear. In my tests, the runtime differences with and without coverage estimator were inside the error bars of the tests. But if I remember correctly I didn't recompile afl-fuzz only disabled the flag for coverage estimation. Therefore, I currently can't say anything about the runtime, also regarding the changes done with the unlikely's...

Jonas-Eppard:
Thanks a lot for the review. I added the requested changes. Some questions arose as seen in the non-resolved conversations.

docs/coverage_estimation.md
vanhauser-thc: Using -D is a bad idea and only something someone should do at the starting of a campaign for a few hours and then terminate it. Hence making code coverage incompatible with -D makes a lot of sense to me :)

src/afl-fuzz-bitmap.c
vanhauser-thc: no I meant what I wrote :)  the fast schedule is the default so it makes more sense it is likely()

vanhauser-thc:
I plan do to a new release today or tomorrow, and I will first merge this into a new branch before merging it to dev so I can properly test this :)

Jonas-Eppard:
> I plan do to a new release today or tomorrow, and I will first merge this into a new branch before merging it to dev so I can properly test this :)

No worries.

I addressed the last changes.
The screen now looks like this, when coverage estimation is enabled:
image
and like this without:
image

vanhauser-thc:
I like it, thanks!

vanhauser-thc:
Screenshot from 2023-08-11 14-04-16

I am trying this and having a question. it say 15-18% coverage, the map density says 18%, so it should be impossible to have below 18% coverage right?

can you give me some insight how you calculate coverage in a few simple sentences without needing to understand your code? :)

vanhauser-thc:
also just encountered during a rerun (`-i -`):

```

[-]  SYSTEM ERROR : Unable to create 'out/default/path_data/time:0'
    Stop location : write_coverage_file(), src/afl-fuzz-stats.c:459
       OS message : No such file or directory
```

Jonas-Eppard:
> Screenshot from 2023-08-11 14-04-16
> 
> I am trying this and having a question. it say 15-18% coverage, the map density says 18%, so it should be impossible to have below 18% coverage right?
> 
> can you give me some insight how you calculate coverage in a few simple sentences without needing to understand your code? :)

Hi,

the coverage estimation works in a way that it keeps track of the amount of times different paths were hit, At the end a list with says how many paths are hit once, twice, three times, ... up to 'AFL_ABUNDANT_CUTOFF' (default 10)

So as a example this would say:

> 132 Paths are hit once
> 52 Paths are hit twice
> ...
> 8 Paths are hit 10 times
> 56 Paths are hit more then 10 times

These numbers will then be used by the estimators (Abundant Coverage Estimator [short ACE]) and this estimator provides an upper bound estimate on the total paths. The total paths describe the amount of paths already hit and an estimate on paths not hit but to be hit.

The Collision probability would cause that two paths have the same counter. So it would say: One path is hit 25 times, but in reality one path is hit 22 times and another 3 times. This causes the estimator to underestimate the amount of total paths, so the percentage would overestimate.... So it would say you found 35%, but you are only at 30%. On only a couple collisions it does not matter, but if the collisions get a lot (>25%) the impact of this error increases.

To the 15-18% coverage this is using two different estimators, one provides an upper bound estimation the other one a lower bound estimation. I have to look it up on Monday which estimators I used exactly...
In your test case, I would be rather confident that the code you are fuzzing currently this estimate for the hit paths hold!


> also just encountered during a rerun (`-i -`):
> 
> ```
> 
> [-]  SYSTEM ERROR : Unable to create 'out/default/path_data/time:0'
>     Stop location : write_coverage_file(), src/afl-fuzz-stats.c:459
>        OS message : No such file or directory
> ```

I haven't used the re-runs since the map with the amount of times different paths are hit is only in memory, so you would have "lost" the progress of the estimators on reruns, The file in question is used to expose the data for the coverage estimator, to enable researches to calculate other estimators. This can be prevented with the define statement of '#define COVERAGE_ESTIMATION_LOGGING 1' in [include/config.h:308](https://github.com/AFLplusplus/AFLplusplus/pull/1830/files#diff-f20ab3eb0041360bbad27bdb1173ada9c2818c430536c89ccd514d822c43a1ec)

Regarding the error on '-i' I delete all the files in the folder 'out/default/path_data' with the method 'delete_files(fn)' from 'src/afl-fuzz-init.c' does this also delete the folder? Or is any other step not done on autoresume?

vanhauser-thc:
> > Screenshot from 2023-08-11 14-04-16
> > I am trying this and having a question. it say 15-18% coverage, the map density says 18%, so it should be impossible to have below 18% coverage right?

from your explanation I get how the number of coverage estimate is calculcated.
but still - it should be impossible that the coverage estimation (the real possible one, not the one you calculcate) could ever be lower than the map density in reality, right?

> > also just encountered during a rerun (`-i -`):
> > ```
> > 
> > [-]  SYSTEM ERROR : Unable to create 'out/default/path_data/time:0'
> >     Stop location : write_coverage_file(), src/afl-fuzz-stats.c:459
> >        OS message : No such file or directory
> > ```
> 
> I haven't used the re-runs since the map with the amount of times different paths are hit is only in memory, so you would have "lost" the progress of the estimators on reruns, The file in question is used to expose the data for the coverage estimator, to enable researches to calculate other estimators. This can be prevented with the define statement of '#define COVERAGE_ESTIMATION_LOGGING 1' in [include/config.h:308](https://github.com/AFLplusplus/AFLplusplus/pull/1830/files#diff-f20ab3eb0041360bbad27bdb1173ada9c2818c430536c89ccd514d822c43a1ec)
> 
> Regarding the error on '-i' I delete all the files in the folder 'out/default/path_data' with the method 'delete_files(fn)' from 'src/afl-fuzz-init.c' does this also delete the folder? Or is any other step not done on autoresume?

no it requires to do a delete_files(afl->out + "path_data") ... I added that.
but this also means that coverage estimation is incompatible with resuming a session. added a check for that too.

Jonas-Eppard:
> from your explanation I get how the number of coverage estimate is calculcated.
but still - it should be impossible that the coverage estimation (the real possible one, not the one you calculcate) could ever be lower than the map density in reality, right?

EDIT: Sorry, I misread. I read that the estimated coverage can't be lower, not the real one...
I'm not quite sure what the map defines exactly. If I remember correctly, it had flags about the instrumentation points and the previous point using XOR.But the real coverage can be higher than the map coverage! So the estimators shine when they describe the end of a fuzzing campaign. The map has more inside into the fuzzing target than the estimators, so I think the map should be more correct.

~~I'm currently not quite sure what the map density says exactly. I thought that the map density is how many bits of the coverage map (from the instrumentation points) were hit. The map itself doesn't say how often the paths are hit. 
The coverage estimator on the other hand uses the amount of times the paths are hit.~~

~~For example, let us assume the map has 10 points and 5 of them are hit. In these 5 hits, we have hit 4 of them once and one of them twice.
Therefore, we have 50% map density, but the coverage estimator would calculate the following;
`Paths hit more then ten times + Paths hit les then ten simes / (1-(Paths hit once/amount of path hits on paths with less then 10 hits)) + normalisation (always >=0)`
This would estimate the amount of total paths. In our example:
`0+5/(1-(4/6)) + ... = 15`~~

~~This would estimate that there is a total of 15 Paths. So the current coverage is 5/15 so 33% which is less than the 50% map density.~~

vanhauser-thc:
The map density shows how many edges (so bytes found in the map, not individual bits in the bytes) have already been found.
So if this shows 30%, the must be the lower bound of your coverage estimate, and the upper bound must be at least that or higher.

i like the idea what you have implemented however in my tests the estimate must be improved because it currently comes up with too low values.

(I also got confused with the part that you striked out, you talk about paths, but paths are not really coverage so if you would calculate numbers of estimated total paths - also a cool thing to know - the value displayed should reflect that and renamed to estimated total existing paths or something like that)

Jonas-Eppard:
> (I also got confused with the part that you striked out, you talk about paths, but paths are not really coverage so if you would calculate numbers of estimated total paths - also a cool thing to know - the value displayed should reflect that and renamed to estimated total existing paths or something like that)

The estimators estimate path coverage. So it estimates the amount of paths and not basic block coverage. If I understood correctly the map shows basic blocks?

Then maybe the UI should be renamed into `path coverage information` and `path coverage:`?

PR #1851, opened 2023-09-03
Skip calculating top_rated_fuzz_p2 with FAST schedule

chinggg:
After #1836 and https://github.com/AFLplusplus/AFLplusplus/commit/26f29fd485efaa08824c27501f82caeea525b5e3, function `update_bitmap_score` became more clear. But I still found redundant calculation of `top_rated_fuzz_p2` when `afl->schedule >= FAST && afl->schedule < RARE`, though it doesn't affect the final result. 
I remove it so the overhead to fetch memory and calculate useless values may be reduced.


vanhauser-thc:
Thanks

vanhauser-thc:
Weird, GitHub is buggy it is not merging, will try again tomorrow

PR #1852, opened 2023-09-05
frida_mode: fix oob when copying details

jiangenj:
Change-Id: I6a19792ee9e174720242602cec4a79134f6218de

### DIFF
--- a/frida_mode/src/lib/lib.c
+++ b/frida_mode/src/lib/lib.c
@@ -44,8 +44,10 @@ static gboolean lib_find_exe(const GumModuleDetails *details,
 
   lib_details_t *lib_details = (lib_details_t *)user_data;
 
-  memcpy(lib_details->name, details->name, PATH_MAX);
-  memcpy(lib_details->path, details->path, PATH_MAX);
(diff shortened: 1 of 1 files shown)


WorksButNotTested:
lgtm. Thanks

frida_mode/src/lib/lib.c
```diff
@@ -44,8 +44,10 @@ static gboolean lib_find_exe(const GumModuleDetails *details,
 
   lib_details_t *lib_details = (lib_details_t *)user_data;
 
-  memcpy(lib_details->name, details->name, PATH_MAX);
-  memcpy(lib_details->path, details->path, PATH_MAX);
+  strncpy(lib_details->name, details->name, PATH_MAX);
+  strncpy(lib_details->path, details->path, PATH_MAX);
+  lib_details->name[PATH_MAX] = '\0';
+  lib_details->path[PATH_MAX] = '\0';
```
vanhauser-thc, line 50: Is the variable path_max+1? Otherwise this is an off by one

WorksButNotTested: Yes it is. The struct is defined just above on line 30.

PR #1867, opened 2023-09-18
Target oddness parameter writing to 'fuzzer_stats'.

CourteousSleet:
Added a new 'afl_state_t' variable named 'is_target_odd' that has an 'u8' type and takes only two values: '1' if target is odd and '0' if it is not. Then this info is written to 'fuzzer_stats' file. The purpose of my PR is that on my job we use AFL++ in an automated pipeline that has a step on which many (>700) fuzzing targets are generated. And some of them are odd and this requires an automated check. So I considered this PR is more preferable than getting oddness info from an AFL++ UI. Also I would like to mention my coworker in this PR, if it is okay. @nataraj-hates-MS-for-stealing-github


nataraj-hates-MS-for-stealing-github:
I would add my comment here...

General idea: we have some targets automatically generated using [Futag](https://github.com/ispras/Futag). Some of them are trivial (has only one possible path). We need to sort them out, in order not to waste time on them. This patch is an a idea of @CourteousSleet how to do that. Me I do not see any better solutions. May be, @vanhauser-thc, you do.

vanhauser-thc:
I see what you want/need but I do not think you need this PR for that.
You can read that out from fuzzer_stats, the item corpus_found will be 0 for such targets, right?

CourteousSleet:
Yes, you are right. Thank you.

PR #1868, opened 2023-09-19
Show shorter fuzzer name in stat when user's banner is too long

nataraj-hates-MS-for-stealing-github:
I found -T option setting user's defined banner very handy.

I acquired taste, and tried to put all info I need in the banner, but found that it can't be as long as I'd like to.

I came to conclusion that "american fuzzy lop" text is useless waste of precocious space compared to very important info that I want to put in the banner ;-) , "AFL" should do the same job

So this patch put "AFL" instead of "american fuzzy lop" in the stat screen header if user's banner does not fit into the space. So one can use bigger banners. 

Hope I did code styling right (I tried my best), and that this patch will be useful not only to me :-)


nataraj-hates-MS-for-stealing-github:
I do not understand why tests fails... See no error message.

BTW for unpatched version on my local machine they  fails too... May be the problem is not in the patch

src/afl-fuzz-stats.c
```diff
@@ -778,10 +778,29 @@ void show_stats_normal(afl_state_t *afl) {
   if (unlikely(!banner[0])) {
 
     char *si = "";
+    char *fuzzer_name;
```
vanhauser-thc, line 781: pls make it a global in afl-fuzz.c and set it there, use extern here.

vanhauser-thc:
I do not mind if you rename it to AFL, not just when the banner is too long otherwise.

src/afl-fuzz-stats.c
nataraj-hates-MS-for-stealing-github: Sorry... I did not got the idea...

I do not understand why we should do it. The variable is local, and is not needed outside of this scope...

Also I do not understand where is the proper place for this variable in `afl-fuzz.c` I did not found similar variables there...

May be it should be added in `afl_state` structure from  `afl-fuzz.h`? But I still not sure it is good idea...

I am not familiar with AFL++ architecture at all. I guess I need some guidance and explanations here...

nataraj-hates-MS-for-stealing-github:
> I do not mind if you rename it to AFL, not just when the banner is too long otherwise.

If no user banner is set (and it is true in most cases), stat header will look nicer with "american fuzzy lop" I guess. So unless you insist, I would keep both "AFL" and  "american fuzzy lop" the way I did in my first patch version...

vanhauser-thc:
the banner in the UI is rendered about 2 times per second. so the if() about how to construct the banner takes up unnecessary resources as this is something that never changes.
so I want to use your change for a bigger change.

set the banner into a global variable in main() in src/afl-fuzz.c, and import that u8* as extern into the stats source.
saves speed and allows for your feature request.

it is fine to be a global variable because it is not something that would need to be thread safe.

vanhauser-thc:
@nataraj-hates-MS-for-stealing-github  ?

nataraj-hates-MS-for-stealing-github:
a bit occupied right now O_o Sorry.

So will say without code commit.

1. This idea with once-calculated banner seems to be good I can do it.

2. May be we are not ambitious enough here? Sine stat title never changes (I guess it never changes) may be we should snprintf whole title to a static buffer in first iteration, and then just print that buffer twice a second?

3. I still does not understand why we should put  that variable in `afl-fuzz.c` we can make it static right in the pace it is used.  like
```
  static char *fuzzer_name;
``` 
and it will be reused in all other iterations. It is common C trick, as far as I know.

nataraj-hates-MS-for-stealing-github:
I've reread the code more carefully.

Everything is already done there.

There is static variable `banner`  [here](https://github.com/AFLplusplus/AFLplusplus/blob/1c582b6aeec034b2c0ef650abd35352df7d78fa7/src/afl-fuzz-stats.c#L542C15-L542C15)

If it dose [not have non-empty value in it](https://github.com/AFLplusplus/AFLplusplus/blob/1c582b6aeec034b2c0ef650abd35352df7d78fa7/src/afl-fuzz-stats.c#L778)

[then proper content is calculated ](https://github.com/AFLplusplus/AFLplusplus/blob/1c582b6aeec034b2c0ef650abd35352df7d78fa7/src/afl-fuzz-stats.c#L780-L795)

and [snpritf'ed into it](https://github.com/AFLplusplus/AFLplusplus/blob/1c582b6aeec034b2c0ef650abd35352df7d78fa7/src/afl-fuzz-stats.c#L797-L817)

So all this code (and my new code too) is executed only once, and it is not really important how much CPU is wasted...

I guess no more coding is required here, if we speak about CPU efficiency

vanhauser-thc:
oh my bad, hadnt seen the if(unlikely(!banner[0]))

nataraj-hates-MS-for-stealing-github:
Thanks!

PR #1882, opened 2023-10-03
Use sync_id instead of use_banner while building statsd metric messages

### LINKED ISSUE #1881: Incorrect banner value used in statsd reporting
Incorrect `banner` value used while reporting statsd statistics.

When running AFL++ with `AFL_STATSD_TAGS_FLAVOR=<any> AFL_STATSD=1` the banner value that is used in statistics reporting is equal to the name of the executable, while in reality it should be the `sync_id`, or the value directly provided using `-T` option. While the latter one works as expected, the first one is not.  This leads to a situation when multiple instances are reporting statistics, the data can't be differentiated as all the instances have the same `banner` value, which is the name of the executable.

Tested on the latest dev revision: `1f7f1eff4323720997079d468c9a55096ac2fc3e`

**To Reproduce**
1. Run your AFL++ instance with the options mentioned before, e.g., `AFL_STATSD_TAGS_FLAVOR=statsd AFL_STATSD=1 afl-fuzz -i - -o out -S none-1 -- ./xxx`
2. Run Grafana instance or somehow otherwise get the reported data.
3. The banner value will be the name of the executable instead of `sync_id`.
4. When running multiple instances, they all will be using the same `banner` value, thus the instances couldn't be distinguished.

**Expected behavior**
The `banner` value should be the `sync_id` or the value provided by the `-T` option (this one works as expected).

**Potential solution**

As I can see, a potential solution might use `afl->sync_id` instead of `afl->use_banner` in the https://github.com/AFLplusplus/AFLplusplus/blob/d09950f4bb98431576b872436f0fbf773ab895db/src/afl-fuzz-statsd.c#L226-L227

But that would make the `-T` not able to set the banner value for statsd reporting.

**Screen output/Screenshots**

Screenshot 2023-10-03 115214

**Additional context**
Adding `-T none-1` directly sets the banner to be the expected value.

m4drat:
This PR fixes https://github.com/AFLplusplus/AFLplusplus/issues/1881

Note, with this fix `-T` option will not have an influence on the banner that is used in statsd messages.

### DIFF
--- a/src/afl-fuzz-statsd.c
+++ b/src/afl-fuzz-statsd.c
@@ -223,7 +223,7 @@ int statsd_format_metric(afl_state_t *afl, char *buff, size_t bufflen) {
   char tags[MAX_TAG_LEN * 2] = {0};
   if (afl->statsd_tags_format) {
 
-    snprintf(tags, MAX_TAG_LEN * 2, afl->statsd_tags_format, afl->use_banner,
+    snprintf(tags, MAX_TAG_LEN * 2, afl->statsd_tags_format, afl->sync_id,
              VERSION);
 
   }


vanhauser-thc:
Thank you

PR #1883, opened 2023-10-04
Make fuzzer_stats update atomic

eqv:
This writes fuzzer_stats to a temp file and then atomically renames the temp file into fuzzer_stats so that any read on fuzzer_stats will always return a consistent view of the AFL state (otherwise there is a very low change of AFL's write and $tool's read lead to a race condition and yield inconsistent results).


vanhauser-thc:
:)

PR #1893, opened 2023-10-19
Add an env to afl-clang-fast to disable setting rpath if LLVM path isn't recognized

jschwartzentruber:
rpath is not desired, eg. in firefox where we use a custom clang and create a distributable build.


vanhauser-thc:
can you please also add a description to docs/env_variables.md ? thanks!

vanhauser-thc:
thank you!

PR #1894, opened 2023-10-19
Use proper AFL_NYX_AUX_SIZE for nyx_aux_string

choller:
Currently, AFL_NYX_AUX_SIZE doesn't work because AFL++ uses the hardcoded 0x1000 size for `nyx_aux_string` rather than the actually set size. This causes ASan traces to still be truncated to 4K, no matter what AFL_NYX_AUX_SIZE is set to.

The attached fix was validated to work in our setup and produces the proper traces now.

### DIFF
--- a/src/afl-forkserver.c
+++ b/src/afl-forkserver.c
@@ -615,14 +615,20 @@ void afl_fsrv_start(afl_forkserver_t *fsrv, char **argv,
 
     if (getenv("AFL_NYX_AUX_SIZE") != NULL) {
 
+      fsrv->nyx_aux_string_len = atoi(getenv("AFL_NYX_AUX_SIZE"));
+
       if (fsrv->nyx_handlers->nyx_config_set_aux_buffer_size(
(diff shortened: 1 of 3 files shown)


choller:
Cc @schumilo

src/afl-forkserver.c
```diff
@@ -615,15 +615,21 @@ void afl_fsrv_start(afl_forkserver_t *fsrv, char **argv,
 
     if (getenv("AFL_NYX_AUX_SIZE") != NULL) {
 
+      fsrv->nyx_aux_string_len = atoi(getenv("AFL_NYX_AUX_SIZE"));
+
       if (fsrv->nyx_handlers->nyx_config_set_aux_buffer_size(
-              nyx_config, atoi(getenv("AFL_NYX_AUX_SIZE"))) != 1) {
+              nyx_config, fsrv->nyx_aux_string_len) != 1) {
```
vanhauser-thc, line 621: Why not … > 4096 ?

vanhauser-thc:
Lgtm

PR #1899, opened 2023-10-25
Scale encoding

vanhauser-thc:


smoelius:
I doubt it matters, but in case you haven't heard of it: https://docs.substrate.io/reference/scale-codec/

vanhauser-thc:
I have and I only need the compact integer encoding and nothing else. so I rather implement that part myself instead of depending on an external library :)
everything else cmplog can already solve on it's own.

smoelius:
I should have been more clear.

I saw "Scale encoding" and thought the PR might have something to do with Parity's SCALE encoding. After glancing at the files, I concluded there was no relationship. I mentioned the Parity work just in case you were worried about the possibility of confusion. But the chances of that are probably low.

Sorry for the noise.

PR #1904, opened 2023-11-09
Use direct call to write to OpenBSD

jasperla:
The linker on OpenBSD emits a warning when linking this file: warning: syscall() may go away, please rewrite code to use direct calls.

Other `syscall()` invocations in AFLplusplus code aren't compiled on OpenBSD at the moment.


vanhauser-thc:
the purpose of this syscall is to circumvent libc and asan, I fear this will break then on openbsd. is there an alternative like for haiku?

jasperla:
There is no alternative as far as I know, the whole point of removing `syscall(2)` is by forcing application to go through libc. The risk of breaking ASAN on OpenBSD is an acceptable trade-off compared to not having AFL at all.

devnexen:
I think it s safe in asan side, openbsd does not have support for it, not even ubsan unless openbsd did its own asan ?

vanhauser-thc:
Ok :)

PR #1910, opened 2023-11-18
add post_run_target_script

yangzao:
Hi @vanhauser-thc, I added some code which allows AFL++ to run a script each time after it executes the target program. It gets the path of the script from environment variable "AFL_POST_RUN_TARGET_SCRIPT". Do you think the code is fine or if I need to change anything?

Thanks!


vanhauser-thc:
Hi,
I tried to point you on discord into the right direction - it should be a custom mutator function.
You try to use an environment variable with a command to be executed. This has a huge performance impact. For your specific case where you might only have 1 exec/s that does not make a big difference, however for other uses this would be an unacceptable slow down.
The correct way is to add another custom mutator function like `afl_custom_fuzz_send`.
This way you can specify a python script or native shared library to do whatever you need to do. You could execute a command from there, but also just put the code you need in the python/shared lib and enjoy a much lower performance penality.

yangzao:
Thanks! I will modify the code.

PR #1912, opened 2023-11-22
Output afl-clang-fast stuffs only if necessary

SonicStark:
TL;DR: If `afl-clang-fast`/`afl-clang-fast++` were always there, calling them may finally fall back into `gcc` rather than `clang`, which is confusing.

When `gcc` and `clang` is available in current system while `llvm-config` not, building AFLplusplus can still be OK and `afl-gcc`, `afl-clang` and `afl-clang-fast` can be found there. However if `afl-clang-fast` were called after that, weird things would happen: https://github.com/AFLplusplus/AFLplusplus/blob/61e27c6b54f7641a168b6acc6ecffb1754c10918/src/afl-cc.c#L1578-L1611

Here `strncmp(callname, "afl-clang-fast", 14)` would **never** work and `compiler_mode` would keep `UNSET`, because `LLVM_MAJOR` is empty without a valid `llvm-config`: https://github.com/AFLplusplus/AFLplusplus/blob/61e27c6b54f7641a168b6acc6ecffb1754c10918/GNUmakefile.llvm#L46 https://github.com/AFLplusplus/AFLplusplus/blob/61e27c6b54f7641a168b6acc6ecffb1754c10918/GNUmakefile.llvm#L293-L295

Then if env `AFL_CC_COMPILER` and `AFL_LLVM_*` is not set and `--afl-...` options are not used, `afl-cc` can bypass https://github.com/AFLplusplus/AFLplusplus/blob/61e27c6b54f7641a168b6acc6ecffb1754c10918/src/afl-cc.c#L2035-L2041 and go to https://github.com/AFLplusplus/AFLplusplus/blob/61e27c6b54f7641a168b6acc6ecffb1754c10918/src/afl-cc.c#L2043-L2063

Obviously `have_gcc` is `1` because we have `as` confirmed here: https://github.com/AFLplusplus/AFLplusplus/blob/61e27c6b54f7641a168b6acc6ecffb1754c10918/src/afl-cc.c#L1542-L1547

That is to say, `afl-clang-fast` would finally run like `afl-gcc`. Isn't this counter-intuitive? I have met several people who used AFL++ for the first time, and in the above situation, they instrumented their programs and found that the target binary was very different from the one build by `clang` or `afl-clang` (especially in the linking of runtimes). This did make them confused for quite a while:joy:


vanhauser-thc:
I get your point and agree.
People execute afl-clang-fast and what is actually executed underneath could be something different.
afl-cc should be so flexible to do that, but afl-clang-fast, afl-clang-lto, afl-gcc-fast, afl-gcc and afl-clang should be doing what their name is.
would you mind changing your PR so that instead afl-clang-fast is created if afl-llvm-pass.so is present, and afl-gcc-fast if  afl-gcc-pass.so?
and if you want to do the extra effort, change afl-cc to bail if afl-clang-fast is the executable name but the necessary pass is not there, and same for afl-gcc-fast, and afl-gcc baling if gcc is not there and afl-clang if there is no clang?

andreafioraldi:
afl-clang-fast can be always used in NATIVE if clang is found btw

vanhauser-thc:
@SonicStark ping

SonicStark:
So sorry for this late replay.

> afl-clang-fast can be always used in NATIVE if clang is found btw

@andreafioraldi was right. When `LLVM_MAJOR < 4`, `afl-clang-fast` should not be built. And when `LLVM_MAJOR >= 4`, it can be always there, with `compiler_mode = CLANG` as its **last** fallback, so if the built passes are all not available, `afl-cc` would go to https://github.com/AFLplusplus/AFLplusplus/blob/ef706ad668b36e65d24f352f5bcee22957f5f1cc/src/afl-cc.c#L2062-L2071 and work as expected. I’ll push a new commit for this later.

> would you mind changing your PR so that instead afl-clang-fast is created if afl-llvm-pass.so is present, and afl-gcc-fast if afl-gcc-pass.so?

As for `afl-gcc-fast`:
https://github.com/AFLplusplus/AFLplusplus/blob/ef706ad668b36e65d24f352f5bcee22957f5f1cc/GNUmakefile.gcc_plugin#L150-L155 https://github.com/AFLplusplus/AFLplusplus/blob/ef706ad668b36e65d24f352f5bcee22957f5f1cc/GNUmakefile.gcc_plugin#L123-L129
It is created just after `afl-gcc-pass.so`, and both of them depend on target `test_deps`, which would run `exit 1` if current `gcc` is not suitable. So that’s not a problem.

> if you want to do the extra effort, change afl-cc to bail if afl-clang-fast is the executable name but the necessary pass is not there, and same for afl-gcc-fast, and afl-gcc baling if gcc is not there and afl-clang if there is no clang?

More graceful sanity checks and fallback actions is needed in `afl-cc`, and I’m glad to do sth for this, maybe in a new PR, if needed. 
But I'm afraid the problem goes beyond that. 
Recently during my off hours, I have trying to bring LTO mode to [aflgo](https://github.com/aflgo/aflgo), hoping this can solve [parallel compiling](https://github.com/aflgo/aflgo/issues/41), [multiple binary targets](https://github.com/aflgo/aflgo/issues/87), [extra-long path](https://github.com/aflgo/aflgo/issues/138), and some other issues. So I read the source code of `afl-cc.c` first (still in progress, notes & updates locate in [afl-cc-rtfsc](https://github.com/SonicStark/afl-cc-rtfsc)), which is really a laborious deal, and some hidden issues came to the light one after another. For example:
https://github.com/AFLplusplus/AFLplusplus/blob/ef706ad668b36e65d24f352f5bcee22957f5f1cc/src/afl-cc.c#L1013 uses `have_instr_list`, but actually it is set as https://github.com/AFLplusplus/AFLplusplus/blob/ef706ad668b36e65d24f352f5bcee22957f5f1cc/src/afl-cc.c#L469-L471 when and only when `process_params` being called at https://github.com/AFLplusplus/AFLplusplus/blob/ef706ad668b36e65d24f352f5bcee22957f5f1cc/src/afl-cc.c#L1143 i.e. later than the using. `have_c` has the same issue, used at https://github.com/AFLplusplus/AFLplusplus/blob/ef706ad668b36e65d24f352f5bcee22957f5f1cc/src/afl-cc.c#L953 but set later until running to  https://github.com/AFLplusplus/AFLplusplus/blob/ef706ad668b36e65d24f352f5bcee22957f5f1cc/src/afl-cc.c#L559 The above two can still be seen in latest `dev` branch. So I suspect that there may be some problems sneaking in current LTO/NATIVE mode, but I haven’t seen any issues reported about this.
One more example, the following two blocks do the same thing, and the later will be executed for sure. So they may be duplicate? https://github.com/AFLplusplus/AFLplusplus/blob/ef706ad668b36e65d24f352f5bcee22957f5f1cc/src/afl-cc.c#L1128-L1137 https://github.com/AFLplusplus/AFLplusplus/blob/ef706ad668b36e65d24f352f5bcee22957f5f1cc/src/afl-cc.c#L398-L407

As far as I'm concerned, `afl-cc.c` is becoming another hulk like `afl-fuzz.c` in old AFL. So I hope to contribute more to `afl-cc`, preventing things from getting worse, but it may take a lot of efforts.

vanhauser-thc:
true. just pushed fixes for that

SonicStark:
UPDATE: I spent a few days to figure out how afl-cc works. And I have refactored it based on v4.08c, with 50 commits or so, in [afl-cc-rtfsc/release-4.08c-dat](https://github.com/SonicStark/afl-cc-rtfsc/tree/main/release-4.08c-dat). The patch for v4.08c is here: [afl-cc-changes.patch](https://github.com/AFLplusplus/AFLplusplus/files/13795050/afl-cc-changes.patch)

It seems that the new afl-cc works well—at least seeming good, with tests passed (`ASAN_BUILD=1`, both llvm-12 & llvm-16):
image

Features overview:
 - Split `afl-cc.c` into 8 more modules (could be more compact perhaps).
 - Hierarchical args parsing and modes choosing.
 - Safe built object finding.
 - Standalone objects linking.
 - Smoothly fix the messes mentioned before.

If you guys have interested in this, I'll push some commits in this PR to migrate the changes to dev branch. BTW I plan to run more regression tests if possible.

:tada:Happy new year!!!

vanhauser-thc:
can you please not split it up into multiple files? I know it is a huge file, but we try to keep all tools in a single file, with the exception of afl-fuzz.

SonicStark:
Oops. I'll merge them finally.
BTW I feel so lucky to have a nice big monitor on my desk🙃

SonicStark:
Looks working well. Further regression tests is needed to be more convincing.

vanhauser-thc:
thanks! I will take a close look at this the next days

vanhauser-thc:
LGTM, thank you

PR #1913, opened 2023-11-22
Pass correct Nyx ID when creating a Nyx runner

choller:
Each Nyx instance requires a numeric id when being created. Nyx internally assumes that id 0 is the parent id and all non-zero ids are child instances that will run what the parent runs. For this purpose, `fsrv->nyx_id` is actually populated with the id used with `-M/-S`. Mistakenly the code currently uses `nyx_bind_cpu_id` though which works by coincidence if you only ever run a single AFL++ instance with `-M` on a device and the CPU with id 0 is not busy already. Running two AFL++ instances with `-M` in parallel will cause a Nyx assertion to fail and reveals the failure.

### DIFF
--- a/src/afl-forkserver.c
+++ b/src/afl-forkserver.c
@@ -679,8 +679,7 @@ void afl_fsrv_start(afl_forkserver_t *fsrv, char **argv,
 
     }
 
-    fsrv->nyx_runner =
-        fsrv->nyx_handlers->nyx_new(nyx_config, fsrv->nyx_bind_cpu_id);
+    fsrv->nyx_runner = fsrv->nyx_handlers->nyx_new(nyx_config, fsrv->nyx_id);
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
@schumilo fine with you?

schumilo:
Patch looks good and yes, this is indeed a silly bug we want to fix. 
I haven't tested the AFL++ companion tools yet, but I would expect the tools to remain functional with this patch applied.  

We should probably also remove `nyx_bind_cpu_id` as it is no longer used. From what I remember, libnyx does not bind a QEMU-Nyx process to a specific CPU-ID anymore (this has changed at some point during development). Instead, this is now up to the fuzzer, which in this case is already handled by AFL++.

vanhauser-thc:
Thanks!

PR #1917, opened 2023-12-01
Stop hardcoding the path /usr/local/lib/afl in afl-ld-lto.c and respect the configured PREFIX

Romain-Geissler-1A:
This is very unrealistic scenarios, but in the strange case where someone would use the environment variable `ALF_PATH=` (ie empty string) really rely on the configure time AFL_PATH rather than an hardcoded `/usr/local/lib/afl` path.


vanhauser-thc:
Makes sense, thanks!

PR #1918, opened 2023-12-02
Add benchmark for Raspberry Pi 5

jakoblell:
This PR adds benchmark results for a Raspberry Pi 5 (with sufficient cooling to prevent thermal throttling).

Please note that the CPU frequency (`cpu_fastest_core_mhz`) has been added manually, for some reason the automatic detection does not work on the Raspberry.


vanhauser-thc:
Thanks, if you can figure out a fix for the CPU setting a PR is appreciated :)

PR #1920, opened 2023-12-04
Removing -Wl,rpath <LLVM_LIBDIR> when using GCC 

vmezzela:
In the scenario where the compiling toolchain is in a custom location, then `afl-cc` sets the option `-Wl,-rpath <LLVM_LIBDIR>` , I guess this is fine as long as we're not using **GCC**.

With this pull request I'm making sure that this option is not added when using **GCC**.


vanhauser-thc:
Thanks

PR #1922, opened 2023-12-05
fixing -Wl,-rpath=<LLVM_LIBDIR>

vmezzela:
In the scenario where we're using **LLVM** installed on a custom toolchain, `afl-cc` sets the option `-Wl,-rpath <LLVM_LIBDIR>`:
```c
cc_params[cc_par_cnt++] = "-Wl,-rpath";
cc_params[cc_par_cnt++] = <LLVM_LIBDIR>;
```

This is not working and I guess the problem is that the option to the linker should be passed using one of the two ways:
- `-Wl,-rpath=<LLVM_LIBDIR>`
- `-Wl,-rpath,<LLVM_LIBDIR>`

Both ways worked fine on my toolchain.


vanhauser-thc:
thank you

PR #1930, opened 2023-12-14
Reformat table in COMPARISON.md to markdown-compatible characters

janisozaur:


vanhauser-thc:
Makes sense but then you also have to do the necessary change in the python script

vanhauser-thc:
@janisozaur ?

janisozaur:
Oh, haven't noticed the ping, sorry. Which script is that?

PR #1940, opened 2023-12-22
llvm 18 build fixes.

devnexen:


devnexen:
@vanhauser-thc gonna love it :) I can kind of see why they removed the get*PtrTy shortcuts, but the starts_with/startswith situation is harder to comprehend .. we ll see.

vanhauser-thc:
oh my what a shit show ... so many modules that need to be changed ...

devnexen:
I think I m done, with the last version from apt.llvm the build and tests pass locally.

instrumentation/afl-llvm-dict2file.so.cc
```diff
@@ -433,15 +433,15 @@ bool AFLdict2filePass::runOnModule(Module &M) {
...
           isStrcmp &=
               FT->getNumParams() == 2 && FT->getReturnType()->isIntegerTy(32) &&
               FT->getParamType(0) == FT->getParamType(1) &&
-              FT->getParamType(0) == IntegerType::getInt8PtrTy(M.getContext());
+              FT->getParamType(0) == IntegerType::getInt8Ty(M.getContext())->getPointerTo(0);
           isStrcasecmp &=
               FT->getNumParams() == 2 && FT->getReturnType()->isIntegerTy(32) &&
               FT->getParamType(0) == FT->getParamType(1) &&
-              FT->getParamType(0) == IntegerType::getInt8PtrTy(M.getContext());
+              FT->getParamType(0) == IntegerType::getInt8Ty(M.getContext())->getPointerTo(0);
```
vanhauser-thc, line 444: Does this not need an ifdef? Or does this work with llvm 11/13 too?

devnexen: Not at all, what the get*PtrTy did is precisely this, they were just shortcuts. llvm 18 feels like "api cleanup" (getFixedSize which calls geFixedValue is the same vein).

vanhauser-thc:
Thank you!

PR #1942, opened 2023-12-25
Fix custom_send link

Xeonacid:
Add a leading '/' to walk in the repo root instead of current dir.


vanhauser-thc:
Thanks that fix is wrong though. You need to add ../ instead otherwise this won’t work when viewed locally

Xeonacid:
> Thanks that fix is wrong though. You need to add ../ instead otherwise this won’t work when viewed locally

Thanks for your review! Fixed now.

vanhauser-thc:
Thank you

PR #1948, opened 2024-01-05
Broken LTO mode and response file support due to #1912

SonicStark:
I still keep on doing some regression tests for #1912 manually by myself. So far I've found two bugs.

## Broken LTO mode

`SanitizerCoverageLTO.so` was taken over by `SanitizerCoveragePCGUARD.so`, which should never happen.

For example:
```text
$ /home/AFLplusplus/afl-clang-lto++ -fsanitize=fuzzer,address -O0 -gdwarf-4 -DUSE_LLVM_LIBFUZZER_STYLE -c -o /home/microBug/build/mbug.o /home/microBug/microBug.cpp

afl-cc++4.10a by Michal Zalewski, Laszlo Szekeres, Marc Heuse - mode: LLVM-LTO-PCGUARD
[+] Found '-fsanitize=fuzzer', replacing with libAFLDriver.a
SanitizerCoveragePCGUARD++4.10a
Note: Found constructor function _GLOBAL__sub_I_microBug.cpp with prio 65535, we will not instrument this, putting it into a block list.
[+] Instrumented 51 locations with no collisions (non-hardened mode) of which are 0 handled and 0 unhandled selects.

$ /home/AFLplusplus/afl-clang-lto -fsanitize=fuzzer,address -O0 -gdwarf-4 -c -o /home/microBug/build/calls.o /home/microBug/calls.c

afl-cc++4.10a by Michal Zalewski, Laszlo Szekeres, Marc Heuse - mode: LLVM-LTO-PCGUARD
[+] Found '-fsanitize=fuzzer', replacing with libAFLDriver.a
SanitizerCoveragePCGUARD++4.10a
[+] Instrumented 11 locations with no collisions (non-hardened mode) of which are 0 handled and 0 unhandled selects.

$ /home/AFLplusplus/afl-clang-lto++ -fsanitize=fuzzer,address -O0 -gdwarf-4 /home/microBug/build/calls.o /home/microBug/build/mbug.o -o /home/microBug/build/bug-san1-dyn1-dbg1-64

afl-cc++4.10a by Michal Zalewski, Laszlo Szekeres, Marc Heuse - mode: LLVM-LTO-PCGUARD
[+] Found '-fsanitize=fuzzer', replacing with libAFLDriver.a
afl-llvm-lto++4.10a by Marc "vanHauser" Heuse <mh@mh-sec.de>
[!] WARNING: No instrumentation targets found.
```

As you can see, afl-cc did not drive clang in LTO mode throughout. I made further investigation on this mess.

First of all, let's see what happened:
1. `afl-cc` got mode "LLVM-LTO-PCGUARD" because `aflcc->compiler_mode` is `LTO` and then `aflcc->instrument_mode` set to `INSTRUMENT_PCGUARD` https://github.com/AFLplusplus/AFLplusplus/blob/ee7d69b8175f31f2efb2b3bf15f2bbb29f61aa14/src/afl-cc.c#L1167-L1182
2. `aflcc->have_c` is true when compiling `calls.o` and `mbug.o`, so it goes to `add_optimized_pcguard`, with no LTO methods involved. When linking there is no `-c` so `SanitizerCoverageLTO.so` kicks in, but instrumentation has already done against each object previously, by `SanitizerCoveragePCGUARD.so`. https://github.com/AFLplusplus/AFLplusplus/blob/ee7d69b8175f31f2efb2b3bf15f2bbb29f61aa14/src/afl-cc.c#L2915-L2936

For the first point, earliest commit introduces the code pieces around `// force CFG` can be traced back to 3 years ago, in commit 7a86149: https://github.com/AFLplusplus/AFLplusplus/blob/7a861498c27997cd7be01a5650d54cff3b87a02e/src/afl-cc.c#L1671-L1684   That's reasonable, because for LTO instrumentation both `afl-llvm-lto-instrumentation.so` and `SanitizerCoverageLTO.so` are available at that time, and the later was recommended. So far the previous pass had gone, while the old code can also work well.

For the second point, earliest commit introduces `have_c` is commit 654f389, and that's the key! It was locally defined in `edit_params` and initialized as 0, first set at https://github.com/AFLplusplus/AFLplusplus/blob/654f389e73c9fd5b7e141b33ea28ab0fdda3178f/src/afl-cc.c#L712 which means the 5 references before all definitely have false logical values, making `if (lto_mode && !have_c)` actually equivalent to `if (lto_mode)`:
https://github.com/AFLplusplus/AFLplusplus/blob/654f389e73c9fd5b7e141b33ea28ab0fdda3178f/src/afl-cc.c#L465
https://github.com/AFLplusplus/AFLplusplus/blob/654f389e73c9fd5b7e141b33ea28ab0fdda3178f/src/afl-cc.c#L485
https://github.com/AFLplusplus/AFLplusplus/blob/654f389e73c9fd5b7e141b33ea28ab0fdda3178f/src/afl-cc.c#L505
https://github.com/AFLplusplus/AFLplusplus/blob/654f389e73c9fd5b7e141b33ea28ab0fdda3178f/src/afl-cc.c#L528
https://github.com/AFLplusplus/AFLplusplus/blob/654f389e73c9fd5b7e141b33ea28ab0fdda3178f/src/afl-cc.c#L564

4 of them were removed in commit fa2b040, and the last one was handed down to this very day——evaluated before setting linker options and loading LTO pass `SanitizerCoverageLTO.so`. However #1912 ensures consistency with actual situation for `have_c` before each use, just right breaks the trick in this aged code piece.🤣

In summary, I think this should be `if (lto_mode)`, unless there is a counter-example can be provided to prove [*warnings during compilation*](https://github.com/AFLplusplus/AFLplusplus/commit/654f389e73c9fd5b7e141b33ea28ab0fdda3178f) exists because of `-c` in LTO mode (at least not met during my test).

## Broken response file support

Because `aflcc->non_dash` detection shadows all "@..." options. The detection is now placed at very end of `parse_misc_params`, and keeps `final_` unchanged to prevent other misses in next parsing steps.

### DIFF
--- a/src/afl-cc.c
+++ b/src/afl-cc.c
@@ -2059,7 +2059,6 @@ void add_lto_passes(aflcc_state_t *aflcc) {
 #endif
 
   insert_param(aflcc, "-Wl,--allow-multiple-definition");
-  insert_param(aflcc, aflcc->lto_flag);
 
 }
 
@@ -2295,10 +2294,6 @@ param_st parse_misc_params(aflcc_state_t *aflcc, u8 *cur_argv, u8 scan) {
 
     SCAN_KEEP(aflcc->have_pic, 1);
 
-  } else if (cur_argv[0] != '-') {
-
-    SCAN_KEEP(aflcc->non_dash, 1);
-
   } else if (!strcmp(cur_argv, "-m32") ||
 
              !strcmp(cur_argv, "armv7a-linux-androideabi")) {
@@ -2373,6 +2368,14 @@ param_st parse_misc_params(aflcc_state_t *aflcc, u8 *cur_argv, u8 scan) {
       final_ = PARAM_DROP;
 
     }
+
+  } else if (cur_argv[0] != '-') {
+
+    /* It's a weak, loose pattern, with very different purpose
+     than others. We handle it at last, cautiously and robustly. */
+
+    if (scan && cur_argv[0] != '@')  // response file support
+      aflcc->non_dash = 1;
 
   }
 
@@ -2912,10 +2915,16 @@ static void edit_params(aflcc_state_t *aflcc, u32 argc, char **argv,
     //     use. insert_param(aflcc, "-flegacy-pass-manager");
     // #endif
 
-    if (aflcc->lto_mode && !aflcc->have_c) {
+    if (aflcc->lto_mode) {
 
-      add_lto_linker(aflcc);
-      add_lto_passes(aflcc);
+      insert_param(aflcc, aflcc->lto_flag);
+
+      if (!aflcc->have_c) {
+
+        add_lto_linker(aflcc);
+        add_lto_passes(aflcc);
+
+      }
 
     } else {
 


src/afl-cc.c
```diff
@@ -2912,7 +2916,7 @@ static void edit_params(aflcc_state_t *aflcc, u32 argc, char **argv,
     //     use. insert_param(aflcc, "-flegacy-pass-manager");
     // #endif
 
-    if (aflcc->lto_mode && !aflcc->have_c) {
+    if (aflcc->lto_mode) {
```
vanhauser-thc, line 2918: why are you removing this have_c check?

SonicStark: Just as what I talked a lot above:
1. Added initially in very early commit 654f389e73c9fd5b7e141b33ea28ab0fdda3178f.
2. Before #1912, `!have_c` always being true when `(lto_mode && !have_c)` being evaluated in execution, because it's used between initialized and the first set. This made the expression evaluated equally as `lto_mode`.
3. After #1912, `have_c` is 1 iff `-c` found in cmdline passed in, for each use in execution. This made LTO mode broken because we add pass, specify linker and add `lto_flag` in this `if (...)` branch. And if `-c` were found, `lto_flag` wouldn't be added.
4. This `have_c` check might be a forgotten useless trick which had been expected to be removed, IMO. Because there had been 5 `(lto_mode && !have_c)` long before, to [*try to remove warnings during compilation*](https://github.com/AFLplusplus/AFLplusplus/commit/654f389e73c9fd5b7e141b33ea28ab0fdda3178f). This is the last one left, and I can't reproduce the warnings due to this check in LTO mode during my test.

vanhauser-thc: I am not sure I understand what you write so let me explain:
if a `-c` is present, then 
```
      add_lto_linker(aflcc);
      add_lto_passes(aflcc);
```
should not be added.

vanhauser-thc:
I think it should look like this:

```
    if (aflcc->lto_mode) {

        if (!aflcc->have_c) {
            add_lto_linker(aflcc);
            add_lto_passes(aflcc);
        }

    } else {
```

src/afl-cc.c
SonicStark: Perhaps I'm too clumsy in describing the mess🤣
Original code piece before #1912 for contrast:
screenshot-20240109-191910
I just divided it into the two functions.

The key point that I can't figure out is: if 
> a `-c` is present

is really exactly equivalent to 
> `have_c` is true

why LTO mode always works well before?

AFAIK, LTO requires `-flto` regardless of whether `-c` is present or not.

SonicStark:
Strip adding `lto_flag` from `add_lto_passes`, IMO.
```c
    if (aflcc->lto_mode) {

        insert_param(aflcc, aflcc->lto_flag);

        if (!aflcc->have_c) {
            add_lto_linker(aflcc);
            add_lto_passes(aflcc);
        }

    } else {
```

SonicStark:
I'm still curious about how to reproduce the [warnings during compilation](https://github.com/AFLplusplus/AFLplusplus/commit/654f389e73c9fd5b7e141b33ea28ab0fdda3178f) caused by  `add_lto_linker(aflcc);` and `add_lto_passes(aflcc);` when `-c` is present. Maybe I should try more versions of llvm 11+

SonicStark:
> I'm still curious about how to reproduce the [warnings during compilation](https://github.com/AFLplusplus/AFLplusplus/commit/654f389e73c9fd5b7e141b33ea28ab0fdda3178f) caused by `add_lto_linker(aflcc);` and `add_lto_passes(aflcc);` when `-c` is present. Maybe I should try more versions of llvm 11+

Well, I just found a new bug might be introduced by commit df0638ab87a37f0a3604f5a95a4164153a3bd582, when testing LTO with LLVM 11:

```text
/usr/lib/llvm-11/bin/clang++ `llvm-config-11 --cxxflags` -fno-rtti -fno-exceptions -fPIC -O3 -funroll-loops -fPIC -Wall -g -I ./include/ -DVERSION=\"++4.10a\" -Wno-variadic-macros -Wno-deprecated-copy-with-dtor -DLLVM_MINOR=0 -DLLVM_MAJOR=11  -Wno-deprecated-declarations -Wl,-znodelete -Wno-writable-strings -fno-rtti -fPIC -std=c++14 -shared instrumentation/SanitizerCoverageLTO.so.cc -o SanitizerCoverageLTO.so `llvm-config-11 --ldflags`  instrumentation/afl-llvm-common.o
warning: unknown warning option '-Wno-deprecated-copy-with-dtor'; did you mean '-Wno-deprecated-copy-dtor'? [-Wunknown-warning-option]
instrumentation/SanitizerCoverageLTO.so.cc:1504:54: error: no member named 'getFixedValue' in 'llvm::TypeSize'; did you mean 'getFixedSize'?
  Array->setAlignment(Align(DL->getTypeStoreSize(Ty).getFixedValue()));
                                                     ^~~~~~~~~~~~~
                                                     getFixedSize
/usr/lib/llvm-11/include/llvm/Support/TypeSize.h:137:12: note: 'getFixedSize' declared here
  uint64_t getFixedSize() const {
           ^
1 warning and 1 error generated.
make[1]: *** [GNUmakefile.llvm:445: SanitizerCoverageLTO.so] Error 1
```

Clang version:
```text
$ clang-11 --version
Ubuntu clang version 11.1.0-++20211011094159+1fdec59bffc1-1~exp1~20211011214622.5
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

LLVM 12+ is not affected.

vanhauser-thc:
in LTO mode, the lto flags have always to be present. it is just the the lto plugins and the linker should not be set.

ignore the llvm 11 issue, I will disable LTO for that version

src/afl-cc.c
vanhauser-thc: LTO always requires -flto..
have_c should always be set if -c is present and never set if -c is not present. 
lto plugin and lto linker information should only be set if NOT have_c

SonicStark:
Now the mess should have been resolved. Thank you for your time!

vanhauser-thc:
looks good, thank you

PR #1952, opened 2024-01-10
Strip `-Wl,-no-undefined` during compilation

nbars:
Make the compiler wrapper stripping `-Wl,-no-undefined` in addition to `-Wl,--no-undefined`. Both flags are accepted by `clang` and used by building systems in the wild (e.g., samba will not build without this fix).

An example showcasing the issue:
```
❯ clang -Wl,-no-undefined
/usr/bin/ld: /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/13.2.1/../../../../lib64/Scrt1.o: in function `_start':
(.text+0x1b): undefined reference to `main'
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
❯ clang -Wl,--no-undefined
/usr/bin/ld: /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/13.2.1/../../../../lib64/Scrt1.o: in function `_start':
(.text+0x1b): undefined reference to `main'
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
```

### DIFF
--- a/src/afl-cc.c
+++ b/src/afl-cc.c
@@ -1962,8 +1962,8 @@ param_st parse_linking_params(aflcc_state_t *aflcc, u8 *cur_argv, u8 scan,
     }
 
   } else if (!strcmp(cur_argv, "-Wl,-z,defs") ||
-
              !strcmp(cur_argv, "-Wl,--no-undefined") ||
+             !strcmp(cur_argv, "-Wl,-no-undefined") ||
              !strcmp(cur_argv, "--no-undefined") ||
              strstr(cur_argv, "afl-compiler-rt") ||
              strstr(cur_argv, "afl-llvm-rt")) {
@@ -3021,4 +3021,3 @@ int main(int argc, char **argv, char **envp) {
   return 0;
 
 }
-


vanhauser-thc:
please note we only accept PRs to dev (see `CONTRIBUTING.md`).
and if you would have read this, you would have not have this conflict as another PR submitted a complete rewrite on afl-cc :-)

For your feature request - why is this something that afl-cc needs? In your example is see nothing that is AFL++ related.

nbars:
> please note we only accept PRs to dev (see CONTRIBUTING.md).
and if you would have read this, you would have not have this conflict as another PR submitted a complete rewrite on afl-cc :-)

I rebased my changes onto the `dev` branch :)

> For your feature request - why is this something that afl-cc needs? In your example is see nothing that is AFL++ related.

`afl-cc` needs to strip `-Wl,--no-undefined` to allow the compilation of, e.g., shared libraries since the AFL++ runtime is not compiled into those, consequently causing an undefined reference error (because `afl_area_ptr` is not defined) if `-Wl,--no-undefined` is not stripped. Since both, `gcc` and `clang`, allow `-Wl,-no-undefined` as an alias of `-Wl,--no-undefined`, AFL++ will fail to compile targets using the alias.

vanhauser-thc:
ok thanks :)

PR #1954, opened 2024-01-11
custom_mutator: add fuzz_run

Xeonacid:
This method can be used if you want to run target yourself, i.e. totally disable forkserver.
This is useful when running a daemon, only start and connect once on init, then send command on each run, result in a similar situation with persistent mode.
You need to set `__AFL_SHM_ID` and `AFL_MAP_SIZE` environ variables in an execve.
Each time called, AFL++ will clear the trace_bits for you, so you don't need to bother with this.
Note: only one `fuzz_run` is allowed, even with many custom mutators.

Both C and Python mutator have been tested locally.
The current implementation is a little dirty since it saved some duplicated variables in `afl_forkserver`.
Any review is highly appreciated!

### DIFF
--- a/src/afl-forkserver.c
+++ b/src/afl-forkserver.c
@@ -229,6 +229,7 @@ void afl_fsrv_init(afl_forkserver_t *fsrv) {
   fsrv->out_dir_fd = -1;
   fsrv->dev_null_fd = -1;
   fsrv->dev_urandom_fd = -1;
+  fsrv->shm_id = -1;
 
   /* Settings */
   fsrv->use_stdin = true;
@@ -238,6 +239,8 @@ void afl_fsrv_init(afl_forkserver_t *fsrv) {
   fsrv->mem_limit = MEM_LIMIT;
   fsrv->out_file = NULL;
   fsrv->child_kill_signal = SIGKILL;
+  fsrv->custom_mutator_data = NULL;
+  fsrv->custom_fuzz_run = NULL;
 
   /* exec related stuff */
   fsrv->child_pid = -1;
@@ -274,6 +277,9 @@ void afl_fsrv_init_dup(afl_forkserver_t *fsrv_to, afl_forkserver_t *from) {
   fsrv_to->child_kill_signal = from->child_kill_signal;
   fsrv_to->fsrv_kill_signal = from->fsrv_kill_signal;
   fsrv_to->debug = from->debug;
+  fsrv_to->custom_mutator_data = from->custom_mutator_data;
+  fsrv_to->custom_fuzz_run = from->custom_fuzz_run;
+  fsrv_to->shm_id = from->shm_id;
 
   // These are forkserver specific.
   fsrv_to->out_dir_fd = -1;
@@ -522,6 +528,9 @@ static void report_error_and_exit(int error) {
 
 void afl_fsrv_start(afl_forkserver_t *fsrv, char **argv,
                     volatile u8 *stop_soon_p, u8 debug_child_output) {
+
+  // Don't start forkserver if we have a custom_fuzz_run.
+  if (fsrv->custom_fuzz_run) { return; }
 
   int   st_pipe[2], ctl_pipe[2];
   s32   status;
@@ -1608,7 +1617,7 @@ afl_fsrv_run_target(afl_forkserver_t *fsrv, u32 timeout,
     enum NyxReturnValue ret_val =
         fsrv->nyx_handlers->nyx_exec(fsrv->nyx_runner);
 
-    fsrv->total_execs++;
+    ++fsrv->total_execs;
 
     switch (ret_val) {
 
@@ -1664,6 +1673,13 @@ afl_fsrv_run_target(afl_forkserver_t *fsrv, u32 timeout,
   memset(fsrv->trace_bits, 0, fsrv->map_size);
   MEM_BARRIER();
 #endif
+
+  if (fsrv->custom_fuzz_run) {
+
+    return fsrv->custom_fuzz_run(fsrv->custom_mutator_data, timeout,
+                                 fsrv->shm_id, fsrv->map_size);
+
+  }
 
   /* we have the fork server (or faux server) up and running
   First, tell it if the previous run timed out. */
--- a/src/afl-fuzz.c
+++ b/src/afl-fuzz.c
@@ -2231,6 +2231,30 @@ int main(int argc, char **argv_orig, char **envp) {
   afl->fsrv.trace_bits =
       afl_shm_init(&afl->shm, afl->fsrv.map_size, afl->non_instrumented_mode);
 
+  if (unlikely(afl->custom_mutators_count)) {
+
+    LIST_FOREACH(&afl->custom_mutator_list, struct custom_mutator, {
+
+      if (el->afl_custom_fuzz_run) {
+
+        if (!afl->fsrv.custom_fuzz_run) {
+
+          afl->fsrv.custom_fuzz_run = el->afl_custom_fuzz_run;
+          afl->fsrv.custom_mutator_data = el->data;
+          afl->fsrv.shm_id = afl->shm.shm_id;
+
+        } else {
+
+          FATAL("Only one afl_custom_fuzz_run is allowed!");
+
+        }
+
+      }
+
+    });
+
+  }
+
   if (!afl->non_instrumented_mode && !afl->fsrv.qemu_mode &&
       !afl->unicorn_mode && !afl->fsrv.frida_mode && !afl->fsrv.cs_mode &&
       !afl->afl_env.afl_skip_bin_check) {
@@ -2282,6 +2306,7 @@ int main(int argc, char **argv_orig, char **envp) {
       afl->fsrv.map_size = new_map_size;
       afl->fsrv.trace_bits =
           afl_shm_init(&afl->shm, new_map_size, afl->non_instrumented_mode);
+      afl->fsrv.shm_id = afl->shm.shm_id;
       setenv("AFL_NO_AUTODICT", "1", 1);  // loaded already
       afl_fsrv_start(&afl->fsrv, afl->argv, &afl->stop_soon,
                      afl->afl_env.afl_debug_child);
--- a/docs/custom_mutators.md
+++ b/docs/custom_mutators.md
@@ -58,7 +58,8 @@ int afl_custom_post_trim(void *data, unsigned char success);
 size_t afl_custom_havoc_mutation(void *data, unsigned char *buf, size_t buf_size, unsigned char **out_buf, size_t max_size);
 unsigned char afl_custom_havoc_mutation_probability(void *data);
 unsigned char afl_custom_queue_get(void *data, const unsigned char *filename);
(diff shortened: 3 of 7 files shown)


src/afl-fuzz.c
```diff
@@ -2282,6 +2306,7 @@ int main(int argc, char **argv_orig, char **envp) {
       afl->fsrv.map_size = new_map_size;
       afl->fsrv.trace_bits =
           afl_shm_init(&afl->shm, new_map_size, afl->non_instrumented_mode);
+      afl->fsrv.shm_id = afl->shm.shm_id;
```
vanhauser-thc, line 2309: this will not work in cmplog mode then. the fork server there needs the correct shmid

src/afl-forkserver.c
```diff
@@ -238,6 +238,7 @@ void afl_fsrv_init(afl_forkserver_t *fsrv) {
   fsrv->mem_limit = MEM_LIMIT;
   fsrv->out_file = NULL;
   fsrv->child_kill_signal = SIGKILL;
+  fsrv->custom_fuzz_run = NULL;
```
vanhauser-thc, line 243: you are not setting the shm_id here, nor copying it over in the init_dup() function

src/afl-forkserver.c
```diff
@@ -1665,6 +1670,14 @@ afl_fsrv_run_target(afl_forkserver_t *fsrv, u32 timeout,
   MEM_BARRIER();
 #endif
 
+  if (fsrv->custom_fuzz_run) {
+
+    fsrv->total_execs++;
```
vanhauser-thc: we prefer `++fsrv->custom_fuzz_run` because it is faster. silly I know but for several compilers this is true.

src/afl-fuzz.c
vanhauser-thc: also the map size is usually different in a cmplog forkserver target

vanhauser-thc:
I left some comments.

However I am not sure I get why this is a feature that is interesting to have. A forkserver will always be faster, so why this?
I do not see how this would be helpful with a daemon, because there you could just modify the source which would be much much much better. And in case this is too much trouble, why does the forkserver not work? it runs before main() hence no files or ports are opened from the server yet etc.
and I am not sure this would work well with qemu/frida mode either.

Xeonacid:
> I left some comments.

Thanks for your review! I will fix them soon.

> However I am not sure I get why this is a feature that is interesting to have. A forkserver will always be faster, so why this? I do not see how this would be helpful with a daemon, because there you could just modify the source which would be much much much better. And in case this is too much trouble, why does the forkserver not work? it runs before main() hence no files or ports are opened from the server yet etc.

Let me introduce my usage more precisely:
I'm fuzzing chromium (full binary instead of some components).
- Modifying the chromium codebase is obviously a headache.
- forkserver is really time-consuming on each chromium startup.

So I choose another approach: Start chromium instance on custom_mutator's init, with a webdriver attached. Each time to run the target, I command the webdriver to visit the corresponding website. This really saves a lot of time on repeatedly startup.

Of course, accepting this PR or not is on you. I'm totally OK with that. :)

> and I am not sure this would work well with qemu/frida mode either.

TBH, I have little knowledge on qemu/frida mode and have not tested them. 😕

vanhauser-thc:
> > However I am not sure I get why this is a feature that is interesting to have. A forkserver will always be faster, so why this? I do not see how this would be helpful with a daemon, because there you could just modify the source which would be much much much better. And in case this is too much trouble, why does the forkserver not work? it runs before main() hence no files or ports are opened from the server yet etc.
> 
> Let me introduce my usage more precisely: I'm fuzzing chromium (full binary instead of some components).
> 
> * Modifying the chromium codebase is obviously a headache.
> * forkserver is really time-consuming on each chromium startup.
> 
> So I choose another approach: Start chromium instance on custom_mutator's init, with a webdriver attached. Each time to run the target, I command the webdriver to visit the corresponding website. This really saves a lot of time on repeatedly startup.

that is very specific. hmmm. if I understand your setup correctly then chromium is running and never exits, the exec you perform is only instructing the chromium instance to load a page?

> > and I am not sure this would work well with qemu/frida mode either.
> 
> TBH, I have little knowledge on qemu/frida mode and have not tested them. 😕

but that is something that then has to be handles as well (even if only to not allow the feature there if untested).

Xeonacid:
> that is very specific. hmmm.

Maybe. Although I think other large software may suffer the same difficulty from modifying source code or heavy forkserver startup overhead.
But after all, this gives user the right to control one fuzz step, which does the same as other custom_mutator's capabilities, i.e. how to run in this PR; how to send or how to mutate in other functions.

> if I understand your setup correctly then chromium is running and never exits, the exec you perform is only instructing the chromium instance to load a page?

Yep!

> > > and I am not sure this would work well with qemu/frida mode either.
> > 
> > 
> > TBH, I have little knowledge on qemu/frida mode and have not tested them. 😕
> 
> but that is something that then has to be handles as well (even if only to not allow the feature there if untested).

Of course. I will dig into this.

src/afl-forkserver.c
Xeonacid: Oh! I missed it.
Fixed. Thanks!

src/afl-forkserver.c
Xeonacid: Fixed. Thanks!

src/afl-fuzz.c
Xeonacid: I think we may have to disable custom_fuzz_run in cmplog mode...
Since we have only one custom_mutator instance, it might not designed to run two binaries (the normal one plus the cmplog one).
Other existing custom_mutator functions may have the same problem? Should we create a separate instance for cmplog?

vanhauser-thc: cmplog is a highly valuable asset to have when fuzzing. not sure how this could be done for your case, might need you to experiment a bit.
it has to be a second process, even if it is the same binary (you can use a cmplog binary also for normal fuzzing, it is `just` slower).

vanhauser-thc:
I am still thinking about if this is something that is good to have in AFL++ as the case is very special. If so it would need a good description of an example setup in the documentation to show why this is useful and how to use it correctly.

vanhauser-thc:
sorry, accidently closed instead of just comment

src/afl-forkserver.c
```diff
@@ -274,6 +277,9 @@ void afl_fsrv_init_dup(afl_forkserver_t *fsrv_to, afl_forkserver_t *from) {
   fsrv_to->child_kill_signal = from->child_kill_signal;
   fsrv_to->fsrv_kill_signal = from->fsrv_kill_signal;
   fsrv_to->debug = from->debug;
+  fsrv_to->custom_mutator_data = from->custom_mutator_data;
+  fsrv_to->custom_fuzz_run = from->custom_fuzz_run;
+  fsrv_to->shm_id = from->shm_id;
```
vanhauser-thc, line 282: btw this is not really correct, it cannot have the same shm_id, so better to set this to -1, sorry.

SonicStark:
@Xeonacid AFAIK, `custom_mutator` focus on how to enrich **Mutator**s in fuzzing. Providing hooks for the **Executor** in a **Mutator** may unexpectedly break the whole fuzzing campaign of AFL++, imo.

FYI: to fuzz such a large target in your circumstances, I would suggest
1. Write a suitable proxy between the target and AFL++, as the harness. See [utils/afl_proxy](https://github.com/AFLplusplus/AFLplusplus/tree/stable/utils/afl_proxy).
2. Try [libAFL](https://github.com/AFLplusplus/LibAFL) to build a custom fuzzer, with **Mutator** and **Executor** cooperate on your demand.

src/afl-fuzz.c
domenukk: Maybe you want to have a `custom_fuzz_run_cmplog` then?

domenukk:
Does it really make sense to have a pointer to the shared map every time? If anything, it should be one init function, I think.

WRT what a custom mutator can/should do, I don't see a reason why we shouldn't support it here, it's optional after all (but of course using LibAFL is the better approach) (but good point, a proxy can do the trick as well)

vanhauser-thc:
an additional thought - maybe nyx_mode would be a better option? moving the snapshot to the right spot? it is then way faster than forking.

Xeonacid:
> @Xeonacid AFAIK, `custom_mutator` focus on how to enrich **Mutator**s in fuzzing. Providing hooks for the **Executor** in a **Mutator** may unexpectedly break the whole fuzzing campaign of AFL++, imo.

Thanks for your reply! But the existing custom_fuzz_send is unrelated to a "pure" mutator...

Xeonacid:
> Does it really make sense to have a pointer to the shared map every time? If anything, it should be one init function, I think.

Agree. But modifying the existing init function would cause an API break, sadly. Should we create something new like custom_fuzz_run_init? 😕

src/afl-fuzz.c
Xeonacid: > Maybe you want to have a `custom_fuzz_run_cmplog` then?

This would work, but I still have some question:
Other existing custom_mutator functions, which save some internal state, may have the same problem?  The mutator itself might not designed to run two binaries (the normal one plus the cmplog one).
Should we duplicate every function with a `_cmplog` prefix, or create a separate mutator instance for cmplog?
At least, I think this should be documented as a special note if we leave it as-is.

Do you have any idea? Also cc @vanhauser-thc

SonicStark:
>  But the existing custom_fuzz_send is unrelated to a "pure" mutator...

a mutator shouldn't interfere with the execution component, at least, imo.
See [this](https://aflplus.plus/fuzzcon_talk_2021.pdf):
screenshot-1705026932825

vanhauser-thc:
> > But the existing custom_fuzz_send is unrelated to a "pure" mutator...
> 
> a mutator shouldn't interfere with the execution component, at least, imo. See [this](https://aflplus.plus/fuzzcon_talk_2021.pdf): screenshot-1705026932825

I disagree here. the name custom mutator implies that but this is because that was the original goal. But since then it has moved to allow changing other parts of the fuzzer as well.
It makes the fuzzer more flexible without disadvantages so that is a good thing. 

Is this specific feature a good and necessary thing though that it would support more than a very specific target case? I am still torn.

SonicStark:
> the name custom mutator implies that but this is because that was the original goal. But since then it has moved to allow changing other parts of the fuzzer as well.
> It makes the fuzzer more flexible without disadvantages so that is a good thing.

Thanks for your clarification. Maybe it is renamed to sth like `custom_hooks`, one day in future🤣

vanhauser-thc:
I agree that would be best but it would mean a breaking change so rather not :)

vanhauser-thc:
after thinking a lot about this, I think this is too specific. not sure how to make this more general useful though.

Xeonacid:
> after thinking a lot about this, I think this is too specific. not sure how to make this more general useful though.

Then should I close the PR?

vanhauser-thc:
yeah I am closing this. sorry, you put a lot of effort in there. well at least it is useful for you and not wasted :)

PR #1955, opened 2024-01-11
Remove dead code in write_to_testcase

Xeonacid:
The custom_mutators_count check in if case is duplicate with if condition.
The else case is custom_mutators_count == 0, neither custom_mutator_list iteration nor sent check needed.

### DIFF
--- a/src/afl-fuzz-run.c
+++ b/src/afl-fuzz-run.c
@@ -169,20 +169,16 @@ write_to_testcase(afl_state_t *afl, void **mem, u32 len, u32 fix) {
 
     }
 
-    if (unlikely(afl->custom_mutators_count)) {
+    LIST_FOREACH(&afl->custom_mutator_list, struct custom_mutator, {
 
-      LIST_FOREACH(&afl->custom_mutator_list, struct custom_mutator, {
+      if (el->afl_custom_fuzz_send) {
 
-        if (el->afl_custom_fuzz_send) {
+        el->afl_custom_fuzz_send(el->data, *mem, new_size);
+        sent = 1;
 
-          el->afl_custom_fuzz_send(el->data, *mem, new_size);
-          sent = 1;
+      }
 
-        }
-
-      });
-
-    }
+    });
 
     if (likely(!sent)) {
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
I agree that the second FOREACH can be removed.

but the `if () { FOREACH() }` in general is necessary because if you look at the macro:

```
  do {                                                             \
                                                                   \
    list_t    *li = (list);                                        \
    element_t *head = get_head((li));                              \
    element_t *el_box = (head)->next;                              \
    if (!el_box) FATAL("foreach over uninitialized list");         \
    while (el_box != head) {                                       \
                                                                   \
      __attribute__((unused)) type *el = (type *)((el_box)->data); \
      /* get next so el_box can be unlinked */                     \
      element_t *next = el_box->next;                              \
      {block};                                                     \
      el_box = next;                                               \
                                                                   \
    }                                                              \
                                                                   \
  } while (0);
```
this would be more instructions than the `if`, and we want the most speed possible if no custom mutator is present.
Please update your PR.

Xeonacid:
> this would be more instructions than the if, and we want the most speed possible if no custom mutator is present.

If I understand correctly, my confusing words made you misunderstand. Sorry for that. :P

> The custom_mutators_count check in if case is duplicate with if condition.

I'm mentioning the outer if below, i.e.:

```c
u32 __attribute__((hot))
write_to_testcase(afl_state_t *afl, void **mem, u32 len, u32 fix) {

  u8 sent = 0;

  if (unlikely(afl->custom_mutators_count)) {      // <- this one
    // We don't need a if () { FOREACH() } here
    // The if clause must be true
  } else {
  }
}
```

src/afl-fuzz-run.c
```diff
@@ -169,20 +169,16 @@ write_to_testcase(afl_state_t *afl, void **mem, u32 len, u32 fix) {
 
     }
 
-    if (unlikely(afl->custom_mutators_count)) {
```
vanhauser-thc, line 172: but here you remove the if ()

Xeonacid: > but here you remove the if ()

https://github.com/Xeonacid/AFLplusplus/blob/f3d8bfc130b8ab36b22a456efa66e0212d0cd091/src/afl-fuzz-run.c#L98

The outter `if` is still here!
Yeah, owe have two same `if` nested originally!

vanhauser-thc:
ah I could not see the full context. thanks!

Issue #1950, opened 2024-01-08
how to use nyx mode for amd cpu


spobit:
**Describe the bug**

I`m using nyx mode to fuzz, but my cpu is AMD, so, i need use no-pt mode.
but, error happened.

**To Reproduce**
1. change coed to following:
/home/nyx/AFLplusplus/nyx_mode/libnyx/fuzz_runner/src/nyx/params.rs:
```
cmd.push("-machine".to_string());
// cmd.push("kAFL64-v2".to_string());
cmd.push("kAFL64-Q35".to_string());

cmd.push("-cpu".to_string());
// cmd.push("kAFL64-Hypervisor-v1".to_string());
cmd.push("kAFL64-Hypervisor-v2".to_string());
```
2. run cmd: lear; afl-fuzz -i /home/nyx/jc_first_fuzz_sample/in -o /home/nyx/jc_first_fuzz_sample/out -X -- /home/nyx/jc_first_fuzz_sample/nyx
```
main::489
afl-fuzz++4.10a based on afl by Michal Zalewski and a large online community
[+] AFL++ is maintained by Marc "van Hauser" Heuse, Dominik Maier, Andrea Fioraldi and Heiko "hexcoder" Eißfeldt
[+] AFL++ is open source, get it at https://github.com/AFLplusplus/AFLplusplus
[+] NOTE: AFL++ >= v3 has changed defaults and behaviours - see README.md
[+] AFL++ Nyx mode is enabled (developed and mainted by Sergej Schumilo)
[+] Nyx is open source, get it at https://github.com/Nyx-Fuzz
[+] No -M/-S set, autoconfiguring for "-S default"
[*] Getting to work...
[+] Using exploration-based constant power schedule (EXPLORE)
[+] Enabled testcache with 50 MB
[+] Generating fuzz data with a length of min=1 max=1048576
main::1797
main::1805
[*] Trying to load libnyx.so plugin...
[+] libnyx plugin is ready!
main::1816
[+] You have 4 CPU cores and 1 runnable tasks (utilization: 25%).
[+] Try parallel jobs - see /usr/local/share/doc/afl/fuzzing_in_depth.md#c-using-multiple-cores
[*] Setting up output directories...
[+] Output directory exists but deemed OK to reuse.
[*] Deleting old session data...
[+] Output dir cleanup successful.
[*] Checking CPU core loadout...
[+] Found a free CPU core, try binding to #0.
[...]

PR #1963, opened 2024-01-20
Bugfix for Nyx mode

schumilo:
This PR fixes a problem with AFL++ Nyx mode, introduced with a previous commit that caused an issue with support for vanilla KVM builds and targets not requiring Intel PT support enabled. Accordingly, the following issue should now also be resolved with this patch: 

https://github.com/AFLplusplus/AFLplusplus/issues/1950

I've tested this PR with both vanilla KVM and KVM-Nyx, and it seems to work quite well again with both kernels. In addition, all submodules have also been tested with the latest version of Nyx-Testing (which now includes a few more tests for different edges cases): https://github.com/nyx-fuzz/Nyx-Testing


vanhauser-thc:
Thanks!

PR #1965, opened 2024-01-23
replay mode support

quarta-qti:
Hi! This pull request introduces functionality to replay records stored by using AFL_PERSISTENT_RECORD.
It includes the following changes:

- store AFL_PERSISTENT_RECORD crashes, and hangs in the respective folders
- adds the AFL_PERSISTENT_REPLAY define that includes in compiler-rt lib support for replaying records
- includes a self-contained persistent_replay.h header file that consists of function required by the replay functionality, and when included after defining AFL_COMPAT (see persistent mode util) works as a compatibility layer to allow using compilers without afl-cc and include the replay functionality (or a simple persistent loop for what is worth).

To use the persistent replay functionality, the record number must be specified via the AFL_PERSISTENT_REPLAY env var (i.e., RECORD:XXXXX -> AFL_PERSISTENT_REPLAY=XXXXX) and the directory can be specified via AFL_PERSISTENT_DIR, otherwise the default will be considered the current directory ("./").

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -83,13 +83,11 @@
    will be kept and written to the crash/ directory as RECORD:... files.
    Note that every crash will be written, not only unique ones! */
 
-#define AFL_PERSISTENT_RECORD
-
-/* Adds support in compiler-rt to replay persistent records */
-#define AFL_PERSISTENT_REPLAY
+// #define AFL_PERSISTENT_RECORD
 
 /* Adds support in compiler-rt to replay persistent records in @@-style
  * harnesses */
+
 //  #define AFL_PERSISTENT_REPLAY_ARGPARSE
 
 /* console output colors: There are three ways to configure its behavior
--- a/include/config.h
+++ b/include/config.h
@@ -83,7 +83,10 @@
    will be kept and written to the crash/ directory as RECORD:... files.
    Note that every crash will be written, not only unique ones! */
 
-// #define AFL_PERSISTENT_RECORD
+//  #define AFL_PERSISTENT_RECORD
+
+/* Builds compiler-rt with support to replay persistent records */
+//  #define AFL_PERSISTENT_REPLAY
 
 /* console output colors: There are three ways to configure its behavior
  * 1. default: colored outputs fixed on: defined USE_COLOR && defined
--- a/include/config.h
+++ b/include/config.h
@@ -83,10 +83,14 @@
    will be kept and written to the crash/ directory as RECORD:... files.
    Note that every crash will be written, not only unique ones! */
 
-//  #define AFL_PERSISTENT_RECORD
+#define AFL_PERSISTENT_RECORD
 
-/* Builds compiler-rt with support to replay persistent records */
-//  #define AFL_PERSISTENT_REPLAY
+/* Adds support in compiler-rt to replay persistent records */
+#define AFL_PERSISTENT_REPLAY
+
+/* Adds support in compiler-rt to replay persistent records in @@-style
+ * harnesses */
+//  #define AFL_PERSISTENT_REPLAY_ARGPARSE
 
 /* console output colors: There are three ways to configure its behavior
  * 1. default: colored outputs fixed on: defined USE_COLOR && defined
--- a/src/afl-forkserver.c
+++ b/src/afl-forkserver.c
@@ -1796,13 +1796,18 @@ afl_fsrv_run_target(afl_forkserver_t *fsrv, u32 timeout,
 
     fsrv->last_kill_signal = fsrv->child_kill_signal;
 
-#ifndef AFL_PERSISTENT_RECORD
-    return FSRV_RUN_TMOUT;
-#else
-    retval = FSRV_RUN_TMOUT;
-    persistent_out_fmt = "%s/hangs/RECORD:%06u,cnt:%06u";
-    goto store_persistent_record;
+#ifdef AFL_PERSISTENT_RECORD
+    if (unlikely(fsrv->persistent_record)) {
+
+      retval = FSRV_RUN_TMOUT;
+      persistent_out_fmt = "%s/hangs/RECORD:%06u,cnt:%06u";
+      goto store_persistent_record;
+
+    }
+
 #endif
+
+    return FSRV_RUN_TMOUT;
 
   }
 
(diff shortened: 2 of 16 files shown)


domenukk:
Looks super useful!
I think this needs a `make code-format`

vanhauser-thc:
yes this looks great! please do the `make code-format` :)

vanhauser-thc:
@quarta-qti can you please check?

quarta-qti:
Hi! Sorry for the late reply, I was on vacation without my laptop with me and couldn't reply! :) On it!

quarta-qti:
Done. I also added relevant documentation in README.persistent_mode.md, added error handling for when a non-existing record directory is passed, and added another define "AFL_PERSISTENT_REPLAY_ARGPARSE" to selectively enable support for @@-style harnesses to increase compatibility since -afaik- not all platforms allow passing command line arguments to constructors.

include/config.h
```diff
@@ -83,7 +83,14 @@
    will be kept and written to the crash/ directory as RECORD:... files.
    Note that every crash will be written, not only unique ones! */
 
-// #define AFL_PERSISTENT_RECORD
+#define AFL_PERSISTENT_RECORD
```
vanhauser-thc: this has a fuzzing overhead, so I would not want to enable it by default

utils/persistent_mode/persistent_demo_new.c
```diff
@@ -31,17 +31,8 @@
...
-
-  #define __AFL_FUZZ_TESTCASE_LEN fuzz_len
-  #define __AFL_FUZZ_TESTCASE_BUF fuzz_buf
-  #define __AFL_FUZZ_INIT() void sync(void);
-  #define __AFL_LOOP(x) \
-    ((fuzz_len = read(0, fuzz_buf, sizeof(fuzz_buf))) > 0 ? 1 : 0)
-  #define __AFL_INIT() sync()
-
+  #define AFL_COMPAT
+  #include "persistent_replay.h"
```
vanhauser-thc: this includes then a lot more that does into belong to the demo - why?

include/config.h
```diff
@@ -83,7 +83,14 @@
...
+
+/* Adds support in compiler-rt to replay persistent records */
+#define AFL_PERSISTENT_REPLAY
```
vanhauser-thc: also this should not by default on I think. but I a second define? persistent recording only makes sense if also replay is added? so why not removing this one and just use afl_persistent_record?

vanhauser-thc:
@quarta-qti looks good, I left some comments for discussion :)

include/config.h
quarta-qti: Ack! This slipped in inadvertently because of the final testing round, gonna comment it.

include/config.h
quarta-qti: I thought persistent replay functionality might introduce some overhead in the persistent loop, and preferred to let the user choose if they want to keep a non-"replay" compiler-rt while still being able to use the record functionality in afl-fuzz.
Tbh, I did not measure if there's significant overhead added, will do and report back the  results.

If persistent replay adds significant overhead I would rather add another ifdef block to define AFL_PERSISTENT_RECORD if not defined when AFL_PERSISTENT_REPLAY is defined.
```c
#ifdef AFL_PERSISTENT_REPLAY
#ifndef AFL_PERSISTENT_RECORD
#define AFL_PERSISTENT_RECORD
#endif
#endif
```

utils/persistent_mode/persistent_demo_new.c
quarta-qti: I used the demo to test and showcase the  self-contained replay functionality, I'm guessing that it was not the right place to do it, and it's better to keep it as it was originally.

I think it would make sense to either:
  + move the self-contained _replay_ demo into a new `persistent_demo_replay.c`,
  + add back the original defines and wrap the two blocks into a conditional block like `#ifdef AFL_PERSISTENT_REPLAY`

Although not directly, I hope this answers your question. 😛

include/config.h
quarta-qti: On a second thought:
  + Overhead on persistent loop should be extremely minimal. It's pretty much just one more if.
  + Initialization should not be a problem, and can also be skipped with manual forkserver initialization.

I'll just use `unlikely` on the replay mode check, and then use only one define.

utils/persistent_mode/persistent_demo_new.c
vanhauser-thc: yes just put that into persistent_demo_replay.c please :)

include/config.h
vanhauser-thc: sounds good!

quarta-qti:
All done. Cleaned up a bit the persistent loop to keep the area ptr/loc logic intact to allow reusing afl tools with the only difference from the regular persistent loop being it won't raise a STOP signal.
Also moved the compatibility functionality to a new header (afl-record-compat.h) so there's no need for the additional AFL_COMPAT define, and moved the record replay example to it's own dir in utils to keep things neat.

src/afl-forkserver.c
```diff
@@ -1684,7 +1689,7 @@ afl_fsrv_run_target(afl_forkserver_t *fsrv, u32 timeout,
 
   }
 
-#ifdef AFL_PERSISTENT_RECORD
+#ifdef AFL_eERSISTENT_RECORD
```
vanhauser-thc: you introduced a typo here

src/afl-forkserver.c
```diff
@@ -1811,48 +1823,58 @@ afl_fsrv_run_target(afl_forkserver_t *fsrv, u32 timeout,
           (fsrv->uses_crash_exitcode &&
            WEXITSTATUS(fsrv->child_status) == fsrv->crash_exitcode))) {
```
vanhauser-thc, line 1830: from here below to the end of the file the changes seem to change logic/behaviour of non AFL_PERSISTENT_RECORD - can you please have a look?

utils/replay_record/Makefile
```diff
@@ -0,0 +1,8 @@
+all:
```
vanhauser-thc, line 1: This directory needs a short README.md what it actually does, even if it just points the the README.persistent_mode.md file.

vanhauser-thc:
we are getting close to merging :)

src/afl-forkserver.c
ocean1: On it!

quarta-qti: Silly mistake here, it just needed wrapping the persistent record logic in a conditional statement checking if the mode is active. Same goes for the hangs logic before here. Fixed! Thanks for spotting this out!

vanhauser-thc:
looks good now, thanks a lot!

quarta-qti:
Thank you! 😊

PR #1967, opened 2024-01-24
push to stable

vanhauser-thc:

PR #1968, opened 2024-01-25
Improve afl-cc for #1967

SonicStark:
For #1967:
 - Check if it is "-fsanitize=..." before we do sth.
 - Remove improper param_st transfer.
 - Check if too many cmdline params here, each time before insert a new param. This may hurt speed a little, but keeps `cc_params` away from heap-buffer-overflow.

### DIFF
--- a/src/afl-cc.c
+++ b/src/afl-cc.c
@@ -192,14 +192,10 @@ u8 *find_object(aflcc_state_t *, u8 *obj);
 
 void find_built_deps(aflcc_state_t *);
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thank you!

vanhauser-thc:
in your next PR can you please add more comments? eg how parse_fsanitize() works is hard to understand with the states and why.

SonicStark:
> in your next PR can you please add more comments? eg how parse_fsanitize() works is hard to understand with the states and why.

Understood. I'm very glad to share my understandings about how afl-cc works :)

PR #1972, opened 2024-01-28
Enhancement on Deterministic stage

kdsjZh:
Hi, 

We enhance the AFL++'s capability by implementing selective deterministic mutation. To enable this module, just run without `-d` or use `-D.`

Evaluation against AFL++ on fuzzbench can be found at https://fuzzbench.com/reports/experimental/2024-01-24-aflpp-sk2/index.html. (aflplusplus_sk is our implementation. Note that one program didn't build successfully; the aflpp cached the previous results while sk didn't have that cache.)

@vanhauser-thc  :)


src/afl-skip-log.cc
```diff
@@ -0,0 +1,54 @@
...
+
+}
+
+FILE *profile_fd;
+
+void plot_profile_data(afl_state_t *afl, struct queue_entry *q) {
+
+  if (!profile_fd) {
+
+    std::string out_dir = (const char *)afl->out_dir;
```
vanhauser-thc: Why does this need to be a c++ file? I see no point for that. and please merge it into afl-fuzz-stats.c

src/afl-skipdet.c
```diff
@@ -0,0 +1,426 @@
+
```
vanhauser-thc: please rename this to afl-fuzz-skipdet.c

src/afl-fuzz-one.c
```diff
@@ -3430,6 +3483,21 @@ u8 fuzz_one_original(afl_state_t *afl) {
...
+  afl->havoc_prof->total_queued_det += afl->havoc_prof->queued_det_stage;
+  afl->havoc_prof->edge_det_stage = before_havoc_edges - before_det_edges;
+  afl->havoc_prof->edge_havoc_stage =
+      count_non_255_bytes(afl, afl->virgin_bits) - before_havoc_edges;
+  afl->havoc_prof->total_det_edge += afl->havoc_prof->edge_det_stage;
+  afl->havoc_prof->det_stage_time = before_havoc_time - before_det_time;
+  afl->havoc_prof->havoc_stage_time = get_cur_time() - before_havoc_time;
+  afl->havoc_prof->total_det_time += afl->havoc_prof->det_stage_time;
+
+  plot_profile_data(afl, afl->queue_cur);
```
vanhauser-thc, line 3509: do you think plotting this away by default is a good thing? what would be the value to normal users?
maybe it is better to put it within `#ifdef INTROSPECTION ... #endif`?

GNUmakefile
```diff
@@ -460,8 +460,14 @@ src/afl-forkserver.o : $(COMM_HDR) src/afl-forkserver.c include/forkserver.h
 src/afl-sharedmem.o : $(COMM_HDR) src/afl-sharedmem.c include/sharedmem.h
 	$(CC) $(CFLAGS) $(CFLAGS_FLTO) -c src/afl-sharedmem.c -o src/afl-sharedmem.o
 
-afl-fuzz: $(COMM_HDR) include/afl-fuzz.h $(AFL_FUZZ_FILES) src/afl-common.o src/afl-sharedmem.o src/afl-forkserver.o src/afl-performance.o | test_x86
-	$(CC) $(CFLAGS) $(COMPILE_STATIC) $(CFLAGS_FLTO) $(AFL_FUZZ_FILES) src/afl-common.o src/afl-sharedmem.o src/afl-forkserver.o src/afl-performance.o -o $@ $(PYFLAGS) $(LDFLAGS) -lm
+src/afl-skipdet.o : $(COMM_HDR) src/afl-skipdet.c
+	$(CC) $(CFLAGS) $(CFLAGS_FLTO) -c src/afl-skipdet.c -o src/afl-skipdet.o
+
+src/afl-skip-log.o : $(COMM_HDR) src/afl-skip-log.cc
```
vanhauser-thc: please remove the .o entries and pass it to the afl-fuzz lines as .c files. we only what to create .o files for those that are used by multiple tools, e.g. also afl-showmap.

vanhauser-thc:
I left a few nits and questions :) thanks!

vanhauser-thc:
Also please make sure that the code compiles:
```
src/afl-fuzz-one.c:573:7: warning: implicit declaration of function ‘likeyly’; did you mean ‘likely’? [-Wimplicit-function-declaration]
  573 |   if (likeyly(afl->skip_deterministic) || likely(afl->queue_cur->passed_det) ||
      |       ^~~~~~~
      |       likely
```

GNUmakefile
```diff
@@ -532,6 +536,7 @@ endif
 .PHONY: code-format
 code-format:
 	./.custom-format.py -i src/*.c
+	./.custom-format.py -i src/*.cc
```
vanhauser-thc: this can be removed

kdsjZh:
Thanks for the code review.

I fixed the issues below and tested it on Fuzzbench; looks like it is ok now. 

BTW, FuzzBench's aflpp configuration use `-d`, which enforce the default havoc only mode. If we want to use skipdet as default option, we can either remove the `-d` or change to `-D`. (I change the default mode, i.e, without -D/-d, as skipdet)

README.skip.md
```diff
@@ -0,0 +1 @@
+this is developed based on AFL++ 60dc37a8cf09f8e9048e4b6a2204d6c90b27655a
```
vanhauser-thc: I think this README can be removed?
it would be good to have a description about what skipdet does though, who developed it etc. maybe in a file called docs/new_technology.md ? (in which future new tech can be added to.)
does not need to be in this PR, can also be added later.

kdsjZh: I forgot to remove it, I'll add more details to the doc after paper is done.

src/afl-fuzz-state.c
```diff
@@ -102,7 +102,7 @@ void afl_state_init(afl_state_t *afl, uint32_t map_size) {
   afl->stats_update_freq = 1;
   afl->stats_file_update_freq_msecs = STATS_UPDATE_SEC * 1000;
   afl->stats_avg_exec = 0;
-  afl->skip_deterministic = 1;
```
domenukk, line 105: So this changes the default to not skip deterministic stage? (Why) is this a good idea? If so, we need to update docs, right?

vanhauser-thc: this is what this PR is about. it is a new approach and makes afl++ slightly better. yes the documentation then needs to be updated.

kdsjZh: We replace the original deterministic stage to `skipdet`, which only do much less deterministic mutation. Therefore, this change means we use `skipdet` by default and original havoc only if `-d` is set. The vanilla deterministic stage is removed. If you think that make sense I can add the vanilla deterministic stage option back again

kdsjZh: This is an evaluation we ran on fuzzbench. skipdet (aflplusplus_sk) show some improvement over AFL++ so I open this PR. https://fuzzbench.com/reports/experimental/2024-01-24-aflpp-sk2/index.html.

GNUmakefile
```diff
@@ -460,6 +460,10 @@ src/afl-forkserver.o : $(COMM_HDR) src/afl-forkserver.c include/forkserver.h
 src/afl-sharedmem.o : $(COMM_HDR) src/afl-sharedmem.c include/sharedmem.h
 	$(CC) $(CFLAGS) $(CFLAGS_FLTO) -c src/afl-sharedmem.c -o src/afl-sharedmem.o
 
+src/afl-fuzz-skipdet.o : $(COMM_HDR) src/afl-fuzz-skipdet.c
```
vanhauser-thc: as I wrote, please do not create a .o file

GNUmakefile
```diff
@@ -460,6 +460,10 @@ src/afl-forkserver.o : $(COMM_HDR) src/afl-forkserver.c include/forkserver.h
...
+	$(CC) $(CFLAGS) $(CFLAGS_FLTO) -c src/afl-fuzz-skipdet.c -o src/afl-fuzz-skipdet.o
+
+
 afl-fuzz: $(COMM_HDR) include/afl-fuzz.h $(AFL_FUZZ_FILES) src/afl-common.o src/afl-sharedmem.o src/afl-forkserver.o src/afl-performance.o | test_x86
 	$(CC) $(CFLAGS) $(COMPILE_STATIC) $(CFLAGS_FLTO) $(AFL_FUZZ_FILES) src/afl-common.o src/afl-sharedmem.o src/afl-forkserver.o src/afl-performance.o -o $@ $(PYFLAGS) $(LDFLAGS) -lm
```
vanhauser-thc: you just have to add afl-fuzz-skipdet.c to `AFL_FUZZ_FILES`

vanhauser-thc:
you still have to do this:
> you have to add afl-fuzz-skipdet.c to AFL_FUZZ_FILES in GNUmakefile

also note that the CI will always fail because of the determinstic change. I will take care of that and doc update after merging

kdsjZh:
The custom mutator and python mutator is only compatible with havoc only. Could we ignore this check? If we set `-d` flag, it can work in origin havoc mode

Ok, thanks

kdsjZh:
> you have to add afl-fuzz-skipdet.c to AFL_FUZZ_FILES in GNUmakefile

As I understand, the GNUmakefile will scan the afl-fuzz*.c automatically, so it should work without any modification. And it work in my local build.

```AFL_FUZZ_FILES = $(wildcard src/afl-fuzz*.c)```

**Comment by vanhauser-thc** [comment]
> > you have to add afl-fuzz-skipdet.c to AFL_FUZZ_FILES in GNUmakefile
> 
> As I understand, the GNUmakefile will scan the afl-fuzz*.c automatically, so it should work without any modification. And it work in my local build.
> 
> `AFL_FUZZ_FILES = $(wildcard src/afl-fuzz*.c)`

oh right!! and I was surprised that the compilation in the CI did not fail :)
well we have lots of magic and I often forget that it is there :)

**Comment by vanhauser-thc** [comment]
to ensure everything is working as it should I just started another fuzzbench run, will pop up in a few hours at
https://www.fuzzbench.com/reports/experimental/2024-01-28-aflpp-sk/index.html

**Comment by kdsjZh** [comment]
BTW, I found that the aflplusplus_sk's fuzzbench setting didn't have any additional flag, this may let aflpp_sk run as aflpp havoc only.

aflpp inherits ```run_afl_fuzz `` from afl, in which afl will add the `-d` flag. So you may want to add `-D` in additional flags to enable skipdet module. 

In my local build the aflpp_sk configure do havoc only. let me try if `-D` work.

**Comment by kdsjZh** [comment]
well, the afl will only remove `-d` if given `-M/-S`.

```
    if not additional_flags or check_skip_det_compatible(additional_flags):
        command.append('-d')
```

Then the easiest way is to change `-d` to skipdet and `-D` to enforce old havoc. This may confuse the user a bit given the swap. If you want to keep the existing setting, you may need to include another modified run_afl_fuzz in aflpp fuzzer.py

**Comment by vanhauser-thc** [comment]
Can you modify the source so -d does nothing?

**Comment by kdsjZh** [comment]
let me do a local test first and I'll let you know

**Comment by kdsjZh** [comment]
I locally tested a fuzzbench build; now it should work with the current configuration.

**Comment by domenukk** [comment]
Now CI breaks with
```bash
[+] NOTE: AFL++ >= v3 has changed defaults and behaviours - see README.md
[-] PROGRAM ABORT : Using -D determinstic fuzzing is incompatible with AFL_CUSTOM_MUTATOR_ONLY!
         Location : main(), src/afl-fuzz.c:1429
```

kdsjZh:
This is an intended behavior. IMO running the custom mutation test with `-D` option can solve. Marc might help with that.
>also note that the CI will always fail because of the determinstic change. I will take care of that and doc update after merging

domenukk:
If custom mutators now need an extra flag it means we'll break people's workflows.
Should we always disable deterministic when a custom mutator is provided?

kdsjZh:
Thanks for head up, that make sense! I've disabled the det flag when there is the custom mutator env.

vanhauser-thc:
As I said I will handle it when I merge it 
@domenukk @kdsjZh

PR #1975, opened 2024-01-31
Improve afl-cc

SonicStark:
## Full response file support

Various SEGVs can be seen if tested with some malformed rsp files. Corpus at [this repo](https://github.com/SonicStark/afl-cc-rtfsc/tree/main/rsp-file-parse/corpus) can be used to reproduce. So I rewrite this part to for a radical cure.

## More comments in source code

Add docs for each function.

## Remove dead code about allow/denylist

The following code piece
```c
((!strncmp(cur_argv, "-fsanitize=fuzzer-",

           strlen("-fsanitize=fuzzer-")) ||
  !strncmp(cur_argv, "-fsanitize-coverage",
           strlen("-fsanitize-coverage"))) &&
 (strncmp(cur_argv, "sanitize-coverage-allow",
          strlen("sanitize-coverage-allow")) &&
  strncmp(cur_argv, "sanitize-coverage-deny",
          strlen("sanitize-coverage-deny")) &&
  aflcc->instrument_mode != INSTRUMENT_LLVMNATIVE))
```
can be removed, because:
 - `fsanitize_fuzzer_comma` called before just strip fuzzer* from -fsanitize=...
 - we need to drop `-fsanitize-coverage-allowlist=...` or so if not in `LLVMNATIVE` mode, this can be done within another if-statement:
   ```c
   if (!strncmp(cur_argv, "-fsanitize-coverage-", 20) &&
         strstr(cur_argv, "list=")) {...}
   ```
 - `strncmp` on `cur_argv` with `"sanitize-coverage-allow"` seems always return non-zero.

### DIFF
--- a/src/afl-cc.c
+++ b/src/afl-cc.c
@@ -2896,134 +2896,248 @@ static void process_params(aflcc_state_t *aflcc, u8 scan, u32 argc,
     if (PARAM_MISS != parse_linking_params(aflcc, cur, scan, &skip_next, argv))
       continue;
 
+    /* Response file support -----BEGIN-----
+      We have two choices - move everything to the command line or
+      rewrite the response files to temporary files and delete them
+      afterwards. We choose the first for easiness.
+      For clang, llvm::cl::ExpandResponseFiles does this, however it
+      only has C++ interface. And for gcc there is expandargv in libiberty,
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thank you!

PR #1982, opened 2024-02-02
Fix type in AFL_NOOPT env variable in afl-cc help message

kyakdan:
A minor fix to show the correct name of the `AFL_NOOPT` environment variable


vanhauser-thc:
thank you :)

PR #1990, opened 2024-02-07
 Add 7950x3d and 6900hs benchmarks, with and without mitigations

monik3r:
Note that `system` entries are running without mitigations disabled vs `both` where grub options are changed. Mitigations make a pretty large difference on Zen 3+ and Zen 4!


vanhauser-thc:
thanks!

PR #1992, opened 2024-02-08
redqueen: time limit for colorization

ndrewh:
For nondeterministic inputs, or inputs where a large portion of the input bytes break the path (e.g., bytes protected by a checksum), colorization will split the range once for every byte in the input

This is bad: for a 10k input that runs in 100ms, colorization will take upwards of 15 minutes. For larger/slower inputs, it can easily take hours.

This places an upper-bound on the runtime of colorization.

I have no idea what a reasonable timeout for colorization is (I completely made up "3 minutes"). It might be nice to benchmark this (e.g. fuzzbench).

### DIFF
--- a/src/afl-fuzz-redqueen.c
+++ b/src/afl-fuzz-redqueen.c
@@ -107,11 +107,12 @@ static struct range *add_range(struct range *ranges, u32 start, u32 end) {
 
 }
 
-static struct range *pop_biggest_range(struct range **ranges) {
+static struct range *pop_biggest_range(struct range **ranges, int *num_ranges) {
 
   struct range *r = *ranges;
   struct range *rmax = NULL;
   u32           max_size = 0;
+  u32           count = 0;
 
   while (r) {
 
@@ -129,9 +130,11 @@ static struct range *pop_biggest_range(struct range **ranges) {
     }
 
     r = r->next;
+    ++count;
 
   }
 
+  *num_ranges = count;
   return rmax;
 
 }
@@ -332,7 +335,8 @@ static u8 colorization(afl_state_t *afl, u8 *buf, u32 len,
 
   }
 
-  while ((rng = pop_biggest_range(&ranges)) != NULL &&
+  u32 range_count = 0;
+  while ((rng = pop_biggest_range(&ranges, &range_count)) != NULL &&
          afl->stage_cur < afl->stage_max) {
 
     u32 s = 1 + rng->end - rng->start;
@@ -362,6 +366,23 @@ static u8 colorization(afl_state_t *afl, u8 *buf, u32 len,
 
         ranges = add_range(ranges, rng->start, rng->start - 1 + s / 2);
         ranges = add_range(ranges, rng->start + s / 2, rng->end);
+
+        /* Check the number of *remaining* execs for colorization; since
+          each represents a previously split range, and may be split
+          again, we break when emptying the queue would consume *half*
+          the timeout */
+        if (range_count * afl->queue_cur->exec_us >
+            CMPLOG_COLORIZATION_TIME_MAX / 2 * 1000000) {
+
+          if (afl->afl_env.afl_no_ui) {
+
+            WARNF("Colorization taking too long, skipping.");
+
+          }
+
+          goto checksum_fail;
+
+        }
 
       }
 
@@ -1840,7 +1861,7 @@ static void try_to_add_to_dictN(afl_state_t *afl, u128 v, u8 size) {
   for (k = 0; k < size; ++k) {
 
   #else
-  u32 off = 16 - size;
+  u32    off = 16 - size;
   for (k = 16 - size; k < 16; ++k) {
 
   #endif
--- a/include/config.h
+++ b/include/config.h
@@ -85,6 +85,9 @@
 /* Maximum allowed fails per CMP value. Default: 96 */
 #define CMPLOG_FAIL_MAX 96
 
+/* Maximum time (sec) for colorization of a single input */
+#define CMPLOG_COLORIZATION_TIME_MAX 180
+
 /* -------------------------------------*/
 /* Now non-cmplog configuration options */
 /* -------------------------------------*/


src/afl-fuzz-redqueen.c
```diff
@@ -121,9 +122,11 @@ static struct range *pop_biggest_range(struct range **ranges) {
     }
 
     r = r->next;
+    count += 1;
```
vanhauser-thc: `++count` is faster don't ask me why the optimizer does not make it so

src/afl-fuzz-redqueen.c
```diff
@@ -355,6 +359,13 @@ static u8 colorization(afl_state_t *afl, u8 *buf, u32 len,
         ranges = add_range(ranges, rng->start, rng->start - 1 + s / 2);
         ranges = add_range(ranges, rng->start + s / 2, rng->end);
 
+        if (range_count * afl->queue_cur->exec_us >
+            90 * 1000000) {  // 90 seconds left ~= 3 min total
+          WARNF("Colorization taking too long, skipping.");
```
vanhauser-thc: This needs an `if (afl->afl_env.afl_no_ui)` otherwise it destroys the UI.

src/afl-fuzz-redqueen.c
```diff
@@ -355,6 +359,13 @@ static u8 colorization(afl_state_t *afl, u8 *buf, u32 len,
         ranges = add_range(ranges, rng->start, rng->start - 1 + s / 2);
         ranges = add_range(ranges, rng->start + s / 2, rng->end);
 
+        if (range_count * afl->queue_cur->exec_us >
+            90 * 1000000) {  // 90 seconds left ~= 3 min total
```
vanhauser-thc: the 90 needs to be a `#define` in `config.h`
e.g.
`#define CMPLOG_COLORIZATION_TIME_MAX 90`

vanhauser-thc:
sounds useful, left some comments.
I will run this through fuzzbench first to see if this really is in improvement.

src/afl-fuzz-redqueen.c
vanhauser-thc: (the time of the define should reflect the real time, so maybe  `x / 2` would be required for the value then in the if here)

vanhauser-thc:
https://www.fuzzbench.com/reports/experimental/2024-02-08-aflpp/index.html
current state vs. this PR with 3 minutes vs this PR with 15 minutes
results will be ready on Sunday.

src/afl-fuzz-redqueen.c
ndrewh: wtf

vanhauser-thc:
@ndrewh this was the latest result before it was merged into a complete fuzzer benchmark (which is buggy and invalidates some targets):
IMG_6512
what we can see is that the difference is mininmal, although at the beginning of the campaign there was a huge gap where colorless 3 minutes was the best, after 12 hours colorless 15 minutes caught up, and after 20 hours normal afl++ caught up.

So what we need IMHO is a sliding windows, that increases the time coloring is allowed to take the longer the campaign runs - if there are timeouts.

This will likely be complicated to get right, so I would propose make the changes that I asked for, then I will merge it and implement a sliding component on top of it afterwards?

ndrewh:
@vanhauser-thc I've made the changes.

To be honest, I'm actually pleasantly surprised that it made a measurable difference in the scores/ranks at all, even if its very small. I implemented this to address what I thought was an edge case that only one or two targets would hit (I observed it in some nondeterministic SQLite inputs, and I've seen similar behavior in some libxml inputs a while ago -- but most targets don't seem to have inputs that cause mutations to stall for more than a minute or two).

My expectation for this change was actually only that it might eliminate some low outliers on one or two targets. I thought this _might_ improve the average score on some targets, but not really affect the median (which is used in almost all of the summary stats, if I understand correctly). There appears to be _some_ evidence for removing low outliers (e.g. lcms, stb), but it's generally just very noisy. I expect this to make more of a difference for particularly slow targets, or targets running under QEMU. (For some reason I cannot yet explain, I find that certain hybrid fuzzers are more likely to generate inputs that hit this case as well.)

Out of curiosity -- can you check the logs to see how many of these targets actually hit the new timeout? I would have thought not too many of them.

ndrewh:
re: increasing the timeout as the campaign goes on

My intuition is that if you hit the timeout, colorization is not going to give you useful information anyway. If you hit the timeout, it's most likely because almost every byte of the input "breaks" the path (due to non-determinism or a large portion of the input being used in something like a checksum). A better option may be: in the case of a timeout, proceed with an empty set of ranges. I think even if we cannot produce a colorized input, redqueen may still be able to make useful substitutions (though that may just necessitate a timeout _somewhere else_, I have not tested).

My gut says that non-deterministic inputs are definitely a waste of time to colorize, unless we mask off the non-deterministic coverage bits, or something.

vanhauser-thc:
thanks!

vanhauser-thc:
@ndrewh do you think this is a good idea? Or how could it improved?
https://github.com/AFLplusplus/AFLplusplus/commit/6316994857740b31a1d3cb3fad8f3c1d5f84baeb

ndrewh:
imo this is increasing the timeout too quickly (e.g. just 5 nondeterministic inputs that time-out during colorization will cause the timeout to increase, so long as the fail rate is low enough). I am generally very bearish on colorization being helpful on non-deterministic inputs anyway, so I'd want to err on the side of not increasing the timeout.

I think it only makes sense to increase the timeout when we've made it through a whole cycle and are stuck. Maybe at the same time that we increase the cmplog max filesize:

https://github.com/AFLplusplus/AFLplusplus/blob/775861ea94d00672c9e868db329073afd699b994/src/afl-fuzz.c#L2644-L2651

vanhauser-thc:
There is no real cycle anymore and even in vanilla afl it can happen that after 3 days it is still in the first cycle. 
if you want you can propose what you think could work better by sending a pr to that colorless branch

Issue #1981, opened 2024-02-02
INSTALL.md instructions for Mac missing accounting for different homebrew paths (I think)


seanm:
The INSTALL.md instructions for Mac say:

```
# Depending on your MacOS system + brew version it is either
export PATH="/opt/homebrew/opt/llvm/bin:$PATH"
# or
export PATH="/usr/local/opt/llvm/bin:$PATH"
# you can check with "brew info llvm"

export PATH="/usr/local/opt/coreutils/libexec/gnubin:/usr/local/bin:$PATH"
```

Regarding the last line: I don't have a `/usr/local/opt/coreutils/libexec/gnubin:` directory.

I think this should have an "or" like the lines above, ie:

```
export PATH="/opt/homebrew/opt/coreutils/libexec/gnubin:/usr/local/bin:$PATH"
# or
export PATH="/usr/local/opt/coreutils/libexec/gnubin:/usr/local/bin:$PATH"
```

vanhauser-thc:
ok done.
in the future please send a pull request with the recommended changes.
None of us is a mac user hence keeping the information for mac (or the implementation) correct with the many changes that apple does with every release is difficult.

seanm:
>in the future please send a pull request with the recommended changes.

I would have, but I created an Issue first because I was not sure if I was correct or not. Even with the change, the instructions do not work for me (see https://github.com/AFLplusplus/AFLplusplus/issues/1865)

>None of us is a mac user hence keeping the information for mac (or the implementation) correct with the many changes that apple does with every release is difficult.

Yeah, Apple's fast and aggressive changes can be annoying, but in fairness, this seems to be because of homebrew, not Apple.

seanm:
I believe I've found your change here: https://github.com/AFLplusplus/AFLplusplus/commit/602eceed8b56eef62d673a54b6011541bf1ab60a#diff-93adeedb62ae4160af1d8713c9f15796d7b6312d45a0148447538fe185cd7347

It doesn't seem correct to me; at least, it is not what I was suggesting.  (But as I still don't have afl working properly on macOS, I can't be sure...)

I was thinking something like this: https://github.com/AFLplusplus/AFLplusplus/pull/1995

PR #1995, opened 2024-02-09
Fixed issue #1981: document PATH correctly based on homebrew version

seanm:
- removed reference to M1 as M2 and M3 exist now too.
- Also use current name and spelling of "macOS".

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -96,9 +96,9 @@ These build options exist:
 
 e.g.: `make LLVM_CONFIG=llvm-config-14`
 
-## MacOS X on x86 and arm64 (M1)
+## macOS on x86_64 and arm64
 
-MacOS has some gotchas due to the idiosyncrasies of the platform.
+macOS has some gotchas due to the idiosyncrasies of the platform.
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
LGTM, thanks!

seanm:
@vanhauser-thc relatedly... In INSTALL.md, it says about macOS: "To build AFL, install llvm (and perhaps gcc) from brew and follow the general instructions for Linux"

The general instructions for Linux include:

```
...
make distrib
sudo make install
```

but the macOS section has:

```
...
gmake
cd frida_mode
gmake
cd ..
sudo gmake install
```

Is this difference required?  Thanks.

vanhauser-thc:
The difference makes things easier but is not absoltely required, nyx, qemu and unicorn cannot build on macOS.

domenukk:
Unicorn should work fine on macOS?

seanm:
> The difference makes things easier but is not absoltely required, nyx, qemu and unicorn cannot build on macOS.

I think it would make the macOS instructions simpler if this difference wasn't there... I'll make a PR for discussion...

PR #1998, opened 2024-02-14
afl-persistent-config: Use GRUB_CMDLINE_LINUX instead of GRUB_CMDLINE_LINUX_DEFAULT

maxammann:
The latter is often overwritten in images used in cloud setups. For example DigitalOcean sets `GRUB_CMDLINE_LINUX_DEFAULT="console=tty1 console=ttyS0"` in `/etc/default/grub.d/`

It is kind of unclear which one to choose for the usecase. From my research it seems like GRUB_CMDLINE_LINUX is used for user config and `GRUB_CMDLINE_LINUX_DEFAULT` for optional improving flags.

Related: https://bugs.launchpad.net/vmbuilder/+bug/1692471

### DIFF
--- a/afl-system-config
+++ b/afl-system-config
@@ -54,7 +54,7 @@ if [ "$PLATFORM" = "Linux" ] ; then
   echo
   dmesg | grep -E -q 'noibrs pcid nopti' || {
     echo It is recommended to boot the kernel with lots of security off - if you are running a machine that is in a secured network - so set this:
-    echo '  /etc/default/grub:GRUB_CMDLINE_LINUX_DEFAULT="ibpb=off ibrs=off kpti=0 l1tf=off mds=off mitigations=off no_stf_barrier noibpb noibrs pcid nopti nospec_store_bypass_disable nospectre_v1 nospectre_v2 pcid=on pti=off spec_store_bypass_disable=off spectre_v2=off stf_barrier=off srbds=off noexec=off noexec32=off tsx=on tsx_async_abort=off arm64.nopauth audit=0 hardened_usercopy=off ssbd=force-off"'
+    echo '  /etc/default/grub:GRUB_CMDLINE_LINUX="ibpb=off ibrs=off kpti=0 l1tf=off mds=off mitigations=off no_stf_barrier noibpb noibrs pcid nopti nospec_store_bypass_disable nospectre_v1 nospectre_v2 pcid=on pti=off spec_store_bypass_disable=off spectre_v2=off stf_barrier=off srbds=off noexec=off noexec32=off tsx=on tsx_async_abort=off arm64.nopauth audit=0 hardened_usercopy=off ssbd=force-off"'
     echo
   }
   echo If you run fuzzing instances in docker, run them with \"--security-opt seccomp=unconfined\" for more speed.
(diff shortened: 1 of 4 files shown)


vanhauser-thc:
I tested on my Debian and the difference between GRUB_CMDLINE_LINUX_DEFAULT and GRUB_CMDLINE_LINUX is that GRUB_CMDLINE_LINUX is active for every entry including recovery, and GRUB_CMDLINE_LINUX_DEFAULT for normal + advanced but without recovery.

Now with your cloud provider they might overwrite GRUB_CMDLINE_LINUX_DEFAULT, but maybe there are others that overwrite GRUB_CMDLINE_LINUX?
Maybe the best it to add to both. that will duplicate the entries but would ensure it will likely work everywhere

maxammann:
If you add it to both then the flags will be added twice. I don't know if that is problematic. Adding to both seems fine. Applying it to recovery seems fine given that you should operate in a throw-away environment.

vanhauser-thc:
do you want to change the PR?

maxammann:
Done, tested on a fresh cloud VM

afl-system-config
```diff
@@ -54,7 +54,7 @@ if [ "$PLATFORM" = "Linux" ] ; then
   echo
   dmesg | grep -E -q 'noibrs pcid nopti' || {
     echo It is recommended to boot the kernel with lots of security off - if you are running a machine that is in a secured network - so set this:
-    echo '  /etc/default/grub:GRUB_CMDLINE_LINUX_DEFAULT="ibpb=off ibrs=off kpti=0 l1tf=off mds=off mitigations=off no_stf_barrier noibpb noibrs pcid nopti nospec_store_bypass_disable nospectre_v1 nospectre_v2 pcid=on pti=off spec_store_bypass_disable=off spectre_v2=off stf_barrier=off srbds=off noexec=off noexec32=off tsx=on tsx_async_abort=off arm64.nopauth audit=0 hardened_usercopy=off ssbd=force-off"'
+    echo '  /etc/default/grub:GRUB_CMDLINE_LINUX="ibpb=off ibrs=off kpti=0 l1tf=off mds=off mitigations=off no_stf_barrier noibpb noibrs pcid nopti nospec_store_bypass_disable nospectre_v1 nospectre_v2 pcid=on pti=off spec_store_bypass_disable=off spectre_v2=off stf_barrier=off srbds=off noexec=off noexec32=off tsx=on tsx_async_abort=off arm64.nopauth audit=0 hardened_usercopy=off ssbd=force-off"'
     echo
```
vanhauser-thc: Please reverse this.  I think GRUB_CMDLINE_LINUX_DEFAULT is the better option, it is just for (some?) cloud linux setups it does not work. or do you think different?
same for best_practices.md and fuzzing_in_depth.md

maxammann: I see `GRUB_CMDLINE_LINUX_DEFAULT` as the place for good defaults that do not change the behavior of the system (e.g. quiet splash; changing the console type during boot; visual changes during boot). `GRUB_CMDLINE_LINUX` is there user adjustments go. The afl-persistent-config applies user options to improve fuzzing.

But this is just my interpretation or vague documentation. In reality the only difference is whether you want to have the settings in recovery mode or not. I would claim that recovery mode is not relevant in a fuzzing setup.

vanhauser-thc: >  In reality the only difference is whether you want to have the settings in recovery mode or not. I would claim that recovery mode is not relevant in a fuzzing setup.

I agree and that is why I proposed the change. GRUB_CMDLINE_LINUX ends up in recovery mode config, GRUB_CMDLINE_LINUX_DEFAULT does not.

maxammann: I'll revert that :) sounds good

vanhauser-thc:
thanks!

PR #1999, opened 2024-02-15
Fixed #1865: create symlinks for afl-clang-fast

### LINKED ISSUE #1865: On a fresh macOS 13 VM, install instructions produce an `afl-clang` but no `afl-clang-fast`
**Describe the bug**
On a fresh macOS 13 VM, install instructions produce an `afl-clang` but no `afl-clang-fast`

**To Reproduce**
I followed the INSTALL.md instructions carefully, and even redid everything from scratch on a fresh VM, but still I only get an `afl-clang` but no `afl-clang-fast`. I kept the entire command line interaction.

**Expected behavior**
The docs say `afl-clang-fast` is the one we should use, so I'd expect the instructions to result in it being built.

**Screen output/Screenshots**
I have the the whole interaction starting from installing brew, but it's too big to paste here apparently...

seanm:


vanhauser-thc:
yes this would create the afl-clang-fast symlink - always. even if it cannot work because no llvm is installed, and that is why it is not done here (when compiling):
https://github.com/AFLplusplus/AFLplusplus/blob/61ceef64b10cc8da0bba79c3f97ae223d2095fc5/GNUmakefile.llvm#L408
and here (when installing):
https://github.com/AFLplusplus/AFLplusplus/blob/61ceef64b10cc8da0bba79c3f97ae223d2095fc5/GNUmakefile.llvm#L519

when afl-clang-fast is not created then it means that your compilation of the llvm components of afl++ did not succeed as they should. that is what I have been trying to tell you :)

We must find out *why* your setup is not able to build afl++ correctly.

seanm:
I may have made a discovery...

It seems like these settings:

```bash
sysctl kern.sysv.shmmax=8388608
sysctl kern.sysv.shmseg=48
sysctl kern.sysv.shmall=98304
```

Must be done ***before*** building AFL. They are not just important for doing the actual work of fuzzing, but are seemingly also required when *building* AFL, because it seems to run some kind of test during the build that expects these to be set already.

INSTALL.md does not mention this requirement at all. :(

@vanhauser-thc do you think this could explain it?  I will try in a fresh VM after dinner... and will update this PR if so.

domenukk:
I think we should make the build not require that much share mem.... Can you try to track down where this happens? Probably some testcase or something..

vanhauser-thc:
I didn’t know but then again on my test Mac I used afl persistent config so I would not have noticed.
It can’t hurt to put it in INSTALL.md

seanm:
> I think we should make the build not require that much share mem....

I agree; it would probably also complicate building afl++ by something like homebrew.

>Can you try to track down where this happens? Probably some testcase or something..

See https://github.com/AFLplusplus/AFLplusplus/pull/1999/commits/7aceab1c14b4a117bbd679c9f8af2a21e24db2e8

> I didn’t know but then again on my test Mac I used afl persistent config so I would not have noticed.

Exactly. Recall, the first step of [my issue](https://github.com/AFLplusplus/AFLplusplus/issues/1865) always said "On a fresh macOS 13 VM".

I've updated commits here with changes to INSTALL.md.

docs/INSTALL.md
```diff
@@ -101,6 +101,36 @@ e.g.: `make LLVM_CONFIG=llvm-config-14`
...
+        shmmax: 4194304 (max shared memory segment size)
+        shmmin:       1 (min shared memory segment size)
+        shmmni:      32 (max number of shared memory identifiers)
+        shmseg:       8 (max shared memory segments per process)
+        shmall:    1024 (max amount of shared memory in pages)
+```
+
+To temporarily change your settings to something usable with AFL++,
+run these commands:
+
```
vanhauser-thc: Wouldn’t it be better to recommend to execute afl-system-config instead?

src/afl-sharedmem.c
```diff
@@ -282,7 +282,9 @@ u8 *afl_shm_init(sharedmem_t *shm, size_t map_size,
   if (shm->shm_id < 0) {
 
     PFATAL("shmget() failed, try running afl-system-config");
-
+#ifdef __APPLE__
```
vanhauser-thc: the pfatal message above tells already what to do it doesn’t need this message

vanhauser-thc:
I like most of the changes please see my two comments

docs/INSTALL.md
seanm: Yeah, looks like!

A shame that INSTALL.md does not mention it at all, I didn't know it existed.

Will update PR...

vanhauser-thc: Nothing is perfect what we should strive for perfection :)

seanm: PR now updated.

vanhauser-thc:
thank you!

seanm:
> thank you!

No, thank you! I'm happy to now actually be able to do some fuzzing of my code!  I'll probably be back with more questions soon. :)

Issue #2001, opened 2024-02-16
On macOS, `ld`'s `rpath` paramater does not want an `=`


seanm:
I continue to struggle to get a working afl++ on macOS...

Perhaps it's my hack for https://github.com/AFLplusplus/AFLplusplus/issues/1865 or https://github.com/AFLplusplus/AFLplusplus/issues/2000 but I find myself in a situation where trying to build a trivial Objective-C++ executable fails to link:

```
% afl-clang-fast++ test.m -framework Cocoa
afl-cc++4.20a by Michal Zalewski, Laszlo Szekeres, Marc Heuse - mode: LLVM-PCGUARD
SanitizerCoveragePCGUARD++4.20a
[+] Instrumented 1 locations with no collisions (non-hardened mode) of which are 0 handled and 0 unhandled selects.
ld: unknown options: -rpath=/opt/homebrew/Cellar/llvm/17.0.6_1/lib
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

Looking at the `ld` man page on macOS, it seems that it should be specified as `-rpath /opt/homebrew/Cellar/llvm/17.0.6_1/lib` (without the equals sign).

This seems to be a difference between Mac & Linux, I've found a few other general discussions of it, ex:

https://forum.dlang.org/thread/svgizdajgmwaqzfuvfyk@forum.dlang.org

https://github.com/conan-io/conan/issues/7715

In afl++, this seems to have been added here: https://github.com/AFLplusplus/AFLplusplus/pull/1912/commits/572a08f0884d5caa656f6cc342f694779d6c9199

I will make a PR with a possible fix.

vanhauser-thc:
On Linux, it works with space, comma and equal sign.
if this continues to make a problem we could switch to a comma.
thanks.

PR #2002, opened 2024-02-16
issue #2001: fix passing rpath to linker on macOS

seanm:
Seems on macOS, `ld` does not want an `=` when specifying `-rpath`.


seanm:
@vanhauser-thc  Not sure this is the best way to fix this, but it finally results in an `afl-clang-fast++` that can actually compile and link!

vanhauser-thc:
Yes, I think this is necessary. thanks for the PR!

PR #2003, opened 2024-02-18
replaced unicornafl with unicorn

Resery:
The submodule of unicorn cannot be imported through unicornafl.*_const here. If we want to use the *_const module, we should reference `from unicorn.*_const import *` like this instead of importing the entire contents of the *_const module via unicornafl。


vanhauser-thc:
@domenukk  ?

domenukk:
I don't quite get this, we do `from unicorn import *` inside unicornafl, so _in theory_ shouldn't the modules be there?
cc @wtdcode

Resery:
> I don't quite get this, we do `from unicorn import *` inside unicornafl, so _in theory_ shouldn't the modules be there? cc @wtdcode

Simple test:
```
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from unicornafl import *
>>> from unicornafl.arm_const import *
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'unicornafl.arm_const'
>>>
```

The version of unicornafl:
```
$ python3 -m pip list | grep unicornafl
unicornafl                    2.0.2
```

Sorry, upload img is too troublesome, so I use text.

wtdcode:
The fix is correct but not __ideal__. Suggested fix:

```python
try:
    from unicornafl.x86_const import * # compatibility with unicornafl v1
except ImportError:
    from unicorn.x86_const import * # v2
```

The root cause is that unicornafl2 removes the duplicate constant definitions totally because Unicorn2 is fast evolving with fixing missing constants.

domenukk:
The try except doesn't sound clean either? Is there a way to make this future-proof then?

wtdcode:
To make it future-proof, this fix should be correct because unicornafl __shall__ no longer duplicate defs itself.

domenukk:
Are you sure we won't end up with different unicornafl and unicorn libraries and it'll mess everything up?

wtdcode:
Uncornafl (using uncorn2) will no longer ship these definitions as the author here suggests because it doesn’t make sense. Besides, Unicorn so far keeps ABI compatibility including enum values. In other words, even with unicornafl1+unicorn1, this fix should also work.
________________________________

PR #2005, opened 2024-02-19
unicornafl: Fix dump_regs() type errors in pwndbg dumper

hyrathon:
TypeError will occur as gdb api return some strange type of values that json can't serialize, this would fix this issue

The regs may be  `<gdb.Value object at xxx>` or None, this would break the serialize process


vanhauser-thc:
note we only accept PRs to dev, fixed that for you.
looks good to me, WDYT @domenukk ?

unicorn_mode/helper_scripts/unicorn_dumper_pwndbg.py
```diff
@@ -111,12 +111,14 @@ def dump_regs():
     reg_state = {}
     for reg in pwndbg.gdblib.regs.all:
         reg_val = pwndbg.gdblib.regs[reg]
+        if not reg_val:
```
domenukk: This will mean reg_val of `0` will never be dumped, I don't think this is correct? This should likely be `if reg_val is None`? Or what's the meaning of this?

hyrathon: Yes, you are right

domenukk: Do you have time to fix it?

hyrathon: > Do you have time to fix it?

I have fixed it.

domenukk:
Thanks!

PR #2008, opened 2024-02-20
opt the free

Resery:
1. There isn't a need to check all chunks when address == 0
2. If the address is not in chunks, the program may want to free an object that doesn't exist. There may be a "double-free" or "invalid-free" vulnerability. (This patch is from the repo named "Battelle/afl-unicorn")


vanhauser-thc:
LGTM but in all cases of unicorn @domenukk is most knowledgeable :)

domenukk:
Thank you!

PR #2009, opened 2024-02-20
impl uaf chcker

Resery:
By adding a list of freed chunks, add the chunk to the list during free, check whether the allocated block is in the freed chunk list during malloc, and if so, remove the chunk from the freed chunk list, in __check_mem_access check whether the address is in the freed chunk list. This enables the detection of uaf.

### DIFF
--- a/unicorn_mode/helper_scripts/unicorn_loader.py
+++ b/unicorn_mode/helper_scripts/unicorn_loader.py
@@ -90,7 +90,7 @@ class UnicornSimpleHeap(object):
     _chunks_freed = [] # List of all freed chunks
     _debug_print = False  # True to print debug information
 
-    def __init__(self, uc, debug_print=False):
+    def __init__(self, uc, debug_print=False， uaf_check=False):
         self._uc = uc
         self._debug_print = debug_print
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
I see the value of it, and the overhead I think is limited.
@domenukk ?

unicorn_mode/helper_scripts/unicorn_loader.py
```diff
@@ -179,6 +187,14 @@ def __check_mem_access(self, uc, access, address, size, value, user_data):
                     # Force a memory-based crash
                     uc.force_crash(UcError(UC_ERR_READ_PROT))
 
+        for chunk in self._chunks_freed:
```
domenukk: If we have to walk all historical allocations it may be worth it to use something like an intervaltree (like https://en.wikipedia.org/wiki/Interval_tree )

domenukk:
Target-dependent, the overhead may not be too small, for example if the target does a lot of allocations, maybe we should make this configurable?

Resery:
Can we change the allocation strategy of the heap? We can map the heap space during initialization. There is no need to map/unmap again when malloc or free. We can set a freed flag bit at the guard page of the chunk. Use this flag bit to check the use-after-free. I think this method is better.

domenukk:
Up to you, I'll be happy to accept contributions, but I would personally just use libafl_qemu that already fixes all of these problems - or _maybe_ write this stuff in rust on top of unicornafl ... writing a "good" allocator in python seems a bit pointless?

Resery:
Thanks for your reply. I think you are right. I'll close this pr soon.

Resery:
Thanks for your reply. I think you are right. Please close this pr.

domenukk:
I would suggest we hide the uaf check behind a flag to `__init__` (`uaf_check=True`) so people can use it, it could still come in handy for someone

Resery:
I have updated the patch. Thank you.

domenukk:
Thanks!

PR #2010, opened 2024-02-20
bugfix: free a chunk with a size of 0, it will cause 1 byte oob.

Resery:
Malloc does not check the size. Generally, malloc(0) should return 0 but there will return two pages. Free will use is_buffer_in_chunk to check whether the address is in the chunk. At that time, the chunk.data_addr == total_size . Free pass address and "1" to is_buffer_in_chunk. So cause 1 byte out-of-bound.

### DIFF
--- a/unicorn_mode/helper_scripts/unicorn_loader.py
+++ b/unicorn_mode/helper_scripts/unicorn_loader.py
@@ -101,6 +101,10 @@ class UnicornSimpleHeap(object):
         #    - Allocate at least 1 4k page of memory to make Unicorn happy
         #    - Add guard pages at the start and end of the region
         total_chunk_size = UNICORN_PAGE_SIZE + ALIGN_PAGE_UP(size) + UNICORN_PAGE_SIZE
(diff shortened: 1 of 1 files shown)


unicorn_mode/helper_scripts/unicorn_loader.py
```diff
@@ -101,6 +101,10 @@ def malloc(self, size):
         #    - Allocate at least 1 4k page of memory to make Unicorn happy
         #    - Add guard pages at the start and end of the region
         total_chunk_size = UNICORN_PAGE_SIZE + ALIGN_PAGE_UP(size) + UNICORN_PAGE_SIZE
+
+        if size == 0:
+            retrun 0
```
vanhauser-thc: Typo. Please test your changes

Resery: Done

vanhauser-thc:
Lgtm @domenukk  ?

PR #2012, opened 2024-02-22
Fix bug where `-t 1000+` may result in enormous timeouts

clesmian:
Using the afl-fuzz option `-t` with `+` results in timeouts that are 1000 times larger than intended, because of a missing conversion between microseconds and milliseconds

### DIFF
--- a/src/afl-fuzz.c
+++ b/src/afl-fuzz.c
@@ -2496,8 +2496,7 @@ int main(int argc, char **argv_orig, char **envp) {
           if ((afl->queue_buf[entry]->exec_us/1000) > max_ms)
             max_ms = afl->queue_buf[entry]->exec_us/1000;
       
-      if (max_ms > afl->fsrv.exec_tmout)
-        afl->fsrv.exec_tmout = max_ms;
+      afl->fsrv.exec_tmout = max_ms;
       afl->timeout_given = 1;
 
     }
--- a/src/afl-fuzz.c
+++ b/src/afl-fuzz.c
@@ -2493,8 +2493,8 @@ int main(int argc, char **argv_orig, char **envp) {
 
       for (entry = 0; entry < afl->queued_items; ++entry)
         if (!afl->queue_buf[entry]->disabled)
-          if (afl->queue_buf[entry]->exec_us > max_ms)
-            max_ms = afl->queue_buf[entry]->exec_us;
+          if ((afl->queue_buf[entry]->exec_us/1000) > max_ms)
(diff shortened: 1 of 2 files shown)


clesmian:
I closed this for now, as this fix is incomplete. I will update and reopen this PR

clesmian:
I implemented one version to work around the issue described in #2013 that just doesn't use any timeout during calibration if `-t ...+` was specified. This comes at the expense of potentially indefinite calibration runs.

I quickly tested that it works on Linux on amd64, but I don't know whether it works this way for all architecture/OS combinations.

vanhauser-thc:
thanks for your effort!

> I implemented one version to work around the issue described in #2013 that just doesn't use any timeout during calibration if `-t ...+` was specified. This comes at the expense of potentially indefinite calibration runs.

this is not the right approach. the right approach is the one documented:

>   * The '+' feature of the `-t` option means to auto-calculate the timeout with the value given being the maximum timeout

clesmian:
I've actually read this sentence multiple times, rereading it now, I realize that I misunderstood it before. I removed the relevant commit from the PR

src/afl-fuzz.c
```diff
@@ -2493,10 +2493,11 @@ int main(int argc, char **argv_orig, char **envp) {
...
       for (entry = 0; entry < afl->queued_items; ++entry)
         if (!afl->queue_buf[entry]->disabled)
-          if (afl->queue_buf[entry]->exec_us > max_ms)
-            max_ms = afl->queue_buf[entry]->exec_us;
-
-      afl->fsrv.exec_tmout = max_ms;
+          if ((afl->queue_buf[entry]->exec_us/1000) > max_ms)
+            max_ms = afl->queue_buf[entry]->exec_us/1000;
+      
+      if (max_ms > afl->fsrv.exec_tmout)
```
vanhauser-thc: the if needs to be removed. we always want the timeout to be set based on max_ms

vanhauser-thc: If you want you can add
```
if (max_ms * 1.2 < afl->fsrv.exec_tmout)
  max_ms *= 1.2;
else
  max_ms = afl->fsrv.exec_tmout;
```
I think it better to add 20% to the timeout as a safety margin

clesmian: I agree. In addition I added a lower boundary of at least 5 milliseconds to avoid setting the timeout to 0 in case of very fast initial seeds.

vanhauser-thc:
thank you!

Issue #2007, opened 2024-02-19
Preserve filename extension of files in the input directory upon mutation and testing


seanm:
Sometimes code examines a filename extension and behaves differently based on it. For example, I have seen code like:

```
if (filename extension == .jpg)
   call jpeg reader
else if (filename extension == .png)
   call png reader
else
   do nothing
```

In order to accommodate such code, if would be nice if AFL++ preserved the file extension of the initial file(s) given in the input directory.

It seems that the actual file passed to the executable being testing is always `/path/to/output-directory/default/.cur_input`, i.e. the extension has been removed. In the example above, this means it always falls into the final `else` and nothing interesting is tested.

domenukk:
It'll be far easier to do two or more runs, one for each filename extension (and use `-e` to set that)

seanm:
> and use `-e` to set that

Ah, I missed that flag, thanks for pointing it out!

Couldn't / shouldn't the default behaviour be for AFL++ to detect the extension automatically though?  i.e. why make the human pass `-e`?

vanhauser-thc:
it is a corner case that a target looks for a filename extension.
and the normal way is setting up a harness for effective fuzzing, usually doubles the speed, better coverage, more bugs found. (like a LLVMFuzzerTestOneInput function or modifying the main.c to set up a specific environment and using AFL_INIT/AFL_FUZZ_INIT etc.)

seanm:
> it is a corner case that a target looks for a filename extension.

It's actually very common, at least for the libraries I'm using.

>and the normal way is setting up a harness for effective fuzzing, usually doubles the speed, better coverage, more bugs found.

That's maybe the normal way once you're a fuzzing pro. :) But when you're just getting started, it would be a nicer experience IMHO to have the extension auto-detected.

Anyway, I've added `-e` and confirm my executable is now passed `/path/to/output-directory/default/.cur_input.jpg` so that's nice.

But the resulting files in the `crashes` folder still do not have any extension.  Is that expected?

vanhauser-thc:
Yes that is expected :)

seanm:
> Yes that is expected :)

Thanks, good to know.

Would it not be preferable to give them the extension specified with `-e`?

vanhauser-thc:
The clear answer is `maybe`. Yes it would make this corner case simpler to reproduce. on the other hand there are tools that work on the queue / crashes / timeout folders, and that might break them maybe.

[...]

PR #2014, opened 2024-02-22
Issue #2007: add filename extension to /crashes files

seanm:
This is very helpful for code that inpects a file name extension when determining what code to run.

It's also useful for applications that constrain the user to choose files by extension.

### DIFF
--- a/src/afl-forkserver.c
+++ b/src/afl-forkserver.c
@@ -2003,7 +2003,7 @@ afl_fsrv_run_target(afl_forkserver_t *fsrv, u32 timeout,
     if (unlikely(fsrv->persistent_record)) {
 
       retval = FSRV_RUN_TMOUT;
-      persistent_out_fmt = "%s/hangs/RECORD:%06u,cnt:%06u";
+      persistent_out_fmt = "%s/hangs/RECORD:%06u,cnt:%06u%s%s";
       goto store_persistent_record;
 
     }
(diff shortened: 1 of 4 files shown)


vanhauser-thc:
you must do that to all file saves, not only crashes, so queue and hangs too. (not persistent though)

seanm:
> you must do that to all file saves, not only crashes, so queue and hangs too. (not persistent though)

If it's something you actually will merge, I can do that...

vanhauser-thc:
Without the queue change I can’t test it fully

vanhauser-thc:
I tested casr, afl-cmin and afl-cov and they all work.
so I am merging this.
note though that some tools do not support file extensions, e.g. afl-cmin and afl-showmap.

seanm:
Thanks!

fuzzmr:
I am getting this compiler error when I try to enable AFL_PERSISTENT_RECORD in config.h

```
src/afl-forkserver.c: In function ‘afl_fsrv_run_target’:
src/afl-forkserver.c:2083:16: error: ‘afl’ undeclared (first use in this function)
 2083 |                afl->file_extension ? "." : "",
      |                ^~~
```

I'm not sure how afl-forkserver.c previously had or obtains a reference to the 'afl' state variable.

vanhauser-thc:
thanks, fixed in 951a0e52254d873dd0f1a3a80d9acda44563edd5 in dev

seanm:
Sorry for the breakage, and thanks for the fix!

PR #2018, opened 2024-02-27
Add -ldl flag

hyrathon:
The linker flags lacks a -ldl so the dlopen series of func symbols can't be found


vanhauser-thc:
thank you

PR #2019, opened 2024-02-28
Load autodictionary when using new forkserver

zidel:
Fixes a bug where the new fork server would decrement dict_size until zero then try to use it as the upper bound for the number of bytes to pass to add_extra_func, causing it to never store any of the tokens.


vanhauser-thc:
thank you!

PR #2026, opened 2024-03-06
Fix delayed pcmap writing for code coverage with pc-table

choller:
When using pc-table and a pcmap (e.g. from nyx preload code) to measure code coverage, modules can be loaded (and __sanitizer_cov_pcs_init runs) before the __afl_pcmap_ptr is actually available. For this purpose, the mechanism was previously delaying this mapping step until the pcmap was available. This logic broke with the refactoring for dynamic coverage filtering and this patch restores the old behavior. With the patch, the pc filtering parts of the code will only run on the last loaded module while we still iterate over all available modules every time the callback is made. We mark modules as mapped only if we had the pcmap already when iterating.

### DIFF
--- a/instrumentation/afl-compiler-rt.o.c
+++ b/instrumentation/afl-compiler-rt.o.c
@@ -1617,7 +1617,7 @@ void __sanitizer_cov_pcs_init(const uintptr_t *pcs_beg,
 
       }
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thank you :)

PR #2027, opened 2024-03-06
Add optional handling of Nyx InvalidWriteToPayload event

choller:
This adds `AFL_NYX_HANDLE_INVALID_WRITE` to allow AFL++ to ignore the InvalidWriteToPayload event. If such events occur sporadically, it is most likely due to OOM in the guest where the kernel is reclaiming memory and therefore touching the payload buffer. I decided to count these as crashes, hoping that they remain visible with this approach (but with a zero-size log). However, this happens quite rarely in my local setup so it is difficult to test.

Without this variable set, AFL++ will just die and leave qemu running in the background. Long term, we might want to change the default behavior here as well.


vanhauser-thc:
> Without this variable set, AFL++ will just die and leave qemu running in the background. Long term, we might want to change the default behavior here as well.

but it is wrong to keep it like this. this should kill the target (qemu). can you please add this?

src/afl-forkserver.c
```diff
@@ -1840,6 +1840,8 @@ afl_fsrv_run_target(afl_forkserver_t *fsrv, u32 timeout,
       case Timeout:
         return FSRV_RUN_TMOUT;
       case InvalidWriteToPayload:
+        if (!!getenv("AFL_NYX_HANDLE_INVALID_WRITE")) { return FSRV_RUN_CRASH; }
+
         /* ??? */
         FATAL("FixMe: Nyx InvalidWriteToPayload handler is missing");
```
vanhauser-thc, line 1846: If you switch ` FATAL`  to `NYX_PRE_FATAL` it should be enough.

choller: Thanks, I'm testing this right now locally.

choller:
Fwiw, I just managed to reproduce with the new code (and the env var set). The log will contain something like:

`Payload buffer write attempt at RIP: ffffffff8dc26847`

So that's even nicer.

src/afl-forkserver.c
vanhauser-thc: great, ping when you have a result :)

vanhauser-thc:
@choller @schumilo  any idea how to solve this gracefully?

vanhauser-thc:
@choller @schumilo please continue to look for a fix ... merging now.

PR #2029, opened 2024-03-11
support parsing of llvm rc minor version

ocean1:
As described in title, support parsing "rc" llvm minor versions that the makefile currently fails to parse.

Thanks,
D.


vanhauser-thc:
is this enough or does it need `'s/rc.*//'` ?

ocean1:
You are right! There are "rcX" versions too, it requires `s/rc.*//` or `s/rc[0-9]*//` (pushed a change with the first one).
Thanks!

vanhauser-thc:
Thanks!

PR #2030, opened 2024-03-11
Replay record loop fix

ocean1:
Needed to fix record compat loop to replay correct number of inputs (and always at least one input) since the cycle cnt will get immediately decremented before first iteration.

Note also that the recent addition of the file extension in saving input in fsrv_run_target in afl-forkserver.c:
```c
               afl->file_extension ? "." : "",
               afl->file_extension ? (const char *)afl->file_extension : "");
```
Breaks compilation, since afl state is not passed to fsrv, passing afl state to fsrv_run_target is a change that will touch several components and will require creating a state variable in the tools, so I did not proceed with fixing this for now.
(But can proceed with the required changes if you want :)).

D.

### DIFF
--- a/include/afl-record-compat.h
+++ b/include/afl-record-compat.h
@@ -24,7 +24,7 @@ unsigned char fuzz_buf[FUZZ_BUF_SIZE];
 
 int __afl_persistent_loop(unsigned int max_cnt) {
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thank you

PR #2033, opened 2024-03-12
Add -l option for adjustable block deletion performance in tmin

### LINKED ISSUE #2032: tmin: Proposing del_len option for improved speed
**Issue**
Block deletion in the tmin stage is a time-consuming task that scales with the size of the file.

**Background**
https://github.com/AFLplusplus/AFLplusplus/blob/775861ea94d00672c9e868db329073afd699b994/src/afl-tmin.c#L427-L434

As `del_len`, the unit for reducing unnecessary blocks, decreases by powers of two, the operation takes longer. However, there are instances
where speed is more critical than achieving minimal minimization.

https://github.com/AFLplusplus/AFLplusplus/blob/775861ea94d00672c9e868db329073afd699b994/src/afl-tmin.c#L483-L486

Currently, the minimum value of `del_len` is 1. However, increasing the minimum value would result in faster speeds.

**proposal**
Therefore, I propose allowing del_len to be specified as an option.

Alternatively, a method could be implemented to terminate the process if there is no reduction of even 1 byte, or if the reduction is less than n bytes, in a given cycle.

kiddo-pwn:
- Introduce the -l option to set min block deletion length using powers of 2 (e.g., 1, 2, 4, 8, 16, ...).
- This enables a trade-off between minimization thoroughness and speed.
- Adjusting del_len_limit allows for faster processing, as doubling it roughly halves the minimization time.
closes #2032

### DIFF
--- a/src/afl-tmin.c
+++ b/src/afl-tmin.c
@@ -82,7 +82,7 @@ static u8 crash_mode,                  /* Crash-centric mode?               */
     remove_shm = 1,                    /* remove shmem on exit?             */
     debug;                             /* debug mode                        */
 
-static u32 del_len_limit;              /* Minimum block deletion length     */
+static u32 del_len_limit = 1;          /* Minimum block deletion length     */
 
 static volatile u8 stop_soon;          /* Ctrl-C pressed?                   */
 
@@ -423,7 +423,6 @@ next_pass:
 
   del_len = next_pow2(in_len / TRIM_START_STEPS);
   stage_o_len = in_len;
-  if (!del_len_limit) { del_len_limit = 1; }
 
   ACTF(cBRI "Stage #1: " cRST "Removing blocks of data...");
 
@@ -1070,7 +1069,7 @@ int main(int argc, char **argv_orig, char **envp) {
 
         del_len_limit = atoi(optarg);
 
-        if (del_len_limit < 1 || del_len_limit >= TMIN_MAX_FILE) {
+        if (del_len_limit < 1 || del_len_limit > TMIN_MAX_FILE) {
 
           FATAL("Value of -l out of range between 1 and TMIN_MAX_FILE");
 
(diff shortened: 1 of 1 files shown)


domenukk:
Thank you! Please direct all PRs towards dev

kiddo-pwn:
Oops, realized that PRs need to be made to the dev branch. Noticed a issue and force-pushed a fix.

src/afl-tmin.c
```diff
@@ -1055,6 +1060,24 @@ int main(int argc, char **argv_orig, char **envp) {
...
+        if (del_limit_given) { FATAL("Multiple -l options not supported"); }
+        del_limit_given = 1;
+
+        if (!optarg) { FATAL("Wrong usage of -l"); }
+
+        if (optarg[0] == '-') { FATAL("Dangerously low value of -l"); }
+
+        del_len_limit = atoi(optarg);
+
+        if (del_len_limit < 1 || del_len_limit >= TMIN_MAX_FILE) {
```
vanhauser-thc: so 0 would be a valid value?

src/afl-tmin.c
```diff
@@ -82,6 +82,8 @@ static u8 crash_mode,                  /* Crash-centric mode?               */
     remove_shm = 1,                    /* remove shmem on exit?             */
     debug;                             /* debug mode                        */
 
+static u32 del_len_limit;              /* Minimum block deletion length     */
```
vanhauser-thc: should be initialized with 1 as otherwise it is 0?

vanhauser-thc:
left 2 comments, please check

src/afl-tmin.c
kiddo-pwn: The existing `del_len` decreases in this manner: (the largest power of 2 less than the file size), 16, 8, 4, 2, 1, ... . So a value of 0 seems to be invalid.

src/afl-tmin.c
kiddo-pwn: Aha. I think it's safer to initialize it 1.

kiddo-pwn:
fixed these issues. please check.

vanhauser-thc:
Lgtm thanks!

PR #2034, opened 2024-03-13
Adds stats tracking for time spend actually mutating & running test i…

fbeqv:
This branch introduces tracking time spend on mutating & running inputs. It introduces the new field `actual_fuzz_time  : 11` in fuzzer_stats that tracks the total number of seconds spend on fuzzing (excluding things such as syncing, initializing, calibrating etc). 

This currently does not affect statsd nor the UI. Only the fuzzer_stats file is updated

### DIFF
--- a/src/afl-fuzz-run.c
+++ b/src/afl-fuzz-run.c
@@ -409,6 +409,7 @@ u8 calibrate_case(afl_state_t *afl, struct queue_entry *q, u8 *use_mem,
   u32 use_tmout = afl->fsrv.exec_tmout;
   u8 *old_sn = afl->stage_name;
 
+  u64 calibration_start_us = get_cur_time_us();
   if (unlikely(afl->shm.cmplog_mode)) { q->exec_cksum = 0; }
 
   /* Be a bit more generous about timeouts when resuming sessions, or when
@@ -503,6 +504,9 @@ u8 calibrate_case(afl_state_t *afl, struct queue_entry *q, u8 *use_mem,
     (void)write_to_testcase(afl, (void **)&use_mem, q->len, 1);
 
     fault = fuzz_run_target(afl, &afl->fsrv, use_tmout);
+
+    // update the time spend in calibration after each execution, as those may be slow
+    update_calibration_time(afl, &calibration_start_us);
 
     /* afl->stop_soon is set by the handler for Ctrl+C. When it's pressed,
        we want to bail out quickly. */
@@ -650,6 +654,7 @@ abort_calibration:
 
   if (!first_run) { show_stats(afl); }
 
+  update_calibration_time(afl, &calibration_start_us);
   return fault;
 
 }
@@ -669,10 +674,14 @@ void sync_fuzzers(afl_state_t *afl) {
   afl->stage_max = afl->stage_cur = 0;
   afl->cur_depth = 0;
 
+  u64 sync_start_us = get_cur_time_us();
   /* Look at the entries created for every other fuzzer in the sync directory.
    */
 
   while ((sd_ent = readdir(sd))) {
+
+    // since sync can take substantial amounts of time, update time spend every iteration
+    update_sync_time(afl, &sync_start_us);
 
     u8  qd_synced_path[PATH_MAX], qd_path[PATH_MAX];
     u32 min_accept = 0, next_min_accept = 0;
@@ -861,6 +870,9 @@ void sync_fuzzers(afl_state_t *afl) {
 
   if (afl->foreign_sync_cnt) read_foreign_testcases(afl, 0);
 
+  //add time in sync one last time
+  update_sync_time(afl, &sync_start_us);
+
   afl->last_sync_time = get_cur_time();
   afl->last_sync_cycle = afl->queue_cycle;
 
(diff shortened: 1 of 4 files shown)


src/afl-fuzz.c
```diff
@@ -2852,7 +2852,14 @@ int main(int argc, char **argv_orig, char **envp) {
 
       }
 
+      if(unlikely(afl->start_fuzzing_time == 0)){
```
vanhauser-thc: but basically this is  afl->start_time, just a tad later (and with less than a second of time difference), so I do not think this helps? just use start_time instead?

vanhauser-thc:
what about when restarting fuzzing (` -i -`) the stats file is then loaded, and maybe the fuzz time (and calibration time) should be expanded from the previous run times? wdyt?

while you are working on this PR, have the same as calibration time for trimming and cmplog would be nice. but otherwise I can do that next week.
because run_time - calibration_time - trimming_time - cmplog_time = fuzzing_time

src/afl-fuzz.c
fbeqv: start_time seems to set before both the initial sync() and cull_queue(), on slow harnesses (where syncing importing inputs takes a significant amount of time) the difference would be quite substantial, and that's exactly the condition I'm trying to detect with this change - the instances where the difference is substantial. I do see that this is a kinda awkward distinction though. Let me thing some more about that.

fbeqv: So really, our problem could also be addressed by making a "currently_cycle" (that has value -1 or something while we are loading & syncing, 0 as soon as we enter the actual fuzzing code and then from there on is equal to cycles_done+1. 

making cycles done 0 for different states `afl->queue_cycle ? (afl->queue_cycle - 1) : 0` is kinda making it useless to detect successful startup.

fbeqv:
left some comments

src/afl-fuzz.c
vanhauser-thc: that would work too, but then you do not know for how long the the fuzzer is running. was it running for 1 second because calibration took 59:59 ?

src/afl-fuzz-run.c
```diff
@@ -897,11 +910,10 @@ u8 trim_case(afl_state_t *afl, struct queue_entry *q, u8 *in_buf) {
 
     }
 
-    if (custom_trimmed) return trimmed_case;
+    if (custom_trimmed) { fault = trimmed_case; goto abort_trimming_wo_count; }
```
fbeqv: I'm not sure this was intentional, but the old code path didn't log the trimed size (same for inputs too small to trim). I opted to keep the original behavior, even though I suspect in this case (but not the other), that's not actually intended behavior.

src/afl-fuzz-run.c
```diff
@@ -504,6 +505,9 @@ u8 calibrate_case(afl_state_t *afl, struct queue_entry *q, u8 *use_mem,
 
     fault = fuzz_run_target(afl, &afl->fsrv, use_tmout);
 
+    // update the time spend in calibration after each execution, as those may be slow
+    update_calibration_time(afl, &calibration_start_us);
```
vanhauser-thc, line 509: why would you update every iteration? Also, you update with the calibration_start_us difference every single time, this does not work out well :)

src/afl-fuzz-run.c
```diff
@@ -669,11 +674,15 @@ void sync_fuzzers(afl_state_t *afl) {
...
   afl->cur_depth = 0;
 
+  u64 sync_start_us = get_cur_time_us();
   /* Look at the entries created for every other fuzzer in the sync directory.
    */
 
   while ((sd_ent = readdir(sd))) {
 
+    // since sync can take substantial amounts of time, update time spend every iteration
+    update_sync_time(afl, &sync_start_us);
```
vanhauser-thc, line 684: exactly the same issue here

src/afl-fuzz-run.c
```diff
@@ -946,6 +958,8 @@ u8 trim_case(afl_state_t *afl, struct queue_entry *q, u8 *in_buf) {
 
       fault = fuzz_run_target(afl, &afl->fsrv, afl->fsrv.exec_tmout);
 
+      update_trim_time(afl, &trim_start_us);
```
vanhauser-thc, line 961: same same

src/afl-fuzz-stats.c
```diff
@@ -2438,4 +2444,20 @@ void show_init_stats(afl_state_t *afl) {
 #undef IB
 
 }
+void update_calibration_time(afl_state_t *afl, u64* time){
+  u64 cur =  get_cur_time_us();
+  afl->calibration_time_us = cur-*time;
+  *time = cur;
```
vanhauser-thc, line 2444: ah you overwrite the time parameter, that is why it works.
still - what is the need? 
I would add a fuzz_time to the fuzzer_stats file that calculates the time from run_time  - sync_time - calibration_time - trim_time.

src/afl-fuzz-run.c
vanhauser-thc: what do you mean by "didn't log the trimed size"? the return of trim_case is only to check for forkserver errors.

PR #2038, opened 2024-03-22
bugfix: override directive and recipe echoing in GNUmakefile

SonicStark:
#### Missing `override` directive

Found this when I can't compile AFL++ static with `NO_SPLICING=1`. Because `CFLAGS_OPT` has already been marked at the very beginning:
```makefile
ifdef NO_SPLICING
  override CFLAGS_OPT += -DNO_SPLICING
endif
```

Similar bug on `LDFLAGS`, too.

#### Recipe echoing not disabled

Found this because *Build Summary:* of `source-only` target was printed in a wrong format.


vanhauser-thc:
LGTM, thanks

PR #2043, opened 2024-03-31
src: fix calculation of fuzzing time in statistics

### LINKED ISSUE #1241: Make afl-fuzz not include the time computer was suspended in its statistics
When the computer is suspended during a fuzzing session, the time spent in suspended state is counted as a "run time" on a statistics screen. In case fuzzing was running for several hours and then the computer was suspended for several more hours, on one hand I would prefer not restarting the session, on the other hand, now it may be harder to tell how well a harness performs (corpus count vs. run time). _(I expect some problems are also possible if wall-clock jumps backwards for some reason but I have never experienced this myself)_

As an example, if I...
* start afl-fuzz, keep it working for 1 hour
* then suspend the computer for 1 hour
* then resume my computer

... I expect the status screen to show run time of ~1h instead of ~2h.

As a rather simple fix, it may be worth using `clock_gettime(CLOCK_MONOTONIC, ...)`  instead of `gettimeofday`, if available. Replacing `gettimeofday` in `get_cur_time()` and `get_cur_time_ns()` with `clock_gettime` seems to fix this issue for me, though I'm not very familiar with AFL++ codebase (which functions should return consistent times? are any functions expected to execute as efficiently as possible due to vDSO? etc.).

Another possible use case could be temporarily suspending the container with fuzzers, but I expect that case to require much more complicated heuristics instead of simply asking the kernel for the right clock.

ligurio:
When the computer is suspended during a fuzzing session, the time spent in suspended state is counted as a "run time" on a statistics screen.

The time returned by `gettimeofday(2)` is affected by discontinuous jumps in the system time. It is better using `clock_gettime(2)`.

The patch replace `gettimeofday` with `clock_gettime` [1]. `clock_gettime` uses a CLOCK_MONOTONIC_COARSE clock type, it is faster than CLOCK_MONOTONIC, but still has resolution (~1ms) that is adequate for our purposes. However, CLOCK_MONOTONIC_COARSE is a Linux-specific clock variant, so on macOS it is replaced with CLOCK_MONOTONIC, and with CLOCK_MONOTONIC_FAST on FreeBSD [2].

Closes #1241

1. https://pubs.opengroup.org/onlinepubs/9699919799/functions/clock_gettime.html
2. https://man.freebsd.org/cgi/man.cgi?query=clock_gettime

### DIFF
--- a/src/afl-common.c
+++ b/src/afl-common.c
@@ -34,6 +34,7 @@
 #endif
 #include <string.h>
 #include <strings.h>
+#include <time.h>
 #include <math.h>
 #include <sys/mman.h>
 
@@ -57,6 +58,26 @@ u8  last_intr = 0;
 #ifndef AFL_PATH
   #define AFL_PATH "/usr/local/lib/afl/"
 #endif
+
+/* - Some BSD (i.e.: FreeBSD) offer the FAST clock source as
+ *   equivalent to Linux COARSE clock source. Aliasing COARSE to
+ *   FAST on such systems when COARSE is not already defined.
+ * - macOS has no support of CLOCK_MONOTONIC_COARSE clock type.
+ */
+#if defined (OS_DARWIN) || defined (OS_SUNOS)
+#    define CLOCK_MONOTONIC_COARSE CLOCK_MONOTONIC
+#elif defined (OS_FREEBSD)
+#    define CLOCK_MONOTONIC_COARSE CLOCK_MONOTONIC_FAST
+#endif
+
+/* Convert seconds to milliseconds. */
+#define SEC_TO_MS(sec) ((sec)*1000)
+/* Convert seconds to microseconds. */
+#define SEC_TO_US(sec) ((sec)*1000000)
+/* Convert nanoseconds to milliseconds. */
+#define NS_TO_MS(ns)   ((ns)/1000000)
+/* Convert nanoseconds to microseconds. */
+#define NS_TO_US(ns)   ((ns)/1000)
 
 void *afl_memmem(const void *haystack, size_t haystacklen, const void *needle,
                  size_t needlelen) {
@@ -973,27 +994,27 @@ void read_bitmap(u8 *fname, u8 *map, size_t len) {
 /* Get unix time in milliseconds */
 
 inline u64 get_cur_time(void) {
+  struct timespec ts;
+  int rc = clock_gettime(CLOCK_MONOTONIC_COARSE, &ts);
+  if (rc == -1) {
+    PFATAL("Failed to obtain timestamp (errno = %i: %s)\n",
+           errno, strerror(errno));
+  }
 
-  struct timeval  tv;
-  struct timezone tz;
-
-  gettimeofday(&tv, &tz);
-
-  return (tv.tv_sec * 1000ULL) + (tv.tv_usec / 1000);
-
+  return SEC_TO_MS((uint64_t)ts.tv_sec) + NS_TO_MS((uint64_t)ts.tv_nsec);
 }
 
 /* Get unix time in microseconds */
 
 u64 get_cur_time_us(void) {
+  struct timespec ts;
+  int rc = clock_gettime(CLOCK_MONOTONIC_COARSE, &ts);
+  if (rc == -1) {
+    PFATAL("Failed to obtain timestamp (errno = %i: %s)\n",
+           errno, strerror(errno));
+  }
 
-  struct timeval  tv;
-  struct timezone tz;
-
-  gettimeofday(&tv, &tz);
-
-  return (tv.tv_sec * 1000000ULL) + tv.tv_usec;
-
+  return SEC_TO_US((uint64_t)ts.tv_sec) + NS_TO_US((uint64_t)ts.tv_nsec);
 }
 
 /* Describe integer. The buf should be


vanhauser-thc:
Lgtm thanks

vanhauser-thc:
@ligurio this PR had a huge negative performance impact on AFL++ as I found out last week (and spend several days to pinpoint the exact commit that caused it).

It makes no sense to me why, because when I just use gettimeofday vs. clock_gettime, clock_gettime is 3x faster.
But the benchmark results are clear.

this benchmark is one that shows current AFL++ vs current AFL++ with your PR reverted and only run against those targets where the impact was the highest: https://www.fuzzbench.com/reports/experimental/2024-06-01-aflpp2/index.html
This is what I changed back: https://github.com/AFLplusplus/AFLplusplus/commit/894339c5d7dc615a98423390d30f6b4e032f01f3

Here is one that among others compares 4.20a (from begin of february) to 4.20c (released in may) in a full benchmark:
https://www.fuzzbench.com/reports/experimental/2024-05-30-aflpp/index.html
99.72 vs 90.87 is unbelievable.

I do not understand how this can have a negative impact. 
Do you have any idea?
also @devnexen @domenukk @andreafioraldi @eqv @schumilo @choller @smoelius @DonggeLiu maybe one of you know why this could be.
I checked that they do the same thing - resolution is the same, and they counting seconds etc. correctly ...
I do not get it.

smoelius:
> It makes no sense to me why, because when I just use gettimeofday vs. clock_gettime, clock_gettime is 3x faster.

I'm grasping at straws, but does FuzzBench use virtualized hardware? If so, could the difference have something to do with how these functions are virtualized?

vanhauser-thc:
> > It makes no sense to me why, because when I just use gettimeofday vs. clock_gettime, clock_gettime is 3x faster.
> 
> I'm grasping at straws, but does FuzzBench use virtualized hardware? If so, could the difference have something to do with how these functions are virtualized?

good idea! 
I did a fuzzbench run where on startup I did a count how many get_cur_time and get_cur_time_ns can be done in 15 seconds each, either with clock_gettime (v0) or gettimeofday(v1).
Results are the same as on my native Linux/CPU, clock_gettime is 3x faster on average:
```
get_cur_time v1 count 505972190
get_cur_time v1 count 529923041
get_cur_time_us v1 count 540799005
get_cur_time v1 count 547068640
get_cur_time v1 count 547500024
get_cur_time v1 count 548125259
get_cur_time v1 count 549550145
get_cur_time v1 count 554262027
get_cur_time v1 count 554644837
get_cur_time v1 count 555059292
get_cur_time v1 count 555570907
get_cur_time v1 count 560973303
get_cur_time_us v1 count 562645093
get_cur_time v1 count 562860360
get_cur_time v1 count 564302191
get_cur_time v1 count 569075104
get_cur_time_us v1 count 576976031
get_cur_time_us v1 count 578432726
get_cur_time_us v1 count 578861398
get_cur_time_us v1 count 579721783
get_cur_time_us v1 count 582934839
get_cur_time_us v1 count 583557005
get_cur_time v1 count 583880916
get_cur_time v1 count 584131571
get_cur_time_us v1 count 584382210
get_cur_time v1 count 584705127
get_cur_time_us v1 count 586683469
get_cur_time_us v1 count 589742947
get_cur_time_us v1 count 590910545
get_cur_time_us v1 count 595409599
get_cur_time_us v1 count 601374519
get_cur_time_us v1 count 617744232
get_cur_time_us v1 count 618147671
get_cur_time_us v1 count 618612149
get_cur_time v1 count 620322457
get_cur_time v1 count 623837606
get_cur_time v1 count 624259432
get_cur_time_us v1 count 656294866
get_cur_time_us v1 count 656414806
get_cur_time_us v1 count 658535959
get_cur_time_us v0 count 1517385097
get_cur_time_us v0 count 1531091499
get_cur_time_us v0 count 1532435486
get_cur_time_us v0 count 1570412671
get_cur_time_us v0 count 1575208030
get_cur_time_us v0 count 1592489014
get_cur_time_us v0 count 1599365841
get_cur_time_us v0 count 1619386613
get_cur_time_us v0 count 1620788491
get_cur_time_us v0 count 1625555740
get_cur_time_us v0 count 1630691584
get_cur_time_us v0 count 1634641374
get_cur_time v0 count 1635177469
get_cur_time v0 count 1636488996
get_cur_time_us v0 count 1645649543
get_cur_time v0 count 1652632218
get_cur_time_us v0 count 1654689113
get_cur_time_us v0 count 1657152165
get_cur_time v0 count 1700992508
get_cur_time v0 count 1701119263
get_cur_time v0 count 1718539601
get_cur_time_us v0 count 1719413509
get_cur_time v0 count 1738415692
get_cur_time v0 count 1747903358
get_cur_time v0 count 1762571236
get_cur_time v0 count 1765622461
get_cur_time v0 count 1769040056
get_cur_time v0 count 1777236766
get_cur_time v0 count 1781785780
get_cur_time v0 count 1787238440
get_cur_time v0 count 1789711682
get_cur_time_us v0 count 1795201607
get_cur_time_us v0 count 1802829053
get_cur_time_us v0 count 1827903838
get_cur_time_us v0 count 1836200621
get_cur_time v0 count 1861684215
get_cur_time v0 count 1939951729
get_cur_time v0 count 1966073676
get_cur_time v0 count 1977135356
get_cur_time v0 count 1991362494
```

smoelius:
> good idea!

Sorry it didn't amount to anything. :(

vanhauser-thc:
Not your fault :) sadly it wasn’t the reason

eqv:
Very low probability of being the root cause, but I'm grasping for straws here. Really dumb question: could it be that the time value is somehow more deterministic at startup in fuzzbench and used to seed the prng?

vanhauser-thc:
Time is the time (I guess they use utc). We don’t use time for randomness, /dev/urandom feeds our prng every 100000 usages.
Time is mostly used for stats. I think - not sure - that we also use it to abandon cmplog if it takes too long. But the time is accurate in either method

DonggeLiu:
I am not sure either: I don't recall anything changed in FuzzBench that could cause this.
By looking at each benchmark's statistics under `Sample statistics and statistical significance (code coverage)` in https://www.fuzzbench.com/reports/experimental/2024-06-01-aflpp2/index.html, I feel `afl++` tends to have a higher std than `afl++_reg`, maybe that is the cause?

Also, I noticed that benchmark [`libpcap_fuzz_both-summary`](https://www.fuzzbench.com/reports/experimental/2024-06-01-aflpp2/index.html#libpcap_fuzz_both-summary) tends to have a high crash rate (i.e., only 3 or 5 out of 20 instances finished the experiment), it's result might not be statistically meaningful.

vanhauser-thc:
> I am not sure either: I don't recall anything changed in FuzzBench that could cause this. By looking at each benchmark's statistics under `Sample statistics and statistical significance (code coverage)` in https://www.fuzzbench.com/reports/experimental/2024-06-01-aflpp2/index.html, I feel `afl++` tends to have a higher std than `afl++_reg`, maybe that is the cause?

what is a `higher std`

thanks for giving me the idea to look at the fuzzer_stats file for both.

the one with gettimeofday() has more coverage also in the stats file, **but has more execs per second**.
clock_gettime() is slower but has more coverage. WTF

> Also, I noticed that benchmark [`libpcap_fuzz_both-summary`](https://www.fuzzbench.com/reports/experimental/2024-06-01-aflpp2/index.html#libpcap_fuzz_both-summary) tends to have a high crash rate (i.e., only 3 or 5 out of 20 instances finished the experiment), it's result might not be statistically meaningful.

Thanks for pointing this out to me. As the regression is visible on 80% of the benchmarks on a full benchmark assessment this is not it. But points me to a different problem - why is afl-fuzz crashing? I cannot see anything in the output logs. @DonggeLiu anything you can see from cloud logs or so what the issue is? I would guess OOM

DonggeLiu:
> what is a `higher std`

I meant a higher standard deviation, from here:
image

> Thanks for pointing this out to me. As the regression is visible on 80% of the benchmarks on a full benchmark assessment this is not it. But points me to a different problem - why is afl-fuzz crashing? I cannot see anything in the output logs. @DonggeLiu anything you can see from cloud logs or so what the issue is? I would guess OOM

Nothing useful/strange in the [cloud log](https://pantheon.corp.google.com/logs/query;query=2024-06-01-aflpp2%0Aseverity%3E%3DERROR;cursorTimestamp=2024-06-01T16:51:31.924390863Z;duration=P7D?project=fuzzbench&e=-13802955&mods=logs_tg_prod), the 3 typical errors are:
image

vanhauser-thc:
so to solve this mystery:

the reason of the regression is COARSE - the timing especially for microseconds - but even for milliseconds - is not precise enough. and that f*cks up the time measurement we use in AFL++.

using CLOCK_MONOTONIC (without _COARSE) is faster than gettimeofday(), and does not count leap days, sleeping the system, etc. so the best of both worlds.

this could not have been solved without the generous use of google's fuzzbench @DonggeLiu thanks!

PR #2052, opened 2024-04-12
4.20 release pre-PR

vanhauser-thc:


domenukk:
Just in time for legalization in Germany

docs/INSTALL.md
```diff
@@ -21,7 +21,7 @@ If you want to build AFL++ yourself, you have many options. The easiest choice
 is to build and install everything:
 
 NOTE: depending on your Debian/Ubuntu/Kali/... release, replace `-14` with
-whatever llvm version is available. We recommend llvm 13, 14, 15 or 16.
+whatever llvm version is available. We recommend llvm 13 or newer.
```
domenukk, line 24: Shouldn't we recommend the latest stable version for performance/optimization reasons?

include/t1ha0_ia32aes_b.h
```diff
@@ -0,0 +1,183 @@
+/*
```
domenukk, line 1: Stupid question, looks like this is the zLIB license, compatible with GPL, but so far we have apache everywhere else.. could that pose a problem for users?

vanhauser-thc: Very good question. Will check it out

vanhauser-thc: https://www.apache.org/legal/resolved.html states it’s compatible

PR #2061, opened 2024-04-17
Specify shared memory visibility

jschwartzentruber:
These should be visible for targets where symbols are normally hidden.


vanhauser-thc:
I am on the phone - does this also work with gcc? Gcc and clang quite different how they expect attributes. If there is no clang then gcc is used (afl-gcc-fast environments)

jschwartzentruber:
Yes, I also tested with `afl-gcc-fast` and the attribute is respected.

vanhauser-thc:
Thank you!

PR #2073, opened 2024-04-28
Fix: Removed the redundant `id` field from redqueen.c 

Acture:
Removed the redundant `id` field from redqueen.c from the debug output in the afl-fuzz-redqueen.c file since cmp_header no longer has this field.


vanhauser-thc:
Thank you

PR #2076, opened 2024-05-02
Fix `afl_whatsup` findings timer

louismerlin:
Since https://github.com/AFLplusplus/AFLplusplus/pull/2043 was merged, AFL++ has been using the computer's [monotonic time](https://www.man7.org/linux/man-pages/man3/clock_gettime.3.html) instead of UNIX time for all timing statistics.

`afl_whatsup` is still using UNIX time for its monitoring operations, and is showing wrong results (e.g. `Time without finds : 19845 days`).

This PR makes `afl_whatsup` use `/proc/uptime`, which is equivalent (on Linux) to the monotonic time used by AFL++.

### DIFF
--- a/afl-whatsup
+++ b/afl-whatsup
@@ -111,7 +111,13 @@ if [ -z "$NO_COLOR" ]; then
   RESET="$NC"
 fi
 
-CUR_TIME=`cat /proc/uptime | awk '{printf "%.0f\n", $1}'`
+PLATFORM=`uname -s`
+if [ "$PLATFORM" = "Linux" ] ; then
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
Thanks! Though what for platforms that don’t have that? This will break. Can you please guards this with a uname check?

louismerlin:
Done. I don't have access to another system than `Linux` right now though, so no idea if we can generalize this somehow.

vanhauser-thc:
Thanks! 
@devnexen  maybe you know if we need something else for Darwin or bsd?

PR #2087, opened 2024-05-13
Allow setting different file permissions on command line

fxlb:
Now, we can build with e.g.:
$ make CPPFLAGS=-DDEFAULT_PERMISSION=0644
without these warnings:
./include/config.h:54:9: warning: DEFAULT_PERMISSION macro redefined
  [-Wmacro-redefined]


vanhauser-thc:
hmm I do not like ifdef  around settings the user can do ... I prefer the warning, sorry :)

Issue #2088, opened 2024-05-14
afl++ nyx mode cannot set the size of fuzz data


nj00001:
In afl++, even though I changed the max limit of the input file with the -G option, the data transferred to nyx still uses the default max limit, which causes data truncation

https://github.com/AFLplusplus/AFLplusplus/blob/90fbf59bf16f057e50132b81be759414787ce1a3/src/afl-forkserver.c#L581

So I want to be able to read the `afl->max_length` variable during nyx initialization to properly set the maximum size of the input data
But this variable seems to be unreadable in the `afl_fsrv_start` function
So I want to keep a backup copy of the `afl->max_length` variable in the `afl->fsrv` variable for later use

Here's a screenshot of the problem
![image](https://github.com/AFLplusplus/AFLplusplus/assets/42004790/b459e811-ba85-4223-a3db-5e7b04b1ebb6)

PR #2089, opened 2024-05-14
Fix afl-fuzz -G option not configuring maximum input data size for nyx

nj00001:
Here's a quick fix for the issue https://github.com/AFLplusplus/AFLplusplus/issues/2088#issue-2295042016 , not sure if it's coding compliant, but it works correctly in my tests

before
image

after
image


vanhauser-thc:
thank you!

PR #2091, opened 2024-05-15
Fix bug of `afl-showmap` in `collect_coverage` mode

bet4it:


vanhauser-thc:
What is the bug? Why does removing code fix this without introducing new issues?

bet4it:
The result of `afl-showmap` before and after `afl-cmin` of a corpus should be same, but currently not.

`total` and `highest` were calculated in:
https://github.com/AFLplusplus/AFLplusplus/blob/0892a2245e9a7188e33c16444499c47942d1e56f/src/afl-showmap.c#L323-L329

vanhauser-thc:
thanks. in the future please explain what your PR does.
there is one minor issue, I will fix it myself.

PR #2092, opened 2024-05-16
Fix runtime underflow & -V exiting before syncing

fbeqv:
print_stats sets exit_soon even while syncing, this leaves -V 0 still broken, as we don't finish syncing.

Additionally, the change that introduced the previous -V fix also broke the runtime tracking, as runtime needs to include all time including sync, splice etc. This caused an underflow in the reported runtime. When resetting the runtime in afl-fuzz.c

### DIFF
--- a/src/afl-fuzz-stats.c
+++ b/src/afl-fuzz-stats.c
@@ -321,8 +321,9 @@ void write_stats_file(afl_state_t *afl, u32 t_bytes, double bitmap_cvg,
 #ifndef __HAIKU__
   if (getrusage(RUSAGE_CHILDREN, &rus)) { rus.ru_maxrss = 0; }
 #endif
-  u64 runtime = afl->prev_run_time + cur_time - afl->start_time;
-  if (!runtime) { runtime = 1; }
+  u64 runtime_ms = afl->prev_run_time + cur_time - afl->start_time;
+  u64 overhead_ms = (afl->calibration_time_us + afl->sync_time_us + afl->trim_time_us) / 1000;
+  if (!runtime_ms) { runtime_ms = 1; }
 
   fprintf(
       f,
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
Thanks! So this fixes both issues?

src/afl-fuzz-stats.c
```diff
@@ -375,20 +376,17 @@ void write_stats_file(afl_state_t *afl, u32 t_bytes, double bitmap_cvg,
...
       afl->longest_find_time > cur_time - afl->last_find_time
           ? afl->longest_find_time / 1000
           : ((afl->start_time == 0 || afl->last_find_time == 0)
                  ? 0
                  : (cur_time - afl->last_find_time) / 1000),
-      (runtime -
-       ((afl->calibration_time_us + afl->sync_time_us + afl->trim_time_us) /
-        1000)) /
-          1000,
+      (runtime_ms - MIN(runtime_ms, overhead_ms)) / 1000,
```
vanhauser-thc, line 386: this is the entry for fuzz_time. runtime_ms - overhead_ms makes sense, sure.
but when would runtime_ms be smaller than overhead_ms?

src/afl-fuzz-stats.c
```diff
@@ -632,7 +630,7 @@ void show_stats_normal(afl_state_t *afl) {
 
   cur_ms = get_cur_time();
 
-  if (afl->most_time_key) {
+  if (afl->most_time_key && afl->queue_cycle) {
```
vanhauser-thc, line 633: queue_cycle is 0 on startup and only increases to 1 once a cycle is done.
this cannot be right.

fbeqv: I thought it was increased to 1 right as the first cycle is started, and only when PRINTING we subtract one. to make the first cycle look like 0 too... let me double check.

src/afl-fuzz-stats.c
fbeqv: I'm not entirely sure. Initially my working hypothesis was that it was due to the runtime reset on -V (which this commit also removes). However that turned out not to be enough. I didn't dive into this too much, as it seemed to only show up with -V 0 where runtime_ms was off by like single digit values, and on any serious run the problem went away. My best guess is: runtime is tracked in ms, while the other times are tracked in us, leading to slightly different tracking results.

src/afl-fuzz-stats.c
fbeqv: https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-fuzz-stats.c#L379 i.e. queue cycle should be 0 while loading, and 1 as soon as we are in the first cycle - it's just printed with an offset of 1.

src/afl-fuzz-stats.c
vanhauser-thc: yeah the ms/us issue could be it.

src/afl-fuzz-stats.c
vanhauser-thc: du kannst ja mal mit -V10 starten und einer queue mit inhalt und einem echten target ... der sollte dann nicht nach 10 sekunden aussteigen wie er sollte.

fbeqv: ok question: What's the expected semantic for a real queue (that takes, say 30 seconds to load) and running with -V 0 and -V 10. I am expecting -V 0 to load the queue for 30 seconds and then terminate right away, and similarly -V 10 should load the queue for 30 seconds + do 10 seconds of fuzzing? 
Maybe our tests didn't catch that because the test target cycles within < 1 second. Let me double check.

vanhauser-thc: yes that is my expectation. also that syncing, calibration etc. happen, and V10 still would do 10 seconds of fuzzing after that

fbeqv: let me check if that's what's happening & if not fix it again :P

fbeqv: validated that both: 
```
make; AFL_IMPORT_FIRST=1 AFL_NO_UI=1 ./afl-fuzz -i /tmp/in -o /tmp/syncdir -M fuzzer -V0 ./test-target/libjpeg-turbo/build/djpeg
```

```
make; AFL_IMPORT_FIRST=1 AFL_NO_UI=1 ./afl-fuzz -i /tmp/in -o /tmp/syncdir -M fuzzer -V10 ./test-target/libjpeg-turbo/build/djpeg
```
do the right thing

PR #2093, opened 2024-05-17
push to stable

vanhauser-thc:


src/afl-showmap.c
```diff
@@ -1339,6 +1333,8 @@ int main(int argc, char **argv_orig, char **envp) {
 
   }
 
+  if (collect_coverage) { binary_mode = false; }  // ensure this
```
domenukk, line 1336: Should we assert and FATAL if it's set, instead?

vanhauser-thc: I don’t care, should never happen :)

PR #2102, opened 2024-05-25
fix: testcache hangs for large test cases

### LINKED ISSUE #2101: Testcache hangs indefinitely if the sum of two test cases exceed AFL_TESTCACHE_SIZE
https://github.com/AFLplusplus/AFLplusplus/blob/e7d871c8bf64962a658e447b90a1a3b43aaddc28/src/afl-fuzz-queue.c#L1340-L1350

A couple of fuzzbench benchmarks hang indefinitely in the above loop. I attached a debugger to one and plucked out the following values:

`q_testcase_cache_size` is 0x100000
`q_testcase_max_cache_size` is 0x200000
`q_testcase_cache_count` is 1.
the current `q->len` is also 0x100000
afl->q_testcase_cache is all-NULL expect for a single entry equal to afl->queue_cur

#### Likely explanation

Fuzzbench sets `AFL_TESTCACHE_SIZE` (`q_testcase_max_cache_size`) to 2(MB). If you have a single very large testcase (e.g. 1MB) that is successfully mutated, you will end up with multiple test-cases that are >1MB in the queue. This causes the test cache to be viewed as "full" and it enters the eviction loop above.

For example, from queue for harfbuzz_hb-shape-fuzzer:
```
-rw-r--r-- 2 root root 1048576 May 25 15:59 id:000031,time:0,execs:0,orig:4fa677791ff5c57b4b3c4c824a21d5fa63042d4c
...
-rw------- 1 root root 1048576 May 25 16:10 id:002387,src:000031,time:401241,execs:302991,op:havoc,rep:2
-rw------- 1 root root 1048576 May 25 16:10 id:002388,src:000031,time:401253,execs:303000,op:havoc,rep:1
-rw------- 1 root root 1048551 May 25 16:10 id:002389,src:000031,time:401328,execs:303135,op:havoc,rep:1,+cov
-rw------- 1 root root 1048576 May 25 16:10 id:002390,src:000031,time:401481,execs:303426,op:havoc,rep:2
```

However, because there is only one entry in the test cache, and it is `afl->q_testcase_cache[tid] == afl->queue_cur`, the loop runs forever.

#### Possible fixes

- Allow the testcase cache to temporarily go over the size limit if there is exactly one entry in the cache e.g. by changing
https://github.com/AFLplusplus/AFLplusplus/blob/e7d871c8bf64962a658e447b90a1a3b43aaddc28/src/afl-fuzz-queue.c#L1303-L1305
to
```
    while (unlikely(
        (afl->q_testcase_cache_size + len >= afl->q_testcase_max_cache_size && afl->q_testcase_cache_count > 1) ||
        afl->q_testcase_cache_count >= afl->q_testcase_max_cache_entries - 1)) {
```

- Enforce that testcases are smaller than `afl->q_testcase_max_cache_size / 2`

ndrewh:
This changes when the test cache will be declared "full", which triggers an eviction. This PR prevents evictions (when they would normally happen based on test cache _size_) when there is only one (large) entry in the test cache. This avoids an edge case where we try to evict the only entry in the cache, but that entry is `afl-> queue_cur`.

Fixes #2101 

I tested this 8x one fuzzbench target for several hours, and it fixes the issue and appears to have no ill side-effects.

### DIFF
--- a/src/afl-fuzz-queue.c
+++ b/src/afl-fuzz-queue.c
@@ -1301,7 +1301,8 @@ inline u8 *queue_testcase_get(afl_state_t *afl, struct queue_entry *q) {
     static u32 do_once = 0;  // because even threaded we would want this. WIP
 
     while (unlikely(
-        afl->q_testcase_cache_size + len >= afl->q_testcase_max_cache_size ||
+        (afl->q_testcase_cache_size + len >= afl->q_testcase_max_cache_size &&
+         afl->q_testcase_cache_count > 1) ||
         afl->q_testcase_cache_count >= afl->q_testcase_max_cache_entries - 1)) {
 
       /* We want a max number of entries to the cache that we learn.


vanhauser-thc:
Thanks!

PR #2103, opened 2024-05-27
Fix incorrect comment

Trolldemorted:
The comment here was probably copypasted from `place_input_callback`, and not properly adapted


domenukk:
Thanks! :)

PR #2104, opened 2024-05-27
Make lldb dumper of unicorn_mode work in modern LLDB

Evian-Zhang:
The lldb dumper script in unicorn mode cannot run in modern LLDB any more:

* It uses some python2 syntax which does not supported in LLDB
* It wrongly uses convenience variables which does not exist in command script mode (See [this stackoverflow answer](https://stackoverflow.com/a/75377915/10005095))

This is fixed by this PR. Works on my LLDB in Ubuntu 22.04. LLDB version 18.1.6.


vanhauser-thc:
Thank you!

PR #2106, opened 2024-05-30
feature: Added (generate_libtoken_dict.sh) script for simplified work with `libtokencap`

sigdevel:
Prepared a script (make_dict.sh) for simplified work with libtokencap. The description of the work libtokencap is given in the file [README.md](https://github.com/AFLplusplus/AFLplusplus/blob/stable/utils/libtokencap/README.md) , but to simplify the startup, it will probably be more convenient to add a separate script that can be used at once


vanhauser-thc:
Hi, I do not mind merging this, however
a) add a description to libtokens README.md
b) a user should not need to edit the file, just make it command line parameters

sigdevel:
Yes, great, would be happy to share the scripts used. if you are happy with this example, I will add it to the README on the next commit.

utils/libtokencap/make_dict_v2.sh
```diff
@@ -0,0 +1,49 @@
+#default values
+timeout_sec=5
+LD_PRELOAD_PATH="/home/${USER}/AFLplusplus/utils/libtokencap/libtokencap.so"
```
vanhauser-thc, line 3: this should also be a command line parameter, and not required (if could set LD:PRELOAD themselves)

utils/libtokencap/make_dict_v2.sh
```diff
@@ -0,0 +1,49 @@
...
+    touch "$AFL_TOKEN_FILE"
+    for i in $(find "$target_output" -type f -name "id*"); do
+        LD_PRELOAD="$LD_PRELOAD_PATH" \
+        timeout -s SIGKILL "$timeout_sec" \
+        "$target_bin" "$i"
+    done
+} >"$AFL_TOKEN_FILE"
+
+# sort & remove duplicates
+sort -u "$AFL_TOKEN_FILE" >"$AFL_DICT_FILE"
```
vanhauser-thc: a) you still need to delete the tempfile and b) an echo that prints the generated filename would be good, right?

utils/libtokencap/make_dict_v2.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# initialize vars
+AFL_TOKEN_FILE="${PWD}/temp_output.txt"
+AFL_DICT_FILE="$(basename "$target_output").dict"
+
+#generate token-file
+{
+    touch "$AFL_TOKEN_FILE"
+    for i in $(find "$target_output" -type f -name "id*"); do
+        LD_PRELOAD="$LD_PRELOAD_PATH" \
```
vanhauser-thc: you should check first if LD_PRELOAD is unset before doing this

utils/libtokencap/make_dict_v2.sh
vanhauser-thc: Such a default is a bad idea IHMO, just remove this line

utils/libtokencap/make_dict_v2.sh
```diff
@@ -0,0 +1,49 @@
...
+        :) echo "Option -$OPTARG requires an args" >&2; usage ;;
+    esac
+done
+
+#check options
+if [ -z "$target_output" ] || [ -z "$target_bin" ]; then
+    echo "Error: Missing mandatory opts" >&2
+    usage
+fi
+
```
vanhauser-thc: put a check here that error if both LD_PRELOAD and LD_PRELOAD_PATH are empty.

vanhauser-thc:
just a few nits and then it is fine IMHO

utils/libtokencap/make_dict_v2.sh
sigdevel: I see the additional sorting by the system utility as justified, since we have no guarantee that the user will perform tokenization on the minimized dictionary. Therefore, i think there are two approaches here:
a)modify the libtokencap tool itself - cache values beforehand and then select unique values afterwards ;
b)use a unix-way style aid band for this task

utils/libtokencap/make_dict_v2.sh
sigdevel: seems yes, its a localized point, worth making more universal.

utils/libtokencap/make_dict_v2.sh
sigdevel: okay, its been added to check flags.

utils/libtokencap/make_dict_v2.sh
vanhauser-thc: that is not what I meant but thinking about it, what you did is better

utils/libtokencap/make_dict_v2.sh
vanhauser-thc: No, it makes no sense to keep AFL_TOKEN_FILE. AFL_DICT FILE is the same just deduplicated. so keeping AFL_TOKEN_FILE is pointless.
and you need to inform the user what the filename is that you generated. like `echo Done! captured tokens written to ...` etc.

utils/libtokencap/make_dict_v2.sh
```diff
@@ -0,0 +1,45 @@
...
+#check options
+if [ -z "$target_output" ] || [ -z "$target_bin" ] || [ -z "$LD_PRELOAD_PATH" ]; then
+    echo "Error: Missing mandatory opts" >&2
+    usage
+fi
+
+# initialize vars
+AFL_TOKEN_FILE="${PWD}/temp_output.txt"
+AFL_DICT_FILE="$(basename "$target_output").dict"
```
vanhauser-thc: I am not a fan of the AFL_DICT_FILE naming.
either use a fixed name (e.g. tokens.dict), and/or set it e.g. with an `-f` parameter, or base it on target_bin.
an output directory is usually `out`, `o`, `corpus` or `corpus`, so the generated filename is not helpful IMHO.

vanhauser-thc:
I would recommend to rename the tool. `make_dict_v2.sh` makes no sense to a user, he/she has never seen v1 :)
maybe `generate_libtoken_dict.sh` is better, because if people copy the script to /usr/local/bin they will still know later what this script does.

vanhauser-thc:
sorry this took 3 loops, I finally had time to really look at the script, I am drowning in work and various tasks.

sigdevel:
> I would recommend to rename the tool. `make_dict_v2.sh` makes no sense to a user, he/she has never seen v1 :) maybe `generate_libtoken_dict.sh` is better, because if people copy the script to /usr/local/bin they will still know later what this script does.

Yes, that comment seems logical - I renamed in the dependent files :)

sigdevel:
> sorry this took 3 loops, I finally had time to really look at the script, I am drowning in work and various tasks.



> sorry this took 3 loops, I finally had time to really look at the script, I am drowning in work and various tasks.

That's okay If the tweaks helped the project a little bit - I was pleased.

vanhauser-thc:
@DonggeLiu you can cancel all running benchmarks by me, I found the regression and know what to fix.
btw if I would have the ability to cancel my benchmarks on my own that would be great, it would save resources because sometimes I do not need full runs or fuck one up and the important fuzzer does not run etc.

vanhauser-thc:
looks good, but still the temporary file is not deleted and the naming of the result file is not helpful + no final echo that prints the final filename plus that everything went fine.

another thought: sometimes a target needs command line parameters. anything after '--' (getopt handles this) should be given to the target command (`$*`)

sigdevel:
> looks good, but still the temporary file is not deleted and the naming of the result file is not helpful + no final echo that prints the final filename plus that everything went fine.
> 
> another thought: sometimes a target needs command line parameters. anything after '--' (getopt handles this) should be given to the target command (`$*`)

Hi, based on your review, I've added changes to the file.

vanhauser-thc:
LGTM thank you

PR #2108, opened 2024-06-01
push to stable

vanhauser-thc:


vanhauser-thc:
@smoelius I highly recommend to push the commit in afl.rs to this new stable state as it an large regression in afl-fuzz

smoelius:
> @smoelius I highly recommend to push the commit in afl.rs to this new stable state as it an large regression in afl-fuzz

OK. But can I ask why you don't want to publish a new release?

vanhauser-thc:
Release will come somewhere next week I guess, but I can’t say when for sure yet

PR #2113, opened 2024-06-05
Fix cmplog shared memory size when USEMMAP=1

andreafioraldi:
mmap of cmp_map was using map_size and not sizeof(cmp_map)


vanhauser-thc:
That code must be very old :) apparently nobody used it previously …

PR #2120, opened 2024-06-12
Fix undefined behavior by casting to uint64_t before left shift

visitorckw:
According to the C standard, left-shifting a value by an amount greater than or equal to the width of its promoted type results in undefined behavior. To prevent potential unexpected results, explicitly cast the uint8_t variable type to uint64_t before performing the left shift operation by 56 bits. This ensures the operation is well-defined and adheres to the standard.

Fixes: 40df85d1 ("adjust cmplog header")

### DIFF
--- a/src/hashmap.c
+++ b/src/hashmap.c
@@ -59,7 +59,7 @@ static inline unsigned int hash(uint64_t key) {
 bool hashmap_search_and_add(uint8_t type, uint64_t key) {
 
   if (unlikely(type >= 8)) return false;
-  uint64_t     val = (key & 0xf8ffffffffffffff) + (type << 56);
+  uint64_t     val = (key & 0xf8ffffffffffffff) + ((uint64_t)type << 56);
   unsigned int index = hash(val);
   HashNode    *node = _hashmap->table[index];
   while (node) {


vanhauser-thc:
But it is not an uint8 during the shift because of integer promotion right?
This function is not used it’s from a feature test that turned out to underperform.
Can’t hurt to merge though :)

visitorckw:
Since they are enclosed in parentheses, the 'type' variable remains uint8_t during the shift. I noticed that redqueen no longer uses hashmap, but since the hashmap source code hasn't been removed and this fix is small enough, I decided to submit this PR anyway. Thank you for accepting this PR.

vanhauser-thc:
That assumption is very likely wrong. Integer promotion happens internally, see https://www.geeksforgeeks.org/integer-promotions-in-c/amp/ - a lot of people are not aware about this

visitorckw:
That's interesting. I had never noticed this before. Thank you for teaching me this.
I simply used the following code and tested it with ubsan during compilation. It seems they are promoted to 32-bit int. However, since it shifts by 56 bits, from my understanding, it is still an error.

```c
#include <stdint.h>
#include <limits.h>

int main()
{
    uint8_t type = CHAR_MAX;
    uint64_t key = LLONG_MAX;
    uint64_t val = (key & 0xf8ffffffffffffff) + (type << 56);

    return 0;
}
```
```
$ gcc -fsanitize=undefined test_ub.c
$ ./a.out
test_ub.c:8:55: runtime error: shift exponent 56 is too large for 32-bit type 'int'
```

vanhauser-thc:
The promotion to int32 is well documented in the standard but came from a pre-64 bit time and hence dunno if promotion to 64 bit happens if it’s a 64 bit operation or not - so it seems not :)

jengelh:
>The promotion to int32 is well documented in the standard

The promotion is to 'int'; if the platform has only a 16-bit int, <<15 is the max you can do; promotion to 64-bit happens if (and only if) a platform has a 64-bit int (ILP64/SILP64, pretty much nonexisting today).

PR #2121, opened 2024-06-12
Remove -Wno-shift-count-overflow flag and hashmap from compilation

visitorckw:
After commit 0c9d8e59 ("Fix undefined behavior by casting to uint64_t before left shift"), the -Wno-shift-count-overflow flag is no longer necessary to suppress compilation warnings. This commit removes the flag, allowing the compiler to provide appropriate warnings for any potential future cases of shifting too many bits.

Additionally, after commit eaf4a299 ("make redqueen hashmap not default"), the hashmap is no longer in use. Therefore, this commit also removes the hashmap from the GNUmakefile.


visitorckw:
Additionally, since we are not actually using the hashmap now, I'm wondering if we should remove the hashmap from the Makefile and exclude it from compilation.

vanhauser-thc:
> Additionally, since we are not actually using the hashmap now, I'm wondering if we should remove the hashmap from the Makefile and exclude it from compilation.

yes that is a good idea

vanhauser-thc:
Thank you!

PR #2122, opened 2024-06-12
Fix memory allocation check in aflpp custom mutators

visitorckw:
The memory allocation check in afl_custom_fuzz function was incorrect. The condition was erroneously checking if ptr was non-null, whereas it should return 0 when ptr is null. Correct the condition to properly handle memory allocation failures.

Fixes: 32ffa266 ("max_len support")


vanhauser-thc:
uh damn, an oversight when I fixed something. thanks!

PR #2125, opened 2024-06-14
Fix signed integer overflow in radamsa

visitorckw:
When UBSan was enabled and 'make test' was executed, the following runtime error was observed:

libradamsa.c:26055:29: runtime error: left shift of 16777215 by 8 places cannot be represented in type 'int'

This issue was caused by a left shift operation on a signed integer. The fix involved changing the integer literal to an unsigned integer by adding a 'U' suffix, ensuring the left shift operation produces the correct value without overflow.

Fixes: 8178f4df ("remove radamsa, add radamsa custom mutator")


vanhauser-thc:
Not our code and it’s not recommended to use it but nonetheless better to have this fix, thanks :)

PR #2130, opened 2024-06-19
Auto disable memory limits for FASAN

killerra:
This change automattically disables memory limits of Address Sanitizer Mode is enabled just like if [__SANITIZE_ADDRESS__](https://github.com/AFLplusplus/AFLplusplus/blob/36db3428ab16156dd72196213d2a02a5eadaed11/src/afl-fuzz.c#L1462C1-L1468C4) is defined.


vanhauser-thc:
Thank you!

PR #2132, opened 2024-06-21
Fix typo in unicorn_mode

tylzh97:
in `unicorn_mode/samples`, python example floder's name is `python_simple`, there is a typo in README file.


vanhauser-thc:
Thanks

PR #2133, opened 2024-06-23
Optimize bit counting using __builtin_popcount

visitorckw:
Use the __builtin_popcount intrinsic to optimize the bit counting function if the compiler supports it. This change replaces the manual bit counting algorithm with the more efficient built-in function, which leverages hardware support on compatible processors.

This modification ensures that the code remains backward-compatible by falling back to the original implementation when __builtin_popcount is not available.

### DIFF
--- a/include/afl-fuzz.h
+++ b/include/afl-fuzz.h
@@ -116,6 +116,10 @@
   #include <TargetConditionals.h>
 #endif
 
+#ifndef __has_builtin
+  #define __has_builtin(x) 0
+#endif
+
 #undef LIST_FOREACH                                 /* clashes with FreeBSD */
 #include "list.h"
 #ifndef SIMPLE_FILES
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
Build fails:

src/afl-fuzz-bitmap.c: In function ‘count_bits’:
src/afl-fuzz-bitmap.c:78:18: error: missing binary operator before token "("
   78 | #if __has_builtin(__builtin_popcount)
      |                  ^
make: *** [GNUmakefile:483: afl-fuzz] Error 1

visitorckw:
mmm, it compiles successfully on my Ubuntu 22.04.4 LTS.
Not sure why __has_builtin is causing a compilation error in the CI environment.
I will try to fix it.

visitorckw:
Adding
```
#ifndef __has_builtin
  #define __has_builtin(x) 0
#endif
```
can solve the problem, but I'm not sure which header this code should be placed in. Maybe afl-fuzz.h?

vanhauser-thc:
Thank you!

PR #2134, opened 2024-06-25
Adding of beta QEMU hooking bridge

CowBoy4mH3LL:
# Purpose
To introduce a QEMU native hooking mechanism in QEMUAFL and supporting it through AFL++.

# New features
1. Ability to hook arbitrary addresses in the target binary (except in the first block after entry)
2. Ability to manipulate memory and register from within the hook
3. Ability to return to address of choice after the hook 
4. Ability to keep the hook intact on demand for future calls

# Code changes
1. A `qemu_mode/hooking_bridge` folder that contains
    * code to generate a QEMU TCG plugin that bridges user a hook library with the QEMU usermode runtime.
    * README that instructs how to use it
    * API definitions in `inc/exports.h` 
2. Changes in `qemu_mode/qemuafl/gdbstub.c` to implement the logic described last in this PR 

# Solution/code walkthrough
https://github.com/CowBoy4mH3LL/AFLplusplus/blob/PR_desc/qemu_mode/hooking_bridge/docs/code_walkthrough.md

# Adherence to contribution guidelines
1. Formatted as per recommended custom clang formatter
2. Arch agnostic make

# Testing 
Currently on an in-house app, which we can share if required.
Tested on Ubuntu Jammy, aarch64 and x86_64.

# Going forward
1. Please suggest if further tests are required and how to integrate with AFL native test cases
2. Please suggest if formatting/coding norms are out of place


.gitmodules
```diff
@@ -6,7 +6,7 @@
 	url = https://github.com/AFLplusplus/Grammar-Mutator
 [submodule "qemu_mode/qemuafl"]
 	path = qemu_mode/qemuafl
-	url = https://github.com/AFLplusplus/qemuafl
+	url = https://github.com/CowBoy4mH3LL/qemuafl
```
vanhauser-thc: No :-) if this needs qemuafl changes then put them as a PR to our qemuafl repo

vanhauser-thc:
@CowBoy4mH3LL  please resolve the conflict and the switch from the qemuafl repo

@andreafioraldi  WDYT?

domenukk:
Hooks like this (and more powerful ones) are already implemented in libafl_qemu.
I would argue it's better to work on replacing qemuafl with libafl_qemu / provide a forkserver in libafl_qemu (which is actively maintained and a fairly recent qemu version), rather than splitting the ecosystem with different hook implementations

On the other hand, the patch looks small enough, so we can merge it of course - I just wanted to say it's not clear to me if anyone will actively maintain the features when the land(?)

domenukk:
In any case thank you @CowBoy4mH3LL , this is for sure cool stuff! I don't mean to sound negtive! :)

.gitmodules
CowBoy4mH3LL: @vanhauser-thc  -- Got it -- so to clarify

1. I create a new PR for qemuafl
2. Resolve the conflict -- i.e. reset the submodule path to AFLplusplus qemuafl

Do we see any other changes?

@domenukk  -- I agree on the libafl (I guess the Bar workshop paper) part -- It's like we're patching the legacy part of it :) -- of course, this patch being advantageous for those who are still willing to just go by the pre-packaged AFL++, e.g. where modular testing may not be required, CTFs, etc.

vanhauser-thc: Yes imho that is all that is needed.

vanhauser-thc: I understand Dominik’s opinion on this but the user base for afl++ is currently and will be still for at least a year or longer much much larger than that of libafl

domenukk: I'm not talking about AFL++ vs LibAFL, I'm talking about qemuafl vs libafl_qemu, the former isn't really actively maintained, adding an afl++ forkserver to the latter would be the better time invest

domenukk: https://github.com/AFLplusplus/LibAFL/pull/1983

vanhauser-thc: > I'm not talking about AFL++ vs LibAFL, I'm talking about qemuafl vs libafl_qemu, the former isn't really actively maintained, adding an afl++ forkserver to the latter would be the better time invest

Ah sorry i misunderstood! I totally agree. Is someone on it? Adding a forkserver is easy but supporting the afl qemu envs are likely quite some work

CowBoy4mH3LL: I suspect even if we move away from qemuafl, since libafl_qemu builds on QEMU itself, we can integrate this hooking system into libafl_qemu as well -- infact, going forward I am thinking of doing a quick performance comparison of the two hooking methods -- if the hooking system of this PR comes out to be faster (due to native integration into QEMU) we can keep both

domenukk: I would assume libafl_qemu native hooks will be *a lot* faster since they are optimized for performance, but I would be very keen to see the benchmark! :)

CowBoy4mH3LL: @domenukk -- absolutely, I'll be on it sooner

@vanhauser-thc -- is it OK to create qemuafl PR on master directly? -- I don't see a dev branch

vanhauser-thc: Yes make a pr to master 👍🏻

CowBoy4mH3LL: Deal!!

CowBoy4mH3LL: > I would assume libafl_qemu native hooks will be _a lot_ faster since they are optimized for performance, but I would be very keen to see the benchmark! :)

Saw that in the paper -- but how are hooks called? Do you follow instructions or TCGs, or leverage the plugin ecosystem or is it event based like trap in our case -- Just out of curiosity

andreafioraldi:
isn't this already supported natively in qemu with tcg plugins?
https://www.qemu.org/docs/master/devel/tcg-plugins.html

CowBoy4mH3LL:
> isn't this already supported natively in qemu with tcg plugins? https://www.qemu.org/docs/master/devel/tcg-plugins.html

At a block level -- not at an instruction level -- also without support for memory/register manipulation

vanhauser-thc:
qemuafl pr is merged now please fix this pr :)

CowBoy4mH3LL:
> qemuafl pr is merged now please fix this pr :)

@vanhauser-thc Done!

.gitmodules
vanhauser-thc: @CowBoy4mH3LL you still need to reverse this

qemu_mode/build_qemu_support.sh
```diff
@@ -386,6 +386,19 @@ else
   make -C libqasan CC="$CROSS $CROSS_FLAGS" && echo "[+] libqasan ready"
 fi
 
+#### Hooking support
+if [ "$ENABLE_HOOKING" = "1" ];then
+  echo "[+] ENABLING HOOKING"
```
vanhauser-thc, line 391: is there a reason this is not on by default? performance? if there is no noticable impact it should be on by default. if it does not work when compiled static then check for that.

vanhauser-thc:
left 2 commends.
also note that you change unicorn afl - likely this is a mistake

qemu_mode/build_qemu_support.sh
CowBoy4mH3LL: > is there a reason this is not on by default? performance? if there is no noticable impact it should be on by default.

By default, it is not compiled because it depends on glib which, if not available, will hinder the entire afl-qemu-trace compilation
Even when compiled, the user can choose to ignore hooking by not including hooking_bridge plugin. If they do, two things can be expected which, along with the compilation roadblock mentioned before, made me make this module optional
1. Performance: each block translation needs to be hooked, even though only the "entry" point hook is used -- I have noticed some performance penalty when translation/execution is hooked -- I may experiment with translation hook removal after the "entry" block at a later stage
2. If afl-qemu-trace is used as standalone, say for one-off testing -- the hooking system would impede debugging as it hijacks GDB

> if it does not work when compiled static then check for that.

I have not checked for that, but it will not interfere with a statically compiled AFL++. However, it does not really fit the static compilation use case because the user hooks themselves are to be placed in a shared library at this time. @vanhauser-thc should I still verify and update build_qemu_support accordingly?

.gitmodules
CowBoy4mH3LL: Sorry about that -- reset it now :)

CowBoy4mH3LL:
> also note that you change unicorn afl - likely this is a mistake

This is baffling me -- its the submodule and even though I took at very bad big at removing the gitlink -- I am not sure if that removes the change -- I am just not sure how the change shows up in this repo even -- would really appreciate any pointers :(

vanhauser-thc:
now it looks like you delete the unicornafl repository in the PR.
can you please make a new PR that is based on the current dev branch that only has what is needed? thanks you. and closing this one.

PR #2138, opened 2024-06-29
IJON SET

meowmeowxw:
Hello,
this is an implementation for IJON set that can be used both in frida mode and in source code instrumentation.
At the moment I haven't externed helper IJON functions like: ijon_hashint, ijon_hashstr, ijon_hashmem, ijon_simplehash (https://github.com/RUB-SysSec/ijon/blob/master/llvm_mode/afl-llvm-rt.o.c#L117) and I haven't hashed the __LINE,__FILE number automatically when using source code instrumentation.

Example on how to use it in frida mode (for now):

```javascript
function hashNumbers(x, y) {
    const prime = 31;
    let hash = 17;
    hash = prime * hash + x;
    hash = prime * hash + y;
    return hash & 0xffffffff;
}

Interceptor.attach(ptr('0x00401291'), function(args) {
    // read state variable
    var rbp = this.context.rbp;
    var x_addr = rbp.sub(0x4);
    var y_addr = rbp.sub(0x8);
    var x = Memory.readS32(x_addr);
    var y = Memory.readS32(y_addr);

    // create hash of state
    var hash = hashNumbers(x, y);
    Afl.print(`hash: ${hash}`);

    // Pass the hashed state to IJON.set
    // It will xor RIP with hash => IDX.
    //  and store 1 in the LSB bit of the coverage map at index (IDX % __afl_map_size)
    Afl.IJON.set(this.context.rip & 0xffffffff, hash);
});

Afl.done();
```

Note:

The original work was done for frida mode only: https://github.com/meowmeowxw/AFLplusplus/tree/ijon, and a view of the changes can be seen at: https://github.com/meowmeowxw/AFLplusplus/compare/f2cd5e1d8e8a97ed86990c0eeb6f00e7c289fc44..ijon

The functions implemented were IJON {max, set, inc + helpers} and we tested them against maze/supermario programs. Additional scripts/tests we (me,[Max Gallup](https://github.com/maxgallup), [Alex Birnberg](https://github.com/alexbirnberg)) have done can be found at: https://github.com/maxgallup/fuzz-project.

I think that it is better to do a separate PR for IJON max() since it requires some heavy modifications of the source code.

An advanced use case for IJON's set() function would be to hash multiple variables or the transitions of one or more states/variables. This approach could lead to too many collisions in the coverage map, so it would be better 
to separate the coverage and "state" maps. Additionally, this "state" map could exceed 64-512kb since it would be rarely accessed. Implementing this approach might be better suited for a separate PR at a later stage if it shows potential.

### DIFF
--- a/instrumentation/afl-compiler-rt.o.c
+++ b/instrumentation/afl-compiler-rt.o.c
@@ -2761,11 +2761,5 @@ void __afl_injection_xss(u8 *buf) {
 
 }
 
-void __afl_ijon_set(u32 edge) {
-
-  __afl_area_ptr[edge % __afl_map_size] |= 1;
-
-}
-
 #undef write_error
 
--- a/instrumentation/afl-compiler-rt.o.c
+++ b/instrumentation/afl-compiler-rt.o.c
@@ -2704,7 +2704,7 @@ void __afl_coverage_skip() {
 // mark this area as especially interesting
 void __afl_coverage_interesting(u8 val, u32 id) {
 
-  __afl_area_ptr[id] = val;
+  __afl_area_ptr[id % __afl_map_size] = val;
 
 }
 
--- a/instrumentation/afl-compiler-rt.o.c
+++ b/instrumentation/afl-compiler-rt.o.c
@@ -2761,5 +2761,11 @@ void __afl_injection_xss(u8 *buf) {
 
 }
 
+void __afl_ijon_set(u32 edge) {
+
+  __afl_area_ptr[edge % __afl_map_size] |= 1;
+
+}
+
 #undef write_error
 
--- a/dynamic_list.txt
+++ b/dynamic_list.txt
@@ -30,6 +30,7 @@
   "__afl_selective_coverage_temp";
   "__afl_sharedmem_fuzzing";
   "__afl_trace";
+  "__afl_ijon_set";
   "__cmplog_ins_hook1";
   "__cmplog_ins_hook16";
   "__cmplog_ins_hook2";
--- a/dynamic_list.txt
+++ b/dynamic_list.txt
@@ -30,7 +30,6 @@
   "__afl_selective_coverage_temp";
   "__afl_sharedmem_fuzzing";
   "__afl_trace";
-  "__afl_ijon_set";
   "__cmplog_ins_hook1";
   "__cmplog_ins_hook16";
   "__cmplog_ins_hook2";
--- a/frida_mode/frida.map
+++ b/frida_mode/frida.map
@@ -45,6 +45,7 @@
     js_api_set_stdout;
     js_api_set_traceable;
     js_api_set_verbose;
+    js_api_ijon_set;
 
   local:
     *;
--- a/frida_mode/include/instrument.h
+++ b/frida_mode/include/instrument.h
@@ -22,7 +22,7 @@ extern guint64  instrument_fixed_seed;
 
 extern uint8_t *__afl_area_ptr;
 extern uint32_t __afl_map_size;
-extern void     __afl_ijon_set(uint32_t);
+extern void __afl_coverage_interesting(uint8_t, uint32_t);
 
 extern __thread guint64 *instrument_previous_pc_addr;
 
--- a/frida_mode/include/instrument.h
+++ b/frida_mode/include/instrument.h
@@ -22,6 +22,7 @@ extern guint64  instrument_fixed_seed;
 
 extern uint8_t *__afl_area_ptr;
 extern uint32_t __afl_map_size;
+extern void     __afl_ijon_set(uint32_t);
 
 extern __thread guint64 *instrument_previous_pc_addr;
 
@@ -71,6 +72,8 @@ void instrument_cache_insert(gpointer real_address, gpointer code_address);
 void instrument_cache(const cs_insn *instr, GumStalkerOutput *output);
 void instrument_write_regs(GumCpuContext *cpu_context, gpointer user_data);
 void instrument_regs_format(int fd, char *format, ...);
+
+void ijon_set(uint32_t edge);
 
 #endif
 
--- a/frida_mode/src/instrument/instrument.c
+++ b/frida_mode/src/instrument/instrument.c
@@ -451,7 +451,7 @@ void instrument_regs_format(int fd, char *format, ...) {
 
 void ijon_set(uint32_t edge) {
 
-  __afl_ijon_set(edge);
+  __afl_area_ptr[edge % __afl_map_size] |= 1;
 
 }
 
--- a/frida_mode/src/instrument/instrument.c
+++ b/frida_mode/src/instrument/instrument.c
@@ -451,7 +451,7 @@ void instrument_regs_format(int fd, char *format, ...) {
 
 void ijon_set(uint32_t edge) {
 
-  __afl_area_ptr[edge % __afl_map_size] |= 1;
+  __afl_coverage_interesting(1, edge);
 
 }
 
--- a/frida_mode/src/instrument/instrument.c
+++ b/frida_mode/src/instrument/instrument.c
@@ -449,3 +449,9 @@ void instrument_regs_format(int fd, char *format, ...) {
 
 }
 
+void ijon_set(uint32_t edge) {
+
+  __afl_ijon_set(edge);
+
+}
+
--- a/frida_mode/src/js/api.js
+++ b/frida_mode/src/js/api.js
@@ -326,6 +326,12 @@ class Afl {
     static jsApiGetSymbol(name) {
         return Afl.module.getExportByName(name);
     }
+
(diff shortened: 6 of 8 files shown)


vanhauser-thc:
I think you can remove `frida_mode/test/fasan/.gdb_history`.

but how does it actually work? this seems to set a bit in an edge in the coverage map, but how does this help to solve a value?

meowmeowxw:
The base example that is used in the original paper for `IJON.set()` is a maze like: https://github.com/maxgallup/fuzz-project/blob/main/eval-new-ijon/src/maze-small.c, where a new state (x,y coordinates) does not result in new code executed, and at the end of the maze a bug is triggered.

In this case we can pass the hash of the coordinate of (x,y) to `IJON.set()` , which would let AFL think that new code coverage was unlocked anytime a new state is being found. In this way it guides the fuzzer towards the solution.

In practice, I believe that it can be (roughly) used when a certain order of state transitions is important to trigger a bug, similar to the examples present in SGFuzz (https://www.usenix.org/conference/usenixsecurity22/presentation/ba) where they build a state transition tree dynamically.
In this case it would be possible to record the state transition of 1 or multiple variables and hash it. For example, suppose that there is a variable  that can transition to different states like: `0 -> 1 -> 2`, or  `1 -> 2 -> 1`. The code coverage would probably look the same, since at each state change the code executed will be the same, but the transitions are different.  

IJON reference: https://ieeexplore.ieee.org/document/9152719

instrumentation/afl-compiler-rt.o.c
```diff
@@ -2761,5 +2761,11 @@ void __afl_injection_xss(u8 *buf) {
 
 }
 
+void __afl_ijon_set(u32 edge) {
```
domenukk: This is conceptually very close to `__afl_coverage_interesting(u8 val, u32 id)`, however that one doesn't use modulo. I think the functions should behave consistently here.
Also, should it be named `ijon`, or is there a more descriptive name?
 `https://github.com/AFLplusplus/AFLplusplus/blob/2a489f844b7bc910caad2ab46aa3d86c3e6fbef8/instrumentation/afl-compiler-rt.o.c#L2705

vanhauser-thc: I agree, IMHO the modulo should be in the frida code, and yes, this could use the __afl_coverage_interesting instead.

meowmeowxw: Indeed they are similar :thinking:, but the usage is different. Do you use `__afl_coverage_interesting` to mark a certain code block/instrucion as particularly "interesting" by setting an high value (ex. 128)? 

It could be named `__afl_coverage_set` (in practice) or `__afl_state_set` (conceptually).
Ideally the edge/id is something that can be set by the user (that's why the modulo), however I'm not really sure what are the implications when using source code instrumentation. I will try to do some tests

vanhauser-thc:
OK, sorry yesterday I was working on different things so I didn't fully read all the text.
so yes, this has everything that is needed. as dominik pointed out, we already have a (better) function for this, please switch to that one and put the modulo in the frida part.

About ijon.max() - do you have a diff somewhere to estimate how much code changes that would be?
I would hope there is not much - maybe it is possible to implement it rather like cmplog, so it has it's own stage and just needs to be called in afl-fuzz-one?

meowmeowxw:
For ijon.max() this was a dirty implementation https://github.com/meowmeowxw/AFLplusplus/compare/f2cd5e1d8e8a97ed86990c0eeb6f00e7c289fc44..ijon (There is also ijon-set).
The trick that I've used was to [increase MAP_SIZE](https://github.com/meowmeowxw/AFLplusplus/compare/f2cd5e1d8e8a97ed86990c0eeb6f00e7c289fc44..ijon#diff-f20ab3eb0041360bbad27bdb1173ada9c2818c430536c89ccd514d822c43a1ecR467), and then substract back  IJON_MAP_SIZE in [afl-compiler-rt.o.c](https://github.com/meowmeowxw/AFLplusplus/compare/f2cd5e1d8e8a97ed86990c0eeb6f00e7c289fc44..ijon#diff-2d6ea5de3694b7613a064621d26bbe4fa4a9b8064048d9a2f5208f1bdff8ff21R754) and [src/afl-fuzz.c](https://github.com/meowmeowxw/AFLplusplus/compare/f2cd5e1d8e8a97ed86990c0eeb6f00e7c289fc44..ijon#diff-fd9f999d1db0c28c1a94de562469fe1fa166c4cdd4f70be3d3471ea9316e3d99R2502).
The reason I've used this approach was that [__afl_map_shm()](https://github.com/AFLplusplus/AFLplusplus/blob/2a489f844b7bc910caad2ab46aa3d86c3e6fbef8/instrumentation/afl-compiler-rt.o.c#L348) was full of boilerplate code/using globals, so I wasn't sure how to map a new shared map (quickly).

Additionally the code is checking whether a [new max has been found](https://github.com/meowmeowxw/AFLplusplus/compare/f2cd5e1d8e8a97ed86990c0eeb6f00e7c289fc44..ijon#diff-23e73fde715cbf7b95ad13909e47d6852ce840c00788539021236bc12a0d7e70R35), [save in a array the pointers to the queue entry that are maximised](https://github.com/meowmeowxw/AFLplusplus/compare/f2cd5e1d8e8a97ed86990c0eeb6f00e7c289fc44..ijon#diff-23e73fde715cbf7b95ad13909e47d6852ce840c00788539021236bc12a0d7e70R610), and schedule these entries with n% [probability](https://github.com/meowmeowxw/AFLplusplus/compare/f2cd5e1d8e8a97ed86990c0eeb6f00e7c289fc44..ijon#diff-6014111dd595b4fe80b8a3217b59af6cdb9f6bb24b22943433cade54d450f114R1513)

meowmeowxw:
I think that it is reasonable to use the coverage map directly as in https://github.com/AFLplusplus/AFLplusplus/pull/2138/commits/8fbeeb143989f4ebbcc238a96305e4c534e074c7, instead of relying on `__afl_coverage_interesting()`, since I'm just oring `1` to that edge.
What do you think? @vanhauser-thc @domenukk 

Would you like to have a test case for frida mode? And to have a different function name? 

I'm also not sure how `__afl_coverage_interesting()` can be used as `ijon_set()` to set interesting states in source code instrumentation, because the use case seems different to me :thinking: .

vanhauser-thc:
> I think that it is reasonable to use the coverage map directly as in [8fbeeb1](https://github.com/AFLplusplus/AFLplusplus/commit/8fbeeb143989f4ebbcc238a96305e4c534e074c7), instead of relying on `__afl_coverage_interesting()`, since I'm just oring `1` to that edge. What do you think? @vanhauser-thc @domenukk
> 
> Would you like to have a test case for frida mode? And to have a different function name?
> 

> I'm also not sure how `__afl_coverage_interesting()` can be used as `ijon_set()` to set interesting states in source code instrumentation, because the use case seems different to me 🤔 .

It is slightly different but can be used for the same thing.
you would just do `__afl_coverage_interesting(edge, 1);` in your case.

we use it for source only, but there it is better because it is for non-colliding coverage and therefore this creates a dedicated map entry.
with frida it is a fixed map with colliding coverage so you have the risk that the map entry is also being used for one or more random edges.

For a real IJON implementation I think we need changes to the map (especially for colliding coverage).

I think it helps if I re-read the paper, it has been long ago.

pinging @eqv here so he is aware

meowmeowxw:
> with frida it is a fixed map with colliding coverage so you have the risk that the map entry is also being used for one or more random edges.

I think that this is a limitation of the design of IJON set, where you want to treat a new "state" as a new edge coverage.
To remove collisions inside the coverage map we could also create a separate map just to store the states, and check whether a new state has been found or not in save_if_interesting().

> I think it helps if I re-read the paper, it has been long ago.

Yes let me know if you find new ideas for implementing it, thanks :+1:

eqv:
yeah, the original implementation of ijon_set just reused the existing bitmap - with the mentioned collision risk. Back in the days, everyone was still using `offset = last_rnd>>2 ^ cur_rnd`, so collisions were a thing anyway.

vanhauser-thc:
please switch `ijon_set` to `__afl_coverage_interesting(edge, 1)` 
You can add `% __afl_map_size` to that function instead.

vanhauser-thc:
ping @meowmeowxw  :)

meowmeowxw:
Hi @vanhauser-thc, I've tested a couple of days ago the change `__afl_coverage_interesting(edge, 1)` but it wasn't working properly on the maze example, so I've wanted to push a test for frida mode but didn't have time yet. I will try to do it by the end of the week

meowmeowxw:
Ah actually I should call `__afl_coverage_interesting(1, edge)`, let me quickly test it.

Edit: Ok it works :+1:

vanhauser-thc:
LGTM, thanks!

PR #2143, opened 2024-07-01
Make fallthroughs explicit in afl-fuzz-extras.c

r-barnes:
Using `__attribute__((fallthrough))` makes fallthroughs explicit in a way the compiler can understand. This allows the enablement of `-Wimplicit-fallthrough`.

### DIFF
--- a/src/afl-fuzz-extras.c
+++ b/src/afl-fuzz-extras.c
@@ -455,13 +455,13 @@ void deunicode_extras(afl_state_t *afl) {
 
         case 2:
           if (!afl->extras[i].data[j]) { ++z3; }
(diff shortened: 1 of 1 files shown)


domenukk:
Thanks!
I don't know if we have a min gcc version, but this one needs at least gcc 7. @vanhauser-thc will need to decide :)
https://stackoverflow.com/a/48702576

domenukk:
Also, please direct fixes to the dev branch, thank you!

vanhauser-thc:
It is difficult to find something older than gcc 9 in my experience. And as we need at least clang 13 I don’t mind this

PR #2146, opened 2024-07-07
Update sample_all.sh for incorrect shell syntax

AlbertnQ:
Incorrect shell syntax


domenukk:
Please direct PRs to dev :)

domenukk:
I think _technically_ `$DIR/harness` should also be quoted, right?

PR #2147, opened 2024-07-08
Enabled afl-gcc and afl-clang for ARM

rtandr01d:
Enable afl-gcc and afl-clang for 32-bit ARM. Tested and working on Debian armv7 docker image.

### DIFF
--- a/src/afl-cc.c
+++ b/src/afl-cc.c
@@ -525,7 +525,7 @@ void find_built_deps(aflcc_state_t *aflcc) {
 
   char *ptr = NULL;
 
-#if defined(__x86_64__) || defined(__i386__)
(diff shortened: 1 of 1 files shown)


domenukk:
Thank you! Please direct PRs to dev next time :)

domenukk:
Have you tried fuzzing with it btw? It shouldn't work. The gcc plugin might, but the "normal" afl instrumentation doesn't have arm support AFAIK

vanhauser-thc:
@rtandr01d 

@domenukk is correct - this cannot work because the assembly only supports x86, not arm.
you could add support for it though :)

Closing this for now.

rtandr01d:
Both instrumentation when compiling with afl-clang or afl-gcc and fuzzing binaries with afl-fuzz works.

PR #2149, opened 2024-07-08
Enable afl-gcc and afl-clang for ARM

rtandr01d:
# Enable afl-gcc and afl-clang for 32-bit ARM

This issue was closed before, so I will provide proof that both instrumenting and fuzzing works.
I used an example AFL++ exercise to instrument and fuzz libexif. In addition, I will use the output of afl-fuzz to show coverage.

Note: AFL shows the execution as slow since the docker container utilizes qemu for system emulation

https://github.com/antonio-morales/Fuzzing101/blob/main/Exercise%202/Readme.md

## Compiling:

`root@b3f9e48be050:~/fuzzing_libexif/libexif-libexif-0_6_14-release# make && make install`

2024-07-08T16:21:19,983908100-05:00

`root@b3f9e48be050:~/fuzzing_libexif/exif-exif-0_6_15-release# make && make install`

2024-07-08T16:22:40,980235278-05:00

## Fuzzing:

`afl-fuzz -i $HOME/fuzzing_libexif/exif-samples-master/jpg/ -o $HOME/fuzzing_libexif/out/ -s 123 -- $HOME/fuzzing_libexif/install/bin/exif @@`

2024-07-08T16:10:41,744936228-05:00

## Coverage

2024-07-08T16:19:49,985004511-05:00

### DIFF
--- a/src/afl-cc.c
+++ b/src/afl-cc.c
@@ -525,7 +525,7 @@ void find_built_deps(aflcc_state_t *aflcc) {
 
   char *ptr = NULL;
 
-#if defined(__x86_64__) || defined(__i386__)
+#if defined(__x86_64__) || defined(__i386__ || defined(__arm__))
   if ((ptr = find_object(aflcc, "afl-as")) != NULL) {
 
   #ifndef __APPLE__


vanhauser-thc:
if you look closely at your screenshots you will see `Mode: LLVM-PCGUARD` 

afl-gcc is just a symlink to `afl-cc` that will prefer what you select by symlink but switch to something else if that option is not available. that switching is however not constistant, as sometimes it is better to error and sometimes better to switch.

your patch changed the behaviour of that, that is all.

If you look into `include/afl-as.h` which is used for the GCC/CLANG mode, then you will see that there is no arm support in there.

domenukk:
FWIW I _think_ fuzzing on arm with gcc and the gcc plugin should be possible(?)

vanhauser-thc:
> FWIW I _think_ fuzzing on arm with gcc and the gcc plugin should be possible(?)

It is, also mips etc. the problem is though that for embedded often a custom vendor gcc is used and that does not have modules

rtandr01d:
I see your point with afl-as. Is there place for me to understand the instrumentation better? Its hard for me to see this being an issue when I am able to compile/instrument and successfully fuzz.

domenukk:
You are successfully compiling and fuzzing using clang/LLVM not with gcc. :)

rtandr01d:
So would you say with this change, it is possible to instrument/compile and fuzz ARM binaries with afl-clang?

domenukk:
That's possible without this change or with this change, but the change is not correct / it doesn't enable gcc

rtandr01d:
Without this change, both afl-gcc and afl-clang report the binary is unavailable on the platform as 'have_gcc' and 'have_clang' are set to 0.

rtandr01d:
I think I understand what you say about gcc. Would it be more appropriate to have a PR with this change saying only afl-clang is enabled on ARM?

rtandr01d:
A possibly more accurate version, enabling clang:

Screenshot_20240709_102419

domenukk:
You need to use afl-clang-fast or afl-cc, then it shoudl already work. Feel free to add documentation around that.
afl-clang is actually some legacy thing...

vanhauser-thc:
> You need to use afl-clang-fast or afl-cc, then it shoudl already work. Feel free to add documentation around that. afl-clang is actually some legacy thing...

Yes this is the correct solution!

PR #2152, opened 2024-07-09
Some fixups to the GNUmakefile.llvm

exoosh:
* rely less on the shell and more on GNU make to parse the versions
* fixed retrieval of minor version (for 18.1.8 it gave 8 instead of 1!)
* auto-detection of llvm-config within the supported version range
* replaced backticks by `$(...)` syntax
* tested against `busybox static-sh`, `bash`, `dash` and `csh`

NB: Please advise if there are any rules I failed to observe or so. Didn't notice anything specific to the make files.

### DIFF
--- a/GNUmakefile.llvm
+++ b/GNUmakefile.llvm
@@ -32,6 +32,9 @@ VERSION     = $(shell grep '^$(HASH)define VERSION ' ./config.h | cut -d '"' -f2
 
 SYS = $(shell uname -s)
 
+override LLVM_TOO_NEW_DEFAULT := 18
+override LLVM_TOO_OLD_DEFAULT := 13
+
 ifeq "$(SYS)" "OpenBSD"
   LLVM_CONFIG ?= $(BIN_PATH)/llvm-config
   HAS_OPT = $(shell test -x $(BIN_PATH)/opt && echo 0 || echo 1)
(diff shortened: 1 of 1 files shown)


exoosh:
Oops, I noticed I added a few changes I would like to have for myself but which aren't necessarily desirable for you. Will push again tomorrow.

Either way I'd recommend to replace the backtick syntax for subshells with `$(...)` (`$$(...)` when used inside GNU make's `$(shell ...)`) ... which is POSIX-compliant (since ~1993) and widely supported.

exoosh:
Could someone from the team explain one item to me, please?! What's the purpose of `LLVM_UNSUPPORTED`? It appears to cover the same ground as `LLVM_TOO_OLD`, just with a different limit and in relation to XCode.

I sifted a bit through the history, but I don't quite get it. Also, the regular expression on that line is odd.

To explain, the following uses brace expansion, so requires Bash or possibly Zsh:

```
for i in {0..2}.$RANDOM 3{a..f}{0..8}.$RANDOM; do echo "$i" | grep -E -q '^[0-2]\.|^3.[0-8]\.' && echo "1: $i" || echo "0: $i"; done
```

I am fairly sure that the intent was not to match something like `3f8.31197` as explicitly unsupported.

My guess is that the secondary part of the regex was meant to be `3\.[0-8]\.`. I fixed that in my amended commit.

vanhauser-thc:
please note we only accept PRs to the dev branch as specified in the contributor.md. I will change this for you.

GNUmakefile.llvm
```diff
@@ -32,31 +32,40 @@ VERSION     = $(shell grep '^$(HASH)define VERSION ' ./config.h | cut -d '"' -f2
 
 SYS = $(shell uname -s)
 
+override LLVM_TO_NEW_LIMIT := 18
+override LLVM_TO_OLD_LIMIT := 13
```
vanhauser-thc: both are missing an extra O for _TOO_

vanhauser-thc: ah now I see what you are doing below. I would still rename this to LLVM_TOO_xxx_DEFAULT or something like that, looks odd otherwise :)

vanhauser-thc:
why change something that is working? but I think it does not hurt either.

GNUmakefile.llvm
exoosh: Changes incoming soon. Wow, I don't even know how I missed the double OO  ¯\\\_(ツ)\_/¯

exoosh:
> why change something that is working?

Hmm, it's a fair point. What prompted this initially was the fact that I needed to give `LLVM_CONFIG` and since I had already devised a workaround for figuring that out automatically, I thought: why not fix that? And then I noticed all those version-based checks and thought this could be made more robust.

Also, I think the check in `LLVM_UNSUPPORTED` really was broken.

vanhauser-thc:
yes I am happy for the bug fix and also the llvm-config autodetect. waiting for the CI and then merging. Thanks!

vanhauser-thc:
@exoosh some things are not working:
```
make -f GNUmakefile.llvm install
make[1]: Entering directory '/prg/afl'
/bin/sh: 1: test: -gt: argument expected
/bin/sh: 1: test: -lt: argument expected
```

exoosh:
What shell? Looks as if one of the expected arguments was empty. Can you give me a way to reproduce? Which shell?

vanhauser-thc:
I type just make. My shell is bash

exoosh:
> @exoosh some things are not working:
> 
> ```
> make -f GNUmakefile.llvm install
> make[1]: Entering directory '/prg/afl'
> /bin/sh: 1: test: -gt: argument expected
> /bin/sh: 1: test: -lt: argument expected
> ```

Looking at the changed version again, I notice there is just one instance of `-gt` and `-lt` respectively. Therefore we know the exact locations. However, this indicates that `llvm-config` wasn't found or didn't yield a sensible value for `LLVM_MAJOR`. Could you try to rerun whatever it is you were running but with line 68 uncommented, so we can see all those variable values, @vanhauser-thc ?

Also, did `make` stop after those failures (I imagine not, because otherwise it wouldn't have shown the second line already ...)?

**Edit:** so after spotting the issue (read further below), the problem wasn't `LLVM_MAJOR` but that `LLVM_TO_NEW_LIMIT` and `LLVM_TO_OLD_LIMIT` respectively evaluated to an empty string.

exoosh:
Able to reproduce. Looking into it now.

exoosh:
Damn, of course. My bad entirely. The issue is that the variables on those two lines are misspelled.
image

Give me a few minutes.

PR #2157, opened 2024-07-12
check the sync_id length once

Ninja3047:
The maximum length of sync_id is now checked once in `fix_up_sync` instead of twice with different sizes.


vanhauser-thc:
Yeah makes sense :)

PR #2158, opened 2024-07-13
push to stable

vanhauser-thc:
waiting for a fix from @exoosh


GNUmakefile.llvm
```diff
@@ -245,7 +254,7 @@ endif
 
 AFL_CLANG_FUSELD=
 ifeq "$(LLVM_LTO)" "1"
-  ifeq "$(shell echo 'int main() {return 0; }' | $(CLANG_BIN) -x c - -fuse-ld=`command -v ld` -o .test 2>/dev/null && echo 1 || echo 0 ; rm -f .test )" "1"
+  ifeq "$(shell echo 'int main() {return 0; }' | $(CLANG_BIN) -x c - -fuse-ld=$$(command -v ld) -o .test 2>/dev/null && echo 1 || echo 0 ; rm -f .test )" "1"
```
domenukk, line 257: Too many `$`?

GNUmakefile.llvm
```diff
@@ -300,8 +309,8 @@ endif
 ifneq "$(LLVM_CONFIG)" ""
   CLANG_CFL += -I$(shell dirname $(LLVM_CONFIG))/../include
 endif
-CLANG_CPPFL  = `$(LLVM_CONFIG) --cxxflags` -fno-rtti -fno-exceptions -fPIC $(CXXFLAGS) $(CPPFLAGS) -Wno-deprecated-declarations
-CLANG_LFL    = `$(LLVM_CONFIG) --ldflags` $(LDFLAGS)
+CLANG_CPPFL  = $$($(LLVM_CONFIG) --cxxflags) -fno-rtti -fno-exceptions -fPIC $(CXXFLAGS) $(CPPFLAGS) -Wno-deprecated-declarations
```
domenukk, line 312: Ah seems to be some syntax I don't know :D

domenukk: How does $$ work?

domenukk (approved):

GNUmakefile.llvm
vanhauser-thc: That’s why I don’t want to touch it ;)

exoosh: `$$` is simply escaping the `$` at GNU Make level. I.e. it's passing the `$` to the shell. POSIX shells have supported this -- i.e. subshells with `$(...)` -- since the POSIX 1993 standard, so one would have to attempt to build this in a rather ancient setup (which in turn is at odds with the requirements set out to build AFL++).

PS: If you feel like the backticks are more readable to you and aren't afraid that they can get you into escaping hell (because nesting subshells will be tricky) we can also revert that part.

exoosh:
#2160 should address the issue. Sorry again, folks!

PR #2159, opened 2024-07-13
Replace gettimeofday with clock_gettime

carpintero-de-c:
This PR replaces `gettimeofday` with `clock_gettime` in all source files. `gettimeofday` was deprecated in POSIX 2017 (that was 7 years ago) and has been [removed in POSIX 2024](https://sortix.org/blog/posix-2024/). It's about time we switch away.

### DIFF
--- a/custom_mutators/gramatron/test.c
+++ b/custom_mutators/gramatron/test.c
@@ -8,8 +8,8 @@
 state *create_pda(u8 *automaton_file) {
 
   struct json_object *parsed_json;
-  state *             pda;
-  json_object *       source_obj, *attr;
+  state              *pda;
+  json_object        *source_obj, *attr;
   int                 arraylen, ii, ii2, trigger_len, error;
 
   printf("\n[GF] Automaton file passed:%s", automaton_file);
@@ -41,7 +41,7 @@ state *create_pda(u8 *automaton_file) {
   enum json_type type;
   json_object_object_foreach(source_obj, key, val) {
 
-    state *  state_ptr;
+    state   *state_ptr;
     trigger *trigger_ptr;
     int      offset;
 
@@ -97,12 +97,12 @@ state *create_pda(u8 *automaton_file) {
 
 void SanityCheck(char *automaton_path) {
 
-  state *        pda = create_pda(automaton_path);
+  state         *pda = create_pda(automaton_path);
   int            count = 0, state;
   Get_Dupes_Ret *getdupesret;
-  IdxMap_new *   statemap;
-  IdxMap_new *   statemap_ptr;
-  terminal *     term_ptr;
+  IdxMap_new    *statemap;
+  IdxMap_new    *statemap_ptr;
+  terminal      *term_ptr;
 
   while (count < NUMINPUTS) {
 
@@ -117,12 +117,9 @@ void SanityCheck(char *automaton_path) {
 
 int main(int argc, char *argv[]) {
 
-  char *         mode;
-  char *         automaton_path;
-  char *         output_dir = NULL;
-  struct timeval tv;
-  struct timeval tz;
-  // gettimeofday(&tv, &tz);
+  char *mode;
+  char *automaton_path;
+  char *output_dir = NULL;
   srand(1337);
   if (argc == 3) {
 
--- a/custom_mutators/symqemu/symqemu.c
+++ b/custom_mutators/symqemu/symqemu.c
@@ -196,12 +196,11 @@ void afl_custom_splice_optout(void *data) {
 
 inline u64 get_cur_time(void) {
 
-  struct timeval  tv;
-  struct timezone tz;
+  struct timespec spec;
 
-  gettimeofday(&tv, &tz);
+  clock_gettime(CLOCK_REALTIME, &spec);
 
-  return (tv.tv_sec * 1000ULL) + (tv.tv_usec / 1000);
+  return (spec.tv_sec * 1000ULL) + (spec.tv_nsec / 1000000ULL);
 
 }
 
--- a/frida_mode/test/unstable/unstable.c
+++ b/frida_mode/test/unstable/unstable.c
@@ -14,7 +14,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <sys/time.h>
+#include <time.h>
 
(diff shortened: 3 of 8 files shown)


domenukk:
The question is if performance will be different (?)

carpintero-de-c:
glibc's `gettimeofday` [defers to `clock_gettime`](https://sourceware.org/git/?p=glibc.git;a=blob;f=time/gettimeofday.c;h=bf92756cae906b9afff44d302800babbb1413b4a;hb=HEAD) so there should be no performance difference, if anything there should be an improvement. Edit: [It's the same with musl](https://git.musl-libc.org/cgit/musl/tree/src/time/gettimeofday.c).

domenukk:
Thanks!

vanhauser-thc:
nonono!!!

I had to revert the merge of this PR because switching to gettimeofday decreases the performance. this change was responsible for the regression in 4.21.

carpintero-de-c:
@vanhauser-thc Do you mean switching to `clock_gettime`? But why would it decrease performance, I mean both glibc and musl `gettimeofday` implementations are just wrappers around `clock_gettime` with the same parameters (as I pointed out before in both their sources). Any benchmarks? Thanks.

vanhauser-thc:
yes switching from gettimeofday to clock_gettime results in a much worse overall performance. depending on the clock that is used it is just a small impact to a huge impact.
COARSE clocks are faster but have the huge negative impact. REALTIME is the closest but still has a noticable negative impact.
dunno why. I can only measure this with fuzzbench performance over 23 hours, 21 targets, 20 instances each ...

just testing the calls itself clock_gettime is faster.

carpintero-de-c:
Strange, it doesn't even makes sense because of the underlying implementation of `gettimeofday`. `CLOCK_REALTIME` should really have no impact as it's how libc itself implements `gettimeofday`. Maybe older versions of libc used a custom implementation for `gettimeofday`, and we're seeing the effects of that?

domenukk:
Yes, looking at the [glibc source](https://sourceware.org/git/?p=glibc.git;a=blob;f=time/gettimeofday.c;h=bf92756cae906b9afff44d302800babbb1413b4a;hb=HEAD) that @carpintero-de-c linked, int calls down to `__clock_gettime (CLOCK_REALTIME` so how would it possibly make a difference?

The only difference is that we then add `(spec.tv_nsec / 1000000ULL);` instead of `(tv.tv_usec / 1000);`, could that possibly be faster?

carpintero-de-c:
@domenukk That doesn't seem to be the case either, [dividing by 1000000 produces less instructions than dividing by 1000 on both `gcc` and `clang`](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYMQADlIOAMngMmABy7gBGmMQgAGykAA6oCoS2DM5uHt7xickCAUGhLBFRsZaY1ilCBEzEBGnunj6l5QKV1QR5IeGRMRZVNXUZjX3tgZ2F3dEAlBaorsTI7BwEAJ5xmFhUANSuDEnAQeibtALARyebAG7HRgCkAMwAQjcaAIJX51RcGt8Q70abDFIl2upxYk2eLxuAHYnq9NvDNsRMAQ5gwAZsbgAmB6bFibAD0my%2B33usMhUIAIhC/qdPt96RpfiCAUCabjwa9oWSEYjkaj0VicXjCcSGaSIdCKRxprROABWXiebi8VCcRybBSzeaYDGYu48UgETTS6YAaxAmKhADooVw7nqNF4uF5HV47nd9JxJIrjaRVRxeAoQBpDcbpnBYEg0Cw4nRIuRKNHY/QoshgFIuHw6ARIkGIGFfWFAtVlpwDUXmMRlgB5MLaMpG5WkaNsQTVhi0UscLSkLBhVzARxiWhBptYFiGYDibu8fBI8oXTCjnuYVRlVw5su8QI52Uz/R4MLEEvOLC%2BgjEPAsLekRfEMKJTAUzAToy0QKgGfTKgGYAKABqeCYAA7tWaxKga/CCCIYjsFIMiCIoKjqPuuiZgYRggKYxjmO%2BYRBpA0yoHENgCKOAC01Z3Cqd6XlgBEQNMTSkZ4EAOAMniZv4owFEUegJEkLEcfx2QsR0vHdJmzEVMMwlSZgVgsa0NTiV0URSbJLj1HoCjDKp4zqUxWoLBIMryj6%2B7%2BpsqheNE5HRJImzAMgyBEpIVpcJsEC4IQJC6vqky8I2WiTGaIB3HKVoaHc0QAJxypIkhyhokhcNEdxQtIe7eqQ17EqQSo9v6gbBqGX6kBGiAoKgMZxmQFAQEmdUoMgGHpk6IY0LQObEHmBb7hWJY3oNVa1vW1g3i2jAEO2na%2Bn2A5DrQI43uOk7Tj2c4Nngi7Lrwq7rpuTY7gpvp4ceVanosPYXleN53g%2BSjPq%2BU4fmGfC/gBQGgeBN5QcIojiPB/1IWovq6Ji%2BiTlhZgHvh8BESRKQUVRNGRHRS4IxYCnbXYbEME4WkZFxBP6XxmYCTkqRE5xWSCSkZOSdjikyW0cnM7jDDKSM%2BRqTpmnpLTultIzhkzHMJlcGZHAKgVvpWTZdkOZsrWTkSXgeVFXk%2BUQxD%2BVLQVhtMCCYEwWBRIxpDmhlVqYpIsVOpIUL6vbcr6p6HA5YVKqcCVIbBSaHuYhZRW%2B2VIXTHeSR2JIQA). This is a really strange regression.

vanhauser-thc:
It’s not a speed thing. As I wrote clock time is faster

vanhauser-thc:
you can read it up here: https://github.com/AFLplusplus/AFLplusplus/pull/2159

carpintero-de-c:
@vanhauser-thc Wasn't the whole problem that switching to `clock_gettime` caused a severe performance regression; because of which this was reverted? I'm confused.

PR #2160, opened 2024-07-13
Fixing change from PR#2152, misspelled variable names

exoosh:
Fixing the issue inadvertently introduced during PR #2152. I hadn't changed the names of two uses of variables whose name were misspelled with `TO` instead of `TOO` and @vanhauser-thc also mentioned he preferred `DEFAULT` over `LIMIT`.

Sorry folks!


vanhauser-thc:
ah classic :)

PR #2165, opened 2024-07-21
Fix FASAN operand access on aarch64

meowmeowxw:
Hello,
I've noticed that on android aarch64 the `operand->access` is equal to `CS_AC_WRITE | CS_AC_READ` (3) when there is a store instruction.
Without this patch a store instruction was wrongly considered as a load instruction.


vanhauser-thc:
@worksbutnottested

WorksButNotTested:
Capstone is probably defective in marking its memory operand as read and write. But I guess if we note both a read and write overflow at the same place, we should probably prioritise reporting the oob write. So the PR looks good. Thanks.

PR #2168, opened 2024-07-23
Fix LLVM mode on MacOS

domenukk:
If clang is system clang, but llvm-config is brew, this wouldn't compile since CC is set after the check for system clang.


vanhauser-thc:
well spotted :) now try to fix the rest of MacOS weirdness if you can :D

PR #2173, opened 2024-07-28
To support AFL instrumentation, add default settings of map_size in forkserver.

ahuo1:
Hello, when aflpp uses a binary instrumented by AFL, the binary's returned status variable is 0. 

Therefore, the bitmap size for aflpp is the default 8 /* 1024 /* 1024, which severely affects the running speed. 

For compatibility, I have added a default bitmap size setting in forkserver.( ◠‿◠ )

ps: I noticed that it can be resolved by setting AFL_SKIP_BIN_CHECK. However, after setting AFL_SKIP_BIN_CHECK, aflpp will no longer check persistent mode. Therefore, I think this setting is more reasonable.

### DIFF
--- a/src/afl-forkserver.c
+++ b/src/afl-forkserver.c
@@ -1338,6 +1338,10 @@ void afl_fsrv_start(afl_forkserver_t *fsrv, char **argv,
 
           fsrv->map_size = tmp_map_size;
 
+        } else {
+           
+           fsrv->real_map_size = fsrv->map_size = MAP_SIZE;
+           
         }
 
         if ((status & FS_OPT_AUTODICT) == FS_OPT_AUTODICT) {
@@ -1444,6 +1448,11 @@ void afl_fsrv_start(afl_forkserver_t *fsrv, char **argv,
 
         }
 
+      } else {
+
+         // The binary is most likely instrumented using AFL's tool, and we will set map_size to MAP_SIZE.
+         fsrv->real_map_size = fsrv->map_size = MAP_SIZE;
+         
       }
 
     }


vanhauser-thc:
wouldn't it be easier to just check if map_size is 0 and then set the MAP_SIZE value if so?

vanhauser-thc:
ping @ahuo1

ahuo1:
Oh, I'm very sorry. 
Would it be possible to add a check after the afl_fsrv_get_mapsize function, and if it equals DEFAULT_SHMEM_SIZE, then change the value of afl->fsrv.map_size to MAP_SIZE?

https://github.com/AFLplusplus/AFLplusplus/blob/db23931e7c1727ddac8691a6241c97b2203ec6fc/src/afl-fuzz.c#L2463

ahuo1:
Oh, I overlooked cmplog. I think it would be more convenient to add a check within the afl_fsrv_get_mapsize function.

https://github.com/AFLplusplus/AFLplusplus/blob/db23931e7c1727ddac8691a6241c97b2203ec6fc/src/afl-forkserver.c#L1692

vanhauser-thc:
> Oh, I overlooked cmplog. I think it would be more convenient to add a check within the afl_fsrv_get_mapsize function.
> 
> https://github.com/AFLplusplus/AFLplusplus/blob/db23931e7c1727ddac8691a6241c97b2203ec6fc/src/afl-forkserver.c#L1692

Just push the change and then I can review it :)

ahuo1:
> > Oh, I overlooked cmplog. I think it would be more convenient to add a check within the afl_fsrv_get_mapsize function.
> > https://github.com/AFLplusplus/AFLplusplus/blob/db23931e7c1727ddac8691a6241c97b2203ec6fc/src/afl-forkserver.c#L1692
> 
> Just push the change and then I can review it :)

Alright!

vanhauser-thc:
one thing so. what happens if the map size would be exactly DEFAULT_SHMEM_SIZE by chance?

ahuo1:
Uh, yeah. To deal with it, I think the first commit is more reasonable...

vanhauser-thc:
thank you

PR #2174, opened 2024-07-28
Fixed lsan defaults evaluation

killerra:
There is a [bug](https://github.com/AFLplusplus/AFLplusplus/issues/2176) in the evaluation of the lsan default parameters inside void set_sanitizer_defaults().
This is causing the following behavior:
```
pre-DEFAULT: ASAN_ENV: detect_leaks=false
pre-DEFAULT: LSAN_ENV: (null)
DEFAULTs: ASAN_ENV: detect_leaks=false
DEFAULTs: LSAN_ENV: exitcode=23:fast_unwind_on_malloc=0:print_suppressions=0:detect_leaks=1:malloc_context_size=0:
```

While detect_leaks is set to false in the LSAN parameters detect_leaks is set to 1 causing the leak detection being enabled. This prevents FASAN from correctly working for example.

Also removed a bogus free I've introduced with adding unstable coverage for any instance name in frida, because I didn't know getenv strings should not be free'ed.

### DIFF
--- a/src/afl-common.c
+++ b/src/afl-common.c
@@ -108,7 +108,7 @@ void set_sanitizer_defaults() {
     if (!have_san_options) { strcpy(buf, default_options); }
     if (have_asan_options) {
 
-      if (NULL != strstr(have_asan_options, "detect_leaks=false")) {
+      if (NULL != strstr(have_asan_options, "detect_leaks=0") || NULL != strstr(have_asan_options, "detect_leaks=false")) {
 
         strcat(buf, "exitcode=" STRINGIFY(LSAN_ERROR) ":fast_unwind_on_malloc=0:print_suppressions=0:detect_leaks=0:malloc_context_size=0:");
 
--- a/src/afl-common.c
+++ b/src/afl-common.c
@@ -108,7 +108,7 @@ void set_sanitizer_defaults() {
     if (!have_san_options) { strcpy(buf, default_options); }
     if (have_asan_options) {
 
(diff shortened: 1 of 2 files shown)


src/afl-common.c
```diff
@@ -108,7 +108,7 @@ void set_sanitizer_defaults() {
     if (!have_san_options) { strcpy(buf, default_options); }
     if (have_asan_options) {
 
-      if (NULL != strstr(have_asan_options, "detect_leaks=0")) {
+      if (NULL != strstr(have_asan_options, "detect_leaks=false")) {
```
vanhauser-thc: why this? it is detect_leaks=0 and false does not exist according to the documention https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer

vanhauser-thc:
sorry for taking so long, lots of other work

src/afl-common.c
killerra: It's set here https://github.com/AFLplusplus/AFLplusplus/blob/dev/src/afl-fuzz.c#L1938 and judging from this [documentation](https://github.com/google/sanitizers/wiki/SanitizerCommonFlags) it can be "true" or "false". I guess in the end it doesn't make a difference for ASAN itself.

killerra: So maybe use an integer value everywhere, or handle both cases? Also in the Frida mode documentation is "detect_leaks=false" mentioned.

vanhauser-thc:
thanks!

PR #2180, opened 2024-08-06
Fix `/bin/sh:` errors when building without LLVM

tchebb:
In `GNUmakefile.llvm`, several variables are unconditionally populated by expanding `$(LLVMVER)` and `$(LLVM_CONFIG)` inside shell commands. However, when LLVM is not present, both those variables are empty, meaning that the shell commands end up malformed and emit harmless, but noisy, errors like this one:

```
/bin/sh: --: invalid option
Usage:	/bin/sh [GNU long option] [option] ...
	/bin/sh [GNU long option] [option] script-file ...
GNU long options:
	--debug
	--debugger
	--dump-po-strings
	--dump-strings
	--help
	--init-file
	--login
	--noediting
	--noprofile
	--norc
	--posix
	--pretty-print
	--rcfile
	--rpm-requires
	--restricted
	--verbose
	--version
Shell options:
	-ilrsD or -c command or -O shopt_option		(invocation only)
	-abefhkmnptuvxBCHP or -o option
/bin/sh: line 1: test: -gt: unary operator expected
/bin/sh: line 1: test: -lt: unary operator expected
/bin/sh: line 1: test: -ge: unary operator expected
/bin/sh: line 1: test: -ge: unary operator expected
```

Fix the problem by only populating the "downstream" variables if the upstream ones have values.

This contribution is on behalf of my company.

### DIFF
--- a/GNUmakefile.llvm
+++ b/GNUmakefile.llvm
@@ -48,18 +48,25 @@ else
   LLVM_CONFIG ?= $(call detect_newest,llvm-config)
 endif
 
-override LLVM_RAW_VER        := $(shell $(LLVM_CONFIG) --version 2>/dev/null)
-LLVMVER                      := $(subst svn,,$(subst git,,$(LLVM_RAW_VER)))
-LLVM_MAJOR                   := $(firstword $(subst ., ,$(LLVMVER)))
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
LGTM thanks

PR #2181, opened 2024-08-06
Fix syntax error when compiling without zlib

tchebb:
commit ecb5854be08fa ("add zlib compression for fast resume"), introduced in #2141, added new logic selected at compile-time when zlib is present. Unfortunately, it also broke the existing logic by removing the last line of a multi-line if statement, resulting in a syntax error when zlib isn't present.

Restore the line as it was.

This contribution is on behalf of my company.


vanhauser-thc:
Thanks!

PR #2185, opened 2024-08-13
Fix get_binary_hash compatibility issue in nyx mode

nj00001:
The write_setup_file calculates the hash of the target, and the target path is obtained directly from afl->fsrv.target_path, however, in nyx mode, afl->fsrv.target_path is a directory, which will cause the calculation of the get_binary_hash to fail, because get_ binary_hash does not take into account the case where the argument is a folder. So it makes sense to pass the nyx target configuration to the get_binary_hash function for nyx compatibility reasons

bug fix before

image

### DIFF
--- a/src/afl-fuzz-stats.c
+++ b/src/afl-fuzz-stats.c
@@ -80,7 +80,21 @@ void write_setup_file(afl_state_t *afl, u32 argc, char **argv) {
 
   snprintf(fn2, PATH_MAX, "%s/target_hash", afl->out_dir);
   FILE *f2 = create_ffile(fn2);
-  fprintf(f2, "%p\n", (void *)get_binary_hash(afl->fsrv.target_path));
+
+  u64 target_hash = 0;
+  if (afl->fsrv.nyx_mode) {
+
+    char *tmp = alloc_printf("%s/config.ron", afl->fsrv.target_path);
+    target_hash = get_binary_hash(tmp);
+    free(tmp);
+
+  } else {
+
+    target_hash = get_binary_hash(afl->fsrv.target_path);
+
+  }
+
+  fprintf(f2, "%p\n", (void *)target_hash);
   fclose(f2);
 
   snprintf(fn, PATH_MAX, "%s/fuzzer_setup", afl->out_dir);


vanhauser-thc:
Thanks!
Note that this is not correct. You are hashing the config.ron file, however what you need to hash is the target binary that is set in the file. so you need to parse the ron file, get the right filename, and use that for hashing.

nj00001:
Unfortunately, the real target path information is not written inside the config.ron file. And you can't actually get the target path by parsing any config file, because nyx mode runs on a single config.ron file, and the real target is already installed in the VM!

this is a example

```

#![enable(implicit_some)]
(
    include_default_config_path: "/home/wu/Desktop/nyx-afl/AFLplusplus/nyx_mode/packer/packer/fuzzer_configs/default_config_vm.ron",
    runner: QemuSnapshot((
        qemu_binary: "/home/wu/Desktop/QEMU-Nyx/x86_64-softmmu/qemu-system-x86_64",
        hda: "/home/wu/Desktop/ubuntu-img/win10.qcow2",
        presnapshot: "/dev/shm/nyx-snap/",
        snapshot_path: DefaultPath,
        debug: false,
    )
  ),
  fuzz: (
    write_protected_input_buffer: false,
    input_buffer_size: 131072,
    workdir_path: "/tmp/workdir",
          mem_limit: 6000,          //snapshot_placement: none,
            dict: [
            ],
      ip0: (
        a: 0,
        b: 0,
      ),
      ip1: (
        a: 0,
        b: 0,
      ),
      ip2: (
        a: 0,
        b: 0,
      ),
      ip3: (
        a: 0,
        b: 0,
      ),
  ),
)
  
```

Refer to the following links
https://github.com/nyx-fuzz/Nyx/blob/main/docs/01-Nyx-VMs.md

vanhauser-thc:
hmm the thing is - if there would be a different ron hash but actually the same binary fuzzed then calibration would happen again.
if the same config.ron is used but the target binary has changed then you are fucking up you fuzzing campaign.

so - either you find a way to extract the hash of the target binary (you could popen("special_nyx_tool_script.sh") and read the hash from it in afl-fuzz nyx) - or this will not be possible.

nj00001:
In the case of using nyx full VMs, there doesn't seem to be a way to get the hash value of the binary program inside the VM from the outside. But if only nyx kernel mode is handled, then it can be obtained from fuzz.sh. So it's a bit hard to achieve if you want to be fully compatible with all modes of nyx

vanhauser-thc:
@schumilo do you have an idea how to do that?

vanhauser-thc:
@nj00001 @schumilo any idea how to get this working safely?
if the solution cannot extract the binary hash for all possible solutions that is not a problem IMHO, then it would still work with some/most setups.
but we must work with the hash of the program to be fuzzed.
maybe hash the qcow disk?

schumilo:
Depending on the mode, a qcow disk is not always required / used.  
 
I have two ideas which both, however, would require extending the packer:

1) The packer adds the name or path of the actual target executable to the config.ron, and then AFL does the hashing.

2) Let the packer generate the hash upon generating a Nyx dir and then add the resulting hash to the config.ron file. This way, we do the hash generation only once. This way, we can also include other executables and files that are part of the packed target (such as the harness and tools).

Both options should work fine for most users. And since a packed target is generally automatically generated and no one hopefully replaces the target executable while keeping the old packed dir, I think this approach would be fine. 
For anyone else choosing to use a custom handwritten harness with a manually crafted Nyx dir, we might still want to have a fallback option or just throw an error and require that a path or hash is specified in the config file (which seems reasonable to me).

vanhauser-thc:
I like option 2! :)

nj00001:
https://github.com/AFLplusplus/AFLplusplus/pull/2207#issue-2525932723
This solution is good enough, the automatic calculation of the hash and the manual filling of the hash value is enough to cover all situations

PR #2193, opened 2024-08-21
Offer more explicit core dump handling tip

ncoghlan:
Tweak the core dump handling tip to explicitly allow restoration
of the system's default core dump handling without restarting.

Also switch testing -> experimenting (since even full fuzzing
runs are still a testing activity).

### DIFF
--- a/src/afl-fuzz-init.c
+++ b/src/afl-fuzz-init.c
@@ -2455,12 +2455,7 @@ void check_crash_handling(void) {
         "    To avoid having crashes misinterpreted as timeouts, please \n"
         "    temporarily modify /proc/sys/kernel/core_pattern, like so:\n\n"
 
-        "    core_cmd=\"$(cat /proc/sys/kernel/core_pattern)\"\n"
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
I appreciate the thought. IMHO this will not really help because people will likely forget in which shell/terminal this was set etc. and people should know how to reset it, and if not it is no biggy IMHO.

ncoghlan:
Yeah, I was originally just going to suggest the `>` -> `| sudo tee` change in order to drop the wording about logging in as root.

Adding the part about saving/restoring was then more the result of having just run `cat /proc/sys/kernel/core_pattern` myself. I'll trim it down to remove those extra lines.

ncoghlan:
I'm sightly ambivalent about the change myself, as the current wording did make me stop and think about the potential unintended consequences of adjusting the core dump handling. The `sudo` version, by contrast, is open to copy-and-pasting if passwordless sudo is active (even if just via credential caching).

So no worries if you decide to keep the current wording and just close this PR.

vanhauser-thc:
I think this is good, thanks

PR #2086, opened 2024-05-12
Add `AFL_SHA1_FILENAMES` option

smoelius:
Fixes #2077.

Nits are welcome. For example, I wasn't sure if the prototypes in `afl-fuzz.h` had a prescribed order, so I just added the new functions at the end.

Also, I've tested these changes locally, but the PR contains no tests. What tests would you like me to add?

PR #2202, opened 2024-09-05
Use `permissive_create` in `mark_as_variable`

smoelius:
This appears to be a case I missed in #2086.

I have not experienced failures running locally, but I have in CI: https://github.com/trailofbits/test-fuzz/actions/runs/10282678713/job/28454887232#step:11:1988

I'll propose the same offer I did before: would you like me to write some tests for the `AFL_SHA1_FILENAMES` option?

### DIFF
--- a/src/afl-fuzz-queue.c
+++ b/src/afl-fuzz-queue.c
@@ -367,9 +367,8 @@ void mark_as_variable(afl_state_t *afl, struct queue_entry *q) {
 
   if (symlink(ldest, fn)) {
 
-    s32 fd = open(fn, O_WRONLY | O_CREAT | O_EXCL, DEFAULT_PERMISSION);
-    if (fd < 0) { PFATAL("Unable to create '%s'", fn); }
-    close(fd);
+    s32 fd = permissive_create(afl, fn);
+    if (fd >= 0) { close(fd); }
 
   }
 


vanhauser-thc:
LGTM, thanks.

about the test case - it is something just for your ensemble fuzzer framework. I do not need a test case. if you would like to have one so to ensure you are not affected by a change, sure push a test. note though that it requires a different compiled AFL++ so how to do this so it does not interfere with the real compiled afl-fuzz is quite some work, and will be extra work if something is changed how afl-fuzz is compiled.

vanhauser-thc:
fyi I pushed to stable

smoelius:
Thanks very much for merging and for pushing to stable. 🙏

> note though that it requires a different compiled AFL++

Sorry, I am not following. Why would a test of `AFL_SHA1_FILENAMES` require a different compiled AFL++?

FWIW, my intent would be to verify that AFL++ works under "normal usage" (say, finding a crash in a simple program) with `AFL_SHA1_FILENAMES` enabled.

vanhauser-thc:
ah sorry you are right, it is an env var, not a compile define

smoelius:
Is there an existing "finding a crash in a simple program" (or something similar) test that I might modify?

vanhauser-thc:
there are:

```
$ grep /crash *
test-custom-mutators.sh:    test -n "$( ls out/default/crashes/id:000000* 2>/dev/null )" && {  # TODO: update here
test-custom-mutators.sh:    test -n "$( ls out/default/crashes/id:000000* 2>/dev/null )" && {  # TODO: update here
test-custom-mutators.sh:      test -n "$( ls out/default/crashes/id:000000* 2>/dev/null )" && {  # TODO: update here
test-llvm.sh:    test -n "$( ls out/default/crashes/id:* 2>/dev/null )" && {
test-llvm.sh:    test -n "$( ls out/default/crashes/id:000000* out/default/hangs/id:000000* 2>/dev/null )" && {
```

PR #2205, opened 2024-09-09
Update env_variables.md

ktpss95112:
Update the docs so that it is consistent with the source code.

https://github.com/AFLplusplus/AFLplusplus/blob/stable/include/config.h#L326-L329


vanhauser-thc:
thank you

PR #2206, opened 2024-09-10
Add `AFL_SHA1_FILENAMES` test

smoelius:
Adds a test as discussed in #2202.

The approach is to run the cmplog test twice, once as normal and once with `AFL_SHA1_FILENAMES` enabled.

I would be happy to adjust if this approach is undesirable.

### DIFF
--- a/test/test-llvm.sh
+++ b/test/test-llvm.sh
@@ -269,9 +269,9 @@ test -e ../afl-clang-fast -a -e ../split-switches-pass.so && {
     {
       mkdir -p in
       echo 00000000000000000000000000000000 > in/in
-      AFL_BENCH_UNTIL_CRASH=1 ../afl-fuzz -Z -l 3 -m none -V30 -i in -o out -c 0 -- ./test-cmplog >>errors 2>&1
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
please only one test, not two, otherwise it delays the CI further.
note that if you just test for the existance of files you need AFL_NO_CRASH_README as otherwise the README will be picked up as a false positive...

smoelius:
I'm sorry for being unclear. Technically, there is only one new test. It is the one that produces these log lines:
```text
[*] running afl-fuzz for llvm_mode cmplog with AFL_SHA1_FILENAMES, this will take approx 10 seconds
[+] afl-fuzz is working correctly with llvm_mode cmplog with AFL_SHA1_FILENAMES
```
The loop on on lines 270-290 has this form:
```sh
for sha1_filenames in false true; do ( ... ) done
```
When `sha1_filenames` is false, the loop body should be equivalent to the original cmplog test. The "new" test is when `sha1_filenames` is true.

I was trying to make the changes minimally invasive. But I would be happy to refactor them in any way you prefer.

vanhauser-thc:
Yes I understood you. What I meant to say - please integrate your test into the existing one, do not add an additional one.

smoelius:
Assuming CI passes, please let me know if you would like any additional changes.

vanhauser-thc:
LGTM, thanks

smoelius:
Thank you!

PR #2207, opened 2024-09-14
fix AFL_AUTORESUME=1 for Nyx mode

schumilo:
Pretty much option 2, as discussed here: https://github.com/AFLplusplus/AFLplusplus/pull/2185

The packer creates now a file in the Nyx share dir named `TARGET_HASH`, which holds a hexadecimal encoded 160bit SHA1 hash string. The hash is generated based on all ELF dependencies of the target executable, including the target executable itself. There is also a `--hash_only_target` option to only use the target executable as input for generating the hash.  

Libnyx now exposes a few more functions, of which `nyx_get_target_hash64(cfg)` is probably the most interesting. Basically, it returns either 0 in case no `TARGET_HASH` file was found (or due to any other error) or the first 64 bits of the 160-bit `TARGET-HASH` are returned. 

AFL++ uses the value return by `nyx_get_target_hash64` as an identifier for the fast resume feature. Currently, the fast resume feature is disabled if the hash is zero (and nyx_mode is enabled). But for this case, we might want to print a fancy note to the user, as this usually only happens if there is no `TARGET_HASH` file found in the Nyx share dir.


vanhauser-thc:
Thank you!

PR #2209, opened 2024-09-16
Update fuzzing_in_depth.md for a typo

behnamben:
Documents section 3-c, it is said: "But note that can slow down the start of the first fuzz by quite a lot {of} you have many fuzzers and/or many seeds."
I think {of} should be {if}.


domenukk:
Thank you! I've switched this PR to point to the dev branch which is our evolving branch :)

PR #2212, opened 2024-09-23
Save crash log returned by nyx when AFL_CRASHING_SEEDS_AS_NEW_CRASH e…

nj00001:
Save crash log returned by nyx when AFL_CRASHING_SEEDS_AS_NEW_CRASH environment variable is enabled
nyx also returns a crash log when reporting a crash, both files are saved in the afl++ fuzzing run, whereas with the AFL_CRASHING_SEEDS_AS_NEW_CRASH setting, only the crash file is saved in the dry run phase if crashes are included in the supplied seeds.
This commit adds crash log saving to keep the behavior consistent


vanhauser-thc:
LGTM thanks

PR #2218, opened 2024-10-02
Fix uninitialized alloc_canary in libdislocator

ea:
When random alloc_canary env var option was introduced, a possibility for use of uninitialized alloc_canary value was made. 

In most cases, constructor will be called during shared library load and the alloc_canary would be initialized to either its default value or a randomly generated one if forced by AFL_RANDOM_ALLOC_CANARY env var.

However, in some cases, libraries loaded before libdislocator will make allocations (still using libdislocator's allocation functions) while alloc_canary is still uninitialized. In such cases, canary value is usually NULL.  If such allocated value is then free()'d after libdislocator's constructor has been run, call to free() will fail causing a false positive. This condition usually happens while calling library destructors at process termination. 

The patch ensures the canary value is initialized in all cases, and introduces a destructor that reverts it to default value. 

This  does mean that certain number of early allocations will use the default canary value rather than the random one set afterwards.  This seems like a reasonable tradeoff as I haven't found a surefire way of forcing libdislocator's constructor to run first in all possible cases (if nothing else, libphtread usually  has priority).


domenukk:
Thanks!
I don't 100% understand what the destructor does? Do all `mallocs` that happen before our initializer then run with the reset / post-destructor `free`?

P.S.: please open PRs towards `dev`, changed :)

ea:
mallocs that happen before the init will still use libdislocator's malloc, but with the default value for `alloc_canary`.
If the destructor wasn't there to restore the value to default (if a random one was used), when free would be called on pointers that were allocated with the default value (prior to init) it would crash because it wouldn't see the expected canary value. By restoring it to default one, the frees would again see the expected (default) value. 
This assumes that the order of library destructors follows the order of constructors, just in reverse, which seems to be the case. 
There might be a more general solution, but it is a tricky problem.

domenukk:
I mean, in that case the only safe way to do it is to not set `AFL_RANDOM_ALLOC_CANARY` and get the default value, right?

domenukk:
So _maybe_ the destructor isn't all too useful, but just a random thought.

ea:
I'm not sure what you are suggesting. Either you have the destructor reset the value back to default during unloading, or you will get false positive crashes if AFL_RANDOM_ALLOC_CANARY was set and a library makes allocations during it's constructor execution.

Regardless, the initial issue of  `alloc_canary` being uninitialized would manifest itself as a false positive crash no matter if AFL_RANDOM_ALLOC_CANARY was used or not.

vanhauser-thc:
@ea yes makes sense, thank you!

PR #2226, opened 2024-10-15
push to stable

vanhauser-thc:


docs/Changelog.md
```diff
@@ -35,6 +37,8 @@
     - new runtime (!) variable: `AFL_OLD_FORKSERVER` to use the old vanilla
       AFL type forkserver. Useful for symcc/symqemu/nautilus/etc. with
       AFL_LLVM_INSTRUMENT=CLASSIC
+    - new compile time variable: `AFL_OPT_LEVEL` to set a specific optimization
```
domenukk, line 40: What's the use case? Build systems that somehow don't support release builds?

vanhauser-thc: It is needed for rust (afl.rs) and has this name, so I added it to afl-cc too

domenukk (approved):

PR #2232, opened 2024-10-29
Add JSON Schema dictionary

michaelmior:
I'm working on using AFL++ with JSON Schema and these keywords are a useful supplement to the existing JSON dictionary.


vanhauser-thc:
thank you!

PR #2241, opened 2024-11-08
Add support for post_process in Rust custom mutator

jma-qb:
Hi, I added support for the post_process function in the Rust custom mutator bindings.
The existing examples still work and there is a new example, derived from the lain one, using that function.
I also changed lain to reference the fork hosted under AFL++ rather than the crates.io version.

### DIFF
--- a/custom_mutators/rust/README.md
+++ b/custom_mutators/rust/README.md
@@ -5,7 +5,15 @@ Bindings to create custom mutators in Rust.
 These bindings are documented with rustdoc. To view the documentation run
 ```cargo doc -p custom_mutator --open```.
 
-A minimal example can be found in `example`. Build it using `cargo build --example example_mutator`. 
+A minimal example can be found in `example`. Build it using `cargo build --example example_mutator`.
 
 An example using [lain](https://github.com/microsoft/lain) for structured fuzzing can be found in `example_lain`.
 Since lain requires a nightly rust toolchain, you need to set one up before you can play with it.
+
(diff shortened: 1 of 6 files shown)


domenukk:
Neat, thank you!

custom_mutators/rust/README.md
```diff
@@ -5,7 +5,15 @@ Bindings to create custom mutators in Rust.
 These bindings are documented with rustdoc. To view the documentation run
 ```cargo doc -p custom_mutator --open```.
 
-A minimal example can be found in `example`. Build it using `cargo build --example example_mutator`. 
+A minimal example can be found in `example`. Build it using `cargo build --example example_mutator`.
 
 An example using [lain](https://github.com/microsoft/lain) for structured fuzzing can be found in `example_lain`.
 Since lain requires a nightly rust toolchain, you need to set one up before you can play with it.
+
+An example for the use of the post_process function, using [lain](https://github.com/microsoft/lain) with [serde](https://github.com/serde-rs/serde) and [bincode](https://github.com/bincode-org/bincode) can be found in `example_lain_post_process`.
```
domenukk, line 13: I would personally take a look use flexbuffers instead of bincode, I assume it should be faster

domenukk: (Not benchmarked that myself, though)

jma-qb: Thanks I didn't know about this one, I'll look into it some day :)
So far out of all the solutions I tested I only managed to make bincode work. Lain's serialization function lose some information so my workaround was to serialize with bincode in the fuzz function, deserialize it in post_process to serialize it again in the right way for the target (and use AFL_POST_PROCESS_KEEP_ORIGINAL and AFL_DISABLE_TRIM)...
So the performance is probably terrible but not an issue compared to emulation with unicorn (my usecase).

PR #2242, opened 2024-11-11
Unsetting AFL_LLVM_ALLOWLIST and AFL_LLVM_DENYLIST environment variables in make files

abhigargrepo:
Instrumentation tests, after building AFLplusplus, fails due to AFL_LLVM_ALLOWLIST and AFL_LLVM_DENYLIST environment variables being set in the user environment.
These environment variables hinder the instrumentation tests by not allowing test files to be instrumented by AFLplusplus.
It throws following warnings
     **[!] WARNING: No instrumentation targets found.**
     **Oops, the instrumentation of afl-cc does not seem to be behaving correctly!**
Unsetting these variables in MakeFiles will mitigate such warnings and the keep the build process clean.


vanhauser-thc:
cannot hurt :)

Issue #2244, opened 2024-11-15
Failed to compile argv_fuzz_persistent_demo  with error "undeclared identifier '__AFL_FUZZ_TESTCASE_BUF'"


fripSide:
**IMPORTANT**
1. You have verified that the issue to be present in the current `dev` branch.
2. Please supply the command line options and relevant environment variables,
   e.g., a copy-paste of the contents of `out/default/fuzzer_setup`.

Thank you for making AFL++ better!

**Describe the bug**

When I compile the stable and dev (4.22a) branches of AFLplusplus, it encounters a compile error for the [argv_fuzz_persistent_demo.c](https://github.com/AFLplusplus/AFLplusplus/blob/stable/utils/argv_fuzzing/argv_fuzz_persistent_demo.c). 
But it does not prevent us from building all the other binaries.

The error message is:
```
[!] WARNING: You are using outdated instrumentation, install LLVM and/or gcc-plugin and use afl-clang-fast/afl-clang-lto/afl-gcc-fast instead!
argv_fuzz_persistent_demo.c:10:1: warning: data definition has no type or storage class
   10 | __AFL_FUZZ_INIT();
      | ^~~~~~~~~~~~~~~
argv_fuzz_persistent_demo.c:10:1: warning: type defaults to 'int' in declaration of '__AFL_FUZZ_INIT' [-Wimplicit-int]
argv_fuzz_persistent_demo.c: In function 'main':
argv_fuzz_persistent_demo.c:23:24: error: '__AFL_FUZZ_TESTCASE_BUF' undeclared (first use in this function)
   23 |   unsigned char *buf = __AFL_FUZZ_TESTCASE_BUF;
      |                        ^~~~~~~~~~~~~~~~~~~~~~~
argv_fuzz_persistent_demo.c:23:24: note: each undeclared identifier is reported only once for each function it appears in
argv_fuzz_persistent_demo.c:28:10: warning: implicit declaration of function '__AFL_LOOP' [-Wimplicit-function-declaration]
   28 |   while (__AFL_LOOP(100000)) {
      |          ^~~~~~~~~~
argv_fuzz_persistent_demo.c:30:15: error: '__AFL_FUZZ_TESTCASE_LEN' undeclared (first use in this function)
   30 |     int len = __AFL_FUZZ_TESTCASE_LEN;
      |               ^~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #2245, opened 2024-11-15
Fix the issue#2244

fripSide:
Address the [issue#2244](https://github.com/AFLplusplus/AFLplusplus/issues/2244) to ensure that `argv_fuzz_persistent_demo.c` can be successfully compiled without any error messages


vanhauser-thc:
that is not a fix. yes it does not error anymore, but the issue is that you compile something that needs afl-clang/gcc-fast/lto with outdated afl-gcc/afl-clang. and even if it compiles, it will not work correctly because persistent fuzzing does not work with afl-gcc/afl-clang

fripSide:
Does it work with `afl-clang-fast` ?  Just like [these demos](https://github.com/AFLplusplus/AFLplusplus/tree/stable/utils/persistent_mode).

vanhauser-thc:
> Does it work with `afl-clang-fast` ? Just like [these demos](https://github.com/AFLplusplus/AFLplusplus/tree/stable/utils/persistent_mode).

if you have llvm-dev installed so AFL++ could compile its plugins - sure. just enter "afl-cc" and you see in the help output which instrumentations could be compiled

fripSide:
> > Does it work with `afl-clang-fast` ? Just like [these demos](https://github.com/AFLplusplus/AFLplusplus/tree/stable/utils/persistent_mode).
> 
> if you have llvm-dev installed so AFL++ could compile its plugins - sure. just enter "afl-cc" and you see in the help output which instrumentations could be compiled

The warning message (WARNING: You are using outdated instrumentation) caused by afl-gcc is not related to the current compilation errors. The compilation errors stem from undefined macros, rather than whether you have successfully instrumented.

```
// https://github.com/AFLplusplus/AFLplusplus/blob/stable/utils/argv_fuzzing/argv_fuzz_persistent_demo.c#L10
__AFL_FUZZ_INIT(); // <-------------- __AFL_FUZZ_INIT is not defined

// https://github.com/AFLplusplus/AFLplusplus/blob/stable/utils/argv_fuzzing/argv_fuzz_persistent_demo.c#L23
unsigned char *buf = __AFL_FUZZ_TESTCASE_BUF; // <-------------- __AFL_FUZZ_TESTCASE_BUF is not defined

// https://github.com/AFLplusplus/AFLplusplus/blob/stable/utils/argv_fuzzing/argv_fuzz_persistent_demo.c#L30
int len = __AFL_FUZZ_TESTCASE_LEN; // <-------------- __AFL_FUZZ_TESTCASE_LEN is not defined

```

I do not understand the rationale for closing my issue and pull request while the compiler error remains unresolved. If this demo is no longer needed, it would be better to remove these problematic files (i.e., `argv_fuzz_persistent_demo.c`) from the repository.

PR #2246, opened 2024-11-20
Fix u64's width to exactly 64 bits

McSinyx:
Because consistency is pretty.


vanhauser-thc:
there are platforms that do not have uint64_t - do not ask me which ones exactly but this was an issue that needed to be fixed.

McSinyx:
Thanks, TIL, interesting!  (If anybody recalls the specifics, please share them with me.)

Issue #2247, opened 2024-11-20
UBSAN deprecated "-fsanitize-undefined-trap-on-error" flag alias


StepanGulyaev:
Hello, me and @Fahnenfluchtige have been researching on UBSAN mode in AFL++ and we've found that UBSAN mode is currently using deprecated flag "-fsanitize-undefined-trap-on-error". 

We've also found recent discussions on topic here https://gcc.gnu.org/bugzilla/show_bug.cgi?id=109489 and here https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=2c7cfc7b418564. That issue still appears to be in a dev branch:

https://github.com/AFLplusplus/AFLplusplus/blob/82752fe38db590184f7f1725954f906da397d91c/src/afl-cc.c#L1951

It seems we can just use "-fsanitize-trap=undefined" for same results, but this approach will give us more space when it comes to filtering different types of UBSAN errors, because it gives us freedom to choose arguments. But we've found even better way to manage UBSAN output.

In the UBSAN related AFL documentation (https://github.com/mirrorer/afl/blob/master/docs/notes_for_asan.txt - the really old one, before  UBSAN support even was in AFL) (https://aflplus.plus/docs/notes_for_asan/#5-what-about-ubsan - the newer one) we've also found that inserting "SIGILL" approach comes from inability of UBSAN killing process and give us a crash we needed. Also it didn't give us a verbose message about place of the crash. 

But recent tests on AFL++ 4.22a with afl-clang-lto v14.0.6 have shown that we can get both crash and verbose message using "-fno-sanitize-recover=undefined" or "-fno-sanitize-recover=all" . It seems to be already in use next line

https://github.com/AFLplusplus/AFLplusplus/blob/82752fe38db590184f7f1725954f906da397d91c/src/afl-cc.c#L1952

But  "-fsanitize-undefined-trap-on-error" messes with it and sends "SIGILL" instead.

For example, we take a primitive program, that may or may not trigger UBSAN:
![изображение](https://github.com/user-attachments/assets/2e6fc169-2435-4b01-a198-5dd5decbfdd8)

Firstly we compile it with AFL_USE_UBSAN=1:
![изображение](https://github.com/user-attachments/assets/b76cc876-8ea0-4add-be13-48293bfb9a24)

We get the crash:
![изображение](https://github.com/user-attachments/assets/481581e8-067f-4d86-bf31-1bc52dbe597a)

And reproduce it:

[...]

PR #2252, opened 2024-11-21
Added CFISAN option for verbose output on crash

StepanGulyaev:
Hello. Me and @Fahnenfluchtige added compile option to CFISAN for it to print verbose output on crash instead of "Illegal Instruction". It's pretty similar to the thing we wrote about UBSAN yesterday: https://github.com/AFLplusplus/AFLplusplus/issues/2247. Program crashes as it should and also prints place where during the crash reproduction:

изображение

изображение


vanhauser-thc:
Adding verbose output to a crash that is compiled for fuzzing will impact the fuzzing speed (each crash will take longer).
this should only be activated with a special env var (eg. `AFL_VERBOSE_SANITIZERS`) at compile time.

StepanGulyaev:
Oh, I see. Thank you, we will work on that. I think we should do it only for CFISAN for now.

StepanGulyaev:
Also shouldn't we have same approach with VERBOSE variable on UBSAN? After fix yesterday it always prints verbose output because of -fno-sanitize-recover=all, maybe we should use "-fsanitize-trap=undefined" if VERBOSE variable not set

https://github.com/AFLplusplus/AFLplusplus/blob/3af042d5bf3116a194168f5d1e40311a484959e8/src/afl-cc.c#L1951

vanhauser-thc:
can you please add the env description to docs/env_variables.md ? thank you!

StepanGulyaev:
We'll surely add. Thank you

vanhauser-thc:
thank you!

PR #2257, opened 2024-11-27
Update afl-analyze.c

S0fr:
## To fix the issue with fields splitting in the afl-analysis.

When using afl-analyze in AFL++, it fails to properly divide the fields of the file structure in the same way that AFL's afl-analyze does.
This leads to incorrect analysis results in subsequent steps.

The cause of the issue is that the type declarations for prev_xff, prev_x01, prev_s10, and prev_a10 were directly copied over.
In reality, the hash in AFL++ should be u64 instead of u32.


vanhauser-thc:
Thank you!

PR #2264, opened 2024-12-03
Bump LLVM version to 20 available in prerelease, 19 being latest stable

exoosh:
There are plenty of strings referring to versions older than the stated version minimum of 13. Should these also be updated at some point? I.e. in `GNUmakefile` and `GNUmakefile.llvm` references to "ancient" versions of LLVM exist.

Anyone able to test the build on OpenBSD? It's one of the versions which gives a very very low version as minimum in its output strings and related comments ...

### DIFF
--- a/GNUmakefile.llvm
+++ b/GNUmakefile.llvm
@@ -32,7 +32,7 @@ VERSION     = $(shell grep '^$(HASH)define VERSION ' ./config.h | cut -d '"' -f2
 
 SYS = $(shell uname -s)
 
-override LLVM_TOO_NEW_DEFAULT := 18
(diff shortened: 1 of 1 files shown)


exoosh:
Side-note:
```
ifeq "$(SYS)" "OpenBSD"
  CLANG_LFL += $$($(LLVM_CONFIG) --libdir)/libLLVM.so
```
could an should probably be replaced by:
```
ifeq "$(SYS)" "OpenBSD"
  CLANG_LFL += $(LLVM_LIBDIR)/libLLVM.so
```
because the value is already stored in a variable at the top of the `GNUmakefile.llvm`.

vanhauser-thc:
thank you!

PR #2265, opened 2024-12-04
push to stable

vanhauser-thc:


GNUmakefile.llvm
```diff
@@ -69,7 +69,7 @@ endif
 
 LLVM_STDCXX                  := gnu++11
 LLVM_LTO                     := 0
-LLVM_UNSUPPORTED             := $(shell echo "$(LLVMVER)" | grep -E -q '^[0-2]\.|^3\.[0-7]\.' && echo 1 || echo 0)
+LLVM_UNSUPPORTED             := $(shell echo "$(LLVMVER)" | grep -E -q '^[0-2]\.|^3\.[0-7]\.|^2[0-9]\.' && echo 1 || echo 0)
```
glandium, line 72: This change makes the makefile actively reject LLVM 20.

vanhauser-thc: Because the plugins currently don’t work with 20

PR #2273, opened 2024-12-17
Fix macro syntax error in config.h

vnc0:
Fixed invalid `#define 0666` by specifying the correct macro name `DEFAULT_PERMISSION`.


vanhauser-thc:
Damn thanks :)

PR #2276, opened 2024-12-19
Fix overflow in execs_ps_last_min calculation

### LINKED ISSUE #2275: Overflow in `execs_ps_last_min` calculation
I was looking at statistics of a long running job (I was out of office for over a month and the PC kept on fuzzing...), and noticed unusual numbers for `execs_ps_last_min`. It showed 71574927.35 executions in the last minute.

When looking through the code, it became clear that there is an overflow once the total number of executions gets bigger than 2^32. From https://github.com/AFLplusplus/AFLplusplus/blob/dev/src/afl-fuzz-stats.c#L342-L345

```cpp
    afl->last_avg_execs_saved =
        (double)(1000 * (afl->fsrv.total_execs - afl->last_avg_execs)) /
        (double)(cur_time - afl->last_avg_exec_update);
    afl->last_avg_execs = afl->fsrv.total_execs;
    afl->last_avg_exec_update = cur_time;
```

Here `last_avg_execs` is an `u32`, but it gets the `afl->fsrv.total_execs` assigned which is a 64bit value. In my case, the total number of executions is already 4867595546, so it's over the 2^32.

So I'd say this can probably be fixed by changing `last_avg_execs` to u64  [here](https://github.com/AFLplusplus/AFLplusplus/blob/dev/include/afl-fuzz.h#L750).  And maybe give it a more fitting name, since this isn't the average but the previous `total_execs`

martinus:
last_avg_execs should be 64bit, same as total_execs, otherwise there is an overflow once total_execs reaches 2^32. Which can happen in practice for long-running fuzzing campaigns.

Also renames `last_avg_execs` -> `last_avg_total_execs`

Fixes #2275

### DIFF
--- a/include/afl-fuzz.h
+++ b/include/afl-fuzz.h
@@ -747,7 +747,7 @@ typedef struct afl_state {
                                   up to 256 */
 
   unsigned long long int last_avg_exec_update;
-  u64                    last_avg_execs;
+  u64                    last_avg_total_execs;
   double                 last_avg_execs_saved;
 
 /* foreign sync */
--- a/include/afl-fuzz.h
+++ b/include/afl-fuzz.h
@@ -747,7 +747,7 @@ typedef struct afl_state {
                                   up to 256 */
 
   unsigned long long int last_avg_exec_update;
-  u32                    last_avg_execs;
+  u64                    last_avg_execs;
   double                 last_avg_execs_saved;
 
 /* foreign sync */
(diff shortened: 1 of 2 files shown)


martinus:
Tested it locally with my long running campaign. Here is an excerpt of `fuzzer_stats` before the fix:

```
execs_done        : 4341714917
execs_per_sec     : 22624.83
execs_ps_last_min : 71405411.68
```

And after the fix:

```
execs_done        : 4343186524
execs_per_sec     : 22625.47
execs_ps_last_min : 24116.80
```

vanhauser-thc:
Thank you!

PR #2280, opened 2024-12-26
Fix macOS env optimization

GAP-dev:
### Conditional DYLD_INSERT_LIBRARIES Setting for macOS Compatibility

Previously, the code unconditionally applied the DYLD_INSERT_LIBRARIES setting, even on non-macOS platforms. This could lead to unnecessary operations on platforms where this setting is irrelevant.

This PR the following changes:

1.	Wrapped the DYLD_INSERT_LIBRARIES setting in a #ifdef __APPLE__ directive to ensure it is applied only on macOS.
2.	Preserved the existing LD_PRELOAD setting for all platforms without modification.

### Purpose
- Ensures compatibility with macOS while avoiding unnecessary settings on other platforms.

### DIFF
--- a/src/afl-analyze.c
+++ b/src/afl-analyze.c
@@ -695,14 +695,17 @@ static void set_up_environment(char **argv) {
       ck_free(frida_binary);
 
       setenv("LD_PRELOAD", frida_afl_preload, 1);
+#ifdef __APPLE__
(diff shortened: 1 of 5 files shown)


vanhauser-thc:
thanks :)

PR #2281, opened 2025-01-02
Mutation Chain Tool

maartendekker1998:
Adding a tool I developed to trace the mutation chain of AFL crash files. I needed it to determine some statistics of custom mutators. Example outputs can be found [here](https://github.com/maartendekker1998/easycwmp_fuzzing_results/blob/main/crash_chains_rpc_request.json).

### DIFF
--- a/utils/mutation_chain/mutation_chain.py
+++ b/utils/mutation_chain/mutation_chain.py
@@ -6,8 +6,8 @@
 # Now you can!
 #
 # This tool is developed to support file structures for parallel fuzzing runs using the
-# naming of main/secondary as stated in the AFL docs (fuzzer01, fuzzer02 etc...)
-# In case you want to use it for single instance runs just recreate the directory structure
+# naming of main/secondary nodes as stated in the AFL docs (fuzzer01, fuzzer02 etc...)
+# In case you want to use it for single node runs just recreate the directory structure
 # which is used when parallel fuzzing is used (dump your results in a dir called fuzzer01).
 #
 # author: Maarten Dekker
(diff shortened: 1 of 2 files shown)


utils/mutation_chain/mutation_chain.py
```diff
@@ -0,0 +1,93 @@
+#
+# MUTATION CHAIN COMPUTATION TOOL
+#
+# Ever wondered what the complete history of your AFL crash file looks like?
+# Now you can!
+#
+# This tool is developed to support file structures for parralel fuzzing runs using the
+# naming of master/slaves as stated in the AFL docs (fuzzer01, fuzzer02 etc...)
```
domenukk: main/secondary

utils/mutation_chain/mutation_chain.py
```diff
@@ -0,0 +1,93 @@
+#
```
domenukk, line 2: Should probably have a `#!/usr/bin/env python3`

utils/mutation_chain/mutation_chain.py
```diff
@@ -0,0 +1,93 @@
...
+# MUTATION CHAIN COMPUTATION TOOL
+#
+# Ever wondered what the complete history of your AFL crash file looks like?
+# Now you can!
+#
+# This tool is developed to support file structures for parralel fuzzing runs using the
```
domenukk: parallel

utils/mutation_chain/mutation_chain.py
```diff
@@ -0,0 +1,93 @@
...
+# 
+# Best to echo the output of this tool into a json file for further analysis.
+#
+# author: Maarten Dekker
+
+# import required module
+import os, re, json
+
+# user config section
+afl_output_directory = "INSERT_YOUR_AFL_OUTPUT_DIRECTORY"
```
domenukk: This should be a commandline parameter

utils/mutation_chain/mutation_chain.py
```diff
@@ -0,0 +1,93 @@
...
+
+            elif re.match(file_we_look_for_2_rex, k):
+                retval[k] = compute_mutation_chain(k, current_fuzzer, n+1)
+
+        return retval
+
+    else:
+        return "Reached original"
+
+for dir in os.listdir(afl_output_directory):
```
domenukk: Usually this should be inside a `if __name__ == '__main__'` block (or even a `main` fn) to reuse this in other scripts

utils/mutation_chain/mutation_chain.py
```diff
@@ -0,0 +1,93 @@
...
+            if re.match(file_we_look_for_1_rex, k):
+                retval[k] = compute_mutation_chain(k, current_fuzzer, n+1)
+
+            elif re.match(file_we_look_for_2_rex, k):
+                retval[k] = compute_mutation_chain(k, current_fuzzer, n+1)
+
+        return retval
+
+    else:
+        return "Reached original"
```
domenukk: I would suggest `seed` as value here

domenukk:
I _feel_ like this tool would be more useful if it would (optionally?) print the relationships chain for a single testcase instead of all of them? Especially since all filenames are printed (instead of, for example, using ids for them), this will get very big for a long running fuzzing campaign, right?

maartendekker1998:
> I _feel_ like this tool would be more useful if it would (optionally?) print the relationships chain for a single testcase instead of all of them? Especially since all filenames are printed (instead of, for example, using ids for them), this will get very big for a long running fuzzing campaign, right?

The length/nesting of the json output is dependant on a) the amount of unique crashes and b) the levels of nested mutations-on-mutations. I don't think those are necessarily a function of fuzzing campaign length. Nevertheless, I agree it's a valuable option to have the option to run just one chain computation.

vanhauser-thc:
as this is python I leave the review to @domenukk :-)

utils/mutation_chain/mutation_chain.py
```diff
@@ -0,0 +1,150 @@
...
+        required = False
+    )
+
+    args = parser.parse_args()
+
+    if args.mode == "single":
+        if args.file == None:
+            parser.error("'--mode single' requires the '--file' argument.")
+        elif args.instance == None:
+            parser.error("'--mode single' requires the '--instance' argument.")
```
domenukk: Don't think this is _always_ needed, I would just list all fuzzers and use the first one if instance isn't provided. Just a thought.

utils/mutation_chain/mutation_chain.py
```diff
@@ -0,0 +1,150 @@
...
+
+    parser.add_argument(
+        "--mode", 
+        choices = ['single', 'all'], 
+        help = 'compute chain for one file or all crash files in supplied directory. In single mode the -f argument is required', 
+        required = True
+    )
+
+    parser.add_argument(
+        "--dir",
```
domenukk: Cmin uses `-i` here (in dir), it might be good to reuse the parameter names

utils/mutation_chain/mutation_chain.py
maartendekker1998: I partially agree, I want to enforce the user to use the fuzzer01, fuzzer02, fuzzer0n directory structure to avoid overcomplication. Changed the argument to optional, and if not provided it will default to fuzzer01.

domenukk:
Thanks!

PR #2282, opened 2025-01-02
Enhance compatibility in update_frida_version.sh

vnc0:
Added a case statement to determine the correct `sed` command, ensuring compatibility with both GNU and non-GNU versions of `sed`.


domenukk:
Thank you! For future reference, PRs should be opened on `dev`

vanhauser-thc:
thanks!

PR #2286, opened 2025-01-13
push to stable

vanhauser-thc:


src/afl-fuzz-python.c
```diff
@@ -220,14 +220,60 @@ static py_mutator_t *init_py_module(afl_state_t *afl, u8 *module_name) {
   if (py_module != NULL) {
 
     u8 py_notrim = 0;
+  #if PY_MINOR_VERSION > 12 || PY_MAJOR_VERSION > 3
```
domenukk, line 223: Why or the two? Shouldn't it be `&&`?
Did you try it on python < 3.13?

src/afl-fuzz-python.c
```diff
@@ -220,14 +220,60 @@ static py_mutator_t *init_py_module(afl_state_t *afl, u8 *module_name) {
...
+  #else
     py_functions[PY_FUNC_INIT] = PyObject_GetAttrString(py_module, "init");
+  #endif
     if (!py_functions[PY_FUNC_INIT]) {
 
       WARNF("init function not found in python module");
 
     }
 
+  #if PY_MINOR_VERSION > 12 || PY_MAJOR_VERSION > 3
```
domenukk, line 235: Same question here

src/afl-fuzz.c
```diff
@@ -238,6 +238,7 @@ static void usage(u8 *argv0, int more_help) {
       "immediately,\n"
       "                  -1 = immediately and together with normal mutation.\n"
       "                  Note: this option is usually not very effective\n"
+      "  -u            - enable testcase splicing\n"
```
domenukk, line 241: Why is this not enabled by default?

src/afl-fuzz-python.c
vanhauser-thc: It works without this patch with 3.12 and I don’t want to test where this would break with the new code. This works.
We don’t need && because it won’t compile with python 2, but at some point there will be python 4

src/afl-fuzz.c
vanhauser-thc: See changelog or our discord

PR #2288, opened 2025-01-19
[RFC] Upstream "SAND: Decoupling Sanitization from Fuzzing for Low Overhead"

wtdcode:
Hello!

Our paper "SAND: Decoupling Sanitization from Fuzzing for Low Overhead" just got accepted by ICSE 25 and I'm willing to upstream it. Generally, we design an approach to decouple the sanitization and the fuzzing loop, i.e. only call sanitizers when necessary. This eliminates most overhead of the sanitizers if we filtered inputs properly. Since our approach is rather simple and straightforward, any workflows involving sanitizers can benefit from it.

This branch serves as a "Request For Comments" purpose because I made tons of quirks to make it work, some of which might not align with AFLplusplus conventions. Feel free to correct my understanding or propose better ways.

For usage, see [our main repo](https://github.com/wtdcode/sand-aflpp). Note due to `-a` is occupied already, `-w` is used in this PR.

Note the port might not incomplete at this moment. I will run the port on a few projects we evaluated to ensure the correctness.

## Key Benefits

- Accelerating fuzzing workflows involved with sanitizers, even MSAN!
- Combine any sanitizers as external oracles, even if they are not compatible when compiling, like MSAN and ASAN.
- Orthogonal to fuzzer design and easy to port.

## Main changes

- Build binaries with sanitizer instrumentation and forkservers, but _without pc instrumentation_.

This is crucial for SAND to work, in a way that we call sanitizer binaries but not messing up the bitmaps (and also make the sanitizers run a bit faster). There are several hacks to ensure this (see afl-cc.c). 

- Execution Pattern Selection

One of the core parameters of SAND is the execution pattern, which decide if an input should be sent to sanitizers for further verification. I used an environment variable to determine this. Is it better to add it to CLI arguments?

- Hashing Algorithm

I notice that AFLplusplus drops XXHash in recent versions, but SAND was evaluated against that. Shall we keep the two algorithms at the same time? 

- Variable Bytes

As the paper suggests, SAND performs best on stable targets and thus I mask the corresponding coverage map in `save_if_interesting`. However, I have no idea if this is the correct place to do so. 

- Misc. Changes

Extra fields are printed to `plot_data` to debug the status of SAND. To preserve the consistency of the csv file format, I just always print the fields regardless if extra sanitizers are given. 

### Does it miss bugs?

_In theory_, yes. But in practice, our evaluation proves that no bugs are missing.

### Does it essentially run sanitizers on the queue? What's the difference?

The queue or corpus only saves seeds that increase coverage, however, we have a core insight: Not all bugs-triggering inputs increase coverage. Thus, only running sanitizers within the queue is not enough to catch all bugs.

Instead, SAND run sanitizers on the seeds that has a unique simplified coverage map. This is a superset of the seeds saved in the queue. SAND also supports using coverage map without simplification, at a cost of lower throughput.

### Does it need multi cores?

No. We spin up multiple fork servers in one instance, bound to a single core and schedule the fork servers according to the execution pattern.

### Persistent Mode?

Yeah, SAND support persistent mode out-of-box.


wtdcode:
cc @vanhauser-thc @domenukk @andreafioraldi for reviews.

include/debug.h
```diff
@@ -357,9 +357,9 @@ static inline const char *colorfilter(const char *x) {
   do {                     \
                            \
     if (res < 0)           \
-      PFATAL(x);           \
+      ABORT(x);           \
     else                   \
-      FATAL(x);            \
+      ABORT(x);            \
```
vanhauser-thc: is it necessary to change this? or can you use your own function that does not change the behaviour of afl++?

instrumentation/SanitizerCoverageLTO.so.cc
```diff
@@ -326,9 +326,15 @@ class ModuleSanitizerCoverageLTOLegacyPass : public ModulePass {
...
     };
-
-    return ModuleSancov.instrumentModule(M, DTCallback, PDTCallback);
-
+    
+    if (!getenv("AFL_SAN_NO_INST")) {
+      return ModuleSancov.instrumentModule(M, DTCallback, PDTCallback);
+    } else {
+      if (getenv("AFL_DEBUG")) {
+        DEBUGF("Instrument disabled\n");
```
vanhauser-thc, line 334: nit: "Coverage instrumentation disabled"

src/afl-cc.c
```diff
@@ -2091,10 +2102,11 @@ void add_native_pcguard(aflcc_state_t *aflcc) {
       "pcguard instrumentation with pc-table requires LLVM 6.0.1+"
       " otherwise the compiler will fail");
   #endif
+
   if (aflcc->instrument_opt_mode & INSTRUMENT_OPT_CODECOV) {
 
     insert_param(aflcc,
-                 "-fsanitize-coverage=trace-pc-guard,bb,no-prune,pc-table");
+                "-fsanitize-coverage=trace-pc-guard,bb,no-prune,pc-table");
```
vanhauser-thc: can you please do "make code-format"? this will remove unnecessary indent changes

src/afl-fuzz-bitmap.c
```diff
@@ -452,6 +462,19 @@ void write_crash_readme(afl_state_t *afl) {
 
 }
 
+static void bitmap_set(u8* map, u32 index) {
+  map[index / 8] |= (1u << ( index % 8 ));
+}
+
+// static u8 bitmap_clear(u8* map, u32 index) {
```
vanhauser-thc: remove if not used please :)

src/afl-fuzz-bitmap.c
```diff
@@ -484,6 +507,21 @@ u8 __attribute__((hot)) save_if_interesting(afl_state_t *afl, void *mem,
...
+  u32 cksum_simplified = 0, cksum_unique = 0;
+  u8 san_fault = 0;
+  u8 san_idx = 0;
+  u8 feed_san = 0;
+
+  afl->san_case_status = 0;
+
+  /* Mask out var bytes */
+  if (unlikely(afl->san_binary_length)) {
+    for (u32 i = 0; i < afl->fsrv.map_size; i++) {
```
vanhauser-thc: that looks very expensive for every safe_if_interesting check. rather change the function that detects varbytes and if there are then do this there.

include/debug.h
wtdcode: I can't remember exactly why I changed this. From what I recalled, this change probably related to debugging. I will confirm this.

wtdcode:
Our servers went down since a few days ago and was brought up this afternoon. I will re-run evaluation and check.

src/afl-fuzz-stats.c
```diff
@@ -541,7 +552,7 @@ void maybe_update_plot_file(afl_state_t *afl, u32 t_bytes, double bitmap_cvg,
                  afl->plot_prev_md == afl->max_depth &&
                  afl->plot_prev_ed == afl->fsrv.total_execs) ||
                 !afl->queue_cycle ||
-                get_cur_time() - afl->start_time <= 60000))) {
+                get_cur_time() - afl->start_time <= 1000))) {
```
vanhauser-thc: is this a necessary change? or is this a bugfix?

src/afl-fuzz-bitmap.c
wtdcode: Yeah exactly, this is the worst hack I did due to deadline approaching. If it's the correct place to mask var bytes, I will change it to detect if any var bytes firstly.

src/afl-fuzz-stats.c
wtdcode: Ah nah, we need more dense points for paper plots. I will remove this. Sorry.

src/afl-fuzz.c
```diff
@@ -2554,6 +2577,109 @@ int main(int argc, char **argv_orig, char **envp) {
 
   }
 
+  san_abstraction = getenv("AFL_SAN_ABSTRACTION");
```
vanhauser-thc, line 2591: could you move this into our env walking in afl-fuzz-stats pls?

src/afl-fuzz.c
```diff
@@ -3450,6 +3576,13 @@ int main(int argc, char **argv_orig, char **envp) {
 
   afl_fsrv_deinit(&afl->fsrv);
 
+  
+
+  for (u8 i = 0; i < afl->san_binary_length; i++) {
+    ck_free(afl->san_fsrvs[i].trace_bits);
+    afl_fsrv_deinit(&afl->san_fsrvs[i]); // TODO: Is this necessary? cmplog fksrv seems never deinit-ed?
```
vanhauser-thc: that is a bug, thanks!

src/afl-performance.c
```diff
@@ -425,3 +425,15 @@ char *sha1_hex_for_file(const char *fname, u32 len) {
 
 }
 
+#ifdef _DEBUG
+u32 hash32_xxh32(u8 *key, u32 len, u32 seed) {
```
vanhauser-thc: the reason we use xxh64 and just use 32 bit is that I read that it is faster than xxh32. if you tested that xxh32 is faster for you?

instrumentation/SanitizerCoverageLTO.so.cc
vanhauser-thc: and of course for all occurances :)

src/afl-fuzz-bitmap.c
vanhauser-thc: I would just do it here: https://github.com/AFLplusplus/AFLplusplus/blob/635cd7374e8484afbca3753d0e157287e334ab0d/src/afl-fuzz-run.c#L584

variable bytes are rare

src/afl-performance.c
wtdcode: During our test, different hashing algorithms (we tested xxhash series and murmurhash and a few other candidates) almost has no performance impact, at least less than the random error across repetitions. We only cares about the collision rate and the final implementation here works the best during our evaluation.

include/forkserver.h
```diff
@@ -167,6 +169,7 @@ typedef struct afl_forkserver {
   u8 *shmem_fuzz;                       /* allocated memory for fuzzing     */
 
   char *cmplog_binary;                  /* the name of the cmplog binary    */
+  char *asanfuzz_binary;                /* the name of the ASAN binary      */
```
domenukk, line 172: The original idea of forksever.h was to have one forkserver instance per binary, not to add all binaries to the same forkserver instance :D 
but I guess this already failed with cmplog

vanhauser-thc:
I think we will need some time to evaluate if we conclude that this is beneficial to have or not.
It requires another compilation of the target (many users just compile with asan and use that single binary for fuzzing because they are lazy or do not know how to fuzz effectively).

do before you invest a lot of time handling the nits and wishes to the PR I propose to give us some time :-)

> For usage, see [our main repo](https://github.com/wtdcode/sand-aflpp). Note due to `-a` is occupied already, `-w` is used in this PR.

-w is fine

> * Execution Pattern Selection
> 
> One of the core parameters of SAND is the execution pattern, which decide if an input should be sent to sanitizers for further verification. I used an environment variable to determine this. Is it better to add it to CLI arguments?

depends if there is a clear favorite, in that case IMHO the other(s) should be configurable via an env var. if there is a clear distinction in which situation which mode is better than command line arg.

> * Hashing Algorithm
> 
> I notice that AFLplusplus drops XXHash in recent versions, but SAND was evaluated against that. Shall we keep the two algorithms at the same time?

no, use the afl++ one please. it is super fast.

> * Misc. Changes
> 
> Extra fields are printed to `plot_data` to debug the status of SAND. To preserve the consistency of the csv file format, I just always print the fields regardless if extra sanitizers are given.

the thing is if that is interesting data that should be in the plot file or not at all. if so, then you always print.

thanks! this looks promising IMHO!

include/forkserver.h
wtdcode: We indeed have multiple forkservers here. See https://github.com/AFLplusplus/AFLplusplus/pull/2288/files/efa2120935a5105d1a0561bacf9f2fb3b7b92156#diff-58bda4a9d7e04051b895bec22791293b2c01b03b6adf79d56134f09e5d838333R739

I added this field because I don't want to reuse `cmplog_binary` for clarity. At this moment, I think SAND is not compatible with cmplog but it is perfectly doable. I would do it in a second PR.

wtdcode:
> I think we will need some time to evaluate if we conclude that this is beneficial to have or not. It requires another compilation of the target (many users just compile with asan and use that single binary for fuzzing because they are lazy or do not know how to fuzz effectively).

No worry, I expect several iterations for this and I would like to get more inputs before working on the integration. Also note that, the extra workflow is fully optional so the lazy users can continue as before.

> 
> do before you invest a lot of time handling the nits and wishes to the PR I propose to give us some time :-)
> 
> > For usage, see [our main repo](https://github.com/wtdcode/sand-aflpp). Note due to `-a` is occupied already, `-w` is used in this PR.
> 
> -w is fine
> 
> > * Execution Pattern Selection
> > 
> > One of the core parameters of SAND is the execution pattern, which decide if an input should be sent to sanitizers for further verification. I used an environment variable to determine this. Is it better to add it to CLI arguments?
> 
> depends if there is a clear favorite, in that case IMHO the other(s) should be configurable via an env var. if there is a clear distinction in which situation which mode is better than command line arg.

Okay, then env var is good and enough. Our evaluation shows "simplified trace" balances well in our evaluation.

> 
> > * Hashing Algorithm
> > 
> > I notice that AFLplusplus drops XXHash in recent versions, but SAND was evaluated against that. Shall we keep the two algorithms at the same time?
> 
> no, use the afl++ one please. it is super fast.

But I have no idea if this introduces more collisions. I will evaluate a bit if I have time (or use afl++ one if lacking enough time). Note our implementation achieves 0 collision on 19 out of 20 targets and only 4 in the last target.

> 
> > * Misc. Changes
> > 
> > Extra fields are printed to `plot_data` to debug the status of SAND. To preserve the consistency of the csv file format, I just always print the fields regardless if extra sanitizers are given.
> 
> the thing is if that is interesting data that should be in the plot file or not at all. if so, then you always print.

It depends. It is more like "Advanced Usage", for example you would like to determine which "execution pattern" should be used. I personally would like to preserve these fields in the plot file.

include/afl-fuzz.h
```diff
@@ -610,7 +611,12 @@ typedef struct afl_state {
   u8 *var_bytes;                        /* Bytes that appear to be variable */
 
 #define N_FUZZ_SIZE (1 << 21)
+#define N_FUZZ_SIZE_BITMAP (1 << 29)
   u32 *n_fuzz;
+  u8 *n_fuzz_dup;
+  u8 *classified_n_fuzz;
+  u8 *simplitied_n_fuzz;
```
wtdcode: a typo here for my own record

src/afl-fuzz.c
wtdcode: This has been included in `afl_environment_variables`

src/afl-cc.c
wtdcode: The codebase seems not formatted for a while and running `make code-format` produces lots of unrelated changes so I just formatted the code I editted.

src/afl-fuzz-bitmap.c
wtdcode: SAND needs stable "execution pattern" for every execution and that's why I put it here instead of calibration.

Since this is too expensive, I simply remove these makeups so it's users' responsibilities to avoid unstable targets. I will document it somewhere.

wtdcode:
I have address most reviews and fixed the integration. Now this shall work well after a few quick evaluation

src/afl-fuzz-init.c
```diff
@@ -2332,7 +2332,7 @@ void setup_dirs_fds(afl_state_t *afl) {
         afl->fsrv.plot_file,
         "# relative_time, cycles_done, cur_item, corpus_count, "
         "pending_total, pending_favs, map_size, saved_crashes, "
-        "saved_hangs, max_depth, execs_per_sec, total_execs, edges_found\n");
+        "saved_hangs, max_depth, execs_per_sec, total_execs, edges_found, total_crashes, servers_count, san1_exec...\n");
```
vanhauser-thc: can you please remove the "..." at the end? :-)

src/afl-performance.c
vanhauser-thc: have you already tested this?

vanhauser-thc:
thanks for the changes.
after the results (you do asan vs. san, right? then I will do this version without san vs. current dev state, to ensure the changes have no negative impact on the existing state) this will also need documentation. so adding to fuzzing_in_depth.md (just short and the essential things) and the env vars to env_variables.md, and the full documentation into its own readme in instrumentation maybe? like instrumentation/README.SAN.md?

wtdcode:
> thanks for the changes. after the results (you do asan vs. san, right? then I will do this version without san vs. current dev state, to ensure the changes have no negative impact on the existing state)

We actually have all baselines in our paper: AFL++, AFL++ + MSAN, AFL++ + ASAN/UBSAN, AFL++ + Debloat12/UBSAN (another research efforts to reduce sanitizers overhead). In theory, there shouldn't be any performance regression. However, when I tried to port the code from 4.05c to the dev, I notice lots of things have changed and thus please tell me anything unexpected.

I'm also doing artifact evaluation so I will soon also re-evaluate both our reference implementation and this branch (not full evaluation, maybe just 1, 2 representative targets). I will post updates once it's done. 

> this will also need documentation. so adding to fuzzing_in_depth.md (just short and the essential things) and the env vars to env_variables.md, and the full documentation into its own readme in instrumentation maybe? like instrumentation/README.SAN.md?

I was writing that already and I planned to put it at `docs/SAND.md`. There are many caveats to use SAND properly (like stable targets as mentioned etc) so I expect I will write a bit long and verbose documents (also for artifact evaluation).

src/afl-performance.c
wtdcode: Not yet, will update once done.

wtdcode:
docs done! Feel free to suggest changes about documents.

wtdcode:
To facilitate evaluation and review, I pushed a docker image containing our evaluation seeds and programs.

Try it with:

```bash
docker run --rm -it -v /path/to/tmpfs:/tmpfs lazymio/sand-unifuzz:upstream bash
```

The programs I tried is `imginfo`. Start it with SAND:

```bash
afl-fuzz -b 103 -i /unibench/seeds/general_evaluation/imginfo -o /tmpfs/sand -w /unibench/bin/aflasanubsan_noins/imginfo -- /unibench/bin/justafl/imginfo -f @@
```

For comparison, the AFL++ without any sanitizer is started with:

```bash
afl-fuzz -b 102 -i /unibench/seeds/general_evaluation/imginfo -o /tmpfs/native  -- /unibench/bin/justafl/imginfo -f @@
```

I ran this for some time and the results seem to align with the evaluation in our paper.

src/afl-fuzz-init.c
wtdcode: Done!

src/afl-performance.c
wtdcode: I decided to use AFL++ `hash32` due to lacking time to evaluate this. I recall that there is no significant performance issue.

src/afl-fuzz.c
wtdcode: I put a fix along with the PR

wtdcode:
I should address all concerns and the PR is ready to go. Please have another check =).

vanhauser-thc:
So far does not look like it has a negative impact when not activated: https://www.fuzzbench.com/reports/experimental/2025-01-26-aflpp/index.html

wtdcode:
Thanks for updates. Any blocker now?

docs/SAND.md
```diff
@@ -0,0 +1,55 @@
...
+## Usage
+
+For a normal fuzzing workflow, we have:
+
+1. Build target project with AFL_USE_ASAN=1 to get `target_asan`
+2. Fuzz the target with `afl-fuzz -i seeds -o out -- ./target_asan`
+
+For SAND fuzzing workflow, this is slightly different:
+
+1. Build target project _without_ any sanitizers to get `target_native`, which we will define as "native binary".
```
vanhauser-thc: can you please clarify:
Build the target project with AFL++ but _without_ any sanitizers ...

docs/SAND.md
```diff
@@ -0,0 +1,55 @@
...
+2. Build target project with AFL_USE_ASAN=1 AFL_SAN_NO_INST=1 to get `target_asan`. Do note this step can be repeated for multiple sanitizers, like MSAN, UBSAN etc.
```
vanhauser-thc: but for effectiveness one could use AFL_USE_UBSAN etc. too, right?

wtdcode: Yep, sure.

docs/SAND.md
```diff
@@ -0,0 +1,55 @@
...
+Then you get:
+
+- almost the same performance as `afl-fuzz -i seeds -o out -- ./target_native`
+- and the same bug-finding capability as `afl-fuzz -i seeds -o out -- ./target_asan`
+
+## Tips
+
+### Alternative execution patterns
+
+By default, SAND use the hash value of the simplified coverage map as execution pattern, i.e. if an input has a unique execution pattern, it will be sent to sanitizers for inspection. This shall work for most cases. However, if you are strongly worried about missing bugs, try `AFL_SAN_ABSTRACTION=unique_trace afl-fuzz ...`. Alternatively, you can also have `AFL_SAN_ABSTRACTION=coverage_increase`, which essentially equals to running sanitizers on the corpus.
```
vanhauser-thc: please clarify for coverage_increase that this might loose crashes. for a newbie that implication is likely not obvious.

docs/SAND.md
```diff
@@ -0,0 +1,55 @@
...
+
+The execution pattern evaluated in our papers is targeting the common bugs, as ASAN/MSAN/UBSAN catches. For other bug types, you probably need to define new execution patterns and re-evaluate.
+
+### My throughput is greatly impacted
+
+Generally, this is due to too many inputs going through sanitizers, for example, because of unstable targets. You could check stats from `plot_file` to confirm this. Try to switch execution patterns as stated above.
+
+### Cmplog Compatibility
+
+At this moment, SAND probably is not compatible with cmplog and we will fix this soon.
```
vanhauser-thc: not a comment on the doc but to be to understand - what is incompatible here?

docs/fuzzing_in_depth.md
```diff
@@ -203,6 +203,9 @@ instances, so running more than one address sanitized target would be a waste.
 *IF* you are running a saturated corpus, then you can run up to half of the 
 instances with sanitizers.
 
+An alternative approach, [SAND](./SAND.md) could combine different sanitizers
```
vanhauser-thc: proposal: "An alternative but more effective approach is to use much [SAND](./SAND.md) which can combine different sanitizers with a much higher throughput."

src/afl-fuzz.c
vanhauser-thc: that is not what I meant. get rid of the getenv here and move it over in the env parsing in afl-fuzz-stats.c :)

src/afl-fuzz.c
```diff
@@ -2554,6 +2588,131 @@ int main(int argc, char **argv_orig, char **envp) {
...
+
+  } else {
+
+    WARNF("Unkown abstraction: %s, fallback to unique trace.\n",
+          san_abstraction);
+    afl->san_abstraction = SIMPLIFY_TRACE;
+
+  }
+
+  afl->no_saving_crash_seed = false;
```
vanhauser-thc: false = 0  so this is the default when the afl structure is initialize and hence this line can be removed right?

docs/SAND.md
wtdcode: I once tested cmplog on our 4.05c reference implementation and it didn’t look working. I plan to test and fix cmplog compatibility in next PR.

src/afl-fuzz-bitmap.c
```diff
@@ -452,6 +461,18 @@ void write_crash_readme(afl_state_t *afl) {
 
 }
 
+static void bitmap_set(u8 *map, u32 index) {
```
vanhauser-thc: please add `inline`

src/afl-fuzz-bitmap.c
```diff
@@ -452,6 +461,18 @@ void write_crash_readme(afl_state_t *afl) {
...
+
+  map[index / 8] |= (1u << (index % 8));
+
+}
+
+static u8 bitmap_read(u8 *map, u32 index) {
```
vanhauser-thc: please add `inline`

docs/SAND.md
wtdcode: That means building project with afl-clang-fast(or lto or so) but without AFL_USE_ASAN/MSAN/UBSAN. I will clarify a bit.

src/afl-fuzz.c
wtdcode: Yeah, sure.

vanhauser-thc:
left a view last nits, then we can merge :)
thanks!

src/afl-fuzz.c
wtdcode: I found the field is no long used, probably was added during evaluation. Removed.

docs/fuzzing_in_depth.md
wtdcode: Done

docs/SAND.md
wtdcode: I just tried with `cmplog` and it works! Previously the cmplog forkserver would always crash and it now works. I removed this part.

docs/SAND.md
wtdcode: Done accordingly.

docs/SAND.md
wtdcode: In addition to this clarification, I also add the example of test-instr.c from our repo.

wtdcode:
I'm done with all changes and revised the documents a bit. Please check =).

vanhauser-thc:
LGTM thanks!

PR #2290, opened 2025-01-25
Update FRIDA mode for Frida Gum 16.6.x API compatibility

vnc0:
The frida-gum API [changed a bit](https://frida.re/news/2025/01/09/frida-16-6-0-released/) in 16.6.0, so frida-mode was no longer compatible.

**What changed**
- `GumModuleDetails` is now deprecated. I refactored the functions to use `GumModule` instead.
- `gum_module_find_export_by_name(NULL, ...)` is now `gum_module_find_global_export_by_name`.

**What I tested**
testinstr is working


vanhauser-thc:
@vnc0 thanks!
can I ask you to please make a quick test to see if there are speed differences due the new frida version?
test-instr.c and make a real world more complex target? if they are at least about equal then it is safe to merge. Sometimes new features or regression makes the fuzzing slower

vnc0:
That was a good point. 
Actually, the execution speed of all examples I tested is a bit slower with the new Gum version. 
Unless someone finds a way to optimize, or some future Gum release introduces optimizations, it is better to stick with the old version.

For example, I measured the following with frida_mode/test/png (AFL ++4.31a on MacOS):
**Gum Version 16.0.11** : exec speed : 1511/sec
**Gum Version 16.6.5**  : exec speed : 1186/sec

domenukk:
Makes sense to keep FRIDA up to date in any case, things constantly break on MacOS and others and they're fixed by Ole pretty regularly

vanhauser-thc:
That is like a 35% difference, that is huge. (if the difference is as large - or maybe larger on a large real world project...)

maybe we should use a different approach, that on macos we have the this version by default on linux the current one, but the user can choose either?

like `make newest` or `make best` or something like that

vanhauser-thc:
@vnc0 @domenukk ping

domenukk:
I mean as stated above, I would definitley merge this, the newer Frida version would be more important to me than some execs on some speciality target.
If there really is a performance regression in Frida this should actually be reported upstream instead of staying on an old version.

CC @WorksButNotTested

WorksButNotTested:
I'm not aware of any changes to stalker (the part which is likely to be most performance critical). So unless there is significant performance penalty for typical targets, then if be inclined to bump the version.

vnc0:
> That is like a 35% difference, that is huge. (if the difference is as large - or maybe larger on a large real world project...)
> 
> maybe we should use a different approach, that on macos we have the this version by default on linux the current one, but the user can choose either?
> 
> like `make newest` or `make best` or something like that

I have just pushed a change that conditionally uses the new code depending on the frida version set. 
Once frida's performance issues are resolved, this can be reverted.

Has anyone else reproduced the performance problems at all?

WorksButNotTested:
> pushed a change that conditionally uses the new code depending on the frida version set. Once frida's performance issues are resolved, this can be reverted.
> 
> Has anyone else reproduced the performance problems at all?

If you can bisect and find the version with the regression, it may be possible to work out the issue from changes.

vnc0:
> > pushed a change that conditionally uses the new code depending on the frida version set. Once frida's performance issues are resolved, this can be reverted.
> > Has anyone else reproduced the performance problems at all?
> 
> If you can bisect and find the version with the regression, it may be possible to work out the issue from changes.

I just did that with test/png and to my surprise the performance issue is not due to the new API.
There is a regression when switching from GUM 16.1.11 to [16.2.0](https://frida.re/news/2024/02/16/frida-16-2-0-released/).

16.1.5 -> 1586/sec
**16.1.11 -> 1603/sec
16.2.0 -> 1298/sec**
16.3.0 -> 1311/sec
16.6.6 -> 1309/sec

Edit: I just opened up an issue at frida/frida-gum for this: https://github.com/frida/frida-gum/issues/991

WorksButNotTested:
It looks like the changes in `gum` are relatively minor. Most notably, there are no changes in `stalker` which is the most performance critical part. Very odd!
https://github.com/frida/frida-gum/compare/097dd4...579c096

vanhauser-thc:
@vnc0 thanks a lot!
I think we can merge this now and bump the version once there is a fix.
an automatic "use newest version" switch for macos users would be nice though.

PR #2293, opened 2025-02-07
Fix various spelling errors

drobilla:
I noticed a few spelling errors when reading the documentation, so I fixed those, and decided to do a `codespell` run and fix some in the code as well while I was at it.


drobilla:
I'm not sure what the policy is on changing things in different places, so I broke this up by visibility so it's easy to drop parts if necessary.  There's also many in `ChangeLog` found by `codespell` but I didn't touch because that it's a historical document of sorts.

domenukk:
Nice one, thank you!

PR #2299, opened 2025-02-12
push to stable

vanhauser-thc:


include/coverage-64.h
```diff
@@ -72,7 +79,7 @@ inline void classify_counts(afl_forkserver_t *fsrv) {
 
 }
 
-inline static void classify_counts_mem(u64 *mem, u32 size) {
```
domenukk, line 75: Shouldn't this remain `static` if it's in a header?

vanhauser-thc: I agree but I tried to fix something for a platform I cannot test for easily (s390)

PR #2300, opened 2025-02-13
afl-fuzz: fix -a option

manyhus:
This fixes the option string so -a will be parsed with an argument again, unbreaking the option. The result of the missing : was that optarg would always be NULL and we would abort in stricmp.

Introduced in 1c9925c7d7


vanhauser-thc:
Thanks! I overlooked that when I reviewed the PR that broke this ...

Issue #2298, opened 2025-02-11
Android Frida-mode Exec Speed & Stability Regressions


etanner1:
**Describe the bug**
Recent versions of afl++ have dramatically lower exec speed and stability in Android frida-mode.

In the provided example, no changes were made to the harness. The only difference between runs is using different versions of afl++. Frida version is locked at 16.0.13 between all tests shown here. Separately, I also tried various updated versions of Frida and that did not appear to make a difference.

I followed the Quarkslab frida-mode tutorial (https://blog.quarkslab.com/android-greybox-fuzzing-with-afl-frida-mode.html) that is suggested on the afl++ website here: https://aflplus.plus/docs/tutorials/. Harness and configuration here: https://github.com/quarkslab/android-fuzzing.

Performance using the open-source android-fuzzing wlinked harness and afl.js (supporting screenshots for reference):
afl++ Version | Exec Speed | Stability |
|---|---|---|
| dev-branch (9cb3fe9) | ~100/s | ~75% |
| 4.31c (9cac7ce) | ~100/s | ~75% |
| 4.30c (4f53803) | ~100/s | ~75% |
| 4.21c (9f6b012) | ~30k/s | ~75% |
| 4.10c (775861e) | ~30k/s | ~75% |
| 4.09c (108fb0b) | ~26k/s | 100% |

**To Reproduce**
```
gh repo clone quarkslab/android-fuzzing
cd android-fuzzing
export NDK_PATH=<HERE>
 
adb push apk/qb.blogfuzz.apk /data/local/tmp/
adb push native/lib/libblogfuzz.so /data/local/tmp/
 
pushd jenv
cmake -DANDROID_PLATFORM=31 -DCMAKE_TOOLCHAIN_FILE=${NDK_PATH}/build/cmake/android.toolchain.cmake -DANDROID_ABI=arm64-v8a .
make
adb push libjenv.so /data/local/tmp/
popd
 
pushd wlinked_jni
mkdir -p build && cd build
cmake -DANDROID_PLATFORM=31 -DCMAKE_TOOLCHAIN_FILE=${NDK_PATH}/build/cmake/android.toolchain.cmake -DANDROID_ABI=arm64-v8a ..
make
adb push fuzz /data/local/tmp/
adb push ../afl.js /data/local/tmp/
popd
 
[...]

PR #2301, opened 2025-02-13
Addressed frida-mode exec speed regression on ARM64 for #2298

etanner1:
Fixed a regression in exec speed for persistent frida-mode ARM64 in relation to #2298. Patch provided by @WorksButNotTested. Other CPU architectures have not been patched yet.


vanhauser-thc:
thank you!

PR #2303, opened 2025-02-14
Fix debug prefix for afl-cc, llvm-rt

manyhus:
After the llvm_mode directory was removed in 996986bed5 and compilation started happening from the root, adding llvm_mode to the debug path is incorrect and causes source file lookups to fail when debugging e.g. afl-cc or the llvm pass.


vanhauser-thc:
thank you!

PR #2304, opened 2025-02-15
Also set /usr/bin/c++ or this fails cc-rs

wtdcode:
As title, this updates `g++-${GCC_VERSION}` to `/usr/bin/c++`, or this would fail `cc-rs` like:

```
error occurred in cc-rs: Failed to find tool. Is `c++` installed?
```


vanhauser-thc:
Thanks :)

PR #2307, opened 2025-02-17
feat: use GH hosted ARM runners.

intrigus-lgtm:
Re-enable arm64 docker containers by using GH arm runners.
Also enable CI for arm64.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: "${{ matrix.os }}"
     strategy:
       matrix:
-        os: [ubuntu-24.04, ubuntu-22.04]
+        os: [ubuntu-24.04, ubuntu-22.04, ubuntu-24.04-arm]
     env:
       AFL_SKIP_CPUFREQ: 1
       AFL_I_DONT_CARE_ABOUT_MISSING_CRASHES: 1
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: "${{ matrix.os }}"
     strategy:
       matrix:
-        os: [ubuntu-24.04, ubuntu-22.04]
+        os: [ubuntu-24.04, ubuntu-22.04, ubuntu-24.04-arm, ubuntu-22.04-arm]
```
vanhauser-thc: IMHO this is too much. to save some CPU and power please remove the 22.04-arm one. thanks!

vanhauser-thc:
do you know if it works? so, without segfaults when installing packages? :)

intrigus-lgtm:
> do you know if it works? so, without segfaults when installing packages? :)

The workflow seems to work:
https://github.com/AFLplusplus/AFLplusplus/actions/runs/13377397708/job/37359408242?pr=2307

.github/workflows/ci.yml
intrigus-lgtm: Will do

vanhauser-thc:
thank you!

PR #2308, opened 2025-02-18
Delete initial call to LLVMFuzzerTestOneInput in aflpp driver

tokatoka:
@vanhauser-thc 
as we discussed this call doesn't make sense with how afl++ works


vanhauser-thc:
thanks!

PR #2310, opened 2025-02-19
preparing for LLVM 20 with new sanitizer.

devnexen:
note: no real valuable option atm.

### DIFF
--- a/docs/env_variables.md
+++ b/docs/env_variables.md
@@ -107,6 +107,7 @@ fairly broad use of environment variables instead:
       conditions
     - `AFL_USE_UBSAN=1` - activates the undefined behavior sanitizer
     - `AFL_UBSAN_VERBOSE=1` - outputs detailed diagnostic information when undefined behavior is detected, instead of simply terminating with "Illegal Instruction"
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
does not hurt :-)
do you think it makes sense to already document it? or guard with LLVM_MAJOR >= 20 ?

devnexen:
took doc route :)

domenukk:
Reading through the RTSan doc, I wonder if this will even work, given that it would detect uses of `malloc` and `free`: won't the afl++ runtime be built with RTSan as well in this case?

vanhauser-thc:
> Reading through the RTSan doc, I wonder if this will even work, given that it would detect uses of `malloc` and `free`: won't the afl++ runtime be built with RTSan as well in this case?

afl-compilter-rt.o is pre-built without any sanitizer or coverage instrumentation if you mean that?

vanhauser-thc:
> took doc route :)

another point to document it would be here:

https://github.com/AFLplusplus/AFLplusplus/blob/854776a9c9f57abe90beb5bccab4dc5dfa0b6776/src/afl-cc.c#L2979

:)

vanhauser-thc:
thank you :-)

PR #2312, opened 2025-02-21
Small change to LTO documentation

stock1218:
Hi, I was going through the documentation on LTO and noticed the example commands jump between LLVM version 19 and 15. It was a subtle difference I didn't notice and I wasn't getting afl-clang-lto/++ to compile and had to spend some time investigating. I noticed there was some history switching from LLVM 15 to 19 and I think these just fell through the cracks.

Hopefully this will save some time for someone else also blindly copy and pasting the commands from the documentation :)


vanhauser-thc:
Thanks!

PR #2315, opened 2025-02-27
Fix incorrect comment in unicorn-afl python script.

moonkick64:
The comment is written where the code loads binary. 
I have fixed the comment and added a new comment to adapt the top comment. ( # Load the binary to emulate and map it into memory.)


domenukk:
Thank you!

PR #2319, opened 2025-03-03
Add `fflush(stdout);` before `abort` call

### LINKED ISSUE #2318: `fflush(stdout)` before `abort()`?
Would you consider adding an `fflush(stdout);` just before following `abort` call?
https://github.com/AFLplusplus/AFLplusplus/blob/a0d996558b3a732a4a9f8d0289db4cacdc245ef2/include/debug.h#L334

For reasons that are unimportant, AFL++ was unable to allocate memory in some tests that I was running. AFL++ tried to produce a [helpful error message](https://github.com/AFLplusplus/AFLplusplus/blob/a0d996558b3a732a4a9f8d0289db4cacdc245ef2/include/alloc-inl.h#L76), but it was not shown because the process aborted without flushing `stdout`. Adding an `flush(stdout);` just before the `abort` would prevent this problem from reoccurring.

smoelius:
Fixes #2318


vanhauser-thc:
thank you!

PR #2325, opened 2025-03-05
Add notes about cpu bind in docker

### LINKED ISSUE #2324: Docker prevents cpu binding from working properly
**IMPORTANT**

Thank you for making AFL++ better!

**Describe the bug**
AFL++ will always try to bind a free CPU by default. This is done by scanning the `/proc` to find a free CPU which is not used by any other processes. However, Linux namespaces, especially docker containers, will hide other processes, and only show processes in current namespace. As a result, multiple docker instances of AFL++ will not see each other, and as a result, **they all try to bind to CPU 0**.

There may not be an ideal solution for this, maybe user can pass an environment variable to manually set CPU index to bind as a workaround, or set `--cpuset-cpus` when starting docker container. However, this is not an easy root cause to find, and users are only to find that "why my afl++ runs so slowly in docker". I think it may be better to document this behavior somewhere. 

**To Reproduce**

Just spawn two individual AFL++ docker containers and fuzz.

**Expected behavior**
They don't bind to the same CPU core.

Evian-Zhang:
Close #2324.

### DIFF
--- a/docs/fuzzing_in_depth.md
+++ b/docs/fuzzing_in_depth.md
@@ -496,6 +496,8 @@ Note:
   protection against attacks! So set strong firewall rules and only expose SSH
   as a network service if you use these (which is highly recommended).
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
LGTM, thanks

PR #2326, opened 2025-03-05
Follow up for #2308

tokatoka:
this `dummy_input` is not used anywhere now


vanhauser-thc:
the previous change with sending the dummy data and for sure with this PR the constants would not be compiled in anymore that are needed for the deferred forkserver and the persistent mode.
I made a fix that does this and also removes the dummy input.
thanks though - I had forgotten about that!

PR #2328, opened 2025-03-10
fix(afl-fuzz-init): ensure proper permissions for setting CPU governor

0xXA:
The previous command used tee without sudo, which could fail due to insufficient permissions.


domenukk:
Thank you! Switched the PR to the `dev` branch.

The output sais `"run these commands as root:"` above, so it shouldn't really fail, but I'll leave it up to @vanhauser-thc

vanhauser-thc:
I think it does not hurt. people do not read (enough).

Issue #2333, opened 2025-03-20
Exit code 23 detected as crash even though LSAN disabled


andy-knowles:
**Describe the bug**
A user program that that terminates normally and returns with exit code 23 is detected as a crash by afl-fuzz if the program was compiled with AFL_USE_ASAN=1, even though AFL_USE_LSAN is not set.

afl-forkserver.c does not differentiate between whether ASAN, LSAN or MSAN is enabled when checking if the child exit code matches LSAN_ERROR or MSAN_ERROR.

The "several possible explanations" described by afl-fuzz-init.c after a seed crashes does not detect or describe the possibility of a false positive crash detection due to the exit code matching LSAN_ERROR or MSAN_ERROR.

The fuzzing_in_depth.md does not describe that programs exiting normally with code 23 (or code 86) may have issues with false positive crash detection when combined with santitizers.

**To Reproduce**
1. export AFL_USE_ASAN=1
2. echo "int main() { return 23; }" > nocrash.c
3. ./afl-cc nocrash.c -o nocrash
4. mkdir seeds outputs && echo "hello" > seeds/world
5. ./afl-fuzz -i seeds -o outputs -- ./nocrash
6. Observe afl-fuzz incorrectly detects a crash on the single seed and fails to start.

![Image](https://github.com/user-attachments/assets/bc973800-a74e-4fb1-ae45-17043c881a75)

**Expected behavior**
afl-fuzz should run the single seed file without incorrectly detecting a crash and then start fuzzing.
Magic error codes that fuzzed programs should avoid using should be documented somewhere.

vanhauser-thc:
I understand why you are not happy with this - but from the view from afl-fuzz how would it know that the target exited with error 23 because of LSAN or because of a normal error? afl-fuzz does not know with what parameters, features, sanitizers etc. a target is compiled with. AFL_USE_ASAN is only for compiling a target and therefore is not an indicator for afl-fuzz.

I agree though that this should be mentioned in the docs.

andy-knowles:
[...]

PR #2334, opened 2025-03-24
Better handling of exit codes used by sanitzers

andy-knowles:
Use more fine-grained detection of sanitizer crashes to reduce false positives.
Explain that a seed crash could be a false positive if it exits with a special sanitizer exit code.
Document that some sanitizers use special exit codes, and that this can result in false positive crash detection.


andy-knowles:
Resolves #2333.

vanhauser-thc:
LGTM, thanks!

PR #2338, opened 2025-03-29
Fix frida-mode compliation error for MacOS

r3sting:
This pull request fixes the frida mode compilation error on MacOS:

`GNUmakefile:193: *** unterminated call to function `shell': missing `)'.  Stop.`

It adds a conditional check for the operating system in the makefile in a syntax supported by MacOS's default shell, 'zsh'.


vanhauser-thc:
LGTM, thanks!

Issue #2295, opened 2025-02-10
error: implicit declaration of function ‘classify_counts_mem’ (s390x)


rwmjones:
**IMPORTANT**
1. You have verified that the issue to be present in the current `dev` branch.

YES

Thank you for making AFL++ better!

**Describe the bug**
AFL++ 4.31c fails to compile on s390x architecture:
```
src/afl-fuzz-bitmap.c: In function ‘save_if_interesting’:
src/afl-fuzz-bitmap.c:541:7: error: implicit declaration of function ‘classify_counts_mem’; did you mean ‘classify_counts’? [-Wimplicit-function-declaration]
  541 |       classify_counts_mem((u64 *)afl->san_fsrvs[0].trace_bits,
      |       ^~~~~~~~~~~~~~~~~~~
      |       classify_counts
```

**To Reproduce**
1. Try to build AFL++ on s390x

**Additional context**

Full build log is available here, probably temporarily:

https://kojipkgs.fedoraproject.org//work/tasks/593/129090593/build.log

vanhauser-thc:
is s390x a 32 bit platform?

rwmjones:
It's 64 bit.  (s390 without the 'x' is the 31 [sic] bit version).

However I'm not clear how "serious" AFL++ support is on s390x in general.  In Fedora what I did was to remove the architecture from the build for 4.31.

Edit: I should add that one weird thing about s390x is it's the only big endian architecture still standing.

msmeissn:
same error on i586 btw. 

include/coverage-64.h:inline static void classify_counts_mem(u64 *mem, u32 size) {
src/afl-fuzz-bitmap.c:      classify_counts_mem((u64 *)afl->san_fsrvs[0].trace_bits,


progably needs a counterpart in coverage-32.h

vanhauser-thc:
I pushed a fix not sure if that helps, is it working now?
(`dev` branch)

msmeissn:
[   15s] src/afl-fuzz-bitmap.c: In function ‘save_if_interesting’:
[   15s] src/afl-fuzz-bitmap.c:541:27: error: passing argument 1 of ‘classify_counts_mem’ from incompatible pointer type [-Wincompatible-pointer-types]
[   15s]   541 |       classify_counts_mem((u64 *)afl->san_fsrvs[0].trace_bits,
[   15s]       |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #2348, opened 2025-04-03
Define WORD_SIZE_64 for riscv64

Xeonacid:
#2295 should actually be fixed by a s390x equivalent of this PR.


vanhauser-thc:
thank you

PR #2350, opened 2025-04-03
Patch 1

### LINKED ISSUE #2335: afl-cmin (awk version) treats afl files as input files
**Describe the bug**
the awk version of afl-cmin use [this line](https://github.com/AFLplusplus/AFLplusplus/blob/stable/afl-cmin#L435) to find input files
```awk
cmdline = "(cd "in_dir" && find . \\( ! -name \".*\" -a -type d \\) -o -type f -exec stat "stat_format" \\{\\} + | sort -k1n -k2r) | grep -Ev '^0'"
```

This treats AFL internal files as input. For example,
```
./default/cmdline
./default/fastresume.bin
./default/fuzz_bitmap
./default/fuzzer_setup
./default/fuzzer_stats
./default/plot_data
./default/target_hash
./M1/cmdline
./M1/fastresume.bin
./M1/fuzz_bitmap
./M1/fuzzer_setup
./M1/fuzzer_stats
./M1/plot_data
./M1/queue/.state/auto_extras/auto_000000
./M1/target_hash
```

**Expected behavior**
The above files shouldn't be processed by afl-cmin, especially fastresume.bin, which is huge.

20urc3:
Fix #2335

### DIFF
--- a/afl-cmin
+++ b/afl-cmin
@@ -432,8 +432,7 @@ BEGIN {
   } else {
     stat_format = "-f '%z %N'" # *BSD, MacOS
   }
-  cmdline = "(cd "in_dir" && find . \\( ! -name \".*\" -a -type d \\) -o -type f -exec stat "stat_format" \\{\\} + | sort -k1n -k2r) | grep -Ev '^0'"
-  #cmdline = "ls "in_dir" | (cd "in_dir" && xargs stat "stat_format" 2>/dev/null) | sort -k1n -k2r"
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
this is not a fix, this is a bug. usually people use afl-cmin on directories they collected potential seeds in. with this patch this would not work anymore :-)

20urc3:
My bad, you totally right. I submited a patch 2 that exclude AFL internal files when run on an AFL output directory. This way, regular seed directories will work as before, but when run on an AFL output directory, it will properly exclude the metadata files.

PR #2354, opened 2025-04-05
reduce skipdet_e memory usage

kcwu:
By using bitmaps, the memory requirement for
`q->skipdet_e->skip_eff_map` and `done_inf_map`, which previously scaled with the corpus size, is reduced to one-eighth of its original size.


vanhauser-thc:
as this depends on the file size and is kept forever this will save quite some memory. thanks!

PR #2358, opened 2025-04-07
fix potential q->trace_mini off by 1 overflow

kcwu:
Be careful this change is incompatible with old version of fastresume.bin


vanhauser-thc:
thanks. this needs one more fix for the fastresume.bin version header, I will do it myself after merging.

PR #2363, opened 2025-04-08
fix: ensure correct rescoring using latest queue entry execution

5angjun:
### Problem

The `update_bitmap_score()` function depends on `trace_bits` to determine
if a queue entry should be marked as `top_rated`. However, in the previous
rescoring implementation, `update_bitmap_score()` was called without
re-executing the corresponding queue entry. This meant that `trace_bits`
could belong to a different entry, leading to incorrect `top_rated` selection.

### Fix

This change ensures that each queue entry is executed before calling
`update_bitmap_score()` during rescoring. This guarantees that the
`trace_bits` used match the queue entry being rescored.

### DIFF
--- a/include/afl-fuzz.h
+++ b/include/afl-fuzz.h
@@ -1187,6 +1187,7 @@ void mark_as_redundant(afl_state_t *, struct queue_entry *, u8);
 void add_to_queue(afl_state_t *, u8 *, u32, u8);
 void destroy_queue(afl_state_t *);
 void update_bitmap_score(afl_state_t *, struct queue_entry *);
+void recalculate_all_scores(afl_state_t *);
 void cull_queue(afl_state_t *);
 u32  calculate_score(afl_state_t *, struct queue_entry *);
 
--- a/src/afl-fuzz-queue.c
+++ b/src/afl-fuzz-queue.c
@@ -938,6 +938,21 @@ void update_bitmap_score(afl_state_t *afl, struct queue_entry *q) {
 
 }
 
+void recalculate_all_scores(afl_state_t *afl) {
+
+  u8 *in_buf;
+
+  for (u32 i = 0; i < afl->queued_items; i++) {
+    if (likely(!afl->queue_buf[i]->disabled)) {
+      in_buf = queue_testcase_get(afl, afl->queue_buf[i]);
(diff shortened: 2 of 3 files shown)


5angjun:
This code path is enabled when the environment variable `AFL_CYCLE_SCHEDULES=1` is set

vanhauser-thc:
you are right, this is a bug.
your fix is however bad as it means the whole queue is executed again and again. this needs a better solution/fix :-)
also we only accept PRs to the dev branch.

5angjun:
Thanks for the feedback

You're right 
Re-executing the whole queue isn't ideal.

I'll work on a more efficient solution and update the PR accordingly.

And noted about the `dev` branch
I’ll make sure to rebase and open the PR there instead. Thanks for the heads-up!

vanhauser-thc:
This is my asap fix for the issue: [4ff2673](https://github.com/AFLplusplus/AFLplusplus/commit/4ff2673895c85ac6ac15dffc999e05a43d8d1521)
I still have to think about if this is good enough though (especially when a fast restart is happening), your input is appreciated :-)

5angjun:
Thanks for sharing the quick fix

However, it seems that the previous patch might not fully support the expected rescoring behavior.

I'd appreciate it if you could have a look at my patch.
https://github.com/AFLplusplus/AFLplusplus/pull/2366

kcwu:
This failed to compile with gcc.
```
src/afl-fuzz-queue.c: In function ‘recalculate_all_scores’:
src/afl-fuzz-queue.c:1014:36: error: passing argument 2 of ‘write_to_testcase’ from incompatible pointer type [-Wincompatible-pointer-types]
 1014 |       (void)write_to_testcase(afl, in_buf, afl->queue_buf[i]->len, 1);
      |                                    ^~~~~~
      |                                    |
      |                                    u8 * {aka unsigned char *}
In file included from src/afl-fuzz-queue.c:25:
include/afl-fuzz.h:1262:39: note: expected ‘void **’ but argument is of type ‘u8 *’ {aka ‘unsigned char *’}
 1262 | u32  write_to_testcase(afl_state_t *, void **, u32, u32);
      |                                       ^~~~~~~
```

vanhauser-thc:
fix is already in dev

PR #2366, opened 2025-04-09
fix: correct rescoring logic with minimal executions

5angjun:
Previous scoring logic did not correctly rescore all queue entries.

This patch ensures rescoring works under the updated scheduling logic, while minimizing executions per feedback from PR #2363.

Based on feedback from: https://github.com/AFLplusplus/AFLplusplus/pull/2363

Since `cycle_schedules` is closer to an optional feature than a default one,
I made sure to avoid modifying default-path functions to minimize any impact on performance.

### DIFF
--- a/src/afl-fuzz-state.c
+++ b/src/afl-fuzz-state.c
@@ -117,7 +117,6 @@ void afl_state_init(afl_state_t *afl, uint32_t map_size) {
   afl->virgin_crash = ck_alloc(map_size);
   afl->var_bytes = ck_alloc(map_size);
   afl->top_rated = ck_alloc(map_size * sizeof(void *));
-  afl->top_rated_candidates = ck_alloc(map_size * sizeof(u32));
   afl->clean_trace = ck_alloc(map_size);
   afl->clean_trace_custom = ck_alloc(map_size);
   afl->first_trace = ck_alloc(map_size);
--- a/src/afl-fuzz-state.c
+++ b/src/afl-fuzz-state.c
@@ -106,7 +106,8 @@ void afl_state_init(afl_state_t *afl, uint32_t map_size) {
   afl->switch_fuzz_mode = STRATEGY_SWITCH_TIME * 1000;
   afl->q_testcase_max_cache_size = TESTCASE_CACHE_SIZE * 1048576UL;
   afl->q_testcase_max_cache_entries = 64 * 1024;
-
+  afl->last_scored_idx = -1;
+  
 #ifdef HAVE_AFFINITY
   afl->cpu_aff = -1;                    /* Selected CPU core                */
 #endif                                                     /* HAVE_AFFINITY */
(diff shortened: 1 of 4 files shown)


vanhauser-thc:
I agree that this is the more sane approach, however this makes cycle_schedules pointless, because the overhead of replaying the queue is such an overhead that changing the schedule will just hurt overall fuzzing.

I think it makes more sense to deprecate this feature :-(

WDYT?

5angjun:
I agree with you to some extent.

As demonstrated in the AFLFast paper, I believe that scheduling in fuzzing provides meaningful benefits, especially in terms of coverage.

As you mentioned, replaying the queue can have an impact on the overall fuzzing process.
Therefore, I quantitatively measured the cost of replaying the queue once.

Based on a two-week fuzzing experiment using Unifuzz's target program, I analyzed the results and found that replaying items in the queue took approximately 6.6 seconds, calculated as:
```
queue items count / execs_per_sec = 6.6 sec
```
Given this, the overall impact of replaying the queue appears to be relatively small. So in my opinion, it's still reasonable to keep this feature.

vanhauser-thc:
ok, thanks for the feedback

src/afl-fuzz-state.c
```diff
@@ -116,6 +117,7 @@ void afl_state_init(afl_state_t *afl, uint32_t map_size) {
   afl->virgin_crash = ck_alloc(map_size);
   afl->var_bytes = ck_alloc(map_size);
   afl->top_rated = ck_alloc(map_size * sizeof(void *));
+  afl->top_rated_candidates = ck_alloc(map_size * sizeof(u32));
```
vanhauser-thc: as this is only used if the cycle schedules env var is set, this should only be allocated in that case, right?

vanhauser-thc:
I fuzz targets that have 2 exec/s and a queue of 10.000 - replaying the queue would take hours. but I would not use this env var, so ... :-)

src/afl-fuzz-state.c
5angjun: yes you are right
it is only used when cycle schedules env var is set

5angjun:
Yeah, I totally agree. It really depends on the target program and the fuzzing throughput. 
There’s no clear-cut answer, and it just comes down to what works best for each user.

In some setups, cycle_schedules might help improve coverage, but in others, it could just add unnecessary overhead

src/afl-fuzz-state.c
vanhauser-thc: so will you provide a fix? :-)

5angjun: No prolem I'll soon

5angjun: Fix applied as discussed
please check

vanhauser-thc:
thank you

PR #2369, opened 2025-04-10
add cleanup for top_rated_candidates in afl_state_deinit()

5angjun:
This PR adds missing cleanup logic for `top_rated_candidates`, which was not included in https://github.com/AFLplusplus/AFLplusplus/pull/2366.

When `cycle_schedules` is enabled, each entry in the `top_rated_candidates` array is dynamically allocated and should be properly freed in `afl_state_deinit()` to avoid memory leaks.

Thank you


vanhauser-thc:
honestly we do not care so much because the program then exits which frees the memory anyway but does not hurt :-)

PR #2375, opened 2025-04-12
code-format cache

kcwu:
This change reduces "make code-format" from 3 minutes to 3 seconds if large files are not changed.


kcwu:
this doesn't work if running `.custom-format.py` withtout `-i`. I will revise it.

vanhauser-thc:
looks good, thanks

PR #2382, opened 2025-04-16
fix build for "make PERFORMANCE=1 unit"

kcwu:


vanhauser-thc:
ah the units tests :-)

PR #2391, opened 2025-04-21
fix __AFL_COVERAGE: multiple definition of `__afl_selective_coverage`

### LINKED ISSUE #2390: __AFL_COVERAGE: multiple definition of `__afl_selective_coverage'
**Describe the bug**
I need to use `__AFL_COVERAGE_OFF` macros in more than one file, so I add `__AFL_COVERAGE()` in each file. However, `__AFL_COVERAGE` not only define function prototypes, it also define variable `__afl_selective_coverage`. Then the compiler complaint multiple definition of `__afl_selective_coverage`.

kcwu:
fix #2390


vanhauser-thc:
thanks!

PR #2395, opened 2025-04-23
fix infinite loop when custom mutator rejects smallest_favored

ryberger-nvidia:
When running with custom mutators, afl-fuzz delegates the responsibility of queuing to` afl_custom_queue_get` implemented by the mutator. If any mutator cannot process the input, then it is rejected. After an input is rejected then a new suitable item to queue must be found. Before this PR, that would be `smallest_favored`. However, if `smallest_favored` were rejected, it would not be cleared from its position as  `smallest_favored` meaning it would be attempted to be queued again catching afl-fuzz in an infinite loop.

To fix it, we simply return that we skipped the entry, along with using a `goto abandon_entry` to clean the entry up so that the fuzzer never considers the input again

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ Thank you! (For people sending pull requests - please add yourself to this list
     Ruben ten Hove                        Joey Jiao
     fuzzah                                @intrigus-lgtm
     Yaakov Saxon                          Sergej Schumilo
-    Ziqiao Kong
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
yes that solution is better.
for the future please send PRs to the dev branch, fixed that for you.

Issue #2394, opened 2025-04-23
LLVM 21 support?


oliness:
AFL++ only supports LLVM up to 19, and the bleeding edge is now 21. Can we support 21 easily? If it's a simple change I'm happy to work on it.

vanhauser-thc:
llvm 20 is supported I have not tried 21.
Note that if you make it work now it can still fail next week because it is not freezer yet afaik.
If you want to work on it - sure, send a pr :)

seanm:
I can confirm llvm 20.1.5 is working for me.

PR #2397, opened 2025-04-24
Initial commit for LLVM21 #2394

oliness:
Fixes compiler warnings when building against LLVM21. Uses opaque pointer types as per https://llvm.org/docs/OpaquePointers.html

### DIFF
--- a/GNUmakefile.llvm
+++ b/GNUmakefile.llvm
@@ -69,7 +69,7 @@ endif
 
 LLVM_STDCXX                  := gnu++11
 LLVM_LTO                     := 0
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
I doubt this will be the only change happening in llvm 21 :-) so expect more to come ...
Two things:
1. you have not touched SanitizerCoverageLTO.so.cc - likely it is affected too?
2. you are not working with ifdefs - are you sure this also works with llvm 14? (the oldest SanitizerCoveragePCGUARD supports)

vanhauser-thc:
@oliness ping

vanhauser-thc:
@oliness ping ...

vanhauser-thc:
closed for inactivity

PR #2398, opened 2025-04-25
Minor fixes

kcwu:


src/afl-fuzz.c
```diff
@@ -1555,13 +1555,6 @@ int main(int argc, char **argv_orig, char **envp) {
 
   #endif
 
-  // silently disable deterministic mutation if custom mutators are used
```
vanhauser-thc, line 1558: why are you removing this?

vanhauser-thc:
thanks for the nits, left one comment

src/afl-fuzz.c
kcwu: the same logic is also at line 2193.

vanhauser-thc: ok thanks. please write something like that next time you send a PR, I do not have 100% of the code in my head, plus multiple people have written it - this saves the back and forth :)

PR #2403, opened 2025-04-26
fix power schedules

kcwu:
AFLFast power schedules regressed since v4.31c


vanhauser-thc:
thanks!
I think there are more fixes needed though, will have a thorough look after merging

kcwu:
yeah, I feel the same (need more fixes) in `save_if_interesting`. Anyway, this fix can already recover most of the regression according to my preliminary testing.

PR #2406, opened 2025-04-28
Define WORD_SIZE_64 for more 64-bit arches

maribu:
This enables 64-bit detection for the following additional systems:

- [PowerPC64 (little endian)](https://en.wikipedia.org/wiki/Ppc64)
- [S390x](https://en.wikipedia.org/wiki/S390x)
- [LoongArch64](https://en.wikipedia.org/wiki/LoongArch64)


vanhauser-thc:
thank you!

PR #2409, opened 2025-04-29
Add missing override directive for CFLAGS+= in libdislocator

Scott-Guest:
libdislocator fails to build with a user-provided `CFLAGS`, e.g.
```
make sources-only CFLAGS='-O1'
```
reports
```
...
make[1]: Entering directory '/home/sguest/src/AFLplusplus/utils/libdislocator'
cc -O1  -shared -fPIC libdislocator.so.c -o libdislocator.so 
libdislocator.so.c:76:10: fatal error: config.h: No such file or directory
   76 | #include "config.h"
      |          ^~~~~~~~~~
compilation terminated.
make[1]:` *** [Makefile:31: libdislocator.so] Error 1
...
```
This is fixed by adding an `override` directive to the appropriate `CFLAGS += -I ...`

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ Thank you! (For people sending pull requests - please add yourself to this list
     fuzzah                                @intrigus-lgtm
     Yaakov Saxon                          Sergej Schumilo
     Ziqiao Kong                           Ryan Berger
-    Sangjun Park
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
thanks

PR #2411, opened 2025-05-03
Disable GCC instrumentation for AFL_SAN_NO_INST

Scott-Guest:
AFL_SAN_NO_INST currently only disables the LLVM instrumentation. This guards the GCC plugin load as well.


Scott-Guest:
@wtdcode

wtdcode (approved):
Thanks for catching this =).

vanhauser-thc:
Thanks!

PR #2412, opened 2025-05-04
Environment variable to discriminate the target and the forkserver

alexandredoyen29:
Hi !

A tiny PR that could be useful for persons that use AFL_PRELOAD to LD_PRELOAD some .so before launching the target.

In particular, when using a preload that disable `fork()` calls, the fork server does not work :
```
[-] PROGRAM ABORT : Unable to request new process from fork server (OOM?)
         Location : afl_fsrv_run_target(), src/afl-forkserver.c:1949
```

So, the idea is to be able to create a preload that has this form :
```C
pid_t fork(void)
{
    if (getenv("AFL_I_AM_THE_FORKSERVER") == NULL)
    {
        return 0;
    }
    else
    {
        return real_fork();
    }
}
```

Have a nice day,
Alexandre


domenukk:
`AFL_FORKSERVER_PARENT`?

domenukk:
Also, setting an env variable every forkserver execution is an overhead, couldn't this also be a global variable? I guess you need some way to read the value from the ldpreloaded lib, but could that be a method or something? Just throwing around ideas. In the end it's still a negligible overhead in contrast to the overhead for the actual `fork` call..

alexandredoyen29:
> `AFL_FORKSERVER_PARENT`?

If you want...



> Also, setting an env variable every forkserver execution is an overhead, couldn't this also be a global variable? I guess you need some way to read the value from the ldpreloaded lib, but could that be a method or something? Just throwing around ideas. In the end it's still a negligible overhead in contrast to the overhead for the actual `fork` call..

Yep, it could be an idea. I don't know if changing the value of an environment variable is less expensive than setting and unsetting a new one...

vanhauser-thc:
Some comments:

IMHO the env name is misleading, I would expect it to work the other way around.

For a niche feature we introduce something into the code that is activated for every usage.
unsetenv and setenv are also not cheap calls.
if there is an `if (getenv("FOOBAR") __afl_setenv_fork = 1;` and in the routines that check in front of the getenv/setenv is an `if (unlikely( __afl_setenv_fork)) setenv("");` this would be better, and then I would consider merging this.

alexandredoyen29:
> Some comments:
> 
> IMHO the env name is misleading, I would expect it to work the other way around.
> 
> For a niche feature we introduce something into the code that is activated for every usage. unsetenv and setenv are also not cheap calls. if there is an `if (getenv("FOOBAR") __afl_setenv_fork = 1;` and in the routines that check in front of the getenv/setenv is an `if (unlikely( __afl_setenv_fork)) setenv("");` this would be better, and then I would consider merging this.

OK, I agree with your POV.
Otherwise, we could not do something like this ?
```C
if (unlikely((getenv("AFL_PRELOAD") != NULL)) setenv(...)
if (unlikely((getenv("AFL_PRELOAD") != NULL)) unsetenv(...)
```

vanhauser-thc:
> ```c
> if (unlikely((getenv("AFL_PRELOAD") != NULL)) setenv(...)
> if (unlikely((getenv("AFL_PRELOAD") != NULL)) unsetenv(...)
> ```

no because - again - setenv and unsetenv are very expensive. set an early global variable in alf-compiler-rt.o.c and check on that.
`if (unlikely(__afl_fork_foobar)) setenv(...);`
also impacting all AFL_PRELOAD usages is also maybe not the right move. I think a specific env var makes more sense.

instrumentation/afl-compiler-rt.o.c
```diff
@@ -1046,6 +1046,11 @@ static void __afl_start_forkserver(void) {
       /* In child process: close fds, resume execution. */
 
       if (unlikely(!child_pid)) {  // just to signal afl-fuzz faster
+        if (unlikely(getenv("AFL_PRELOAD_DISCRIMINATE_FORKSERVER_PARENT") != NULL)) {
```
vanhauser-thc: I am not sure we have a language barrier here? for the third time:
I will never merge
`if (unlikely(getenv("AFL_PRELOAD_DISCRIMINATE_FORKSERVER_PARENT") != NULL))`
I will consider merging if you do
`if (unlikely(__afl_forkserver_setenv))` instead and set that variable e.g. at the start of the function.

alexandredoyen29: Oh sorry, I didn't fully understand what you want.
You want I do the `getenv()` at the beginning, and store the result in `__afl_forkserver_setenv`[1], and check back in the `if` ?

[1]:
```C
if (getenv("AFL_PRELOAD_DISCRIMINATE_FORKSERVER_PARENT")) __afl_forkserver_setenv = 1;
```

vanhauser-thc: yes

alexandredoyen29:
For information, why do you expect to use auxiliary variable instead of directly checking the result of `getenv()` ?

instrumentation/afl-compiler-rt.o.c
```diff
@@ -1046,6 +1054,11 @@ static void __afl_start_forkserver(void) {
       /* In child process: close fds, resume execution. */
 
       if (unlikely(!child_pid)) {  // just to signal afl-fuzz faster
+        if (unlikely(__afl_forkserver_setenv)) {
```
vanhauser-thc: yes that is how it should be. can you please move the `if` to just before the return though? the signal to afl-fuzz is delayed otherwise and the unsetenv is not time critical.

alexandredoyen29: Between `close(FORKSRV_FD + 1);` and `return;` ?

src/afl-forkserver.c
```diff
@@ -867,6 +868,11 @@ void afl_fsrv_start(afl_forkserver_t *fsrv, char **argv,
 
   }
 
+  if (getenv("AFL_PRELOAD_DISCRIMINATE_FORKSERVER_PARENT") != NULL)
```
vanhauser-thc: can you move this to afl_fsrv_init please and add the variable to the afl_forkserver_t struct?

instrumentation/afl-compiler-rt.o.c
vanhauser-thc: yes

vanhauser-thc:
and once everything is done the env var has to be added to include/envs.h and described in in docs/env_variables.md :)

vanhauser-thc:
thanks!

PR #2413, opened 2025-05-04
add afl-cmin.py

kcwu:
I added afl-cmin.py first and didn't replace the awk and bash version yet. Maybe we can switch to python version after more positive feedback from users.

### DIFF
--- a/afl-cmin.py
+++ b/afl-cmin.py
@@ -54,130 +54,156 @@ except ImportError:
 parser = argparse.ArgumentParser()
 
 cpu_count = multiprocessing.cpu_count()
-group = parser.add_argument_group('Required parameters')
-group.add_argument('-i',
-                   dest='input',
-                   action='append',
-                   metavar='dir',
-                   required=True,
-                   help='input directory with the starting corpus')
-group.add_argument('-o',
-                   dest='output',
-                   metavar='dir',
-                   required=True,
-                   help='output directory for minimized files')
-
-group = parser.add_argument_group('Execution control settings')
-group.add_argument('-f',
-                   dest='stdin_file',
-                   metavar='file',
-                   help='location read by the fuzzed program (stdin)')
+group = parser.add_argument_group("Required parameters")
 group.add_argument(
-    '-m',
-    dest='memory_limit',
-    default='none',
-    metavar='megs',
-    type=lambda x: x if x == 'none' else int(x),
-    help='memory limit for child process (default: %(default)s)')
-group.add_argument('-t',
-                   dest='time_limit',
-                   default=5000,
-                   metavar='msec',
-                   type=lambda x: x if x == 'none' else int(x),
-                   help='timeout for each run (default: %(default)s)')
-group.add_argument('-O',
-                   dest='frida_mode',
-                   action='store_true',
-                   default=False,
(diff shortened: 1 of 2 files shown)


afl-cmin.py
```diff
@@ -0,0 +1,687 @@
...
+                   required=True,
+                   help='output directory for minimized files')
+
+group = parser.add_argument_group('Execution control settings')
+group.add_argument('-f',
+                   dest='stdin_file',
+                   metavar='file',
+                   help='location read by the fuzzed program (stdin)')
+group.add_argument(
+    '-m',
```
domenukk: Can you run `black` to format the python?

afl-cmin.py
```diff
@@ -0,0 +1,687 @@
...
+    dest='memory_limit',
+    default='none',
+    metavar='megs',
+    type=lambda x: x if x == 'none' else int(x),
```
domenukk: -m 0 should probably also be none (right?)

afl-cmin.py
domenukk: ah wait looks below like you used a formatter(?) 
then why is this indentation so weird here?

afl-cmin.py
```diff
@@ -0,0 +1,687 @@
...
+    if args.crash_only:
+        env['AFL_CMIN_CRASHES_ONLY'] = '1'
+    if args.allow_any:
+        env['AFL_CMIN_ALLOW_ANY'] = '1'
+
+    if input_from_file:
+        p = subprocess.Popen(cmd,
+                             stdout=subprocess.PIPE,
+                             env=env,
+                             bufsize=1048576)
```
domenukk: Maybe the value should be a global/const, especially since you use it twice

vanhauser-thc:
IMHO I would move directly for afl-cmin to call afl-cmin.py if afl-cmin.py can be executed successfully.
Testing is what the dev branch is for :-)

afl-cmin.py
kcwu: Both `afl-cmin` and `afl-cmin.bash` don't do so. I think it is not bad to keep the old behavior.

afl-cmin.py
kcwu: it was formatted by yapf. Anyway, I reformatted it using black and the indent looks better.

afl-cmin.py
kcwu: I didn't do so because the value is not so important IMHO.

kcwu:
> IMHO I would move directly for afl-cmin to call afl-cmin.py if afl-cmin.py can be executed successfully. Testing is what the dev branch is for :-)

done

afl-cmin.py
vanhauser-thc: 0 and none is the same and it is the default

vanhauser-thc:
thanks!

vanhauser-thc:
I will make some changes to afl-cmin and also actually install afl-cmin.py on make install ;-)

vanhauser-thc:
4d9d8aaf163e8d8546abf6b0d9e85eee3140c02b

afl-cmin.py
domenukk: I mean you use it twice, it cries for a variable. That way you can also make clear, why you set it to that number

PR #2418, opened 2025-05-09
Do not include NUL when memmem

Evian-Zhang:
Only programs written by C has strings that terminated with NUL. If the instrumentation logic is written in other languages like Rust, the string is not NUL terminated, thus fails all such comparisons.


vanhauser-thc:
well I would expect that in those languages people add a null byte. but sure, it does not hurt to ignore the null byte either. thanks.

PR #2419, opened 2025-05-09
Allow cmplog with unicorn mode

Evian-Zhang:
If AFLplusplus/unicornafl#37 is merged, then we shall use cmplog with unicorn-mode. Removed certain check preventing this.

Moreover, I found that although `qemu_mode`, `frida_mode`, `cs_mode` are defined as fields in forkserver struct, `unicorn_mode` is defined directly in afl struct. This is strange and makes it hard to check unicorn_mode in forkserver logic. This PR also moves unicorn_mode to be field of forkserver struct, making it more consistent.

### DIFF
--- a/include/afl-fuzz.h
+++ b/include/afl-fuzz.h
@@ -587,7 +587,6 @@ typedef struct afl_state {
       no_arith,                         /* Skip most arithmetic ops         */
       shuffle_queue,                    /* Shuffle input queue?             */
       bitmap_changed,                   /* Time to update bitmap?           */
-      unicorn_mode,                     /* Running in Unicorn mode?         */
       use_wine,                         /* Use WINE with QEMU mode          */
       skip_requested,                   /* Skip request, via SIGUSR1        */
       run_over10m,                      /* Run time over 10 minutes?        */
--- a/include/forkserver.h
+++ b/include/forkserver.h
@@ -153,6 +153,8 @@ typedef struct afl_forkserver {
 
   bool cs_mode;                      /* if running in CoreSight mode or not */
 
+  bool unicorn_mode;                   /* if running in unicorn mode or not */
+
   bool use_stdin;                       /* use stdin for sending data       */
 
   bool no_unlink;                       /* do not unlink cur_input          */
--- a/src/afl-analyze.c
+++ b/src/afl-analyze.c
@@ -78,6 +78,7 @@ static u8 *target_path;
 static u8  frida_mode;
 static u8  qemu_mode;
 static u8  cs_mode;
+static u8  unicorn_mode;
 static u32 map_size = MAP_SIZE;
 
 static afl_forkserver_t fsrv = {0};   /* The forkserver                     */
@@ -813,7 +814,7 @@ static void usage(u8 *argv0) {
 int main(int argc, char **argv_orig, char **envp) {
 
   s32    opt;
-  u8     mem_limit_given = 0, timeout_given = 0, unicorn_mode = 0, use_wine = 0;
+  u8     mem_limit_given = 0, timeout_given = 0, use_wine = 0;
   char **use_argv;
   char **argv = argv_cpy_dup(argc, argv_orig);
 
@@ -960,6 +961,7 @@ int main(int argc, char **argv_orig, char **envp) {
         if (!mem_limit_given) { mem_limit = MEM_LIMIT_UNICORN; }
 
         unicorn_mode = 1;
+        fsrv.unicorn_mode = unicorn_mode;
         fsrv.mem_limit = mem_limit;
         break;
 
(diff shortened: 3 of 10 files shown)


domenukk:
Do we really need a specialized unicorn mode? IIRC it changes basically nothing but set one or two env variables, specifically AFL_SKIP_BIN_CHECK (right?)

domenukk:
Ah this PR adds a bunch of other things, I see...
Other option: can we merge some flags for binary-only instrumentation into one? Does qemu_mode or unicorn mode (or frida mode?) really make a difference for the forkserver? Ideally it should not

Evian-Zhang:
This question is really out of my knowledge. Hope someone could answer this.

wtdcode:
I can take a look but allow me a few days since still on travelling

wtdcode:
Ideally, we do not need any special handling for unicornafl as said previously. It will act like any forkserver targets, like those compiled by afl-cc.

Evian-Zhang:
AFL++ internally has many checks to avoid users from composing unsupported features. For example, when using afl-forkserver, only qemu-mode would use cmplog, where other mode would fail the check (afl-cc use another code so not checked at that location). As a result, it may not be "special handling", it's just some misuse checks that require explicit unicorn mode set.

wtdcode:
> AFL++ internally has many checks to avoid users from composing unsupported features. For example, when using afl-forkserver, only qemu-mode would use cmplog, where other mode would fail the check (afl-cc use another code so not checked at that location). As a result, it may not be "special handling", it's just some misuse checks that require explicit unicorn mode set.

I mean, AFL++ can remove unicorn mode, i.e. “-U”, totally.

domenukk:
For example, I would prefer a "forkserver->force_cmplog" flag that both qemu and unicorn mode set, rather than a bunch of obscure feature flags sprinkled around the codebase. It gets very spaghetti otherwise

domenukk:
unicorn mode likely still needs SKIP_BINCHECK in comparison to afl-cc binaries

wtdcode:
> I mean -U likely still needs SKIP_BINCHECK

No, current uncornafl v3 pass that seamlessly thanks to libafl fsrv implementation

Evian-Zhang:
There may be some confusions here. I totally agree that force_cmplog flag is better, but without inferring unicorn mode by -U cmdline option, how can we set that flag accordingly? At least for now, we do not have a mechanism of guessing target instrumentation by user-provided binaries.

wtdcode:
The idea is that uncornafl spins up fsrv like a normal afl-cc or libafl_cc instrumented binary so that it is very transparent to afl++. Unicornafl only needs to notify afl++ a crash is found by aborting intentioanlly.

wtdcode:
> There may be some confusions here. I totally agree that force_cmplog flag is better, but without inferring unicorn mode by -U cmdline option, how can we set that flag accordingly? At least for now, we do not have a mechanism of guessing target instrumentation by user-provided binaries.

The point is that “unicorn mode” is not necessary. Unicornafl can mock as a pretty normal fsrv targets.

Evian-Zhang:
OK, that makes sense.

Evian-Zhang:
> > I mean -U likely still needs SKIP_BINCHECK
> 
> No, current uncornafl v3 pass that seamlessly thanks to libafl fsrv implementation

It suddenly occurred to me that, if we provide python binding of new unicornafl, then the bin check should be skipped, otherwise the `python` executable will be checked? 🤔

wtdcode:
Good point. I need to check code and see. Maybe we simply alias the qemu mode or so?

________________________________

wtdcode:
> > > I mean -U likely still needs SKIP_BINCHECK
> > 
> > 
> > No, current uncornafl v3 pass that seamlessly thanks to libafl fsrv implementation
> 
> It suddenly occurred to me that, if we provide python binding of new unicornafl, then the bin check should be skipped, otherwise the `python` executable will be checked? 🤔

I looked into the code and we can reuse QEMU mode to skip bin checks.

vanhauser-thc:
@wtdcode please ping me (repeatedly if necessary :-) ) when this is ready to be merged

wtdcode:
> @wtdcode please ping me (repeatedly if necessary :-) ) when this is ready to be merged

This PR may no longer be needed (cc @Evian-Zhang) 

For harness building with new unicornafl in Rust/C/C++, nothing special is needed: https://github.com/AFLplusplus/unicornafl/blob/ea475ce52594ad256e602823f8b2f3d25a295904/.github/workflows/ci.yaml#L74

For harness building witn new unicornafl in Python, I would suggest aliasing unicorn mode to qemu mode to skip bin check. Not really tested yet, but I checked the code and it should work, or with a few lines.

Evian-Zhang:
I agree. @vanhauser-thc Feel free to close this PR if you also think this is unnecessary :)

vanhauser-thc:
I cannot comment if this PR is needed or not - you are the experts :-)
At least it needs a PR to update the unicornafl checkout

wtdcode:
> I cannot comment if this PR is needed or not - you are the experts :-) At least it needs a PR to update the unicornafl checkout

Will let you know once I have tested the python cases.

Evian-Zhang:
I would close this PR. I think the new unicornafl checkout shall be updated after AFLplusplus/unicornafl#43 is merged.

PR #2421, opened 2025-05-12
Rename `AFL_SAN_NO_INST` to `AFL_FSRV_ONLY`

wtdcode:
## Motivation

Recently I'm building another fuzzer which needs to spin up a quite slow binary repeatedly without instrumentation, i.e. just running the binary and collecting outputs elsewhere. This reminds me of the exactly same goal of `AFL_SAN_NO_INST` and I find the name actually is quite misleading. Therefore, this PR renames it to reflect its actual functionality: `AFL_FSRV_ONLY`. I also make SAND tolerate normally built sanitizer enabled binaries.

## Highlights

- Rename `AFL_SAN_NO_INST` to `AFL_FSRV_ONLY`
- Documents of `AFL_FSRV_ONLY` are added as well to illustrate the use cases.
- Now binaries instrumented by sanitizers and AFL++, i.e. without `AFL_FSRV_ONLY` or `AFL_SAN_NO_INST` can be used with SAND too.
- To avoid break changes, `AFL_SAN_NO_INST` now goes to `afl_environment_deprecated` and becomes an alias of `AFL_FSRV_ONLY`


wtdcode:
CI seems dead after 6 hours, is it related to my changes?!

wtdcode:
I can reproduce this locally and it dies on `afl-cmin.py`

```
(default_venv) mio@station ~/o/A/test (rename-afl-san-no-inst)> python3 ../afl-cmin.py -m none -i in -o in2 -- ./test-instr.plain
Hint: install python module "tqdm" to show progress bar
2025-05-13 00:06:22,388 - INFO - use 1 workers (-T)
2025-05-13 00:06:22,389 - INFO - Found 3 input files in 1 directories
2025-05-13 00:06:22,393 - INFO - Remain 3 files after dedup
2025-05-13 00:06:22,393 - INFO - Sorting files.
2025-05-13 00:06:22,395 - INFO - Setting AFL_MAP_SIZE=19
2025-05-13 00:06:22,395 - INFO - Testing the target binary
2025-05-13 00:06:22,398 - INFO - ok, 5 tuples recorded
  File "/home/mio/opensource/AFLplusplus/test/../afl-cmin.py", line 728, in <module>
    main()
  File "/home/mio/opensource/AFLplusplus/test/../afl-cmin.py", line 577, in main
    jobs = list(itertools.batched(enumerate(files), chunk))
AttributeError: module 'itertools' has no attribute 'batched'

```

@kcwu thoughts?

wtdcode:
https://docs.python.org/3/library/itertools.html#itertools.batched

So this requires py3.12+

wtdcode:
Okay, should be fixed by backporting implementation from `more-itertools`

vanhauser-thc:
thanks for fixing afl-cmin.py - weird that it fails the CI because it afl-cmin.py does not execute correctly the afl-cmin script should then execute the awk script.

For `AFL_FSRV_ONLY` I am not that sold on the naming yet. To me it could also mean an option for afl-fuzz.
I have to think about it. AFL_NO_INSTRUMENTATION is long ... maybe AFL_LLVM_NO_INST (and AFL_GCC_NO_INST)...

wtdcode:
> thanks for fixing afl-cmin.py - weird that it fails the CI because it afl-cmin.py does not execute correctly the afl-cmin script should then execute the awk script.
> 
> For `AFL_FSRV_ONLY` I am not that sold on the naming yet. To me it could also mean an option for afl-fuzz. I have to think about it. AFL_NO_INSTRUMENTATION is long ... maybe AFL_LLVM_NO_INST (and AFL_GCC_NO_INST)...

Note there is already `AFL_DISABLE_LLVM_INSTRUMENTATION`, which further disables forkserver injection as documented. To avoid confusion, I would suggest `AFL_NO_PC_INST`. What do you think?

vanhauser-thc:
> > thanks for fixing afl-cmin.py - weird that it fails the CI because it afl-cmin.py does not execute correctly the afl-cmin script should then execute the awk script.
> > For `AFL_FSRV_ONLY` I am not that sold on the naming yet. To me it could also mean an option for afl-fuzz. I have to think about it. AFL_NO_INSTRUMENTATION is long ... maybe AFL_LLVM_NO_INST (and AFL_GCC_NO_INST)...
> 
> Note there is already `AFL_DISABLE_LLVM_INSTRUMENTATION`, which further disables forkserver injection as documented. To avoid confusion, I would suggest `AFL_NO_PC_INST`. What do you think?

you see, there is so much that I forget about things :-)
PC_INST is only partially true because it should also affect gcc plugin, lto, classic etc. and these do not do pc guard. maybe AFL_LLVM_ONLY_FSRV?

wtdcode:
Done. It is common to forget features for such a fancy code base =)

vanhauser-thc:
When we do it for LLVM, we also need it for GCC too, right?

wtdcode:
Done with `AFL_GCC_ONLY_FSRV`

vanhauser-thc:
LGTM, thanks!!

PR #2423, opened 2025-05-13
show stats more frequently when sync foreign

kcwu:
otherwise, the stats might have no updates for hours for large foreign directory

### DIFF
--- a/src/afl-fuzz-init.c
+++ b/src/afl-fuzz-init.c
@@ -658,12 +658,8 @@ void read_foreign_testcases(afl_state_t *afl, int first) {
         munmap(mem, st.st_size);
         close(fd);
 
-        if (st.st_mtime > mtime_max) {
-
-          mtime_max = st.st_mtime;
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
I agree that it can take a long time until stats are updated. this change would update the stats way too often though, like many times per second for fast targets.

we could do, e.g.:
```
time_t last_print = time(NULL);
for (...) {

...

  time_t now = time(NULL);
  if (now - last_print >= 1) {
    last_print = now;
    show_stats(afl);
  }
}
```

WDYT?

kcwu:
I think we can just rely on `show_stats`'s check https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-fuzz-stats.c#L717

vanhauser-thc:
there is quite some stuff happening in show_stats before the check, but as this is in the foreign sync and only one instance should do that in a multi-instance campaign the impact is minimal.
thanks, I will merge this

PR #2425, opened 2025-05-14
push to stable

vanhauser-thc:


vanhauser-thc:
if you can pls have a look dominik :)

afl-cmin
```diff
@@ -1,12 +1,19 @@
 #!/usr/bin/env sh
+THISPATH=`dirname ${0}`
+
+# call afl-cmin.py if it can be executed successfully.
+if $THISPATH/afl-cmin.py --help > /dev/null 2>&1; then
+  exec $THISPATH/afl-cmin.py "$@"
+fi
```
domenukk, line 7: Should we inform the user about the possiblity of the pyhon script here?
Also, when/why would this ever fail?

src/afl-fuzz-bitmap.c
```diff
@@ -494,12 +499,11 @@ u8 __attribute__((hot)) save_if_interesting(afl_state_t *afl, void *mem,
   u8  fn[PATH_MAX];
   u8 *queue_fn = "";
   u8  new_bits = 0, keeping = 0, res, is_timeout = 0, need_hash = 1;
+  u8  classified = 0;
```
domenukk, line 502: Why not use bools here as we do in other places?

src/afl-fuzz-init.c
```diff
@@ -653,17 +653,14 @@ void read_foreign_testcases(afl_state_t *afl, int first) {
         u32 len = write_to_testcase(afl, (void **)&mem, st.st_size, 1);
         fault = fuzz_run_target(afl, &afl->fsrv, afl->fsrv.exec_tmout);
         afl->syncing_party = foreign_name;
+        afl->foreign_file = nl[i]->d_name;
```
domenukk, line 654: This is a use after free, right? `nl[i]` is freed in line 592 above

afl-cmin
vanhauser-thc: if dependencies are not installed - or python is not present :)
I don't think we need to inform users IMHO

src/afl-fuzz-init.c
vanhauser-thc: ah good catch!

vanhauser-thc: fix in [bedb38e](https://github.com/AFLplusplus/AFLplusplus/pull/2425/commits/bedb38e2168c5a6abddd765ea354fb6e44604479)

src/afl-fuzz-init.c
```diff
@@ -589,8 +589,6 @@ void read_foreign_testcases(afl_state_t *afl, int first) {
         u8 *fn2 =
             alloc_printf("%s/%s", afl->foreign_syncs[iter].dir, nl[i]->d_name);
 
-        free(nl[i]);                                         /* not tracked */
```
domenukk, line 592: Couldn't you just have kept the pointer to `d_name` in a var here and kept the free here?
Or ar there other things that can happen?
Also, what about d_name, should it be freed or is it part of the struct or who owns it?

afl-cmin
domenukk: Not sure what would happen with `exec` if python isn't installed?

src/afl-fuzz-init.c
vanhauser-thc: you can not continue to use `nl[i]->d_name` once you have free'd `nl[i]`.
d_name seems not needed to be free'd - this is from the scandir man page:
```
       int
       main(void)
       {
           struct dirent **namelist;
           int n;

           n = scandir(".", &namelist, NULL, alphasort);
           if (n == -1) {
               perror("scandir");
               exit(EXIT_FAILURE);
           }

           while (n--) {
               printf("%s\n", namelist[n]->d_name);
               free(namelist[n]);
           }
           free(namelist);

           exit(EXIT_SUCCESS);
       }
```

domenukk: You're right, sorry

afl-cmin
domenukk: I would just get rid of the awk version and worst case fall back to bash, but we can do that later ;)

vanhauser-thc: if python is not installed the afl-cmin.py --help will fail and then the awk is used.
we should first ensure for 2-3 releases that things run fine with the python script before we remove the awk one. there was already an issue that it did not work for python < 3.12 :-)

PR #2426, opened 2025-05-15
Fix resume for syncing

### LINKED ISSUE #2416: reverse input when resume
**Describe the bug**
Currently, when resume in place, the corpus will be added into queue in reversed order. Why reverse?

https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-fuzz-init.c#L760

The reversing introduced some issues:
 1. in the case of parallel fuzzing, other fuzzers assume the queue order is fixed. If reversed the queue, some corpus may be missed by other fuzzers.
 2. with fastresume, the queue metadata will be back filled. But the into incorrect queue item.

vanhauser-thc:


kcwu:
Looks good to me. BTW, the change to compiler issue is also included in this PR.

PR #2430, opened 2025-05-15
Add env variable SHM_FUZZ_MAP_SIZE

domenukk:
The SHM_FUZZ_MAP_SIZE env variable tells the forkserver about the size the max shm input page has.

This is related to the discussion in:
https://github.com/AFLplusplus/LibAFL/pull/3229

The same is done in LibAFL in
https://github.com/AFLplusplus/LibAFL/pull/3235

### DIFF
--- a/instrumentation/afl-compiler-rt.o.c
+++ b/instrumentation/afl-compiler-rt.o.c
@@ -296,13 +296,13 @@ static void __afl_map_shm_fuzz() {
 #ifdef USEMMAP
 
     // Newer afl-fuzz versions will set a shm_fuzz page size env, else fall back
-    size_t shm_fuzz_page_size = MAX_FILE + sizeof(u32);
-    char *page_size_env = getenv(SHM_FUZZ_PAGE_SIZE_ENV_VAR);
-    if (page_size_env != NULL) {
+    size_t shm_fuzz_map_size = MAX_FILE + sizeof(u32);
+    char *map_size_env = getenv(SHM_FUZZ_MAP_SIZE_ENV_VAR);
+    if (map_size_env != NULL) {
       char* endptr;
       errno = 0;
-      shm_fuzz_page_size = (size_t)strtoul(page_size_env, &endptr, 10);
-      if (errno != 0 || shm_fuzz_page_size == 0) {
+      shm_fuzz_map_size = (size_t)strtoul(map_size_env, &endptr, 10);
+      if (errno != 0 || shm_fuzz_map_size == 0) {
         perror("shm_fuzz page size parsing");
         send_forkserver_error(FS_ERROR_SHM_OPEN);
         _exit(1);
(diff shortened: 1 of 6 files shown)


instrumentation/afl-compiler-rt.o.c
```diff
@@ -308,7 +323,7 @@ static void __afl_map_shm_fuzz() {
     }
 
     map =
-        (u8 *)mmap(0, MAX_FILE + sizeof(u32), PROT_READ, MAP_SHARED, shm_fd, 0);
+        (u8 *)mmap(0, shm_fuzz_page_size, PROT_READ, MAP_SHARED, shm_fd, 0);
```
wtdcode: If this represents the length of input, why not just name it something like `AFL_SHM_FUZZ_FILE_SIZE`? The page size might be confusing and potentially suggest to users that it should be only the host page size.

domenukk: It's not the input size, it's the input size *plus* an int indicating the actual size of the input. So it represent the underlying page.
Happy to use a better name.

domenukk: shm_fuzz_map_size? mapping_size?

wtdcode: I will go for `AFL_SHM_FUZZ_MAP_SIZE`. Sounds much clearer =).

src/afl-fuzz-init.c
```diff
@@ -2915,11 +2915,16 @@ void setup_testcase_shmem(afl_state_t *afl) {
   afl->shm_fuzz = ck_alloc(sizeof(sharedmem_t));
 
   // we need to set the non-instrumented mode to not overwrite the SHM_ENV_VAR
-  u8 *map = afl_shm_init(afl->shm_fuzz, MAX_FILE + sizeof(u32), 1);
+  size_t shm_fuzz_page_size = MAX_FILE + sizeof(u32);
+  u8 *map = afl_shm_init(afl->shm_fuzz, shm_fuzz_page_size, 1);
```
wtdcode: `afl_shm_init` is used at multiple places, if I remember it correctly. I think it would be better to overwrite the map size inside `afl_shm_init` for consistency?

wtdcode: Oh so this is the initial map size and we take the sizes from forkserver afterwards?

wtdcode: Oh, wait, that's coverage map size. Silly me =p.

instrumentation/afl-compiler-rt.o.c
```diff
@@ -294,6 +294,21 @@ static void __afl_map_shm_fuzz() {
     u8 *map = NULL;
 
 #ifdef USEMMAP
+
+    // Newer afl-fuzz versions will set a shm_fuzz page size env, else fall back
+    size_t shm_fuzz_map_size = SHM_FUZZ_MAP_SIZE_DEFAULT;
```
vanhauser-thc, line 299: so we only need this when using USEMMAP? so shm_open does not need that? just to be sure.

domenukk: Looks like it, as far as I understand shm_open just opens the full mapping, whatever size it has

wtdcode: `shm_open` (on Linux) is capped by the `tmpfs` without `ftruncate`. So it can be configured by the kernel directly.

PR #2433, opened 2025-05-16
skip known case if the file is actually coming from us

kcwu:
Assume we have one main node and N secondary nodes in a parallel fuzzing campaign. Every time the main node found a new case, the case will be synced to all secondary nodes. Later when the main node sync, the main node need to run the file again to see if the file is interesting because they are "new" cases on the secondary nodes.

In other words, for one new case, the main node has to run the redundent test N times. This is wasteful and slowed down the progress of main node.

The wasteful issue on secondary nodes is acceptable because we can run more secondary nodes to mitigate the inefficiency. OTOH, increasing the number of secondary nodes slow down the main node further.

### DIFF
--- a/src/afl-fuzz-run.c
+++ b/src/afl-fuzz-run.c
@@ -699,6 +699,35 @@ abort_calibration:
 
 }
 
+bool is_known_case(afl_state_t *afl, u8 *name, void *mem, u32 len) {
+
+  int sync_id_pos;
+  u32 id, src_id;
+  if (sscanf(name, "id:%06u,sync:%n%*[^,],src:%06u", &id, &sync_id_pos,
+             &src_id) != 2)
+    return false;
+
+  if (strncmp(name + sync_id_pos, afl->sync_id, strlen(afl->sync_id)) != 0)
+    return false;
+  if (name[sync_id_pos + strlen(afl->sync_id)] != ',') return false;
+  if (src_id < 0 || src_id >= afl->queued_items) return false;
+
+  struct queue_entry *q = afl->queue_buf[src_id];
+  if (q->len != len) return false;
+
+  if (q->testcase_buf) return memcmp(q->testcase_buf, mem, len) == 0;
+
+  int fd = open((char *)q->fname, O_RDONLY);
+  if (fd < 0) return false;
+  u8 *buf = malloc(len);
+  ck_read(fd, buf, len, q->fname);
+  close(fd);
+  bool result = (memcmp(buf, mem, len) == 0);
+  free(buf);
+  return result;
+
+}
+
 /* Grab interesting test cases from other fuzzers. */
 
 void sync_fuzzers(afl_state_t *afl) {
@@ -896,26 +925,30 @@ void sync_fuzzers(afl_state_t *afl) {
 
         if (mem == MAP_FAILED) { PFATAL("Unable to mmap '%s'", path); }
 
-        /* See what happens. We rely on save_if_interesting() to catch major
-           errors and save the test case. */
+        if (!is_known_case(afl, namelist[o]->d_name, mem, st.st_size)) {
 
-        u32 new_len = write_to_testcase(afl, (void **)&mem, st.st_size, 1);
+          /* See what happens. We rely on save_if_interesting() to catch major
+             errors and save the test case. */
 
-        fault = fuzz_run_target(afl, &afl->fsrv, afl->fsrv.exec_tmout);
+          u32 new_len = write_to_testcase(afl, (void **)&mem, st.st_size, 1);
 
-        if (afl->stop_soon) {
+          fault = fuzz_run_target(afl, &afl->fsrv, afl->fsrv.exec_tmout);
 
-          munmap(mem, st.st_size);
-          close(fd);
+          if (afl->stop_soon) {
 
-          goto close_sync;
+            munmap(mem, st.st_size);
+            close(fd);
+
+            goto close_sync;
+
+          }
+
+          afl->syncing_party = sd_ent->d_name;
+          afl->queued_imported += save_if_interesting(afl, mem, new_len, fault);
+          show_stats(afl);
+          afl->syncing_party = 0;
 
         }
-
-        afl->syncing_party = sd_ent->d_name;
-        afl->queued_imported += save_if_interesting(afl, mem, new_len, fault);
-        show_stats(afl);
-        afl->syncing_party = 0;
 
         munmap(mem, st.st_size);
 
--- a/src/afl-fuzz-run.c
+++ b/src/afl-fuzz-run.c
@@ -701,15 +701,25 @@ abort_calibration:
 
 bool is_known_case(afl_state_t *afl, u8 *name, void *mem, u32 len) {
 
-  int sync_id_pos;
-  u32 id, src_id;
-  if (sscanf(name, "id:%06u,sync:%n%*[^,],src:%06u", &id, &sync_id_pos,
-             &src_id) != 2)
-    return false;
+  static char coming_from_me_str[16 + SYNC_ID_MAX_LEN];
+  static int  coming_from_me_len = 0;
+  if (!coming_from_me_len) {
 
-  if (strncmp(name + sync_id_pos, afl->sync_id, strlen(afl->sync_id)) != 0)
-    return false;
-  if (name[sync_id_pos + strlen(afl->sync_id)] != ',') return false;
+    snprintf(coming_from_me_str, sizeof(coming_from_me_str),
+             ",sync:%s,src:", afl->sync_id);
+    coming_from_me_len = strlen(coming_from_me_str);
+
+  }
+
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
This is a great idea and optimization I have not thought about!

IMHO you over-complicate things though - you just need to check if the sync name matches the instance name.

Through this over-complication you might not a known case, e.g. a secondary synced an item before it was trimmed by the main, then the trims the input and fuzzes it, then syncs to other instances where it is not trimmed yet and because the file lengths are different will not report it as known.

Something like this is fast and easy:
```
bool is_known_case(...) {
  static char *buf = NULL;
  if (!buf) {
    buf = ck_alloc(64);
    snprintf(buf, 64, ",sync:%s,", afl->sync_id);
  }
  return strstr(name, buf) != NULL;
}
```

src/afl-fuzz-run.c
```diff
@@ -860,26 +889,30 @@ void sync_fuzzers(afl_state_t *afl) {
 
         if (mem == MAP_FAILED) { PFATAL("Unable to mmap '%s'", path); }
 
-        /* See what happens. We rely on save_if_interesting() to catch major
-           errors and save the test case. */
+        if (!is_known_case(afl, namelist[o]->d_name, mem, st.st_size)) {
```
vanhauser-thc, line 938: if you use a simplified check as I remarked, you cna move the check upwards into the `if (st.st_size && st.st_size <= MAX_FILE) {` check which then also avoids the mmap.

kcwu:
I didn't use the simplified check because I want to be safe for cases like 
 - user restarted afl (by deleted out directory)
 - afl discard old out directory because previous session is less than 25 minutes

Maybe I can check if file mtime is after my start time. What do you think?

kcwu:
Two more cases to consider:
 - if the file is trimmed and shorter on secondary, we still want to sync it to the main node
 - secondary nodes may run different binary or different instrumentations. files trimmed on secondary nodes may have different bitmap on the main node

So file content comparison is necessary.

vanhauser-thc:
> * user restarted afl (by deleted out directory)
> * afl discard old out directory because previous session is less than 25 minutes

you mean eg. instance A is the master and is restarted, would not re-sync testcases from a secondary?

> * if the file is trimmed and shorter on secondary, we still want to sync it to the main node

no you would not, because the file is already there so testing it will not obtain new coverage, and the one already being there will need to be trimmed anyway when selected.

> * secondary nodes may run different binary or different instrumentations. files trimmed on secondary nodes may have different bitmap on the main node

it is still the item that was originally present on the main instance, hence it does not make a difference what instrumentation is run on the secondary.

> Maybe I can check if file mtime is after my start time. What do you think?

I agree for the case that an instance is restarted yes that would be the better solution

vanhauser-thc:
Thinking more about it the restarting a fuzzer from scratch needs a different handling outside of this pr - please use what i recommended and I will take care of handling restarts

kcwu:
What if the main node is running libjpeg's djpeg and the secondary node is running libjpeg-turbo's djpeg. If a file is trimmed with libjpeg-turbo, the file still should sync back to the main node.

vanhauser-thc:
> What if the main node is running libjpeg's djpeg and the secondary node is running libjpeg-turbo's djpeg. If a file is trimmed with libjpeg-turbo, the file still should sync back to the main node.

I will not stand for such nonsense setups :-)
in a fuzzing campaign you fuzz the same target, maybe differently instrumented in the instances but it must be the same.
If you want to include from a different campaign with a similar target then use the `-F` parameter.

vanhauser-thc:
this fixes the syncing to restarted instances. please have a look.
https://github.com/AFLplusplus/AFLplusplus/pull/2437

kcwu:
Even with #2437, I feel some cases are not handled. Let me think a while.

kcwu:
Assume M is main node, S is secondary node.
- t=0, M start
- t=1, M found `sync_dir/M/queue/id:000101`
- t=2, M found `sync_dir/M/queue/id:000102`
- t=3, S start
- t=4, S sync and copy the above file as `sync_dir/S/queue/id:000201:sync:M,src:000101`
- t=5, S sync and copy the above file as `sync_dir/S/queue/id:000202:sync:M,src:000102`
- t=6, M stop, `sync_dir/M` is deleted
- t=7, M restart from scratch
- t=8, S trimmed `id:000202` and make it shorter
- t=9, M sync from S
    - because `id:000201` mtime (t=4) is before M's start time (t=7), so we will add it to M's queue
    - because `id:000202` mtime (t=8) is after M's start time (t=7), so we will skip it.

Because M is restarted, I expect that it should add `id:000202` to its queue.

vanhauser-thc:
yes, such a corner case would not be synced then.
one could do a "simple" check and obtain the orig: id number and see if it exists already, skip if yes, otherwise sync.

I fear that if a lot of tests are done then this wastes a lot of time for every sync item and in the end this is more costly than not syncing an item that will be likely found at another point in time.

syncing is good, but it is not essential. it is essential to sync from cmplog and compcov  instances but even if one is skipped, as several other corpus items will pass the same edges.

syncing should be best effort, not absolutely guarantee 100%.

vanhauser-thc:
if you have a good idea for detecting when a restarted instance from scratch should sync items and when not and it does not have quite some overhead - shoot :)

kcwu:
This overhead won't increase over time because it is guarded behind `next_min_accept`. One file did the check only once. 

BTW, `strstr` is O(n^2) and may be not faster than my current checks. I can slightly reduce the overhead of my current check. Let me revise it tomorrow. I still prefer to check file content though. IMO, the saving of this PR is much larger than introduced overhead.

vanhauser-thc:
Once the options are on the table I can do a benchmark.

kcwu:
I revised my code.

I'm curious how will you benchmark? I have an idea: create symbolic links to fool the main node.

```
afl-fuzz -o sync_dir -M M ...&
afl-fuzz -o sync_dir -S S1 ...&
for i in $(seq 2 100); do
  ln -s sync_dir/S$i S1
done
```

vanhauser-thc:
IMHO the overhead of comparing the actual file contents is too much and if the length would be the same but different content then this is such a corner case that it is not an issue that it is not synced.

I propose the following which is not perfect but much faster:
```
bool is_known_case(afl_state_t *afl, u8 *name, u32 len) {

  static char coming_from_me_str[SYNC_ID_MAX_LEN + 2];
  static int  coming_from_me_len = 0;
  static int  min_len = 15 + 4 + 6;

  if (!coming_from_me_len) {

    snprintf(coming_from_me_str, sizeof(coming_from_me_str),
             "%s,", afl->sync_id);
    min_len += coming_from_me_len = strlen(coming_from_me_str);

  }

  if (unlikely(len < min_len)) { return false; }
  // we perform a 4 length memcmp so the optimizer can transform it to an integer comparison
  if (likely(memcmp(name + 10, "sync", 4) != 0)) { return false; }
  // we jump over the ':' after 'sync'
  if (unlikely(memcmp(name + 15, coming_from_me_str, coming_from_me_len) != 0)) { return false; }

  int src_id = atoi(name + 15 + coming_from_me_len + 4);
  if (unlikely(src_id >= afl->queued_items)) return false;
  // I am not sure I actually would like to check the length TBH
  if (unlikely(q->len != len)) return false;

  return true;
}
```
(this is untested code, just wrote it down as is into the reply)

kcwu:
How about merge my current PR and you modify on top of my changes?

vanhauser-thc:
I am still discussing this, not deciding :)

kcwu:
IMHO, "4 length memcmp" is little overkill here.

vanhauser-thc:
> IMHO, "4 length memcmp" is little overkill here.

yes, but it works so why not do it :-)

what is your opinion on not perfect syncing, so less time spend in potential testcases and comparing content vs syncing as much as possible?

kcwu:
I found only very corner case need to compare file content.

pseudo code
```
bool is_known_case() {
  if ("sync:$sync_id,src:" not in filename)
    return false;
  if (my_starttime < their_starttime)
    return true;
  if (file_mtime < my_starttime)
    return false;
  if (this is NOT the first time sync from that directory)
    return true; // the file appeared after last sync
  if (src_id >= queued_items)
    return false;
  
  // Now, possible cases:
  // 1. sync from us (this run)
  // 2. sync from my previous run, and I have ever restated, then the file is trimmed after my current run. 
  //    (the case I mentioned earlier https://github.com/AFLplusplus/AFLplusplus/pull/2433#issuecomment-2889026080)
  if (q->len == len && is_file_identical())
    return true;
  else {
    return false; // not sure, return false for safe
  }  
}
```

If we can know queue size of other fuzzers when we start. The check could be further simplified.
```
bool is_known_case() {
  if ("sync:$sync_id,src:" not in filename)
    return false;
  int id = parse...;
  return (id >= queue_size_when_I_start);
}
```

vanhauser-thc:
I will be working on it in a different branch and ping you once I have a PR

PR #2435, opened 2025-05-17
feat: add AFL_QEMU_EXITPOINT for controlled exit via env var

### LINKED ISSUE #1932: QEMU Mode: Stop Fuzzing at a Specific Address
I am currently using AFL++ with QEMU mode for fuzz testing and I am wondering if there is a way to stop fuzzing when a specific address is reached. In some cases, the target programs are quite large, and I am only interested in fuzzing a specific segment of the binary code, typically from the program entry point to a specific address.
I would like to propose a feature that allows users to set a specific address as a fuzzing target in AFL++ QEMU mode.

- The ability to specify the target address through a command-line option or configuration file.
- Ensuring that AFL++ gracefully halts the fuzzing process when the specified address is reached, then repeat fuzzing from the program entry to the specific address.

[persistent mode](https://github.com/AFLplusplus/AFLplusplus/blob/stable/instrumentation/README.persistent_mode.md) allows for repeated execution of a particular section of a reentrant function, but that doesn't satisfy my needs. I want to be able to repeat the fuzzing of a program entry to any given address.

**Additional context**
none.

rajanarahul93:
### Feature: QEMU Mode - Exit Fuzzing at a Specific Address

This PR adds support for stopping fuzzing when execution reaches a specified address in QEMU mode, as proposed in issue [#1932](https://github.com/AFLplusplus/AFLplusplus/issues/1932).

#### 🔧 Implementation Details:
- Introduced environment variable `AFL_QEMU_EXITPOINT` to set the exit address.
- On reaching the specified address during fuzzing, AFL++ logs the hit and gracefully exits.
- Added `setup_afl_qemu_exitpoint()` in `patch_init()`.
- Hooked address check added in `patch_block_trans_cb()`.

This feature is useful for focusing fuzzing between entry point and a custom-defined region of interest in large binaries.

Closes: #1932

### DIFF
--- a/qemu_mode/hooking_bridge/src/patching.c
+++ b/qemu_mode/hooking_bridge/src/patching.c
@@ -4,6 +4,9 @@
 #include <glib.h>
 #include "common.h"
 #include "exports.h"
+#include <stdint.h>
+
+static uint64_t afl_qemu_exitpoint = 0;
 
 void        *handle;
 struct conf *config;
@@ -70,6 +73,11 @@ void patch_block_trans_cb(struct qemu_plugin_tb *tb) {
 
   unsigned long long addr;
   addr = qemu_plugin_tb_vaddr(tb);
+
+  if (afl_qemu_exitpoint && addr == afl_qemu_exitpoint) {
+    fprintf(stderr, "[AFL_QEMU] Exitpoint hit at 0x%llx — exiting now.\n", addr);
+    exit(0);
+  }
 
   if (addr == config->entry_addr) {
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
Hi!
thanks for your PR, however several things:
1. this PR patches the hooking bridge - something that is not default and therefore of limited value
2. (not so important) we only accept PRs to `dev` - not to `stable` directly
3. (not so important) please code format with `make code-format`

so IMHO accepting this PR is not helpful.

What you need to do (and I think this will be equally easy to implement) is to modify the code at https://github.com/aflplusplus/qemuafl instead - just look for the implementation of `AFL_ENTRYPOINT` and do something similar for AFL_EXITPOINT.

advanced implementation: support multiple exit point implementations (addresses, separated with a comma).
and a hint: `_exit()` is faster than `exit()`

rajanarahul93:
Hi!
Thanks a lot for the feedback and suggestions. I see your point about the hooking bridge and understand why this PR isn’t the best fit. I’ll shift my focus to modifying qemuafl as you recommended and will follow the approach used for AFL_ENTRYPOINT to implement AFL_EXITPOINT. I’ll also make sure to use the dev branch and format the code properly next time. Appreciate the helpful tips—especially the one about _exit()!

vanhauser-thc:
if you need any help just create an issue in the unicornafl repo

PR #2437, opened 2025-05-18
Fix sync for restarted instances

vanhauser-thc:


src/afl-fuzz-run.c
```diff
@@ -802,6 +796,48 @@ void sync_fuzzers(afl_state_t *afl) {
...
+    time_t      last_mtime = {};
+    char        id0[PATH_MAX];
+    glob_t      glob_result;
+    struct stat st;
+
+    if (stat(qd_synced_path, &st) == 0) { last_mtime = st.st_mtime; }
+
+    snprintf(id0, sizeof(id0), "%s/id:000000,*", qd_path);
+
+    if (glob(id0, 0, NULL, &glob_result) == 0 && glob_result.gl_pathc == 1 &&
```
kcwu, line 815: Need to globfree

src/afl-fuzz-run.c
```diff
@@ -802,6 +796,48 @@ void sync_fuzzers(afl_state_t *afl) {
...
+        ck_write(id_fd, &min_accept, sizeof(u32), qd_synced_path);
+        close(id_fd);
+        continue;
+
+      }
+
+    } else {
+
+      // something went wrong - this cannot be right, mabye the instance is
+      // restarting, skip
```
kcwu, line 835: don't forget to free namelist, goto close_sync probably better.

src/afl-fuzz-run.c
```diff
@@ -802,6 +796,48 @@ void sync_fuzzers(afl_state_t *afl) {
...
+
+      if (last_mtime && last_mtime < st.st_mtime) {
+
+        // the first entry is newer than when we synced last - instance was
+        // restarted - we have to reset our counter and will skip this instance
+        // this time
+        min_accept = 0;
+        ck_write(id_fd, &min_accept, sizeof(u32), qd_synced_path);
+        close(id_fd);
+        continue;
```
kcwu: how about no "continue"; set next_min_accept=0 and let it continue the below sync?

kcwu (changes requested):

src/afl-fuzz-run.c
vanhauser-thc: yeah true

src/afl-fuzz-run.c
vanhauser-thc: ah yes! had never worked with glob() before ...

src/afl-fuzz-run.c
vanhauser-thc: then I need another jump point and adding that for just one jump source - I rather do the write instead there.

src/afl-fuzz-run.c
```diff
@@ -816,6 +816,7 @@ void sync_fuzzers(afl_state_t *afl) {
         stat(glob_result.gl_pathv[0], &st) == 0) {
 
       // we found exactly one "id:000000,*" file and obtained its mtime
+      globfree(glob_result);
```
kcwu: globfree(&glob_result); ?

kcwu:
looks good to me now

PR #2444, opened 2025-05-20
Better sync

vanhauser-thc:
@kcwu  WDYT?

### DIFF
--- a/src/afl-fuzz-run.c
+++ b/src/afl-fuzz-run.c
@@ -908,19 +908,13 @@ void sync_fuzzers(afl_state_t *afl) {
     // It could be that the target syncing instance was restarted, check!
     time_t      last_mtime = 0;
     char        id0[PATH_MAX];
-    glob_t      glob_result = {};
     struct stat st;
 
     if (stat(qd_synced_path, &st) == 0) { last_mtime = st.st_mtime; }
 
-    snprintf(id0, sizeof(id0), "%s/id:000000,*", qd_path);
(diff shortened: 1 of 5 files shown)


src/afl-fuzz-run.c
```diff
@@ -851,17 +913,19 @@ void sync_fuzzers(afl_state_t *afl) {
...
-        stat(glob_result.gl_pathv[0], &st) == 0) {
+    if (likely(glob(id0, 0, NULL, &glob_result) == 0 &&
+               glob_result.gl_pathc == 1 &&
+               stat(glob_result.gl_pathv[0], &st) == 0)) {
 
       // we found exactly one "id:000000,*" file and obtained its mtime
       globfree(&glob_result);
 
-      if (last_mtime && last_mtime < st.st_mtime) {
+      if (unlikely(last_mtime && last_mtime < st.st_mtime)) {
```
kcwu: BTW, id:000000 is usually hard-linked of input file and thus its mtime is not the fuzzer start time.

kcwu (approved):

src/afl-fuzz-run.c
vanhauser-thc: yeah true. Then I will use the `cmdline` file instead

vanhauser-thc: otherwise you think it is fine?

kcwu: Yes, looks good to me

PR #2446, opened 2025-05-22
fix save_if_interesting

kcwu:
The value of `classified`, `bits_new`, and `cksum`, were not always correctly maintained.
 1. In the past, `afl->queue_top->exec_cksum` was always assigned when `add_to_queue`, however it became conditional since cd5764170595. This doesn't change correctness because calibrate_case() will calculate the checksum. However, this mean one calibration run is wasted.

 2. Sometimes `classified` is set incorrectly. For example, this code snippet 
 
    ```
    new_bits = has_new_bits_unclassified(afl, afl->virgin_bits);
    classified = 1;
    ```
    should be changed to
    ```
    new_bits = has_new_bits_unclassified(afl, afl->virgin_bits);
    if (new_bits) classified = 1;
    ```

This commit fixed above issues and use macros to make the code easier to understand. This should prevent to forget to set classified in the future (like the bug fixed by 30c93d132166).

The macros also defers the calculations to where the values are really needed. This could save cpu if the code returns earlier. For example, if a case is timeout first and not timeout the second time, the current code does classify_counts, which is not always needed.

### DIFF
--- a/include/afl-fuzz.h
+++ b/include/afl-fuzz.h
@@ -1216,7 +1216,6 @@ u8 *describe_op(afl_state_t *, u8, size_t);
 #endif
 u8 save_if_interesting(afl_state_t *, void *, u32, u8);
 u8 has_new_bits(afl_state_t *, u8 *);
-u8 has_new_bits_unclassified(afl_state_t *, u8 *);
 #ifndef AFL_SHOWMAP
 void classify_counts(afl_forkserver_t *);
 #endif
--- a/src/afl-fuzz-bitmap.c
+++ b/src/afl-fuzz-bitmap.c
@@ -255,7 +255,8 @@ inline u8 has_new_bits(afl_state_t *afl, u8 *virgin_map) {
  * on rare cases it fall backs to the slow path: classify_counts() first, then
  * return has_new_bits(). */
 
-inline u8 has_new_bits_unclassified(afl_state_t *afl, u8 *virgin_map) {
+static inline u8 has_new_bits_unclassified(afl_state_t *afl, u8 *virgin_map,
+                                           bool *classified) {
 
   /* Handle the hot path first: no new coverage */
   u8 *end = afl->fsrv.trace_bits + afl->fsrv.map_size;
@@ -272,6 +273,7 @@ inline u8 has_new_bits_unclassified(afl_state_t *afl, u8 *virgin_map) {
 
 #endif                                                     /* ^WORD_SIZE_64 */
   classify_counts(&afl->fsrv);
+  *classified = true;
   return has_new_bits(afl, virgin_map);
 
 }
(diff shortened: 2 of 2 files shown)


vanhauser-thc:
I think the refactoring is a good idea.
I am not a fan of macros as this makes changing the code harder.
How about functions with the inline identifier?

domenukk:
+1 on using functions here.
Just pass the parameters you need to the fn and set the bool to true manually in the next line.
One extra line, but way more maintainable.

Or even just pass in the pointer to the bool...

kcwu:
I'm not fan of macros as well. I didn't use functions because I feel it ugly --- need to pass `classified` to function `calculate_cksum_if_necessary`.
```c
void calculate_cksum_if_necessary(u64 *cksum, bool *cksumed, afl_state_t *afl, bool *classified) {
    if (!*cksumed) {
      classify_if_necessary(afl, classified);
      *cksum = hash64(afl->fsrv.trace_bits, afl->fsrv.map_size, HASH_CONST);
      *cksumed = true;
    }
}
```
WDYT?

domenukk:
How about creating a `postprocess_state` struct (or similar), that contain `chksum`, `cksumed` and `classified` as members, then you just need to pass a single pointer to the struct on the stack. Might even be faster than using multiple parameters.

So:
```c
void calculate_cksum_if_necessary(afl_state_t *afl, postprocess_state_t *pp_state) {
      if (!pp_state->cksumed) {
        classify_if_necessary(afl, pp_state);
        pp_state->cksum = hash64(afl->fsrv.trace_bits, afl->fsrv.map_size, HASH_CONST);
        pp_state->cksumed = true;
    }
}
```

(Feel free to rename the struct to something more sensible)

vanhauser-thc:
> I'm not fan of macros as well. I didn't use functions because I feel it ugly --- need to pass `classified` to function `calculate_cksum_if_necessary`.
> 
> ```c
> void calculate_cksum_if_necessary(u64 *cksum, bool *cksumed, afl_state_t *afl, bool *classified) {
>     if (!*cksumed) {
>       classify_if_necessary(afl, classified);
>       *cksum = hash64(afl->fsrv.trace_bits, afl->fsrv.map_size, HASH_CONST);
>       *cksumed = true;
>     }
> }
> ```
> 
> WDYT?

yes the compiler can easily optimize this.

> How about creating a `postprocess_state` struct (or similar), that contain `chksum`, `cksumed` and `classified` as members, then you just need to pass a single pointer to the struct on the stack. Might even be faster than using multiple parameters.

please not, that is overcomplicating it and makes it more difficult for the compiler to optimize :-)

domenukk:
> please not, that is overcomplicating it and makes it more difficult for the compiler to optimize :-)

Not at all?

vanhauser-thc:
> > please not, that is overcomplicating it and makes it more difficult for the compiler to optimize :-)
> 
> Not at all?

His solution is fine, yours is over complicating it imho :)

kcwu:
done. PTAL

vanhauser-thc:
thank you!

PR #2458, opened 2025-05-28
Update FRIDA

WorksButNotTested:


frida_mode/src/lib/lib.c
```diff
@@ -39,7 +39,6 @@ typedef struct {
 static guint64 text_base = 0;
 static guint64 text_limit = 0;
 
-  #ifdef GUM_16_6_PLUS
```
domenukk, line 42: So we drop support for older frida versions?? Makes a lot of sense, just double checking

WorksButNotTested: Yeah. Not sure that there's much point keeping the old versions of frida. If people want them, they can always run an older afl++. If it becomes a problem it can always be reinstated later i guess, the changes aren't that dramatic.

PR #2461, opened 2025-06-03
Let user pass their own CPU_TARGET in test-pre.sh

justuswilhelm:
The target system might be different from the host system. For example, you can fuzz Linux binaries compiled for *mipsel*, while your host is *x86_64*.

Some of the tests depend on specific platforms to run correctly. For example, the afl-fuzz qemu_mode cmplog test only works on Intel or ARM systems. The `SYS` variable is populated using `uname -m` and the test cases then consult this variable to decide whether to run the test or not.

If you want to test afl-fuzz for qemu_mode on mipsel, you might want to make sure that Intel or ARM tests don't run. With this patch, you can supply your own `SYS` environment variable and skip platform specific tests.

This allows you to add tests for *mipsel* or other niche platforms in the future as well.

Sample usage:

```
$ cd qemu_mode && env CPU_TARGET=mipsel ./build_qemu_support.sh
$ cd ../test && env SYS=mipsel ./test-qemu-mode.sh
[*] Using user supplied environment variable SYS=mipsel
[*] starting AFL++ test framework ...
[*] Testing: qemu_mode
...
```

### DIFF
--- a/test/test-pre.sh
+++ b/test/test-pre.sh
@@ -17,9 +17,9 @@ test -z "" 2>/dev/null || { echo Error: test command not found ; exit 1 ; }
 GREP=`type grep > /dev/null 2>&1 && echo OK`
 test "$GREP" = OK || { echo Error: grep command not found ; exit 1 ; }
 echo foobar | grep -qE 'asd|oob' 2>/dev/null || { echo Error: grep command does not support -q and/or -E option ; exit 1 ; }
-test -e ./test-all.sh || cd $(dirname $0) || exit 1
+test -e ./test-all.sh || cd $(dirname "$0") || exit 1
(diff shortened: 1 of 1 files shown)


domenukk:
I wonder if just `SYS` as env variable might be prone to collisions for some users? Like, maybe use `AFL_TEST_SYS` instead, or something?

vanhauser-thc:
Or use CPU_TARGET?

justuswilhelm:
> Or use CPU_TARGET?

Thanks for getting back so quickly :) `CPU_TARGET` sounds the most consistent. I've pushed a new version.

vanhauser-thc:
Lgtm thanks

PR #2462, opened 2025-06-04
Modify the UID and GIDs of fuzzing target

graux-pierre:
Hello,

This commit adds two new environment variables (`AFL_FORKSRV_UID` and `AFL_FORKSRV_GID`) that allow, when running the target using the fork server, to use specific user and group IDs.

Example of Linux session:
```bash
# Add required permissions to AFL binary
setcap "cap_setuid,cap_setgid,cap_chown+ep" afl-fuzz

# Fuzz the target using custom UID (1001) and GIDs (1001 main group, 23 and 42 supplementary groups)
AFL_FORKSRV_UID=1001 AFL_FORKSRV_GID=1001,23,42 afl-fuzz [AFL options and target]
```

Under the hood, the fork server calls `setregid`, `setgroups` and `setreuid` right before launching the target using `execv`. Unfortunately when the UID and GIDs of afl-fuzz and the forkserver are different, the communication between them is faulty: files and shared memory have default permission rights set to 600 and UID/GID set to the ones of afl-fuzz. When only the GID is modified this isn't problematic (access granted using the common UID) but when the UID is modified we need to set permission rights to 660 so the fork server can access entities using the common group ID. In case both are modified, we even need to change the group ownership of the entities so afl-buzz binary can access them using its UID and the fork server using its GID.

This new ownership scheme constitutes the most of the commit since the default permission and the potential new group has to be passed to all function creating communication entities (files and shared memory). I have modified the creation of the entities related to the forkserver and left the other one intact but I might have missed cases and thus I think a review by an experimented AFL developer is required.

Finally, I have used the constructions shown bellow several times. If there's macro more handy that automatically deals with errno please tell me so I can improve the code.

```C
WARNF("Syscall failed: %s.", strerror(errno));
FATAL("Syscall failed: %s.", strerror(errno));
```

Thank you very much,

Pierre Graux


src/afl-fuzz-init.c
```diff
@@ -2240,20 +2267,32 @@ void setup_dirs_fds(afl_state_t *afl) {
...
 
-  if (afl->sync_id && mkdir(afl->sync_dir, 0700) && errno != EEXIST) {
+  if (afl->sync_id && mkdir(afl->sync_dir, afl->dir_perm) && errno != EEXIST) {
 
     PFATAL("Unable to create '%s'", afl->sync_dir);
 
   }
 
-  if (mkdir(afl->out_dir, 0700)) {
+  printf("out_dir = %s\n", afl->out_dir);
```
vanhauser-thc: that is a debug printf I guess?

vanhauser-thc:
Looks like a niche issue but it does not hurt, so why not :-)
When reading through the change it looks like the code does not allow to move to UID 0 (e.g. afl-fuzz is started with uid 1000, but the target process needs to run as root, so CAP_SETUID is given to the process).
What do you think about adding this? (so the default would need to be -1 not 0.

src/afl-fuzz-init.c
graux-pierre: Indeed, I forgot this one. I will remove it next commit.

graux-pierre:
I've tested this behavior and it's working. The code uses two variables for handling the uid : the actual value and one boolean that indicates when the uid value is meaningful. Thus, the default value 0 can be used as any other value. I have done this for clarity purpose. I can use only one variable if you prefer.

I'm currently working on patching the code for the docker image (the failing CI checks).

src/afl-fuzz-init.c
vanhauser-thc: still waiting for that ... :)

vanhauser-thc:
I think you only need to fix utils/afl_network_proxy .

graux-pierre:
I have fixed the docker build but I don't manage to setup a working afl server and client (even on the stable branch) so I haven't checked that the UID/GID modification works in that case. I have also removed the debug `printf` and added a check.

vanhauser-thc:
Thank you!

PR #2463, opened 2025-06-05
Update test code and documentation for afl-fuzz qemu_mode on mipsel

justuswilhelm:
Thanks again for merging my previous two pull requests. In this pull request I propose two new patches that make it easier to get started with black-box fuzzing for *mipsel* binaries.

Getting persistent mode to work requires a lot of tinkering (especially if using a target different from the host), so adding more documentation and test code should help others get started as well.

First, I've added a new (optional) environment variable called `CPU_TARGET_CC` to `test-qemu-mode.sh` that allows you to specify a cross-compiler. I was able to verify that the persistent mode tests work on *mipsel*, so I've also updated the two `test` calls further down in the same file to pass when the `SYS` environment variable is set to `mipsel`. 

Then, I've updated the persistent hook documentation to explain how to use the persistent hook on *mipsel*.

I'm open to feedback, so please let me know if I need to make any changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,6 +108,8 @@ utils/persistent_mode/persistent_demo
 utils/persistent_mode/persistent_demo_new
 utils/persistent_mode/persistent_demo_new_compat
 utils/persistent_mode/test-instr
+utils/qemu_persistent_hook/mipsel_test
+utils/qemu_persistent_hook/test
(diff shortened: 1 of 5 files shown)


vanhauser-thc:
looks good to me, thanks :)

justuswilhelm:
Thank you so much :)

PR #2464, opened 2025-06-06
WARNF for setting scheduling for main fuzzer instead of FATAL

alexandredoyen29:
Hi, thank you for making AFL++ better and better.

Here is a tiny PR that replaces the FATAL by WARNF when using `-p exploit` on the main fuzzing node (As suggested on Discord).

Best regards,


vanhauser-thc:
Thanks

PR #2465, opened 2025-06-08
instrumentation: drop removed TODO flag from GCC plugins

thesamesam:
TODO_verify_il was removed in GCC trunk by 9739ae9384dd7cd3bb1c7683d6b80b7a9116eaf8, so drop it here.


vanhauser-thc:
I hope it is not a good flag to have for older versions but otherwise I guess people would report the plugin to fail ... thanks

PR #2466, opened 2025-06-09
refactor!: drop python2

e-kwsm:
<https://www.python.org/doc/sunset-python-2/>

> We have decided that January 1, 2020, was the day that we sunset Python 2. That means that we will not improve it anymore after that day, even if someone finds a security problem in it. You should upgrade to Python 3 as soon as you can.


vanhauser-thc:
thanks!

PR #2468, opened 2025-06-11
UnicornAFL v3 integration

wtdcode:
This PR enables integration with UnicornAFL mode by enabling cmplog.

I finally decided to keep Unicorn mode instead of reusing QEMU mode as previously proposed because the harness of UnicornAFL is still slightly different from QEMU harness.

For UnicornAFL C/CPP/RUST bindings harness, `-U` is not necessary, but enabling that is also harmless since it essentially skips the binary check.

For UnicornAFL Python bindings harness, `-U` is compulsory to make it work. This is because the `argv[0]` is `python` instead of the real harness.

I will further update samples and scripts under `unicorn_mode`


include/forkserver.h
```diff
@@ -146,6 +146,8 @@ typedef struct afl_forkserver {
   bool use_fauxsrv;                     /* Fauxsrv for non-forking targets? */
 
   bool qemu_mode;                       /* if running in qemu mode or not   */
+  
+  bool unicorn_mode;                    /* if running in unicorn mode or not*/
```
domenukk, line 152: Can we get around this extra flag? Ideally merge it with existing flags of the same meaning

vanhauser-thc: other modes have their own flag too - so why not?

wtdcode: https://github.com/AFLplusplus/AFLplusplus/pull/2468/files#diff-d5539013e7d028b9fc3b6e581f284c5ebedc4cd8a7151b2c11cd0aeaabf071a0R42

We don't have a pointer to `afl` here so we must keep it to fsrv.

domenukk: > other modes have their own flag too - so why not?

The other modes should also not have extra flags, it should be a flag for feature xyz IMHO

Evian-Zhang: My previous PR (#2419) contains some codes that remove this unnecessary variable. Maybe you could sync some codes from there :)

vanhauser-thc: > > other modes have their own flag too - so why not?
> 
> The other modes should also not have extra flags, it should be a flag for feature xyz IMHO

then it is on us to refactor that for all of them, not in this PR - IMHO

domenukk: Fair

vanhauser-thc:
any idea when unicornv3 will land?
I will release a new version of AFL++ very soon, should this be in it or rather in the next release?

wtdcode:
On academic travel so no eta unfortunately =/

Next release is probably better.

wtdcode:
The branch is now ready for review and the integration is complete, both building and samples.

cc

@Evian-Zhang

Note that there is a significant change: Except Rust bindings, all other bindings now require a dynamically linked `libunicorn.so` instead of statically link against `libunicorn.a`. I would assume that does not affect your workflow and just let you know.
 
@vanhauser-thc 
@domenukk 
Could check if there are any unintended changes =).

wtdcode:
CI & Dockerfile should be fixed too =).

Now unicornafl no longer maintains a fork of forkserver implementation but uses LibAFL forkserver implementation instead, which shall make further maintaince much easier.

unicorn_mode/build_unicorn_support.py
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env python
```
domenukk: python3

unicorn_mode/build_unicorn_support.py
```diff
@@ -0,0 +1,154 @@
...
+
+if libs_path.exists():
+    print("[!] Cleaning previous artifacts...")
+    shutil.rmtree(libs_path)
+if include_path.exists():
+    print("[!] Cleaning previous headers...")
+    shutil.rmtree(include_path)
+
+if not (cwd.parent / "afl-showmap").exists():
+    print("[!] Please compile AFL++ firstly.")
```
domenukk: first

unicorn_mode/README.md
```diff
@@ -105,11 +104,8 @@ An example use of AFL-Unicorn mode is discussed in the paper Unicorefuzz:
 As for the QEMU-based instrumentation, unicornafl comes with a sub-instruction
 based instrumentation similar in purpose to laf-intel.
 
-The options that enable Unicorn CompareCoverage are the same used for QEMU. This
-will split up each multi-byte compare to give feedback for each correct byte:
-
-* `AFL_COMPCOV_LEVEL=1` to instrument comparisons with only immediate values.
-* `AFL_COMPCOV_LEVEL=2` to instrument all comparison instructions.
+`UNICORN_AFL_CMPCOV=1` will split up each multi-byte compare to give feedback for
```
domenukk, line 107: Just use the normal `AFL_` env here? No reason to add a new one, IMHO.
Anyway, unicorn-specific env variables should start with `AFL_UNICORN` or maybe just `AFL_UC` to be consistent with AFL_QEMU, etc.

unicorn_mode/build_unicorn_support.py
```diff
@@ -0,0 +1,154 @@
...
+def detect_from_env_or_file(target: str):
+    if target in os.environ:
+        return os.environ[target]
+    elif (cwd / target).exists():
+        with open(cwd / target, mode="r") as f:
+            return f.read().strip()
+    else:
+        return None
+    
+if sys.platform == 'win32':
```
domenukk: We don't support windows in AFL++ at all

unicorn_mode/build_unicorn_support.py
```diff
@@ -0,0 +1,154 @@
...
+    print("[!] TLDR: `curl https://sh.rustup.rs -sSf | sh -s -- -y`")
+    exit(1)
+
+unicornafl_version = detect_from_env_or_file("UNICORNAFL_VERSION")
+if not unicornafl_version:
+    print("[!] No valid UNICORNAFL_VERSION found")
+    exit(1)
+
+try:
+    run_cmd("git status", cwd)
```
domenukk, line 73: The old sh file used to work with a tar archive instead of a git repo, too.
Not sure if we still want to / can support this.
The rust stuff would need to be vendored, too, so I guess it is more work (and honestly not worth it, IMHO).
Just pointing this out.

unicorn_mode/build_unicorn_support.py
```diff
@@ -0,0 +1,154 @@
...
+print(f"[*] We will checkout unicornafl {unicornafl_version}")
+run_cmd(f"git fetch --all && git checkout {unicornafl_version}", unicornafl_path)
+
+print(f"[*] Now building unicornafl python bindings")
+venv = in_venv()
+if not venv:
+    print(f"[!] A python venv is highly recommended! We will try to add --user for you.")
+
+if not shutil.which("maturin"):
+    print(f"[!] No maturin, will install maturin firstly")
```
domenukk: first

unicorn_mode/build_unicorn_support.py
```diff
@@ -0,0 +1,154 @@
...
+    if not venv:
+        run_cmd("pip install --user maturin")
+    else:
+        run_cmd("pip install maturin")
+
+print(f"[*] Now building unicornafl with maturin")
+run_cmd("maturin develop --release", unicornafl_path, True)
+print(f"[*] Python bindings built, now testing...")
```
domenukk, line 113: I think we were using `[+]` for this, instead of `[*]`, or the other way around :D

unicorn_mode/build_unicorn_support.py
```diff
@@ -0,0 +1,154 @@
...
+            if "linked_libs" in ln_json and any(["unicorn" in x for x in ln_json['linked_libs']]):
+                if "out_dir" in ln_json:
+                    out_dir = Path(ln_json['out_dir'])
+                    shutil.copytree(out_dir / "lib", libs_path, dirs_exist_ok=True)
+                    shutil.copytree(out_dir / "include", include_path, dirs_exist_ok=True)
+                    print(f"[*] Copied from {out_dir.absolute()}")
+
+
+print(f"""[*] All done! You have compiled unicornafl without any issue.
+    You could start using python bindings by `import unicornafl`.
```
domenukk: You can now start...

unicorn_mode/build_unicorn_support.py
wtdcode: This mostly supports Docker build where .git is not copied and unicornafl is empty.

unicorn_mode/build_unicorn_support.py
```diff
@@ -0,0 +1,154 @@
...
+print(f"[*] Now building unicornafl with maturin")
+run_cmd("maturin develop --release", unicornafl_path, True)
+print(f"[*] Python bindings built, now testing...")
+
+with tempfile.TemporaryDirectory() as tmpdir:
+    dst_file = Path(tmpdir) / "test-instr0"
+    print("[*] Testing a rather simple python harness")
+    run_cmd(f"../afl-showmap -U -m none -t 2000 -o {dst_file.absolute()} -- python3 ./samples/python_simple/simple_test_harness.py ./sample_inputs/sample1.bin", None, True)
+
+    if dst_file.exists():
```
wtdcode, line 120: I forget an exit here for failed case.

unicorn_mode/README.md
vanhauser-thc: why does it need to be an env? Using the `-c0` command line parameter should activate this (and set this env). this is how it works for qemu and frida mode too

unicorn_mode/build_unicorn_support.py
vanhauser-thc: I think * for information and + for notice level? fine for me as it is

unicorn_mode/README.md
wtdcode: That “-c” enables cmplog but not compcov.

@Evian-Zhang could you elaborate the difference?

unicorn_mode/samples/speedtest/rust/src/main.rs
```diff
@@ -87,8 +88,40 @@ fn main() {
...
+        _input: &[u8],
+        _persistent_round: u64,
+    ) -> bool {
+        unicorn_result.is_err()
+    }
+}
+
 fn fuzz(input_file: &str) -> Result<(), uc_error> {
-    let mut uc = Unicorn::new(Arch::X86, Mode::MODE_64)?;
+    let mut uc = Unicorn::new_with_data(Arch::X86, Mode::MODE_64, UnicornFuzzData::default())?;
```
domenukk, line 124: Why not make `new` have the default data and leave the API as is here?

unicorn_mode/samples/speedtest/rust/src/main.rs
```diff
@@ -87,8 +88,40 @@ fn main() {
     fuzz(input_file).unwrap();
 }
 
+struct FuzzCB;
+
+impl<'a, D> UnicornAflExecutorHook<'a, D> for FuzzCB {
+    fn place_input(
+        &mut self,
+        uc: &mut Unicorn<'a, unicornafl::executor::UnicornFuzzData<D>>,
```
domenukk, line 96: If the second parameter is always of type UnicornFuzzData, this should just take D as generic

domenukk:
Did the python API stay exactly the same?

unicorn_mode/README.md
Evian-Zhang: This is designed to keep the same behavior as the old cpp version. See [codes around this line](https://github.com/dotCirill/unicornafl/blob/4e08c3a31c6e0abacf5e66132c719e7e71039394/unicornafl.cpp#L585).

domenukk: Ah that must have slipped through review back then. Good time to change it, IMHO.

If you think it's already in use somewhere, you could still accept the old UNICORN_ env variable as well (and potentially print a depreciation warning).

The UNICORN_ variant was never mentioned in the docs, fwiw:
https://aflplus.plus/docs/env_variables/

wtdcode:
> Did the python API stay exactly the same?

Yes (see samples)

unicorn_mode/samples/speedtest/rust/src/main.rs
wtdcode: I do not really get it. `UnicornFuzzData` is needed for unicornafl internally and `D` is left for user to encapsulate any type of data.

unicorn_mode/samples/speedtest/rust/src/main.rs
wtdcode: `Unicorn::new` creates a `Unicorn<'_, ()>`, which means that it carries nothing.

unicorn_mode/README.md
vanhauser-thc: oh sorry, my bad, read CMPLOG. yes for COMPCOV please keep as is :-)

vanhauser-thc:
Hey @domenukk I was OOO for three weeks and a lot of catch up to do - you know unicorn better than me anyway, can you merge it please once you think everything is fine?

unicorn_mode/samples/speedtest/rust/src/main.rs
domenukk: Yes so the type here should be Unicorn<'a, D>

domenukk: Ah wait, you need UnicornFuzzData because it's unicornafl specific, not for unicorn directly? Sad

unicorn_mode/samples/speedtest/rust/src/main.rs
domenukk: Aah again because of Unicorn vs UnicornAfl... I guess?

unicorn_mode/samples/speedtest/rust/src/main.rs
wtdcode: I can’t recall why we need this type in the public interface.

@Evian-Zhang did this refactoring. Could you elaborate a bit more?

wtdcode: If I recall it correctly, this is mostly for hook data here: https://github.com/AFLplusplus/unicornafl/blob/6da4c5315333f0d6f87a4a40e5bd3570261dc5ef/src/executor.rs#L107

Without this, we have to do some bad unsafe pointer operations with D I think because D is shared across all hooks. Generally, it seems a design flaw from Unicorn bindings and could be improved once unicorn-engine-sys is published with stable APIs.

unicorn_mode/samples/speedtest/rust/src/main.rs
wtdcode: Yeah. But we will move to unicorn-engine-sys once it is published.

unicorn_mode/samples/speedtest/rust/src/main.rs
Evian-Zhang: Yes, this structure is used to store the internal data, since in the design of unicorn api, in call backs we can only access the unicorn handle and such data, so we have to store things like prev_loc there.

wtdcode: Looked into Unicorn bindings a bit more and unfortunately it is hard to obtain `Unicorn<'_, UnicornFuzzData<D>>` from `Unicorn<'_, D>` because `D` will be dereferenced in the hooks (note each hook callback data is defined as `UcHook<D, F>`). It will cause UB for hooks registered in advance if we transmute the passed in `Unicorn` type.

vanhauser-thc:
@domenukk and @wtdcode  what is the current status? :-)

wtdcode:
> @domenukk and @wtdcode what is the current status? :-)

Sorry for late. Engaged in my research recently and will get some time before this weekend to fix.

wtdcode:
Should be fine now.

unicorn_mode/build_unicorn_support.py
```diff
@@ -0,0 +1,153 @@
...
+
+# https://stackoverflow.com/questions/1871549/how-to-determine-if-python-is-running-inside-a-virtualenv
+def in_venv():
+    return sys.prefix != sys.base_prefix
+
+def run_cmd(cmd: str, cwd: Path = None, quiet: bool = False):
+    if not cwd:
+        cwd = Path(__file__).parent
+    if quiet:
+        print(f"[*] Runing queitly: \"{cmd}\" under workding directory {cwd}")
```
domenukk: typo

unicorn_mode/build_unicorn_support.py
```diff
@@ -0,0 +1,153 @@
...
+
+print(f"[*] Now building unicornafl python bindings")
+venv = in_venv()
+if not venv:
+    print(f"[!] A python venv is highly recommended! We will try to add --user for you.")
+
+if not shutil.which("maturin"):
+    print(f"[!] No maturin, will install maturin first")
+    if not venv:
+        run_cmd("pip install --user maturin")
```
domenukk: shouldn't this possibly be `python3 -m pip`? At least on some distributions(?)

wtdcode: This is actually tricky. For example, on `manylinux`, `python3 -m pip` will definitely be rejected and the expected usages are like `python3.10 -m pip` or `python3.11 -m pip` or `pipx`. Generally, both `pip` and `python3 -m pip` are not correct and I would say we should not install anything if a venv is not present nowadays, if that is acceptable.

vanhauser-thc: I dont think this will be working (requiring venv). usually people do `make binary-only`  and would not know that they need a venv and then it fails. Also for a docker image build this would just not work.

wtdcode: For docker build I shipped a brand new venv activated by default, which is a common practice and works out-of-box.

Regarding `make binary-only`, this is more tricky and let me take a reference from pwndbg: https://github.com/pwndbg/pwndbg/blob/e43542852aecce2f1d4288d698a108241d4179dd/setup.sh#L212 It might be more appropriate to create one for users.

*Previous unicornafl simply installs itself via setup.py which is very hard to manage afterwards (like upgrade or remove cleanly etc) and already deprecated for a quite a long time.

vanhauser-thc:
I would like to do a new release as soon (= a 1-2 weeks after) this is merged. what is still required for merging?

wtdcode:
> I would like to do a new release as soon (= a 1-2 weeks after) this is merged. what is still required for merging?

I don't see any except the discussion about the venv above.

wtdcode:
Generally we should choose between:

- Create a `.venv` for users if there is none. Print this information in the final output of the script.
- Best efforts by trying to install unicornafl to current python3 environment. Honestly this won't work reliably in any way but we can handle case by case, or simply wait for users' reports.

vanhauser-thc:
creating a .venv if none is active is then fine with me I think. @domenukk ?

wtdcode:
Done both ways, and default to create a `.venv` for users. Also added an environment variable to control the behavior, in case someone really wants to install to a non-venv python.

vanhauser-thc:
@domenukk  wdyt?

PR #2469, opened 2025-06-13
Fix persistent qemu_mode test base address on aarch64

refi64:
qemu_mode/README.persistent.md documents the different base address, but the test still only had the cases for x86 and x64.


vanhauser-thc:
Thanks :)

PR #2472, opened 2025-06-16
Fix a cmplog rtn instrumentation check if all vectors are empty.

am009:
Fix a cmplog rtn instrumentation check if all vectors are empty.

The logic here seems to be: if all vectors are empty, then no instrumentation is inserted. I suspect some negations (`!`) might have been accidentally omitted.

In [this commit](https://github.com/AFLplusplus/AFLplusplus/commit/ed10f3783bd8fab33ab5750f56bf87ed008f28ed), support for `memcmp`, `strcmp`, and `strncmp` was added, which introduced this piece of code.

image-20250616201742353

In [earlier commit](https://github.com/AFLplusplus/AFLplusplus/commit/1c19804834d2ea4f169be0a99b8ce493a2f10167), the condition in the `if` statement checked whether all arrays were empty:

image-20250616202050186


vanhauser-thc:
yes that is wrong, thanks!

PR #2473, opened 2025-06-20
Adding CI for testing from llvm 14 to llvm 20

tokatoka:
Adding CI in `check-compiler-passes` jobs.
I used ubuntu22.04 because ubuntu24.04 doesn't have package from llvm 14 up to llvm 16

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
           sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-${{ matrix.version }} 200
       - name: install llvm-tools 
         # because llvm 18.1.8 (installed by the script above) is broken for us
-        if matrix.version == '18'
+        if: matrix.version == '18'
         run: sudo apt install -y llvm-18 clang-18 llvm-config-18 libclang-rt-18-dev
       - name: install clang-rt (for llvm 15)
         # because ubuntu-22.04 already has this package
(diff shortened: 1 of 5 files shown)


tokatoka:
it doesn't work with llvm 18.1.8 specifically but i think this is a bug from LLVM?
it does work with llvm 18.1.3

tokatoka:
clang 18.1.8 puts the whole command arguments in string literal 😅 .
and then this target is mis-treated as the deferred forkserver binary

image

vanhauser-thc:
> clang 18.1.8 puts the whole command arguments in string literal 😅 . and then this target is mis-treated as the deferred forkserver binary
> 
> image

I don’t really understand the bug :)

vanhauser-thc:
Good idea to have this. Wouldn’t it be better to just have one ci for all compiler versions? Would be much faster. 
Also some packages can be removed eg gcc/g++

tokatoka:
> I don’t really understand the bug :)

llvm 18.1.8 doesn't work because this specific version puts the string literal of the whole command into the binary.
but we memmem() the binary to see if the defered forkserver signature is there, so even if the target is not for deferred forkserver, afl-fuzz thinks that the binary is deferred forkserver by mistake.

(But this is just that that specific llvm version is broken)

>  Wouldn’t it be better to just have one ci for all compiler versions?

but isn't that slower? running them in a concurrent way is fast

tokatoka:
@vanhauser-thc 
this one is ready

vanhauser-thc:
Thanks!

PR #2474, opened 2025-06-21
Remove macros for old compiler versions

tokatoka:


tokatoka:
sorry it has conflicts. i'll make new one..

PR #2475, opened 2025-06-21
Remove macros for old compiler versions from cmplog passes

tokatoka:
I removed all macros which are about llvm versions <= 13 from cmplog passes

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -345,6 +345,11 @@ llvm:
 	-$(MAKE) -j$(nproc) -f GNUmakefile.llvm
 	@test -e afl-cc || { echo "[-] Compiling afl-cc failed. You seem not to have a working compiler." ; exit 1; }
 
+llvm-build-test:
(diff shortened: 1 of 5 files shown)


GNUmakefile
```diff
@@ -345,6 +345,11 @@ llvm:
 	-$(MAKE) -j$(nproc) -f GNUmakefile.llvm
 	@test -e afl-cc || { echo "[-] Compiling afl-cc failed. You seem not to have a working compiler." ; exit 1; }
 
+llvm-build-test:
```
tokatoka, line 348: added new target. because old `make llvm` will ignore error (thus ci won't fail for broken code)

vanhauser-thc:
LGTM, thanks!
can you please have an eye out on new LLVM releases and add them once they are supported? e.g. 21 so far works beside some deprecation warnings

tokatoka:
> can you please have an eye out on new LLVM releases and add them once they are supported? e.g. 21 so far works beside some deprecation warnings

OK

PR #2481, opened 2025-06-24
replace select() by poll()

kcwu:
poll() has lower overhead than select().
This improves my benchmark/benchmark.py speed by 0.8% on linux


vanhauser-thc:
I will need to do further tests with this, so I will merge it after the release. thanks!

vanhauser-thc:
merging now into 4.34a

PR #2499, opened 2025-07-19
Don't clear the screen when printing stats

### LINKED ISSUE #2498: Don't clear the screen when printing the banner
When fuzzing starts, the banner that afl-fuzz updates clears the logs it just printed, and possibly even terminal history unrelated to the afl-fuzz command.

Instead of clearing the screen, just print the banner and scroll the terminal down (either `[#S` or `[#T` is used for this, I'm not sure which one of the top of my head).

This could be gated behind an environment variable.

k4lizen:
Closes: https://github.com/AFLplusplus/AFLplusplus/issues/2498

The way most applications (btop/htop etc) do this stats-type screen is switch to an alternate screen buffer on entry, and switch back on exit. Doing the same seems to work pretty well.

That being said it might be smarter to do this switch right after "[+] All set and ready to roll!" and return back on exit, instead of swapping buffers on each display. If theres a nice "entrypoint" and "exitpoint"-type place to put these swaps, let me know.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -117,3 +117,4 @@ utils/plot_ui/afl-plot-ui
 vuln_prog
 argv_fuzz_demo
 argv_fuzz_persistent_demo
+.cache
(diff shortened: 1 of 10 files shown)


k4lizen:
It would be nice if we could add one final print of stats to the normal buffer before we exit, so the latest stats can persist in the terminal scrollback.

PR #2504, opened 2025-07-25
FIx isssue AFLplusplus#2503:Incorrect guard offset calculation for __afl_coverage_interesting calls and comparison instructions in FunctionGuardArray.

85lx:
I have submitted the issue https://bgithub.xyz/AFLplusplus/AFLplusplus/issues/2503 using my partner's account 'bigchengz'.
FIx isssue https://bgithub.xyz/AFLplusplus/AFLplusplus/issues/2503:Incorrect guard offset calculation for __afl_coverage_interesting calls and comparison instructions in FunctionGuardArray.
Description:
I've identified a buffer layout issue in the PCGUARD instrumentation mode, The FunctionGuardArray size calculation is correct but the offset calculations for special instrumentation guards is incorrect, which mabe leads to a buffer overflow.
Problem Analysis:
In instrumentation/SanitizerCoveragePCGUARD.so.cc:969-976, Since cnt_hidden_sel_inc is always 0 now，the FunctionGuardArray size is correctly calculated as:

AllBlocks.size() + (first + cnt_cov + cnt_sel_inc - skip_blocks)
However, the offset calculations for special calls and comparison instructions don't account for skip_blocks .

Expected Layout:

[0 to AllBlocks.size() - skip_blocks - 1] -> Used basic blocks
[AllBlocks.size() - skip_blocks to AllBlocks.size() - skip_blocks + cnt_cov - 1] -> Special calls
[AllBlocks.size() - skip_blocks + cnt_cov to ...] -> Comparison instructions
Current Layout:

[0 to AllBlocks.size() - skip_blocks - 1] -> Used basic blocks
[AllBlocks.size() - skip_blocks to AllBlocks.size() - 1] -> Unused space (wasted)
[AllBlocks.size() + 1 to AllBlocks.size() + cnt_cov] -> Special calls
[AllBlocks.size() + cnt_cov to ...] -> Comparison instructions
Specific Issues:

Special call (_afl_coverage_interesting ) offset calculation in instrumentation/SanitizerCoveragePCGUARD.so.cc:1017-1021:
Current: (++special + AllBlocks.size()) * 4
Should be: (special++ + AllBlocks.size() - skip_blocks) * 4
Comparison instructions offset calculation in instrumentation/SanitizerCoveragePCGUARD.so.cc:1117-1120:
Current: (cnt_cov + local_selects++ + AllBlocks.size())
Should be: (cnt_cov + local_selects++ + AllBlocks.size() - skip_blocks)

### DIFF
--- a/instrumentation/SanitizerCoveragePCGUARD.so.cc
+++ b/instrumentation/SanitizerCoveragePCGUARD.so.cc
@@ -1017,7 +1017,7 @@ bool ModuleSanitizerCoverageAFL::InjectCoverage(
         Value *GuardPtr = IRB.CreateIntToPtr(
             IRB.CreateAdd(
                 IRB.CreatePointerCast(FunctionGuardArray, IntptrTy),
-                ConstantInt::get(IntptrTy, (++special + AllBlocks.size()) * 4)),
+                ConstantInt::get(IntptrTy, (special++ + AllBlocks.size() - skip_blocks ) * 4)),
             Int32PtrTy);
 
         LoadInst *Idx = IRB.CreateLoad(IRB.getInt32Ty(), GuardPtr);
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thanks!
skipped_blocks is so far always 0, dunno why, so this issue never came up afaik

85lx:
@vanhauser-thc 
Could you please help me figure this out? Or let me know if I'm misunderstanding something.
According to the current code logic, `skip_blocks` is the sum of the number of calls to `__afl_coverage_interesting` (cnt_cov) and the number of basic blocks that contain comparison and selection instructions which do not affect branch jumps (cnt_sel). How could it be 0?
If skipped_blocks is always 0, doesn't that mean the instrumentation logic related to __afl_coverage_interesting, cmpInst, and SelectInst is never triggered?

vanhauser-thc:
Skip blocks is the number of blocks that do not need to be instrumented by pcguard as they are already instrumented through select etc. but so far this never happened.

85lx:
@vanhauser-thc Thank you for your reply! It was very helpful and greatly appreciated.

PR #2514, opened 2025-08-07
Allow to disable the build of QEMU support and unicorn

fabiankeil:
While at it, remove leading spaces from a line and remove a duplicated 'frida_mode' in the help target output.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -322,6 +322,12 @@ ifdef TEST_MMAP
 	LDFLAGS += -Wno-deprecated-declarations
 endif
 
+ifeq "$(ARCH)" "aarch64"
+  ifdef NO_UNICORN_ARM64
+	NO_UNICORN=1
+  endif
+endif
+
 .PHONY: all
 all:	test_x86 test_shm test_python ready $(PROGS) llvm gcc_plugin test_build all_done
 	-$(MAKE) -C utils/aflpp_driver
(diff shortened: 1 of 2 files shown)


GNUmakefile
```diff
@@ -383,7 +383,7 @@ help:
 	@echo "HELP --- the following make targets exist:"
 	@echo "=========================================="
 	@echo "all: the main AFL++ binaries and llvm/gcc instrumentation"
-	@echo "binary-only: everything for binary-only fuzzing: frida_mode, nyx_mode, qemu_mode, frida_mode, unicorn_mode, coresight_mode, libdislocator, libtokencap"
+	@echo "binary-only: everything for binary-only fuzzing: frida_mode, nyx_mode, qemu_mode, unicorn_mode, coresight_mode, libdislocator, libtokencap"
```
vanhauser-thc, line 392: why this change? if frida mode is not built, then this is a bug and needs to be added

GNUmakefile
```diff
@@ -414,7 +414,8 @@ help:
 	@echo "NO_UTF - do not use UTF-8 for line rendering in status screen (fallback to G1 box drawing, of vanilla AFL)"
 	@echo NO_NYX - disable building nyx mode dependencies
 	@echo "NO_CORESIGHT - disable building coresight (arm64 only)"
-	@echo NO_UNICORN_ARM64 - disable building unicorn on arm64
```
vanhauser-thc, line 417: do not remove those please - they are there as they are used

vanhauser-thc:
thanks, left a few questions

GNUmakefile
fabiankeil: frida_mode is listed twice on the same line. Once seemed sufficient to me.

GNUmakefile
fabiankeil: Do you propose to add NO_UNICORN in addition to NO_UNICORN_ARM64?

GNUmakefile
vanhauser-thc: ah right, sorry

GNUmakefile
vanhauser-thc: if you want to have NO_UNICORN then yes it needs both. you can make it simper by setting NO_UNICORN if NO_UNICORN_ARM64 is set and then working with NO_UNICORN as is in the makefile

vanhauser-thc:
Lgtm, thanks

PR #2517, opened 2025-08-14
Fix LLVM 20 Memory Allocation Crash in SanitizerCoveragePCGUARD Pass

tobgaertner:
# Fix LLVM 20 Memory Allocation Crash in SanitizerCoveragePCGUARD Pass

## Problem
AFL++ crashes with LLVM 20 when building with the SanitizerCoverage pass enabled. The crash occurs in `SmallVectorTemplateBase::grow()` during pass registration:

```
Stack dump:
0. Program arguments: /usr/local/llvm-20/bin/clang -cc1 -triple x86_64-unknown-linux-gnu -emit-obj [...]
1. <eof> parser at end of file
#3 0x0000000002075198 llvm::SmallVectorTemplateBase<std::function<void (llvm::PassManager<llvm::Module, llvm::AnalysisManager<llvm::Module>>&, llvm::OptimizationLevel, llvm::ThinOrFullLTOPhase)>, false>::grow(unsigned long)
#8 0x00007f77df762e64 llvm::PassBuilder::registerOptimizerEarlyEPCallback(...)
#12 0x00007f77df75ba5e llvmGetPassPluginInfo::$_0::operator()(llvm::PassBuilder&) const /opt/AFLplusplus/instrumentation/SanitizerCoveragePCGUARD.so.cc:228:13
```

## Root Cause
The issue is in the `llvmGetPassPluginInfo()` function where the lambda callback for LLVM 20 includes the `ThinOrFullLTOPhase Phase` parameter but doesn't properly handle it. This causes undefined behavior when LLVM tries to register the pass callback, leading to memory allocation failures.

## Solution
This PR fixes the issue by:

1. **Properly handling the `ThinOrFullLTOPhase` parameter** in LLVM 20 builds
2. **Adding phase-aware pass registration** to avoid conflicts during LTO phases
3. **Maintaining backward compatibility** with older LLVM versions

## Changes
- Modified `instrumentation/SanitizerCoveragePCGUARD.so.cc` to properly handle LLVM 20's `ThinOrFullLTOPhase` parameter
- Added conditional logic to only register the pass during appropriate compilation phases
- Added comments explaining the LLVM 20 compatibility fix

## Testing
- ✅ Builds successfully with LLVM 20.1.8 on Amazon Linux 2
- ✅ No memory allocation crashes during pass registration
- ✅ AFL++ instrumentation works correctly
- ✅ Backward compatibility maintained with LLVM 16-19

## Impact
This fix enables AFL++ to work properly with LLVM 20, which is important for:
- Modern Linux distributions shipping LLVM 20
- Users building AFL++ from source with latest LLVM
- CI/CD systems using recent LLVM versions

Fixes the crash reported in multiple issues where AFL++ fails to build with LLVM 20.

### DIFF
--- a/instrumentation/SanitizerCoveragePCGUARD.so.cc
+++ b/instrumentation/SanitizerCoveragePCGUARD.so.cc
@@ -217,38 +217,36 @@ class ModuleSanitizerCoverageAFL
 
 }  // namespace
 
-extern "C" ::llvm::PassPluginLibraryInfo LLVM_ATTRIBUTE_WEAK
-llvmGetPassPluginInfo() {
-
-  return {LLVM_PLUGIN_API_VERSION, "SanitizerCoveragePCGUARD", "v0.2",
-          /* lambda to insert our pass into the pass pipeline. */
-          [](PassBuilder &PB) {
-
+extern "C" ::llvm::PassPluginLibraryInfo LLVM_ATTRIBUTE_WEAK llvmGetPassPluginInfo() {
+  return {
+    LLVM_PLUGIN_API_VERSION, 
+    "SanitizerCoveragePCGUARD", 
+    "v0.2",
+    [](PassBuilder &PB) {
 #if LLVM_VERSION_MAJOR >= 16
-            PB.registerOptimizerEarlyEPCallback([](ModulePassManager &MPM,
-                                                   OptimizationLevel  OL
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thanks! (all PRs need to be directed to the dev branch)
lets see if things still work but should :-)

Issue #2518, opened 2025-08-14
ASAN/UBSAN errors in afl_mutate


TERESH1:
Checked on the last commit of the **dev** branch (e630b6dbe149e116683b5266eb198358d48ea4c4)

**Describe the bug**
`afl_mutate` contains too many errors (stripped log):
- UBSAN:
```
include/afl-mutations.h:2014:31: runtime error: load of misaligned address 0x55d015cef2b1 for type 'u16', which requires 2 byte alignment
include/afl-mutations.h:2014:29: runtime error: store to misaligned address 0x55d015cef2b1 for type 'u16', which requires 2 byte alignment
include/afl-mutations.h:1912:50: runtime error: store to misaligned address 0x55d015cef2b1 for type 'u16', which requires 2 byte alignment
include/afl-mutations.h:1899:50: runtime error: store to misaligned address 0x55d015cef2b1 for type 'u16', which requires 2 byte alignment
include/afl-mutations.h:2000:29: runtime error: load of misaligned address 0x55d015cef2b1 for type 'u16', which requires 2 byte alignment
include/afl-mutations.h:2000:29: runtime error: store to misaligned address 0x55d015cef2b1 for type 'u16', which requires 2 byte alignment
include/afl-mutations.h:1986:31: runtime error: load of misaligned address 0x55d015cef2b1 for type 'u16', which requires 2 byte alignment
include/afl-mutations.h:1986:29: runtime error: store to misaligned address 0x55d015cef2b1 for type 'u16', which requires 2 byte alignment
include/afl-mutations.h:1972:29: runtime error: load of misaligned address 0x55d015cef2b1 for type 'u16', which requires 2 byte alignment
include/afl-mutations.h:1972:29: runtime error: store to misaligned address 0x55d015cef2b1 for type 'u16', which requires 2 byte alignment
include/afl-mutations.h:2042:31: runtime error: load of misaligned address 0x55d015cef2b1 for type 'u32', which requires 4 byte alignment
include/afl-mutations.h:2042:29: runtime error: store to misaligned address 0x55d015cef2b1 for type 'u32', which requires 4 byte alignment
include/afl-mutations.h:2070:31: runtime error: load of misaligned address 0x55d015cef2b1 for type 'u32', which requires 4 byte alignment
include/afl-mutations.h:2070:29: runtime error: store to misaligned address 0x55d015cef2b1 for type 'u32', which requires 4 byte alignment
[...]

PR #2519, opened 2025-08-14
Fix ASAN/UBSAN errors in afl_mutate

TERESH1:
#2518

### DIFF
--- a/include/types.h
+++ b/include/types.h
@@ -161,6 +161,50 @@ typedef int128_t s128;
                                                \
   })
 
+#define EXTRACT16(_s, _o)      \
+  ({                           \
+                               \
+    u8 *s = (u8 *)(_s) + (_o); \
+    u16 _ret = s[1];           \
+    _ret = (_ret << 8) | s[0]; \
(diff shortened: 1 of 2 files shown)


include/types.h
```diff
@@ -161,6 +161,50 @@ typedef int128_t s128;
                                                \
   })
 
+#define EXTRACT16(_s, _o)      \
+  ({                           \
+                               \
+    u8 *s = (u8 *)(_s) + (_o); \
```
vanhauser-thc, line 167: shouldn't this be
```
u8 *s = (u8 *)((_s) + (_o));
```
and same for all others?

vanhauser-thc:
thanks! left one comment.

include/types.h
TERESH1: Depending on the behavior we want. In this patch, it is only used with `u8 *buf` and there are no problems. 

But when a type of a different size (`u16`, `u32`) is used, the `*s` pointer will point at the offset for that type. For example:

https://godbolt.org/z/qGTr1zhrd

Issue #2534, opened 2025-08-26
Unicornv3 build issues


vanhauser-thc:
@wtdcode  @domenukk 

Building the prerequisites for unicorn results in a fail. I know this is not unicorn, but users will run into this issue, so either the build script needs further checks to ensure the environment is fine, or this needs documentation.
**The reason this fails here is that my rust version is too old, something that will be common I guess.**
The problem description is nested deep in the messages. Hope to cope more user-friendly with this?

```
[*] Runing quietly: "/prg/afl/unicorn_mode/.venv/bin/maturin develop --release" under workding directory /prg/afl/unicorn_mode/unicornafl

[!] Failed with:
b'Collecting unicorn>=2.1.3\n  Downloading unicorn-2.1.3-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (4.2 kB)\nDownloading unicorn-2.1.3-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (16.4 MB)\n   \xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81\xe2\x94\x81 16.4/16.4 MB 41.3 MB/s eta 0:00:00\nInstalling collected packages: unicorn\nSuccessfully installed unicorn-2.1.3\n'
[...]

PR #2535, opened 2025-08-26
Unicorn Mode CI Fix

wtdcode:
Solve #2534


vanhauser-thc:
one questions, one comment:

Q: is the minimum version stable? or is this something you will need to touch every few months? with libafl for example this is a very moving target

C: running rustup can lead to unhappy users (e.g. disk space low ...) I would prefer to abort and let the user run rustup, WDYT?

wtdcode:
> one questions, one comment:
> 
> Q: is the minimum version stable? or is this something you will need to touch every few months? with libafl for example this is a very moving target

Not stable at all. libafl actually almost always chasing the latest rust stable version, mostly implicitly. It was even relying on nightly version just a few months ago and we just made it work (fast!) with stable rust toolchains. 

> 
> C: running rustup can lead to unhappy users (e.g. disk space low ...) I would prefer to abort and let the user run rustup, WDYT?

Due to the reason above, actually it is hard to safeguard with correct rustc version and I would say it is safe to always `rupst uprade`. Perhaps with a minimal profile to reduce disk space.

unicorn_mode/build_unicorn_support.py
```diff
@@ -59,15 +67,31 @@ def detect_from_env_or_file(target: str):
...
-
+    exit(2)
+else:
+    if not detect_rustc_version():
+        print("[!] Your rustc seems too old to build unicornafl and libafl")
+        
+        if shutil.which("rustup"):
+            print("[*] We will help you upgrade your rust version")
+            run_cmd("rustup upgrade", None, False)
+            run_cmd("rustup default stable", None, False)
```
domenukk: We shouldn't do this, just notify the user about what they have to do. 
Some users may manually install rust from apt, too

wtdcode: Okay, make sense. Will put an error in later building process.

vanhauser-thc: thanks!

vanhauser-thc:
Also important and can be fixed in this PR: our CI tests.
test/test-unicorn-mode.sh tests that unicorn is working correctly.
**IMPORTANT**: I already updated the script so that the python venv is loaded, so sync to the dev branch first.
But I do not know how to modify the tests so they work.

vanhauser-thc:
@wtdcode ping :-) any idea when you can finish this PR? we have no CI for the unicorn stuff ...

wtdcode:
The persistent testing sample is weirdly failing. I need to check a bit.

wtdcode:
There is a bug in Unicorn Rust bindings I can confirm.

vanhauser-thc:
@wtdcode @domenukk 
we have some important fixes in the dev branch that need to be pushed to stable.
but I dont want to push to stable until we test unicorn in the CI (and it passes).

when can this be done? @domenukk can you help?
If this takes too long I am thinking about copying the state into a branch and revert the unicorn stuff until it is ready to be remerged.

Thanks!

wtdcode:
> @wtdcode @domenukk we have some important fixes in the dev branch that need to be pushed to stable. but I dont want to push to stable until we test unicorn in the CI (and it passes).
> 
> when can this be done? @domenukk can you help? If this takes too long I am thinking about copying the state into a branch and revert the unicorn stuff until it is ready to be remerged.
> 
> Thanks!

Mostly due to unicorn experiencing a refactoring, see we just created this: https://github.com/unicorn-engine/unicorn-engine-rs

I will fix this weekend.

wtdcode:
Sorry, I should have updated the status:

Mostly samples now passing (locally) and the things left to be done:

- Fix unicorn bindings
- Add CI workflows

wtdcode:
I just noticed that `test-unicorn-mode.sh` is already included in the workflow so that makes the work much easier... Sorry I thought I had to write a new workflow.

I also added cmplog test and now it should be complete, let's wait for CI =).

vanhauser-thc:
thank you for the update!

wtdcode:
https://github.com/AFLplusplus/AFLplusplus/actions/runs/17857797340/job/50780797835?pr=2535#step:5:278

@vanhauser-thc Seems it is skipped again, any idea?

wtdcode:
I tried to add `binary-only` to `tests` makefile target. Let me know if this is not expected.

wtdcode:
https://github.com/AFLplusplus/AFLplusplus/actions/runs/17874892342/job/50834890879?pr=2535#step:9:1669

CI seems buggy here by terminating too early.

vanhauser-thc:
> https://github.com/AFLplusplus/AFLplusplus/actions/runs/17874892342/job/50834890879?pr=2535#step:9:1669
> 
> CI seems buggy here by terminating too early.

yes this is a bug. fixed it in dev.

vanhauser-thc:
ah damn, you based your state from stable not dev.

does `test/test-unicorn-mode.sh` work for you?

wtdcode:
I thought I have synced?

vanhauser-thc:
I have merged and will test if it works now. if not then lets do another pr ...

wtdcode:
Okay, I'm available in next few hours.

vanhauser-thc:
it works partially:

```
[*] Loading python venv for Unicorn
[*] Testing: unicorn_mode
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'setuptools'
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'setuptools'
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'setuptools'
[-] Error: Python setup-tools not found. Run 'sudo apt-get install python-setuptools'.
[*] running afl-fuzz for unicorn_mode (persistent), this will take approx 25 seconds
[+] afl-fuzz is working correctly with unicorn_mode (persistent)
[*] Using python binary 
[-] we cannot test unicorn_mode for python because it is not present
```

the reason is that the setuptools have not been installed into venv

wtdcode:
Oh okay, but it should be there unless like uv?

vanhauser-thc:
when the venv is created then maturin is installed, but not setuptools. I just added this. now it should work.

wtdcode:
https://github.com/AFLplusplus/AFLplusplus/actions/runs/17880902917/job/50848438479#step:9:1678

Seems working! Going to bed =).

vanhauser-thc:
yes, thanks for your work!

PR #2537, opened 2025-09-05
Fix __sanitizer_cov_trace_pc_guard implementation

jschwartzentruber:
`__GNUC__` is set for clang, which was causing `__sanitizer_cov_trace_pc_guard` to use the wrong implementation.

 `#if (LLVM_VERSION_MAJOR < 9)` was true because `llvm-config.h` was never included.


vanhauser-thc:
Thanks

PR #2538, opened 2025-09-06
Fix: Add default case for cmplog i686 (#2536)

### LINKED ISSUE #2536: Problems with the cmplog for i686
When building a fuzz target for pcre2 with cmplog with the -m32 flag, the linking error __cmplog_ins_hookN appears. I found that this error appears due to the fact that [cmplog-instructions-pass](https://github.com/AFLplusplus/AFLplusplus/blob/eadc8a2a7e0fa0338802ee6254bf296489ce4fd7/instrumentation/cmplog-instructions-pass.cc#L532) does not have restrictions like [cmplog-switches-pass](https://github.com/AFLplusplus/AFLplusplus/blob/eadc8a2a7e0fa0338802ee6254bf296489ce4fd7/instrumentation/cmplog-switches-pass.cc#L307) for x86 builds. This causes the __cmplog_ins_hookN call to be added, but there is no such function in [x86 runtime of afl](https://github.com/AFLplusplus/AFLplusplus/blob/eadc8a2a7e0fa0338802ee6254bf296489ce4fd7/instrumentation/afl-compiler-rt.o.c#L2168).

I think there are two ways to solve this problem:
1) Add to cmplog-instructions-pass.cc conditional compilation as in cmplog-switches-pass.cc
2) Rewrite the [block](https://github.com/AFLplusplus/AFLplusplus/blob/eadc8a2a7e0fa0338802ee6254bf296489ce4fd7/instrumentation/cmplog-instructions-pass.cc#L395) like this:

```c
switch (max_size) {
        case 8:
          break;
        case 9 ... 16:
          cast_size = 16;
          break;
        case 17 ... 32:
          cast_size = 32;
          break;
        case 33 ... 64:
          cast_size = 64;
          break;
        case 80:
        case 128:
          cast_size = max_size;
          break;
        default:
          cast_size = 128;
}
```

akshay-juneja:
Fixes #2536

This pull request addresses the #2536 Problems with the cmplog for i686 issue by implementing the proposed solution. It adds a default case to the switch (max_size) block in instrumentation/cmplog-instructions-pass.cc .

### DIFF
--- a/instrumentation/cmplog-instructions-pass.cc
+++ b/instrumentation/cmplog-instructions-pass.cc
@@ -18,7 +18,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
-
+#include <cstdint>
 #include <iostream>
 #include <list>
 #include <string>
@@ -392,25 +392,50 @@ bool CmpLogInstructions::hookInstrs(Module &M, DomTreeCallback DTCallback) {
       }
 
       // do we need to cast?
-switch (max_size) {
+#if INTPTR_MAX == INT32_MAX
+    /* 32-bit code */
+    switch (max_size) {
         case 8:
-          break;
+            break;
         case 9 ... 16:
-          cast_size = 16;
-          break;
+            cast_size = 16;
+            break;
         case 17 ... 32:
-          cast_size = 32;
-          break;
+            cast_size = 32;
+            break;
         case 33 ... 64:
-          cast_size = 64;
-          break;
+            cast_size = 64;
+            break;
         case 80:
+            break;
         case 128:
-          cast_size = max_size;
-          break;
+            cast_size = max_size;
+            break;
         default:
-          cast_size = 128;
-}
+            cast_size = 128;
+            break;
+    }
+#else
+    /* original code */
+    switch (max_size) {
+        case 8:
+            break;
+        case 9 ... 16:
+            cast_size = 16;
+            break;
+        case 17 ... 32:
+            cast_size = 32;
+            break;
+        case 33 ... 64:
+            cast_size = 64;
+            break;
+        case 80:
+            break;
+        case 128:
+            cast_size = max_size;
+            break;
+    }
+#endif
 
       // XXX FIXME BUG TODO
       if (is_fp && vector_cnt) { continue; }
--- a/instrumentation/cmplog-instructions-pass.cc
+++ b/instrumentation/cmplog-instructions-pass.cc
@@ -392,19 +392,25 @@ bool CmpLogInstructions::hookInstrs(Module &M, DomTreeCallback DTCallback) {
       }
 
       // do we need to cast?
-      switch (max_size) {
-
+switch (max_size) {
         case 8:
-        case 16:
-        case 32:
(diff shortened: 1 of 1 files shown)


akshay-juneja:
Hi, I just submitted this PR. After some thought, I realize I should have done more to confirm the proposed solution before submitting the code. I didn't verify the fix on a local machine. I'm new to this, and I apologize for the oversight. I am committed to learning how to properly contribute. Can you please guide me on how to best test this change?

vanhauser-thc:
well you should check that this fix works for you, right?
Also your fix affects everyone - so also the majority of people who use 64 bit systems. you should keep the original code, plus your new code, switch them at compile time with an `#ifdef`

akshay-juneja:
Thank you for the detailed feedback. I appreciate you taking the time to review my PR. I will get to work on properly testing the fix and implementing the #ifdef solution as you suggested.

akshay-juneja:
@vanhauser-thc  What is the correct preprocessor macro I should use for the i686 architecture in this project?

vanhauser-thc:
you can do:
```
#include <stdint.h>
#if INTPTR_MAX == INT32_MAX
    /* 32-bit code */
#else
    /* original code */
#endif
```

akshay-juneja:
@vanhauser-thc  I've implemented the conditional logic for the switch statement as you suggested. I'm now working on setting up a 32-bit environment to test the fix. I'll provide another update once I have verified that it resolves the issue. Thank you for your guidance!

vanhauser-thc:
looks good to me, now check if it works for you :-)

koltiradw:
@vanhauser-thc @andreafioraldi  @akshay-juneja 
I still think it's better to add a condition to the compile time like [here](https://github.com/AFLplusplus/AFLplusplus/blob/eadc8a2a7e0fa0338802ee6254bf296489ce4fd7/instrumentation/cmplog-switches-pass.cc#L307).  Because the logic for choosing the instrumentation in these two plugins is the same.

akshay-juneja:
@vanhauser-thc Hello. I have successfully built the project and am now ready to test the fix. Could you provide the specific command needed to reproduce the cmplog error on a 32-bit system so I can verify the fix?

akshay-juneja:
@koltiradw Thanks for the suggestion, I'll look into that after I have tested and verified my fix.

vanhauser-thc:
> @vanhauser-thc @andreafioraldi @akshay-juneja I still think it's better to add a condition to the compile time like [here](https://github.com/AFLplusplus/AFLplusplus/blob/eadc8a2a7e0fa0338802ee6254bf296489ce4fd7/instrumentation/cmplog-switches-pass.cc#L307). Because the logic for choosing the instrumentation in these two plugins is the same.

thanks for pointing that out - this is a bug. that define does not exist (it comes from config.h)

koltiradw:
ah, i see, sry
then maybe the same patch as [here](https://github.com/AFLplusplus/AFLplusplus/pull/2533/commits/1803f43b4230b5d7b66e5c26ff4d9672b55243ce)

vanhauser-thc:
thank you!

PR #2540, opened 2025-09-10
IJON Full Implementation

vi3tL0u1s:
# IJON Full Implementation

This is a complete implementation of all IJON features for source code instrumentation in AFL++.

**Base commit:** `6b6cc9c1`

**Note to maintainers**:
This PR is based on a commit, that is not the latest dev. If possible, please consider creating a new branch for it.

---

#### `IJON_STACK_MIN(x)`

---

### Memory Layout

To avoid collisions, this implementation uses static shared memory for the coverage bitmap. For targets with bitmaps <= 65k, the following layout is used:

```
Virtual Address Space Layout:
                                        
0x7f1234560000  ┌─────────────────────────────────────┐
                │ AFL++ Shared Memory Region          │
                │ (Total: 69,632 bytes)               │
                │                                     │
0x7f1234560000  ├─────────────────────────────────────┤ <- __afl_area_ptr
                │ Coverage Bitmap (afl_area)          │   (fuzzer: trace_bits)
                │ Size: 65,536 bytes (MAP_SIZE)       │   (target: __afl_area_ptr)
                │ Type: u8[65536]                     │
                │                                     │
                │ [0x0000] = edge_hits[0]             │
                │ [0x0001] = edge_hits[1]             │
                │ ...                                 │
                │ [0xFFFE] = edge_hits[65534]         │
                │ [0xFFFF] = edge_hits[65535]         │
                │                                     │
0x7f1234570000  ├─────────────────────────────────────┤ <- __afl_ijon_bits
                │ IJON Max Values (afl_max)           │   (fuzzer: ijon_bits) 
                │ Size: 4,096 bytes (512 × 8)         │   (target: __afl_ijon_bits)
                │ Type: u64[512]                      │     Static Offset: +65536
                │                                     │
                │ [0] = max_value_slot_0              │
                │ [1] = max_value_slot_1              │
                │ ...                                 │
                │ [25] = 15240170669                  │ <- Tracked value
                │ ...                                 │
                │ [351] = 520011065792645             │ <- Tracked value  
                │ ...                                 │
                │ [458] = 964077327750                │ <- Tracked value
                │ ...                                 │
                │ [511] = max_value_slot_511          │
                │                                     │
0x7f1234571000  └─────────────────────────────────────┘
```

For targets with bitmaps > 65k, a normal dynamic coverage bitmap is used with the IJON max feedback array following it.

---

### Building AFL++ with IJON

```bash
make clean
make LLVM_CONFIG=llvm-config-18 source-only
```

### Compiling Target Programs

When using IJON instrumentation in AFL++, it is required to invoke `__AFL_INIT()` at the beginning of your target program’s `main()` function:
```c
int main(int argc, char **argv) {
    __AFL_INIT();
    // Your code here
}
```
Adding `__AFL_INIT()` ensures:
- Proper initialization of the AFL++ runtime environment
- Correct setup of the coverage bitmap and IJON feedback regions
- Improved compatibility across targets

Set the environment variables during compilation:
```bash
AFL_LLVM_IJON=1 CC=afl-clang-fast CXX=afl-clang-fast++ make
```

#### Basic Fuzzing
```bash
AFL_IJON=1 AFL_IJON_HISTORY_LIMIT=1000 afl-fuzz -i input_dir -o output_dir -- ./target
```

#### For Large Targets (bitmap > 65k)
```bash
echo test > test_input.txt && AFL_DUMP_MAP_SIZE=1 ./target

AFL_MAP_SIZE=10000 AFL_IJON_HISTORY_LIMIT=1000 AFL_IJON=1 afl-fuzz -S worker -i input_dir -o output_dir -- ./target
```

### Environment Variables

- **`AFL_LLVM_IJON=1`**: Enables IJON instrumentation during compilation
- **`AFL_IJON=1`**: Enables IJON feedback during fuzzing  
- **`AFL_IJON_HISTORY_LIMIT=N`**: Sets the maximum number of IJON max-value inputs stored on the host (default: 20)
- **`AFL_MAP_SIZE=N`**: Sets a custom coverage map size for large targets (bitmap > 65k)

---

## Performance (Super Mario Bros. Level 1.1, ijon_max(pos_y/16, world_pos))

Test environment: Ubuntu 20.04, 16 GB RAM, 8 cores

| Run | IJON AFL | IJON AFL++ |
| ---- | -------- | ---------- |
| 1 | 1 h 19 min | 15 min |
| 2 | 50 min | 20 min |
| 3 | 31 min | 21 min |
| 4 | 1 h 22 min | 23 min |
| 5 | 2 h 14 min | 25 min |
| AVG | 1 h 16 min | 20.8 min |

Overall, IJON AFL++ is ~3.7x faster on average (76.0 min -> 20.8 min).

## Performance (Maze, IJON_SET(ijon_hashint(x, y)))

Test environment: Ubuntu 20.04, 16 GB RAM, 8 cores

| Run | Easy Small (AFL ijon) | Easy Small (AFL++ ijon) | Easy Big (AFL ijon) | Easy Big (AFL++ ijon) | Hard Small (AFL ijon) | Hard Small (AFL++ ijon) | Hard Big (AFL ijon) | Hard Big (AFL++ ijon) |
| --- | ------------------------ | -------------------------- | ---------------------- | ------------------------ | ----------------------- | ------------------------- | --------------------- | ----------------------- |
| 1 | 1 min 56 s | 26 s | 15 min 32 s | 54 s | 40 s | 8 s | 22 s | 35 s |
| 2 | 1 min 21 s | 11 s | 10 min 56 s | 2 min 13 s | 25 s | 14 s | 11 min 6 s | 19 s |
| 3 | 1 min 53 s | 42 s | 18 min 18 s | 48 s | 5 min 8 s | 11 s | 10 min 33 s | 53 s |
| 4 | 3 min 25 s | 24 s | 29 min 32 s | 1 min 20 s | 1 min 12 s | 10 s | 2 min 11 s | 20 s |
| 5 | 2 min 28 s | 21 s | 10 min 34 s | 49 s | 19 s | 9 s | 4 min 29 s | 26 s |
| 6 | 2 min 30 s | 24 s | 11 min 49 s | 1 min 24 s | 2 min 37 s | 12 s | 11 min 16 s | 33 s |
| 7 | 1 min 5 s | 22 s | 12 min 18 s | 4 min 6 s | 38 s | 8 s | 10 min 9 s | 32 s |
| 8 | 42 s | 14 s | 11 min 41 s | 56 s | 52 s | 9 s | 5 min 47 s | 18 s |
| 9 | 12 min 18 s | 37 s | 9 min 36 s | 2 min 36 s | 56 s | 9 s | 5 min 24 s | 19 s |
| 10 | 1 min 32 s | 15 s | 12 min 7 s | 47 s | 1 min 5 s | 15 s | 17 min 10 s | 24 s |
| AVG | 2 min 55 s | 23.6 s (x7.42) | 14 min 14 s | 1 min 35 s (x9) | 1 min 23 s | 10.5 s (x7.9) | 7 min 50 s | 27.9 s (x16.85) |


src/afl-fuzz-queue.c
```diff
@@ -684,6 +684,11 @@ void add_to_queue(afl_state_t *afl, u8 *fname, u32 len, u8 passed_det) {
 
   q->fname = fname;
   q->len = len;
+
+  // Debug: Track queue entry creation
+  if (getenv("AFL_DEBUG")) {
```
vanhauser-thc: use `afl->debug` instead

src/afl-fuzz-queue.c
```diff
@@ -1550,6 +1555,11 @@ inline u8 *queue_testcase_get(afl_state_t *afl, struct queue_entry *q) {
   u32    len = q->len;
   double weight = q->weight;
 
+  // Debug: Log queue entry access for debugging
+  if (getenv("AFL_DEBUG")) {
```
vanhauser-thc: dito

src/afl-fuzz-one.c
```diff
@@ -338,6 +339,27 @@ u8 fuzz_one_original(afl_state_t *afl) {
   u8  a_collect[MAX_AUTO_EXTRA];
   u32 a_len = 0;
 
+  /* IJON: If we're doing IJON, skip deterministic stages and go directly to havoc */
+  if (afl->is_doing_ijon) {
```
vanhauser-thc: please insert `unlikely`

src/afl-fuzz-one.c
```diff
@@ -2032,6 +2054,12 @@ u8 fuzz_one_original(afl_state_t *afl) {
 
   }
 
+  /* IJON stage name override */
+  if (afl->is_doing_ijon) {
```
vanhauser-thc: please insert `unlikely`

src/afl-fuzz-one.c
```diff
@@ -3430,12 +3458,20 @@ u8 fuzz_one_original(afl_state_t *afl) {
 /* we are through with this queue entry - for this iteration */
 abandon_entry:
 
+  /* IJON queue protection only - memory cleanup handled normally */
+  if (afl->is_doing_ijon) {
```
vanhauser-thc: please insert `unlikely`

src/afl-fuzz-one.c
```diff
@@ -3430,12 +3458,20 @@ u8 fuzz_one_original(afl_state_t *afl) {
...
+
+  }
+
   afl->splicing_with = -1;
 
   /* Update afl->pending_not_fuzzed count if we made it through the calibration
      cycle and have not seen this entry before. */
 
-  if (!afl->stop_soon && !afl->queue_cur->cal_failed &&
+  if (!afl->is_doing_ijon && !afl->stop_soon && !afl->queue_cur->cal_failed &&
```
vanhauser-thc: please add an unlikely around everything :-)

src/afl-fuzz-one.c
```diff
@@ -3450,7 +3486,9 @@ u8 fuzz_one_original(afl_state_t *afl) {
 
   }
 
-  ++afl->queue_cur->fuzz_level;
+  if (!afl->is_doing_ijon) {
```
vanhauser-thc: please insert `unlikely`

src/afl-fuzz-one.c
```diff
@@ -6172,6 +6210,85 @@ u8 fuzz_one(afl_state_t *afl) {
...
 
+  /* IJON execution path - variables for havoc stage (declare at function level) */
+  u32 len = 0, temp_len = 0;
+  u8 *in_buf = NULL, *out_buf = NULL, *orig_in = NULL;
+  u64 havoc_queued = 0, orig_hit_cnt = 0;
+  u32 perf_score = 100;
+  u8 ret_val = 0;
+  s32 fd = -1;
+
+  /* IJON max tracking: Check if we should use IJON input (80% chance) */
```
vanhauser-thc, line 6218: same as all of the above please

src/afl-fuzz-run.c
```diff
@@ -102,6 +103,67 @@ fsrv_run_result_t __attribute__((hot)) fuzz_run_target(afl_state_t      *afl,
 
   }
 
+  /* Check for new IJON max values after execution */
+  if (afl->ijon_state && afl->ijon_bits) {
```
vanhauser-thc: please insert `unlikely`

src/afl-fuzz.c
```diff
@@ -3650,5 +3721,121 @@ int main(int argc, char **argv_orig, char **envp) {
 
 }
 
+
+#if 0  /* REMOVED - Dead code from previous callback approach */
```
vanhauser-thc: remove if not needed anymore?

vanhauser-thc:
@eqv  FYI

vanhauser-thc:
@vi3tL0u1s thanks a lot for this!

this will be quite some effort to merge, because not only did you base this on an older commit (understandable, took quite some time to implement), but you also ran a code formatter on the whole code base - never ever do this in a PR because this will always create conflicts.

the easiest would be if you check out the current dev branch and copy over all files that you did not need to change - this will resolve all conflicts.
please also remove your config.h in the root of the repository and re-create it as a symlink to include/config.h

I left a few nits to reduce the minimal overhead to non-ijon fuzzing further.

include/debug.h
```diff
@@ -418,8 +419,29 @@ static inline const char *colorfilter(const char *x) {
   do {                                                         \
                                                                \
     s32 _len = (s32)(len);                                     \
+    if (getenv("AFL_DEBUG")) {                                 \
```
vanhauser-thc: with all the changes I overlooked these previously.
can you please put an `#if 0` here? otherwise the impact is quite severe, same for after the `read`

vi3tL0u1s: Thank you so much for reviewing this! Yes, I'll update it accordingly.
My apologies for messing up the format earlier. I'm working on fixing it now.

instrumentation/afl-compiler-rt.o.c
```diff
@@ -1978,6 +2003,43 @@ void __sanitizer_cov_trace_pc_guard_init(uint32_t *start, uint32_t *stop) {
 
 #endif  // __AFL_CODE_COVERAGE
 
+  // IJON SUPPORT: Expand map size when AFL_IJON=1 to ensure AFL tools
+  // read the full map that will contain IJON coverage points
+  if (getenv("AFL_IJON")) {
+    // CONDITIONAL EXPANSION: Different logic for ≤65k vs >65k maps
```
vanhauser-thc: thinking about this, this will not work.
pc_guard_init can be called multiple times during startup, and this will then break.
This should rather be moved into the starting section of the forkserver, WDYT?

instrumentation/afl-compiler-rt.o.c
```diff
@@ -1978,6 +2003,43 @@ void __sanitizer_cov_trace_pc_guard_init(uint32_t *start, uint32_t *stop) {
...
+    if (__afl_map_size <= 65536) {
```
vanhauser-thc: Also for pcguard we have a variable map - always. And your case does not cover LTO.
so what you do here serves no purpose :-)

If we have a fixed size map then your code is stuck. but IMHO that is not a problem - we can just disable ijon support for classic instrumentation (colliding coverage).

For pcguard/lto - Just add the amount of memory you need to __afl_final_loc. No need for setting AFL_MAP on the alf-fuzz side, as the forkserver + afl-fuzz communication ensures that the map has the right size.
You can set a flag in the forkserver communication to make it work correctly in afl-fuzz, e.g with
```
include/types.h:#define FS_OPT_IJON 0x04000000
```
(just look how e.g. FS_OPT_NEWCMPLOG is used)

src/afl-forkserver.c
```diff
@@ -308,6 +308,12 @@ void afl_fsrv_init(afl_forkserver_t *fsrv) {
   /* exec related stuff */
   fsrv->child_pid = -1;
   fsrv->map_size = get_map_size();
+
+  if (getenv("AFL_IJON")) {
```
vanhauser-thc: this is an empty `if`?

instrumentation/afl-compiler-rt.o.c
vi3tL0u1s: You're absolutely right about this, and I might be wrong in my current understanding, thank you for pointing it out!

Current Working Mechanism (as I understand it):
#### Target Side (afl-compiler-rt.o.c):
- During pc_guard_init, __afl_final_loc expands to include the IJON space.
    - Small maps (<=65k): __afl_final_loc = MAP_SIZE - 1 (65535).
    - Large maps (>65k): __afl_final_loc += MAP_SIZE_IJON_BYTES (adds 4096 bytes).
    - Reports the expanded size to the fuzzer via fork server communication.
    - At runtime, the target uses IJON offset calculations for IJON-specific calls.
#### Fuzzer Side (afl-fuzz.c):
- Receives the map size from the target (already includes the IJON expansion).
- Allocates shared memory based on the received size:
    - Small maps: Uses fixed MAP_SIZE_TOTAL (69,632 bytes).
    - Large maps: Uses map_size + MAP_SIZE_IJON_BYTES.
- Calculates the IJON offset and sets the pointer accordingly.
 ```c
  /* UNIFIED SHARED MEMORY: Use static total size for IJON mode */
  size_t shm_size = afl->fsrv.map_size;
  if (getenv("AFL_IJON")) {
    /* CONDITIONAL ALLOCATION: Fixed for ≤65k, dynamic for >65k */
    if (afl->fsrv.map_size <= 65536) {
      /* PRESERVE CURRENT BEHAVIOR: Use fixed MAP_SIZE_TOTAL for small maps */
      shm_size = MAP_SIZE_TOTAL;  // 65536 + 4096 = 69632
      OKF("IJON enabled: fixed layout for map size %u (total: %zu bytes)",
          afl->fsrv.map_size, shm_size);
    } else {
      /* DYNAMIC BEHAVIOR: Calculate based on actual map size for large maps */
      shm_size = afl->fsrv.map_size + MAP_SIZE_IJON_BYTES;
      OKF("IJON enabled: dynamic layout for map size %u (total: %zu bytes)",
          afl->fsrv.map_size, shm_size);
    }
  }
 ```
#### Why It Currently Works (based on my assumption):
- Both sides coordinate the expansion consistently:
    - The target expands the map size.
    - The fuzzer allocates the same expanded size.
    - IJON pointers are computed correctly on both ends.

That said, I may have overlooked the severity of the issue you highlighted. I'll revisit this carefully.

instrumentation/afl-compiler-rt.o.c
vi3tL0u1s: Thank you for this explanation! I'll need a bit of time to review your suggestions carefully!

vi3tL0u1s: Ah, I see now, adding `#define FS_OPT_IJON 0x04000000` makes a lot of sense. This seems like a neat step to remove the need for manually setting `AFL_DUMP_MAP_SIZE=1` and calculating `AFL_MAP_SIZE` beforehand.

vanhauser-thc: I can help you here - it is not that obvious necessarily

vi3tL0u1s: Thank you so much for offering to help! It's getting quite late in my timezone, so I'll take a break for now and will be back tomorrow to continue working on this.

vi3tL0u1s: I've added the `FS_OPT_IJON` flag to improve fork server communication in the latest commit, focusing on the large map case. I'd be grateful if you could review this update.

#### Updated Working Mechanism (Large Bitmap)
##### Target Side (afl-compiler-rt.o.c)
- During `pc_guard_init`, the target checks if IJON is enabled via the AFL_IJON environment variable.
- For large maps (>65k):
    - Dynamically adds `MAP_SIZE_IJON_ENTRIES` (512) to `__afl_final_loc`.
    - Sets the `FS_OPT_IJON` flag during the forkserver handshake.
    - Reports the expanded map size to the fuzzer automatically.
##### Fuzzer Side (afl-fuzz.c)
- Detects the `FS_OPT_IJON` flag during forkserver communication.
- Enables IJON mode by setting `fsrv->use_ijon = 1`.
- Dynamically allocates `map_size + MAP_SIZE_IJON_BYTES` for large maps.
- Calculates the IJON offset and sets the memory pointers accordingly.

vanhauser-thc:
@vi3tL0u1s your PR removes various fixes that came into AFL++ after you started your work.
If you give me write permissions to your fork, I will include those fixes.

vi3tL0u1s:
> @vi3tL0u1s your PR removes various fixes that came into AFL++ after you started your work. If you give me write permissions to your fork, I will include those fixes.

Thank you so much! I've added you to my fork. Really appreciate your help and support!

instrumentation/afl-llvm-pass.so.cc
```diff
@@ -181,6 +181,36 @@ PreservedAnalyses AFLCoverage::run(Module &M, ModuleAnalysisManager &MAM) {
 
   skip_nozero = getenv("AFL_LLVM_SKIP_NEVERZERO");
   use_threadsafe_counters = getenv("AFL_LLVM_THREADSAFE_INST");
+
```
vanhauser-thc: IMHO IJON should only be available in PCGUARD and LTO, not this file.

src/afl-fuzz-init.c
```diff
@@ -2377,6 +2378,16 @@ void setup_dirs_fds(afl_state_t *afl) {
   if (mkdir(tmp, afl->dir_perm)) { PFATAL("Unable to create '%s'", tmp); }
   ck_free(tmp);
 
+  /* IJON max tracking directory (only if IJON is enabled) */
+
+  if (getenv("AFL_IJON")) {
```
vanhauser-thc: this should be moved to afl-fuzz after the forkserver and use fsrv->use_ijon instead

vanhauser-thc:
backported all fixes.
added `IJON-TODO` with what still needs to be done and tested.
moved AFL_IJON_HISTORY_LIMIT

IJON-TODO
```diff
@@ -0,0 +1,8 @@
+* No need for AFL_IJON with afl-fuzz (enable through forkserver)
+* No IJON support in afl-llvn-pass.so.cc
+* Ensure PCGUARD + LTO work fine with IJON (need AFL_IJON env)
+  => SanitizeCoveragePCGUARD.cc + SanitizeCoverageLTO.cc
+  - afl-compiler-rt
+  - afl-fuzz different map size handling needed for ijon parts?
```
vi3tL0u1s: Hi @vanhauser-thc, just to confirm, did you mean that we should remove the fixed memory map size for targets with bitmaps <=65k?

I've tried implementing a unified approach that drops the 65k threshold entirely:

Before (mixed approach):
- <=65k maps: fixed MAP_SIZE_TOTAL (69,632 bytes)
- larger maps: dynamic map_size + MAP_SIZE_IJON_BYTES

After (unified approach):
- All maps: dynamic map_size + MAP_SIZE_IJON_BYTES

This way everything uses consistent allocation logic with no special cases. The changes include:
- Fuzzer side: unified allocation in `afl-fuzz.c`
- Target runtime: unified offset calculation in `afl-compiler-rt.o.c`
- IJON processing: removed `shared_data_t` fixed structure
- Forkserver communication: already working correctly

Does this line up with the direction you had in mind? I think we are ignoring the collision issue, but it makes sense to have a unified approach for all map sizes.

vi3tL0u1s: If that's correct, I'll push the mentioned commit upstream.

vanhauser-thc: Yes, no fixed memory map support. (so also removing the code form afl-llvm-pass.so.c)

The code you added to PCGUARD and LTO will need quite some changes by me because yours does not compile on all support LLVM version we support. I will take care of that.

vanhauser-thc: yes: All maps: dynamic map_size + MAP_SIZE_IJON_BYTES

vanhauser-thc: question about the MAP_SIZE_IJON_BYTES part of the map - do they need to be zeroed from the memory in every single fuzz attempt or only before the ijon mode starts or ends for a queue entry (or during ijon mode)?

vanhauser-thc:
Sorry I do not know a lot about ijon works.

Beside my question in https://github.com/AFLplusplus/AFLplusplus/pull/2540#discussion_r2343923590 - I have another one:
The MAP_SIZE_IJON_BYTES part of the map - why is it a fixed size? You know you could make it variable sized if necessary?

IJON-TODO
vi3tL0u1s: I think for IJON, the MAP_SIZE_IJON_BYTES region is not cleared between fuzzing attempts. The values are meant to persist so that the fuzzer can track the global maximum of 512 variables across queue entries, repeated executions, and the entire fuzzing campaign.

vi3tL0u1s: I have removed the code from `afl-llvm-pass.so.c` in commit `ed34bb8`. Thank you for your confirmation! I’ll push the commit now.

vanhauser-thc:
Is this WIP or outdated?
```
src/afl-fuzz-one.c:6214:16: warning: unused variable
      'temp_len' [-Wunused-variable]
 6214 |   u32 len = 0, temp_len = 0;
      |                ^~~~~~~~
src/afl-fuzz-one.c:6216:7: warning: unused variable
      'havoc_queued' [-Wunused-variable]
 6216 |   u64 havoc_queued = 0, orig_hit_cnt = 0;
      |       ^~~~~~~~~~~~
src/afl-fuzz-one.c:6216:25: warning: unused variable
      'orig_hit_cnt' [-Wunused-variable]
 6216 |   u64 havoc_queued = 0, orig_hit_cnt = 0;
      |                         ^~~~~~~~~~~~
src/afl-fuzz-one.c:6217:7: warning: unused variable
      'perf_score' [-Wunused-variable]
 6217 |   u32 perf_score = 100;
      |       ^~~~~~~~~~
src/afl-fuzz-one.c:6218:6: warning: unused variable
      'ret_val' [-Wunused-variable]
 6218 |   u8 ret_val = 0;
      |      ^~~~~~~
5 warnings generated.
```

vi3tL0u1s:
> Sorry I do not know a lot about ijon works.
> 
> Beside my question in [#2540 (comment)](https://github.com/AFLplusplus/AFLplusplus/pull/2540#discussion_r2343923590) - I have another one: The MAP_SIZE_IJON_BYTES part of the map - why is it a fixed size? You know you could make it variable sized if necessary?

You're right, the MAP_SIZE_IJON_BYTES region is currently fixed. I followed the original IJON paper's design of 512 variables x 8 bytes each = 4096 bytes. This keeps the implementation consistent with the paper for now!

vi3tL0u1s:
> Is this WIP or outdated?
> 
> ```
> src/afl-fuzz-one.c:6214:16: warning: unused variable
>       'temp_len' [-Wunused-variable]
>  6214 |   u32 len = 0, temp_len = 0;
>       |                ^~~~~~~~
> src/afl-fuzz-one.c:6216:7: warning: unused variable
>       'havoc_queued' [-Wunused-variable]
>  6216 |   u64 havoc_queued = 0, orig_hit_cnt = 0;
>       |       ^~~~~~~~~~~~
> src/afl-fuzz-one.c:6216:25: warning: unused variable
>       'orig_hit_cnt' [-Wunused-variable]
>  6216 |   u64 havoc_queued = 0, orig_hit_cnt = 0;
>       |                         ^~~~~~~~~~~~
> src/afl-fuzz-one.c:6217:7: warning: unused variable
>       'perf_score' [-Wunused-variable]
>  6217 |   u32 perf_score = 100;
>       |       ^~~~~~~~~~
> src/afl-fuzz-one.c:6218:6: warning: unused variable
>       'ret_val' [-Wunused-variable]
>  6218 |   u8 ret_val = 0;
>       |      ^~~~~~~
> 5 warnings generated.
> ```

Yes, this is still WIP. These warnings come from the IJON implementation, and I'm in the process of cleaning them up!

vanhauser-thc:
> > Sorry I do not know a lot about ijon works.
> > Beside my question in [#2540 (comment)](https://github.com/AFLplusplus/AFLplusplus/pull/2540#discussion_r2343923590) - I have another one: The MAP_SIZE_IJON_BYTES part of the map - why is it a fixed size? You know you could make it variable sized if necessary?
> 
> You're right, the MAP_SIZE_IJON_BYTES region is currently fixed. I followed the original IJON paper's design of 512 variables x 8 bytes each = 4096 bytes. This keeps the implementation consistent with the paper for now!

Does one IJON macro use one u64 entry in the ijon map?
If yes what happens if a user uses more than 512?

vi3tL0u1s:
> > > Sorry I do not know a lot about ijon works.
> > > Beside my question in [#2540 (comment)](https://github.com/AFLplusplus/AFLplusplus/pull/2540#discussion_r2343923590) - I have another one: The MAP_SIZE_IJON_BYTES part of the map - why is it a fixed size? You know you could make it variable sized if necessary?
> > 
> > 
> > You're right, the MAP_SIZE_IJON_BYTES region is currently fixed. I followed the original IJON paper's design of 512 variables x 8 bytes each = 4096 bytes. This keeps the implementation consistent with the paper for now!
> 
> Does one IJON macro use one u64 entry in the ijon map? If yes what happens if a user uses more than 512?

Only the `IJON_MAX` macro makes use of the IJON map directly. Other macros such as `IJON_STATE`, `IJON_SET`, and `IJON_INC` do not consume entries in the IJON map; instead, they modify the normal coverage map by marking new indices to represent state changes in different ways.

It is possible for a user to specify more than 512 `IJON_MAX` entries, in which case collisions will occur in the fixed map. However, this is generally acceptable in practice, as the IJON design philosophy emphasizes selective tracking of critical variables rather than unbounded use.

vi3tL0u1s:
Sorry, I forgot to update some of the logic for `IJON_SET` and `IJON_INC`. I'm working on fixing this and will include the changes in the next commit.

instrumentation/afl-compiler-rt.o.c
```diff
@@ -2999,3 +3090,382 @@ void __afl_injection_xss(u8 *buf) {
...
+
+  // Use actual available shared memory size when AFL_IJON=1
+  u32 available_map_size = 0;
+  if (&__afl_ijon_enabled != NULL) {
+    // Use target's calculated size for standalone mode
+    available_map_size = __afl_map_size;
+  }
+
+// BUG TODO -> THIS IS WRONG ... what is this meant to do?
+  u32 coverage_id = combined_hash % available_map_size;
```
vanhauser-thc: this is NOT ok - you cannot overwrite just any coverage map entry. As we have non-colliding coverage, you are overwriting the byte for a specific edge.

So my question - do specific macros need a coverage map entry? If so, this can be done, it is trivial.

vi3tL0u1s: I see now, since AFL++ guarantees a non-colliding coverage bitmap, overwriting an index in the bitmap is indeed not correct. That said, this is how IJON_SET was originally implemented in AFL.

The idea behind IJON_SET is to hash the location together with the value of a variable and use that as a new index in the bitmap, then set the hit count to 1. If we were to change this to use incremental map entries, the challenge is that we don't always know the possible range of values for a variable, and each distinct value would require its own entry. This could easily lead to an explosion in the number of entries. I'm not entirely sure what the best approach would be for handling this in AFL++.

vi3tL0u1s: But again, if we follow the IJON philosophy of selective tracking (not injecting too many IJON_SET and IJON_INC macros), then adding more entries could still make sense in practice. WDYT?

vanhauser-thc:
After looking into the code I now better understand how this works.

A problem that I see is that with a variable map size ijon functions set, inc and xor_state need a large map to not even collide with own ijon functions.
If it is a small target with eg. 100 edges, then 25% will likely be reached by an input, and any of these three ijon functions will map into the 100 edges.

I think this needs a different way to solve his.
How about that if ijon is used then 64k + 4k extra map space is reserved? the three ijon functions map into the 64k area, the min/max stay in the 4k map.

The disadvantage is that it slows the fuzzing down because of the +64kb map that needs to be cleaned for every fuzz run, and processed for coverage afterwards. but on the plus side it would not impact the real coverage of the target
WDYT?

vi3tL0u1s:
Thank you for the feedback regarding the risk of IJON_SET overwriting coverage map entries, you're absolutely right that this is an important issue to address.

I agree that reserving a fixed 64k + 4k map size is a clean and effective approach, and it can be integrated with minimal changes. The performance overhead from the larger map should be acceptable, IMHO.

I had previously explored an alternative design where the coverage bitmap was expanded dynamically by 15% to create space for IJON_SET and IJON_INC entries. The idea was to: expand the target's map size by 15% beyond its actual coverage needs; hash IJON_SET entries into this expanded region to prevent collisions with regular coverage; and synchronize this expansion with the fuzzer through the forkserver protocol. However, this approach did not work reliably in practice because the expanded region still exhibited a high number of collisions for small targets (particularly problematic for IJON_INC), and the 15% factor itself was arbitrary with no strong justification.

Given these issues, I think the fixed 64k + 4k solution is the better choice moving forward.

vanhauser-thc:
> Given these issues, I think the fixed 64k + 4k solution is the better choice moving forward.

OK, I will implement this today and correct the map handling everywhere.

eqv:
I think it's true that ijon would need a certain minimum map size. Having min/max be their own area is also quite sensible. 
Having Set/Inc etc live in their own map should also work as expected.
But ijon_state is kinda annoying to combine with collision free coverage. In theory you could make a dynamic number of individual collision free bitmaps (i.e. one per state variable), but that doesn't sound efficient to me (though a sane-ish implementation could be adding just one extra pointer indirection: have an array of size max_states of bitmap pointers that all point to a read only reserved memory region & when we segfault on one of those entries, we allocate the actual pages needed for this state entry. In the normal case this "just" adds one pointer lookup through state_map_array[i=0] when the state changes. But the complexity seems dubious. I haven't found a lot of use in the ijon_state feature in first place tbh. It would probably require bigger integrations in scheduling etc to avoid spamming the queue too badly with trivial copies. 

*edit* In hindsight: you probably don't even need to have a page fault handler: just allocate 16 or 32 copies of the collision free bitmap ahead of time with empty pages. No changes to the instrumentation needed, just swapping the bitmap pointer when calling ijon_state to the correct copy ... (and obviously you'd need some 16/32 bit bitmap so that the novely check code can be checking which maps to look at, so we don't scan all the pointless maps that never got touched.)
  
Alternatively, you could use the original collision free map range for the case where state == 0 and then use the remaining 64k map to do old school hashed mapping with the state variable added in & hope for the best in terms of collisions?

vanhauser-thc:
I pushed the change with the fixed 64kb ijon map and adapted the code in afl-fuzz as well.
I will have to check if cmplog and SAN still work as expected though - that is on my TODO list.

instrumentation/afl-compiler-rt.o.c
```diff
@@ -2999,3 +3081,346 @@ void __afl_injection_xss(u8 *buf) {
...
+  __afl_ijon_state = (__afl_ijon_state ^ val) % state_modulo;
+
+  if (__afl_debug) {
+    fprintf(stderr, "[IJON_STATE] State: %u, modulo: %u, layout: unified\n",
+            __afl_ijon_state, state_modulo);
+  }
+}
+
+void ijon_push_state(uint32_t x) {
+  // TODO: is this correct? xor - move x into val - then xor again?
```
vanhauser-thc: can you have a look that this is fine? it looks weird to me but on the other hand I don't know that much about ijon :-)

vi3tL0u1s: Let me clarify how `IJON_STATE` and `IJON_CTX` are supposed to work and why they look a bit unusual at first glance.

`IJON_STATE` uses `ijon_xor_state()`
- Persistent: updates the global __afl_ijon_state until explicitly changed or reset.
- Effect: all subsequent edges are recorded as coverage_map[state xor edge_id]
- Example:
```c
void process_input(char *input) {
    if (input[0] == 'A') {
        IJON_STATE(1);  // __afl_ijon_state = 1
        path_A();       // Let's say this has edges: 100, 101, 102
                       // Actually recorded at: (100 xor 1), (101 xor 1), (102 xor1)
                       // = 101, 100, 103
    } else {
        IJON_STATE(2);  // __afl_ijon_state = 2  
        path_B();       // Same function, same original edges: 100, 101, 102
                       // Actually recorded at: (100 xor 2), (101 xor 2), (102 xor 2)  
                       // = 102, 103, 100
    }
}
```

`IJON_CTX(x)` uses `ijon_xor_state()` twice
- Scoped: temporarily xor in a hash (from __LINE__, __FILE__), evaluates x, and then restores the old state.
- Effect: only edges inside x are influenced, then state resets automatically.
- Example:
```c
void process_data(int value) {
    // State = 0 initially
    
    int result1 = IJON_CTX(expensive_function(value));
    // During expensive_function(): state = hash(__LINE__, __FILE__)
    // After IJON_CTX: state restored to 0
    
    int result2 = IJON_CTX(expensive_function(value + 1));  
    // Different line = different hash = different state context
    // After IJON_CTX: state restored to 0
    
    common_cleanup();  // Always executed with state = 0
}
```
Compiler side (in SanitizerCoveragePCGUARD.so.cc):
Both rely on the same mechanism - LLVM inserts an xor with the current IJON state:
```c
// Apply IJON state-aware coverage if enabled
if (ijon_enabled && AFLIJONState) {
  LoadInst *IJONStateVal = IRB.CreateLoad(Int32Ty, AFLIJONState);
  // Apply IJON formula: state XOR coverage_index
  CoverageIndex = IRB.CreateXor(IJONStateVal, CoverageIndex);
}
```

Key difference
- IJON_STATE: persistent until next call.
- IJON_CTX: temporary, auto-restores (scoped).

Both rely on `ijon_xor_state()` which currently has the buffer overrun issue we discussed - the state modulo needs to match `__afl_map_size` to ensure the XOR results stay within map bounds.

vi3tL0u1s: Sorry for the confusion about ijon_push_state(). This function is currently unused.
I initially added it while experimenting with IJON_CTX, but the final implementation uses automatic XOR reversal instead of manual push/pop semantics. I'll clean this up in the next commit.

instrumentation/SanitizerCoveragePCGUARD.so.cc
```diff
@@ -1706,7 +1706,10 @@ void ModuleSanitizerCoverageAFL::InjectCoverageAtBlock(Function   &F,
       ModuleSanitizerCoverageAFL::SetNoSanitizeMetadata(IJONStateVal);
       // Apply IJON formula: state XOR coverage_index
       // This makes every coverage point context-aware based on current IJON state
-      CoverageIndex = IRB.CreateXor(IJONStateVal, CoverageIndex);
+      Value *XorResult = IRB.CreateXor(IJONStateVal, CoverageIndex);
+      // Ensure result stays within map bounds to prevent buffer overruns
+      Value *MapSize = ConstantInt::get(Int32Ty, MAP_SIZE);
+      CoverageIndex = IRB.CreateURem(XorResult, MapSize);
     }
```
vi3tL0u1s, line 1760: With the code changes here, I'm trying to partially avoid the collision issue @eqv mentioned.
- State Range Limitation: IJON state is constrained to [0, 65535] via MAP_SIZE_IJON_MAP modulo.
- Bounded XOR Results: CreateURem(XorResult, MAP_SIZE) keeps all results within the standard 65k coverage map.
- Collision Reduction: This doesn't eliminate collisions entirely, but it significantly lowers the probability compared to the previous unbounded xor.

vi3tL0u1s:
There's a mismatch between the fuzzer and runtime IJON offset calculations, which is causing memory access issues and breaking IJON functionality. I'm working on a fix.

vanhauser-thc:
oh yeah you are right I had no time to properly test it. I just pushed a feature that disables ijon for cmplog and SAN forkservers.

instrumentation/SanitizerCoveragePCGUARD.so.cc
```diff
@@ -1643,7 +1698,21 @@ void ModuleSanitizerCoverageAFL::InjectCoverageAtBlock(Function   &F,
...
+    
+    // Apply IJON state-aware coverage if enabled
+    if (ijon_enabled && AFLIJONState) {
+      LoadInst *IJONStateVal = IRB.CreateLoad(Int32Ty, AFLIJONState);
+      ModuleSanitizerCoverageAFL::SetNoSanitizeMetadata(IJONStateVal);
+      // Apply IJON formula: state XOR coverage_index
+      // This makes every coverage point context-aware based on current IJON state
+      Value *XorResult = IRB.CreateXor(IJONStateVal, CoverageIndex);
+      // Ensure result stays within map bounds to prevent buffer overruns
+      Value *MapSize = ConstantInt::get(Int32Ty, MAP_SIZE);
```
vanhauser-thc: This is wrong - the map is not 64kb but the total number of edges in the target. what you could do is reading the value of the global variable `__afl_cov_map_size` instead.

instrumentation/SanitizerCoverageLTO.so.cc
```diff
@@ -2256,6 +2302,18 @@ void ModuleSanitizerCoverageLTO::InjectCoverageAtBlock(Function   &F,
...
+    
+    // Apply IJON state-aware coverage if enabled
+    if (ijon_enabled && AFLIJONState) {
+      LoadInst *IJONStateVal = IRB.CreateLoad(Int32Tyi, AFLIJONState);
+      ModuleSanitizerCoverageLTO::SetNoSanitizeMetadata(IJONStateVal);
+      // Apply IJON formula: state XOR coverage_index
+      // This makes every coverage point context-aware based on current IJON state
+      Value *XorResult = IRB.CreateXor(IJONStateVal, val);
+      // Ensure result stays within map bounds to prevent buffer overruns
+      Value *MapSize = ConstantInt::get(Int32Tyi, MAP_SIZE);
```
vanhauser-thc: same, load __afl_cov_map_size instead

vi3tL0u1s:
> There's a mismatch between the fuzzer and runtime IJON offset calculations, which is causing memory access issues and breaking IJON functionality. I'm working on a fix.

This is still WIP!

vanhauser-thc:
I fixed the ijon map disaster i created :-)

this big seems to be yours though:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000055555555b7fc in ijon_update_max_dynamic (self=0x5555555ddd50, data=0x5555555edf10 'A' <repeats 26 times>, "\n", len=0x1b, shared=<optimized out>) at src/afl-fuzz-ijon.c:299
299	    if (shared->ijon_max_area[i] > 0) {
```

vi3tL0u1s:
> I fixed the ijon map disaster i created :-)
> 
> this big seems to be yours though:
> 
> ```
> Program terminated with signal SIGSEGV, Segmentation fault.
> #0  0x000055555555b7fc in ijon_update_max_dynamic (self=0x5555555ddd50, data=0x5555555edf10 'A' <repeats 26 times>, "\n", len=0x1b, shared=<optimized out>) at src/afl-fuzz-ijon.c:299
> 299	    if (shared->ijon_max_area[i] > 0) {
> ```

Oh, I already fixed the memory mismatch issue in commit a67de2d and verified it works correctly with both Mario and maze. Apologies for not leaving a comment earlier. Could you take a look at the commit logic and let me know if we should adopt it directly or adapt it to your recent changes?

The Core Problem:
- Target-side calculation: Used __afl_final_loc + MAP_SIZE_IJON_MAP (actual coverage size + offset)
- Fuzzer-side calculation: Used map_size - MAP_SIZE_IJON_BYTES (total map size – IJON area size)

The Fix Applied:
- Target-Side Changes (afl-compiler-rt.o.c):
```c
// FIXED: Use actual coverage size instead of hardcoded MAP_SIZE
// For ijon_max:
u32 ijon_offset = __afl_final_loc + MAP_SIZE_IJON_MAP;

// For ijon_set/ijon_inc:
__afl_area_ptr[__afl_final_loc + coverage_id] = val;  // Instead of __afl_cov_map_size
```

- Fuzzer-Side Changes (afl-fuzz-ijon.c, afl-fuzz-run.c):
```c
// FIXED: Use real_map_size (from target) instead of aligned map_size
setup_dynamic_shared_access(fsrv->trace_bits, fsrv->real_map_size);

// FIXED: Calculate consistent offset
u32 coverage_size = map_size - 1 - MAP_SIZE_IJON_MAP - MAP_SIZE_IJON_BYTES;
u32 ijon_offset = coverage_size + MAP_SIZE_IJON_MAP;
```

Memory Layout After Fix:
- [Coverage Data: 0 … __afl_final_loc-1]
- [IJON SET/INC: __afl_final_loc … __afl_final_loc+65535]
- [IJON MAX/MIN: __afl_final_loc+65536 … __afl_final_loc+69631]

Key Technical Details

- __afl_final_loc: Actual coverage size from the target
- real_map_size: Actual map size reported by the target via forkserver
- map_size: Fuzzer's aligned version (may exceed real_map_size)
- Consistent calculation: Both sides now use the same approach

Outcome: This fix ensures IJON data is written and read from consistent memory locations, restoring correct guided fuzzing behavior across all IJON variants (ijon_max, ijon_set, ijon_inc, etc).

vanhauser-thc:
this crash is with the most current commit

vi3tL0u1s:
> this crash is with the most current commit

Thanks for confirming! Just to clarify, the fix I mentioned was in the commit before your most recent one! I had already addressed the mismatch there and tested it with Mario and maze. I'll review again against the latest commit to see where the difference comes in.

vanhauser-thc:
1. the crash only occurs when ijon is used together with SAN (-w). otherwise it works fine.

2. there is a display bug (and not sure what else might be affected) - visible in the UI. The count coverage is wrong: `17602192.34 bits/tuple`. 

3. the map density value in the UI (bitmap_cvg in fuzzer_stats file) is calculating the coverage including the extra 64kb from ijon.

=> 1. needs to be fixed
=> 2. needs to be fixed
=> 3. I think we have to live with this

vi3tL0u1s:
Thanks so much for the detailed breakdown, that really helps! I hadn't considered the SAN interaction before. I'll start working on fixes for those!

vanhauser-thc:
fixed the 1. SAN -w crash, was just a one-liner

vanhauser-thc:
fixed 2. the bits values as well.

What else is there to do?

IMHO ijon stats could be added to the trim/eff line ... `trim/eff/ijon` ... but this will need to gather some useful statistics
image

vi3tL0u1s:
> fixed 2. the bits values as well.
> 
> What else is there to do?
> 
> IMHO ijon stats could be added to the trim/eff line ... `trim/eff/ijon` ... but this will need to gather some useful statistics image

Thank you for the incredibly quick fixes!

I think adding IJON stats to the display is a great idea. I am thinking we could show IJON_SET/INC as locations_hit/total_locations.

vi3tL0u1s:
ijon_max doesn't seem to be working well with Mario at the moment - I'll take another look and try to fix it.

vi3tL0u1s:
With the latest commit, I fixed the IJON offset calculation issue. The root cause was that the fuzzer-side offset calculation didn't account for the updated map_size behavior, which now excludes MAP_SIZE_IJON_BYTES.

vi3tL0u1s:
Sorry, I realized today there was a 4k misalignment in the IJON offsets, and I've pushed a fix for it. I also added IJON stats to the UI, showing the number of new max values found vs. the total number of max entries so far (max value is 512 ijon_max entries). For IJON_SET and IJON_INC, I don't think we really need to display them in the UI since they’re more like coverage stats rather than progress indicators.

vi3tL0u1s:
This is the current UI.
image

vanhauser-thc:
in case a session is resumed (`-i -` -o out) does this work?
also if a session is restarted but overwritten (`-i input -o out`) does this work too (so is the ijon folder removed)?

the stats: we try to show what new queue entries have been found by the technique. does this somehow apply to ijon? 
it should be :   `numbery_of_new_queue_entries_found / total_executions_in_ijon_function`.

vanhauser-thc:
And are you still working on this or can this be removed?
```
src/afl-fuzz-ijon.c:296:7: warning: variable 'non_zero_slots' set but not used [
-Wunused-but-set-variable]
  296 |   int non_zero_slots = 0;
      |       ^
1 warning generated.

```

vi3tL0u1s:
> And are you still working on this or can this be removed?
> 
> ```
> src/afl-fuzz-ijon.c:296:7: warning: variable 'non_zero_slots' set but not used [
> -Wunused-but-set-variable]
>   296 |   int non_zero_slots = 0;
>       |       ^
> 1 warning generated.
> ```

This can be removed. I'm working on fixing IJON fastresume and will clean it up in the next commit.

vanhauser-thc:
cool, then we are close to merging :-)
maybe we can just copy over the IJON docs into docs/IJON.md with little modifications?

vanhauser-thc:
notice that there are two resumes - the fast resume that just loads a dump from disk if the binary has not changed, and one that re-reads and re-calibrates all data in case the binary has changed.

vi3tL0u1s:
> notice that there are two resumes - the fast resume that just loads a dump from disk if the binary has not changed, and one that re-reads and re-calibrates all data in case the binary has changed.

After adding the fastresume logic, the second resume works fine with my simple test cases. I'll continue with the remaining tasks tomorrow.

vi3tL0u1s:
In recent commits, I added an IJON max stat. Example: 3.95% (92/2332) = successful IJON max updates vs. total IJON max executions. This focuses only on IJON max, since it reflects real progress (finding new maximums). Other operations (SET/INC/STATE) are coverage-oriented, where it's hard to track which inputs in the queue actually triggered them.
image

vanhauser-thc:
> In recent commits, I added an IJON max stat. Example: 3.95% (92/2332) = successful IJON max updates vs. total IJON max executions. This focuses only on IJON max, since it reflects real progress (finding new maximums). Other operations (SET/INC/STATE) are coverage-oriented, where it's hard to track which inputs in the queue actually triggered them. image

this is definitely better.
but a user would expect this to have created 92 queue entries though. and only showing one value out of several ijon features is also misleading or what do you think?
If there is no value that can be displayed that helps the user to see how successful (or not) ijon is, then we should not add this. If there are value that are helpful for the user they could also be displayed differently, eg as a percentage value.

It would be great to have a value that shows the user how much ijon helps to the fuzzing. dunno what that could be or look like.

vi3tL0u1s:
I think it's hard to show this directly. To make it work, we'd need two steps:
1. Extend IJON state tracking to monitor:
- New bits set in the IJON bitmap (SET ops)
- New increments from zero (INC ops)
- New state transitions (STATE ops)
- New maximum values (MAX ops)
2. Add queue correlation tracking to mark which inputs triggered IJON discoveries and count how many queue entries came from IJON guidance.

Step 1 is feasible, but step 2 would be quite hard and could add significant overhead. Without step 2, though, it won't really show any meaningful progress. Maybe this could be something for a future TODO? If that's the case, maybe we don't need to show IJON max progress either.

vanhauser-thc:
it is fine to leave that as a future todo.
It could be overhead yes, but as ijon itself is a huge overhead the overall impact should not be that high :-)
thanks!

vanhauser-thc:
with the help of AI I updated the documentation :-)
is everything done, ready and tested?

vi3tL0u1s:
> with the help of AI I updated the documentation :-) is everything done, ready and tested?

Yes, everything LGTM. If it's fine to merge, I can remove the IJON-TODO in the next commit.

vanhauser-thc:
You might want to update the statistics in the IJON.md at some point, due to the changes to the map etc. I expect the results to be slower than before. still faster than original ijon though.
merging now.
thanks for your effort!!

vi3tL0u1s:
@vanhauser-thc Thank you so much for your guidance throughout this pull request! I've learned a lot! I'll update the stats in the IJON.md in my next commit.

vanhauser-thc:
FYI I found various bugs.
One issue I have not fixed yet is persistent mode support.
Also with fastresume there are issues so I have temporarily disabled it

vi3tL0u1s:
Hi @vanhauser-thc, thanks for the update! Could you share more details on the issues with fastresume? Having the specific test cases would also be really helpful. So far, I've only tested on Mario and maze, which may not be general enough. I'll set up a separate branch in the same fork to work on the fixes.

vi3tL0u1s:
I've noticed your recent commits! I'll look into them accordingly!

vanhauser-thc:
The biggest issue is IJON_STATE. It results in afl++ disabling all coverage.
In each step in a test case calibration the xor value is switched resulting in variable bytes, which are then ignored by afl-fuzz.
When calibrating the coverage bytes need to be stable. Also the same test case executed later should have the same coverage.

I think the solution is to reset the state xor variable before every run

vanhauser-thc:
so I fixed ijon_state and verified that everything is working. plus some changes to make ijon work also for the rust afl.rs crate for which I am submitting a pr.

next step is re-enabling the fast resume stuff.
besides bugs in afl-comopiler-rt and afl-fuzz and afl-fuzz-ijon I saw that you seem to adjusted the already adjusted map in the fast resume and that made things fail.

vanhauser-thc:
FYI I had to change the ijon feature detection in the plugins because it didn’t work for macOS

vi3tL0u1s:
Hi @vanhauser-thc, thanks for the update!
I don't have access to a macOS system, so I can't properly test the changes you added. On the Linux side though, I tested ijon with Mario and can confirm it's working fine with no regressions.

PR #2542, opened 2025-09-22
Update docs/IJON.md

vi3tL0u1s:
Update docs/IJON.md as requested to complete missing data in the performance tables.


vanhauser-thc:
noticeably slower but still much better. plus now less interfering with non-colliding coverage.
thanks!

PR #2543, opened 2025-09-22
Add AFL_COMPILER_LAUNCHER to wrap compiler invocations with custom binary

nbars:


src/afl-cc.c
```diff
@@ -1517,10 +1520,24 @@ void add_real_argv0(aflcc_state_t *aflcc) {
...
     }
 
-    aflcc->cc_params[0] = alt_cc;
+    compiler_path = alt_cc;
 
   }
 
+  if (getenv("AFL_USE_CCACHE")) {
+    // Shift all existing parameters to make room for ccache
+    for (u32 i = aflcc->cc_par_cnt; i > 0; i--) {
```
domenukk, line 1536: this needs to check `MAX_PARAMS_NUM`

src/afl-cc.c
```diff
@@ -1517,10 +1520,24 @@ void add_real_argv0(aflcc_state_t *aflcc) {
...
+      aflcc->cc_params[i] = aflcc->cc_params[i - 1];
+    }
+
+    // Insert ccache as the first parameter
+    aflcc->cc_params[0] = "ccache";
```
domenukk: Isn't that a bit ... overly specific?

domenukk: Maybe AFL_CC_BIN_PREFIX=ccache or something? Or are there less intrusive ways?
In any case, it should be documented in the envs md file

vanhauser-thc:
LGTM, thanks

PR #2546, opened 2025-09-25
Fix IJON fastresume functionality

vi3tL0u1s:
I am also working on removing unnecessary logics!


src/afl-fuzz.c
```diff
@@ -2641,8 +2641,11 @@ int main(int argc, char **argv_orig, char **envp) {
 
     }
 
-    afl->fsrv.map_size -= MAP_SIZE_IJON_BYTES;
-    afl->fsrv.real_map_size -= MAP_SIZE_IJON_BYTES;
+    // Only adjust map_size if not doing fastresume (fastresume will handle map_size correctly)
+    if (!fast_resume) {
```
vanhauser-thc, line 2647: why not for fast-resume?

src/afl-fuzz.c
```diff
@@ -2654,10 +2657,11 @@ int main(int argc, char **argv_orig, char **envp) {
     afl->ijon_state = new_ijon_min_state(max_dir);
     ck_free(max_dir);
 
-    setenv("AFL_NO_IJON", "1", 1);
+    // setenv("AFL_NO_IJON", "1", 1);
```
vanhauser-thc: no this is essential. we do not need ijon extra map for cmplog and asan

src/afl-fuzz.c
```diff
@@ -2864,52 +2868,50 @@ int main(int argc, char **argv_orig, char **envp) {
...
-              // hasn't changed)
-              save_ijon_state_for_fastresume(
-                  saved_ijon_state.ijon_offset, saved_ijon_state.map_size,
-                  saved_ijon_state.real_map_size,
-       saved_ijon_state.target_map_size);
+      if (saved_ijon_state.is_initialized) {
 
-            }
+        // Enable IJON for fastresume (will be confirmed by forkserver handshake later)
+        afl->fsrv.use_ijon = 1;
```
vanhauser-thc: that is after the forkserver handshake - not before

src/afl-fuzz.c
vi3tL0u1s: Sorry, I missed this comment yesterday.

During a fresh session, we subtract the IJON bytes from the map size to reserve space for IJON. In fast-resume, however, the saved state already includes the adjusted map sizes from the original run. If we apply the subtraction again, we'd end up double-adjusting, which shrinks the map incorrectly. This check makes sure the adjustment only happens once, keeping the save/load behavior consistent.

vanhauser-thc:
is this done and you have verified that it is working?
Or is there still something missing or needs testing?

src/afl-fuzz-run.c
```diff
@@ -156,18 +156,22 @@ fsrv_run_result_t __attribute__((hot)) fuzz_run_target(afl_state_t      *afl,
...
-
-            }
-
-      */
-      dynamic_shared_access_t *shared_access = setup_dynamic_shared_access(
-          fsrv->trace_bits, effective_map_size, effective_real_map_size);
+        // Use saved IJON offset directly during fastresume for consistent offset calculation  
+        dynamic_shared_access_t *shared_access;
+        if (has_saved_ijon_state()) {
+          ijon_fastresume_state_t *saved_state = get_saved_ijon_state();
```
vanhauser-thc: Why this section here? this is executed after every single fuzz.

vanhauser-thc:
note that I have pushed code formatting

vanhauser-thc:
please remove not-needed commented out code.

is this done and you have verified that it is working?
Or is there still something missing or needs testing?

vi3tL0u1s:
I've cleaned up the architecture: the IJON shared_access setup was moved out of the code path in run_target() into proper one-time initialization, added ijon_shared_access to afl_state_t for persistent storage, and removed leftover debug code. This avoids repeated setup on every fuzz execution while keeping everything fully functional. Tests confirmed it works correctly with Mario+CmpLog+IJON and Maze+ASAN+IJON across multiple fastresume sessions.

vanhauser-thc:
OK I will merge this and then push to stable so hopefully some people test it

PR #2554, opened 2025-10-04
include/forkserver.h: add missing include

maribu:
Without `#include <fcntl.h>`, `mode_t` is undeclared (at least on musl based systems).


vanhauser-thc:
Thanks

PR #2555, opened 2025-10-04
instrumentation: only assume execinfo.h provided on glibc linux systems

maribu:
The current preprocessor check assumes that on every Linux based system the execinfo.h header is present. But in fact this will only work on glibc based systems.

This changes the condition to check for both Linux and glibc.


maribu:
Note: I did not try this on glibc based systems. Before merging, it is probably best to check that the new check indeed unlocks backtracing support on glibc systems.

vanhauser-thc:
Thanks. Please direct PRs to the dev branch in the future.

maribu:
Ah, sorry, I just trusted the defaults :/

I'll take care in the future

PR #2564, opened 2025-10-15
Update afl-fuzz-ijon.c. Use cur_state file to save the maximum value

nj00001:
When using the ijon-max method, a file cur_state containing historical maximum values ​​is also saved in max_dir, which helps humans understand the fuzz progress by viewing the file.

### DIFF
--- a/docs/IJON.md
+++ b/docs/IJON.md
@@ -142,6 +142,21 @@ make clean
 make LLVM_CONFIG=llvm-config-18 source-only
 ```
 
+### IJON mode debug build for afl-fuzz
+
+Note that this does not affect any ijon functionality, it just creates an extra file.
+
+```bash
+make clean
+CFLAGS="-DDUMP_IJON_STATE" make afl-fuzz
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
this should only be written if `afl->debug` is set, right?

nj00001:
> this should only be written if `afl->debug` is set, right?

When the AFL_DEBUG=1 environment variable is set, there may be a lot of messy information in the terminal, and qemu will also print ijon related logs to the terminal, but directly adding another environment variable to control the generation of this file may be too much. If you consider optimizing performance, there is indeed room for improvement. By monitoring the changes in self->num_updates, you can update this file only when there is a new overwrite. Do you have any ideas?

vanhauser-thc:
This should not be a default. Please add a define in the c file that you can enable if you need it

nj00001:
How about this change?

vanhauser-thc:
LGTM, thanks

PR #2565, opened 2025-10-16
Fix handling of skipped custom mutations

renatahodovan:
According to the documentation, when `afl_custom_fuzz` returns 0, the corresponding mutation result should be skipped. The implementation correctly runs the mutant only if the return value is greater than 0. However, one step earlier it also checks whether `mutation_buffer` is null, and if so, aborts fuzzing of the entire current test case instead of skipping only the current mutant.

Since `mutation_buffer` can legitimately be null when the custom mutator signals that the mutation should be discarded, the buffer check should only be performed when the mutant is intended to be kept. The commit fixes this.


vanhauser-thc:
thanks

PR #2570, opened 2025-10-29
push to stable

vanhauser-thc:


src/afl-forkserver.c
```diff
@@ -555,6 +559,41 @@ static void afl_fauxsrv_execv(afl_forkserver_t *fsrv, char **argv) {
...
+
+        if (python_pid < 0) { PFATAL("GUI mode fork failed."); }
+
+        if (!python_pid) {  // New python interactor
+          ACTF("Non-forkserver exec'ing, with PID = %ld\n", (long)parent_pid);
+          char parent_pid_str[16];
+          sprintf(parent_pid_str, "%d",
+                  (int)parent_pid);  // Convert pid_t to a string
+
+          char *pargs[] = {"/usr/bin/python3",
```
domenukk: This will certainly fail on some platforms. Really, this whole thing should not be a part of the forkserver IMHO, but should be passed in through env variables or otherwise be handled in a script that gets called by forkserver..

src/afl-forkserver.c
```diff
@@ -2078,6 +2130,44 @@ fsrv_run_result_t __attribute__((hot)) afl_fsrv_run_target(
...
+
+      if (fsrv->gui_python_pid < 0) { PFATAL("GUI mode fork failed."); }
+
+      if (!fsrv->gui_python_pid) {
+
+        char child_pid_str[16];
+        sprintf(child_pid_str, "%d",
+                (int)fsrv->child_pid);  // Convert pid_t to a string
+
+        char *pargs[] = {"/usr/bin/python3",
```
domenukk: Same here. (And why does this need to be copy&paste at all?)

src/afl-forkserver.c
vanhauser-thc: @Trowlett0 - I had overlooked this

src/afl-forkserver.c
vanhauser-thc: @Trowlett0 dito

src/afl-forkserver.c
domenukk: @Trowlett0  I thought about this some more; we already link in python for the python mutators, so can't we just use that one here?

vanhauser-thc: It might be there. But it is possible that python is present but not compiled in - eg if python-dev was not installed

domenukk: Sure, but also, people who want to use python can just compile it in? It's a bit of a corner case I think

Trowlett0: Submitted fixes in PR #2573. I went with a separate script that’s called by the forkserver, mainly because I wasn’t sure how to correctly reference or reuse the python that’s linked in for the python mutators. I'm happy to update the PR to use the existing linked python, I just need a pointer on how I would access it properly.

PR #2572, opened 2025-11-04
Integrate FrameShift

hgarrereyn:
This PR adds support for FrameShift (https://arxiv.org/pdf/2507.05421)

It can be enabled at runtime by setting the env var `AFL_FRAMESHIFT_ENABLED=1`.

## Overview

FrameShift analysis runs once per newly discovered corpus entry in order to try to identify relation fields (i.e. size and offset fields) in the input format. If any are discovered, subsequent resizing mutations (inserts and deletes) will be tracked during the splicing and havoc phase in order to be able to fixup the corresponding relation fields.

The max analysis time per new input is controlled by the compile time definition `FRAMESHIFT_TIME_BUDGET_MS` (by default 2 seconds).

## Evaluation

We evaluated AFL++ with and without FrameShift (on AFL++ v2.1c) on 16 targets and found that enabling FrameShift could confer statistically significant coverage increases of up to 60% (10 x 48 hour runs, p < 0.01) on a variety of targets, such as `lcms`, `bloaty`, `ms-tpm-20-ref`, `woff2`, `libpcap`, and `openssl` (refer to our paper for a full table). We also found that enabling FrameShift decreased time-to-trigger bugs on the Magma bug benchmark by about an hour on average during a 24 hour run (unpublished). Our paper eval ran AFL++ in FuzzBench configuration (with cmplog and dict2file).

On text-based formats, we don't expect FrameShift to find anything useful and the extra analysis time will likely slow down fuzzing. On the `libxml2` target for example, enabling FrameShift decreased coverage by about 12% when starting from a seeded corpus.

## Practical Usage Suggestions

Enabling FrameShift is likely the most useful when your target input format expects binary-serialized data that contains size or offset fields, especially when they are nested (i.e. most binary formats).

Since FrameShift analysis runs once-per corpus entry, analysis time will be largest either when starting with a large seed corpus or for targets with lots of coverage that grow quickly (i.e. `harfbuzz`/`freetype2`). Keep in mind that the initial startup speed may be slower with FrameShift enabled since it will spend time running analysis instead of directly fuzzing. We expect that in very long fuzz runs (i.e. a week or longer), the analysis time will amortize out as the coverage growth plateaus.

## Code Changes

The core algorithm and implementation is in [src/afl-fuzz-frameshift.c](src/afl-fuzz-frameshift.c).

The main changes are:

- A FrameShift mode which is enabled with `AFL_FRAMESHIFT_ENABLED=1`
- The `queue_entry` structure is extended with a new `fs_meta_t *fs_meta` field pointing to the learned relation(s) for the input, and a `u8 fs_status` indicator.
- When a new input is processed for the first time inside `fuzz_one_original`, it will run the `frameshift_stage` on the input to try to learn relations.
-  `havoc` and `splice` mutations are tracked and before running the input, the new relation values (if any) are re-serialized into the input.

If you are running with `AFL_NO_UI=1`, the logs will contain statistics about frameshift at the end of each line:
```
[*] Fuzzing test case #122 <snip> FS (t=0:00:00:15, st=63950, avg=504 ms, found=20/30)...
```
Here:
- `t`: total analysis runtime of the FrameShift stage
- `st` (search tests) number of times FrameShift ran a test input dynamically
- `avg` average analysis time per input (upper bounded by `FRAMESHIFT_TIME_BUDGET_MS`)
- `found` number of inputs FrameShift found at least one relation in over number of tested inputs

If the `found` ratio is very low, this likely means you are either starting from very unstructured data (i.e. an empty corpus) or you are fuzzing a format without size/offset fields.


src/afl-fuzz-one.c
```diff
@@ -433,20 +433,51 @@ u8 fuzz_one_original(afl_state_t *afl) {
...
-        afl->saved_crashes != 0 ? cRED : "", afl->saved_crashes, cRST,
-        get_fuzzing_state(afl), afl->fuzz_mode ? "exploit" : "explore",
-        afl->queue_cur->perf_score, afl->queue_cur->weight,
-        afl->queue_cur->favored, afl->queue_cur->was_fuzzed,
-        afl->queue_cur->exec_us,
-        likely(afl->n_fuzz) ? afl->n_fuzz[afl->queue_cur->n_fuzz_entry] : 0,
-        afl->queue_cur->bitmap_size, afl->queue_cur->is_ascii, time_tmp);
-    fflush(stdout);
+
+    if (afl->afl_env.afl_frameshift_enabled) {
```
vanhauser-thc: please change to:
`if (unlikely(afl->afl_env.afl_frameshift_enabled)) {`

src/afl-fuzz-one.c
```diff
@@ -530,6 +561,20 @@ u8 fuzz_one_original(afl_state_t *afl) {
...
+
+  if (unlikely(afl->afl_env.afl_frameshift_enabled &&
+               afl->queue_cur->fs_status == 0)) {
+
+    frameshift_stage(afl);
+
+  }
+
+  // Frameshift: reload the original input meta
+  if (afl->afl_env.afl_frameshift_enabled) { fs_clone_meta(afl); }
```
vanhauser-thc: can you please change this to:

```
if (unlikely(afl->afl_env.afl_frameshift_enabled)) {

    if (unlikely(afl->queue_cur->fs_status == 0)) { frameshift_stage(afl); }
    fs_clone_meta(afl);

}
```

src/afl-fuzz-one.c
```diff
@@ -2191,6 +2236,9 @@ u8 fuzz_one_original(afl_state_t *afl) {
              afl->queue_cur->fname, afl->queue_cur->is_ascii, use_stacking);
 #endif
 
+    // Frameshift: save the current input meta
+    if (afl->afl_env.afl_frameshift_enabled) { fs_save(afl->fs_curr_meta); }
```
vanhauser-thc: `unlikely` please

src/afl-fuzz-one.c
```diff
@@ -2574,6 +2622,13 @@ u8 fuzz_one_original(afl_state_t *afl) {
             afl_swap_bufs(AFL_BUF_PARAM(out), AFL_BUF_PARAM(out_scratch));
             temp_len += clone_len;
 
+            // Frameshift tracking
+            if (afl->afl_env.afl_frameshift_enabled) {
```
vanhauser-thc: dito

src/afl-fuzz-one.c
```diff
@@ -2625,6 +2680,13 @@ u8 fuzz_one_original(afl_state_t *afl) {
...
```
vanhauser-thc: dito

src/afl-fuzz-one.c
```diff
@@ -2802,6 +2864,13 @@ u8 fuzz_one_original(afl_state_t *afl) {
 
           temp_len -= del_len;
 
+          // Frameshift tracking
+          if (afl->afl_env.afl_frameshift_enabled) {
```
vanhauser-thc: dito

src/afl-fuzz-one.c
```diff
@@ -2859,6 +2928,13 @@ u8 fuzz_one_original(afl_state_t *afl) {
...
```
vanhauser-thc: dito

src/afl-fuzz-one.c
```diff
@@ -2898,6 +2974,13 @@ u8 fuzz_one_original(afl_state_t *afl) {
           afl_swap_bufs(AFL_BUF_PARAM(out), AFL_BUF_PARAM(out_scratch));
           temp_len += clone_len;
 
+          // Frameshift tracking
+          if (afl->afl_env.afl_frameshift_enabled) {
```
vanhauser-thc: dito

src/afl-fuzz-one.c
```diff
@@ -3039,6 +3122,13 @@ u8 fuzz_one_original(afl_state_t *afl) {
             afl_swap_bufs(AFL_BUF_PARAM(out), AFL_BUF_PARAM(out_scratch));
             temp_len += (new_len - old_len);
 
+            // Frameshift tracking
+            if (afl->afl_env.afl_frameshift_enabled) {
```
vanhauser-thc: dito

src/afl-fuzz-one.c
```diff
@@ -3132,6 +3222,13 @@ u8 fuzz_one_original(afl_state_t *afl) {
           memcpy(out_buf + insert_at, ptr, extra_len);
           temp_len += extra_len;
 
+          // Frameshift tracking
+          if (afl->afl_env.afl_frameshift_enabled) {
```
vanhauser-thc: dito

src/afl-fuzz-one.c
```diff
@@ -3190,6 +3287,13 @@ u8 fuzz_one_original(afl_state_t *afl) {
...
```
vanhauser-thc: dito

src/afl-fuzz-one.c
```diff
@@ -3303,6 +3407,13 @@ u8 fuzz_one_original(afl_state_t *afl) {
           afl_swap_bufs(AFL_BUF_PARAM(out), AFL_BUF_PARAM(out_scratch));
           temp_len += clone_len;
 
+          // Frameshift tracking
+          if (afl->afl_env.afl_frameshift_enabled) {
```
vanhauser-thc: dito

src/afl-fuzz-one.c
```diff
@@ -3323,6 +3434,9 @@ u8 fuzz_one_original(afl_state_t *afl) {
     temp_len = len;
     memcpy(out_buf, in_buf, len);
 
+    // Frameshift: restore the original input meta
+    if (afl->afl_env.afl_frameshift_enabled) { fs_restore(afl->fs_curr_meta); }
```
vanhauser-thc: dito

src/afl-fuzz-one.c
```diff
@@ -3382,6 +3496,9 @@ u8 fuzz_one_original(afl_state_t *afl) {
     u8                 *new_buf;
     s32                 f_diff, l_diff;
 
+    // Frameshift: reload the original input meta
+    if (afl->afl_env.afl_frameshift_enabled) { fs_clone_meta(afl); }
```
vanhauser-thc: dito

src/afl-fuzz-one.c
```diff
@@ -3432,6 +3549,15 @@ u8 fuzz_one_original(afl_state_t *afl) {
     if (unlikely(!out_buf)) { PFATAL("alloc"); }
     memcpy(out_buf, in_buf, len);
 
+    // Frameshift tracking
+    if (afl->afl_env.afl_frameshift_enabled) {
```
vanhauser-thc: dito

src/afl-fuzz-queue.c
```diff
@@ -762,6 +762,9 @@ void add_to_queue(afl_state_t *afl, u8 *fname, u32 len, u8 passed_det) {
 
   q->skipdet_e = (struct skipdet_entry *)ck_alloc(sizeof(struct skipdet_entry));
 
+  q->fs_meta = NULL;
+  q->fs_status = 0;
+
```
vanhauser-thc: it is freshly allocated so all 0/null, so you can remove this. (and while doing this please remove the already present
```
  q->trace_mini = NULL;
  q->testcase_buf = NULL;
  ```
  as well :-) thanks

File: src/afl-fuzz-run.c [hunk 17]
```diff
@@ -1463,6 +1463,13 @@ u8 __attribute__((hot)) common_fuzz_stuff(afl_state_t *afl, u8 *out_buf,
 
   u8 fault;
 
+  if (afl->afl_env.afl_frameshift_enabled && afl->fs_curr_meta) {
```
vanhauser-thc: dito

src/afl-fuzz-state.c
```diff
@@ -152,6 +152,13 @@ void afl_state_init(afl_state_t *afl, uint32_t map_size) {
   afl->havoc_prof =
       (struct havoc_profile *)ck_alloc(sizeof(struct havoc_profile));
 
+  afl->frameshift_index_buffer = NULL;
+  afl->fs_curr_meta = NULL;
+  afl->fs_stats.found = 0;
+  afl->fs_stats.searched = 0;
+  afl->fs_stats.search_tests = 0;
+  afl->fs_stats.total_time_ms = 0;
+
```
vanhauser-thc, line 160: please remove because freshly allocated

vanhauser-thc:
I will need to do a fuzzbench run on this (dunno where, asking around currently).
what would be the most beneficial setup for frameshift? empty seeds or saturated corpus? with or without dictionary?

include/envs.h
```diff
@@ -124,7 +124,7 @@ static char *afl_environment_variables[] = {
     "AFL_NO_FASTRESUME", "AFL_SAN_ABSTRACTION", "AFL_LLVM_ONLY_FSRV",
     "AFL_GCC_ONLY_FRSV", "AFL_SAN_RECOVER",
     "AFL_PRELOAD_DISCRIMINATE_FORKSERVER_PARENT", "AFL_FORKSRV_UID",
-    "AFL_FORKSRV_GID", "AFL_COMPILER_LAUNCHER", NULL};
+    "AFL_FORKSRV_GID", "AFL_COMPILER_LAUNCHER", "AFL_FRAMESHIFT_ENABLED", NULL};
```
domenukk: Why not just `AFL_FRAMESHIFT`?

domenukk: Also we should add info about the env into the docs I think

hgarrereyn:
Most beneficial setup is likely empty corpus, running in fuzzbench configuration (cmplog + dictionaries).

We saw coverage improvements both from empty corpus and from a single seed corpus. The main adversarial setup would be if there are e.g. thousands of seed files which would incur a big initial analysis overhead.

I have access to a 128-core server, would be happy to run a full FuzzBench eval if you'd like.

vanhauser-thc:
> Most beneficial setup is likely empty corpus, running in fuzzbench configuration (cmplog + dictionaries).
> 
> We saw coverage improvements both from empty corpus and from a single seed corpus. The main adversarial setup would be if there are e.g. thousands of seed files which would incur a big initial analysis overhead.
> 
> I have access to a 128-core server, would be happy to run a full FuzzBench eval if you'd like.

if you can, then please do!

this would be the setup:

[experiment.tar.gz](https://github.com/user-attachments/files/23364814/experiment.tar.gz)

vanhauser-thc:
> Most beneficial setup is likely empty corpus, running in fuzzbench configuration (cmplog + dictionaries).
> 
> We saw coverage improvements both from empty corpus and from a single seed corpus. The main adversarial setup would be if there are e.g. thousands of seed files which would incur a big initial analysis overhead.
> 
> I have access to a 128-core server, would be happy to run a full FuzzBench eval if you'd like.

were you able to run it?

hgarrereyn:
Not yet, the server is occupied with some other experiments til friday

vanhauser-thc:
@domenukk did a run for me, only 14 targets had at least 5 successful runs for both fuzzers:

|benchmark      |better |type|
|-|-|-|
|bloaty         |fs     |binary|
|curl           |orig++ |asc|
|freetype       |orig+  ||
|harfbuzz       |orig   ||
|lcms           |orig++ ||
|libpcap        |fs+    |binary|
|libxml2        |fs     |asc|
|mbedtls        |orig   |binary|
|openssl        |fs+    |binary|
|openthread     |orig+  ||
|sqlite3        |orig   |asc|
|stbi           |fs+    ||
|vorbis         |orig   |binary|
|woff2          |fs++   |binary|

one + means it is noticable better, two ++ mean is is much better.
setup: empty corpus, no dictionary

overall they are equal in stats roughly, however on an individual target it can make a huge difference.
it is not as easy as ascii vs binary it seems (see libxml2 or freetype).

hgarrereyn:
Nice! These results correlate almost exactly with what we saw in our eval, except for `lcms` where we saw the opposite effect, AFL++(FS) had more than 60% coverage benefit from an empty corpus over AFL++ (p < 0.01, 24 hours, 10 runs), I'd be curious to figure out what happened there.

We also saw negative effects for freetype and harfbuzz -- I'm pretty sure this is due to how large those targets are. In our eval, the corpus grew to tens of thousands of files really quickly (almost two orders of magnitude bigger than other targets), since FrameShift runs on every new corpus input, this means you can eat up most of the time just doing analysis. I would think that in a long enough fuzz campaign, FrameShift would catch up but we haven't actually tested that. It could be interesting in addition to a per-input analysis time limit to impose a sort of global frameshift analysis scaling limit -- like it can never exceed more than 10% of the total fuzz time so far. That limit might help it retain some performance in these settings on shorter runs.

For libxml we saw no statistically significant effect (either way) from an empty corpus, but we did see a negative effect from a seeded corpus. In the case where FrameShift doesn't find valid relation fields (as in xml), the analysis tests its doing basically just bias the mutations. I.e. its effectively doing a bunch of small integer manipulation on every small field size in the input. I noticed for xml, this has the effect of frequently messing with newline characters (i.e. testing if they are valid field sizes of value 10).

vanhauser-thc:
> Nice! These results correlate almost exactly with what we saw in our eval, except for `lcms` where we saw the opposite effect, AFL++(FS) had more than 60% coverage benefit from an empty corpus over AFL++ (p < 0.01, 24 hours, 10 runs), I'd be curious to figure out what happened there.
> 
> We also saw negative effects for freetype and harfbuzz -- I'm pretty sure this is due to how large those targets are. In our eval, the corpus grew to tens of thousands of files really quickly (almost two orders of magnitude bigger than other targets), since FrameShift runs on every new corpus input, this means you can eat up most of the time just doing analysis. I would think that in a long enough fuzz campaign, FrameShift would catch up but we haven't actually tested that. It could be interesting in addition to a per-input analysis time limit to impose a sort of global frameshift analysis scaling limit -- like it can never exceed more than 10% of the total fuzz time so far. That limit might help it retain some performance in these settings on shorter runs.
> 
> For libxml we saw no statistically significant effect (either way) from an empty corpus, but we did see a negative effect from a seeded corpus. In the case where FrameShift doesn't find valid relation fields (as in xml), the analysis tests its doing basically just bias the mutations. I.e. its effectively doing a bunch of small integer manipulation on every small field size in the input. I noticed for xml, this has the effect of frequently messing with newline characters (i.e. testing if they are valid field sizes of value 10).

Could you please write down a human readable rule set when frameshift should be beneficial and when not ?

hgarrereyn:
As a rule of thumb, targets which will likely benefit:
1. are binary formats with size/offset fields, especially nested ones
2. have corpora that don't grow too quickly (i.e. <= 10k after 24 hours)
3. don't contain sync markers (i.e. not _frameshift-resiliant_)

---

Examples:

**Best Case**:
These targets fit all of the above points. With FrameShift, you might get anywhere ~1-80% coverage increase depending on other factors.

E.g. `bloaty` (various object formats), `openssl` (DER-encoded ASN.1), `libpcap` (PCAP), `woff2` (WOFF), `libjpeg` (JPEG), `libpng` (PNG), `ms-tpm-20-ref` (TPM)

**Binary formats with fast-growing corpus**:
These targets would likely benefit from preserving relation fields, but the corpora grows so quickly that a lot of time is spent on analysis. For short runs, FrameShift is likely neutral/negative, in long runs (where analysis cost is amortized) FrameShift _may_ turn positive.

E.g. `freetype2`, `harfbuzz`

**Binary formats with sync markers**:
Formats that contain sync markers allow parsers to find data even when it is shifted in the input. This breaks FrameShift's analysis precisely because the target does not actually depend on size/offset fields. FrameShift is likely neutral or negative. See https://en.wikipedia.org/wiki/Ogg#Page_structure "capture pattern" for an example in OGG formats

E.g. `vorbis` (OGG), `openh264` (H.264)

**Non-binary formats**
FrameShift is _probably_ not useful for text-based formats, likely neutral or negative effect. Sometimes we see small positive effects, likely from coincidentally biasing the mutation space (during analysis) in a way that happens to be good for particular targets.

E.g. `libxml2` (XML)

vanhauser-thc:
> Binary formats with fast-growing corpus

that sounds rather easy to take care of - just do FS on max 30 per hour for example, or only on "favorites" in the queue etc

hgarrereyn:
Agree it should be possible to set up a threshold that helps on worst-case targets.

Do you happen to have the raw data / logs from the fuzzbench run? I think we could use that to pick a reasonable threshold

vanhauser-thc:
> Agree it should be possible to set up a threshold that helps on worst-case targets.
> 
> Do you happen to have the raw data / logs from the fuzzbench run? I think we could use that to pick a reasonable threshold

https://drive.google.com/file/d/15bYhzdRhcjIOTINUClMIZIF1lRgugS6d/view?usp=drive_web
please download asap :)

hgarrereyn:
@vanhauser-thc the link seems private, could you add me -- harrisonmichaelgreen@gmail.com

domenukk:
You should have access now

hgarrereyn:
Downloaded thanks!

hgarrereyn:
I compiled the data from that fuzzbench run and correlated it with the frameshift overhead (pulled from the logs) and corpus size. This table shows average line coverage for both tools, and number of successful runs (in parenthesis). Coverage ratio is `afl++fs / afl++` along with a Mann-Whitney U test p-value (there is low statistical power here due to the low number of successful runs). Overhead is the average overhead of frameshift across the run. Max corpus is the largest final corpus (from the frameshift runs).

_Note that `lcms - orig++` in the table posted earlier is likely a typo; the results here show frameshift gets roughly 10% more line coverage on average._

| Target | AFL++ | FS | Coverage Ratio | Overhead | Max Corpus |
|--------|--------|--------|--------|--------|--------|
libjpeg-turbo | 986 (3) | 2711 (8) | 2.75 (p=0.051) | 0.04 | 3704
woff2 | 1951 (7) | 2795 (6) | 1.43 (p=0.012) | 0.02 | 2498
proj4 | 14921 (8) | 17753 (4) | 1.19 (p=0.004) | 0.10 | 8268
stb | 2164 (6) | 2514 (5) | 1.16 (p=0.052) | 0.02 | 2439
lcms | 2135 (5) | 2339 (9) | 1.10 (p=0.688) | 0.01 | 796
openssl | 4926 (6) | 5123 (7) | 1.04 (p=0.764) | 0.02 | 1328
bloaty | 5488 (5) | 5634 (9) | 1.03 (p=0.606) | 0.03 | 2510
libxml2 | 13636 (6) | 13952 (8) | 1.02 (p=0.852) | 0.12 | 17649
libpcap | 5005 (5) | 5083 (7) | 1.02 (p=0.530) | 0.06 | 3235
libxslt | 11554 (6) | 11734 (4) | 1.02 (p=0.257) | 0.05 | 12080
zlib | 985 (8) | 986 (4) | 1.00 (p=0.657) | 0.01 | 553
openh264 | 14337 (2) | 14319 (6) | 1.00 (p=0.857) | 0.13 | 16325
systemd | 451 (4) | 450 (6) | 1.00 (p=0.205) | 0.02 | 862
re2 | 5466 (4) | 5430 (4) | 0.99 (p=0.309) | 0.02 | 3897
sqlite3 | 20829 (6) | 20675 (7) | 0.99 (p=0.945) | 0.09 | 7060
mbedtls | 5467 (6) | 5407 (6) | 0.99 (p=0.375) | 0.02 | 633
harfbuzz | 16556 (7) | 15994 (5) | 0.97 (p=0.639) | 0.20 | 16637
openthread | 11993 (6) | 11400 (8) | 0.95 (p=0.002) | 0.03 | 860
freetype2 | 25314 (7) | 23874 (6) | 0.94 (p=0.534) | 0.24 | 17510
curl | 22514 (8) | 20860 (7) | 0.93 (p=0.463) | 0.14 | 7591
vorbis | 2321 (6) | 2000 (5) | 0.86 (p=0.329) | 0.04 | 1505

I've also plotted the average overhead of frameshift over time across the different targets here:

overhead

# Takeaways

Generally it seems like for targets where FrameShift is effective, it can do so with minimal overhead (<= ~5%).

There are three targets which have a significant overhead and also perform worse than baseline: `freetype2`/`harfbuzz`/`curl`.

`vorbis` is an interesting case where it has low overhead but performs much worse. In the eval we ran for the paper (10 instances x 48 hours) we saw +1% coverage from empty corpus and -0.9% coverage (p<0.001) on the seeded corpus. I believe the large coverage loss here is fuzzer noise.

`proj4` is an interesting case where it has large overhead (10%) but does better (statistically significant). I think this is likely a case like libxml2 where the types of inputs tested during the analysis itself is what's driving coverage increase, not necessarily any learned structure. (i.e. this is probably a coincidental effect of mutation space biasing).

In any case, however, it looks like allocating a fixed upper limit of say 5% total runtime to FrameShift analysis might be able to preserve the benefit on good targets while reducing the negative impact on bad targets. I'll try to push an implementation of this later today.

hgarrereyn:
Ok I added a `AFL_FRAMESHIFT_MAX_OVERHEAD` flag that lets you control the maximum overhead (as a percentage of runtime) for frameshift, default is currently `0.05` (5%).

I'm currently running a mini eval with:
- baseline AFL++
- AFL++FS (5%)
- AFL++FS (10%)
- AFL++FS (100%)

On `libjpeg`/`woff2`/`harfbuzz`/`freetype2`.

Will report results in ~24 hours.

src/afl-fuzz-one.c
```diff
@@ -530,6 +569,33 @@ u8 fuzz_one_original(afl_state_t *afl) {
...
+      /* FrameShift has not run on this input. */
+
+      /* Check frameshift overhead budget before running analysis.
+         Use total runtime since start as the reference. */
+      u64 total_runtime_ms =
+          afl->prev_run_time + get_cur_time() - afl->start_time;
+      double max_overhead = afl->afl_env.afl_frameshift_max_overhead;
+      u64    allowed_ms = (u64)((double)total_runtime_ms * max_overhead);
+
+      if (afl->fs_stats.total_time_ms < allowed_ms) { frameshift_stage(afl); }
```
vanhauser-thc: you could also allow just favorite entries - or maybe both, time budget and favorite .

hgarrereyn: Could be interesting to try -- what's the criteria for determining what a "favorite" is in AFL++?

vanhauser-thc: to be the best queue entry for at least one edge ("best" is execution speed and size)

vanhauser-thc:
Any updates?

hgarrereyn:
Sorry, got sidetracked with other work.

In my small eval, setting the overhead to 10% got the best results empirically on the four benchmarks, although it was too noisy to get statistically significant results either way. I made that the default in the latest commit.

If you have the bandwidth for another experiment, you could try evaluating this limited version of FrameShift against baseline AFL++. Potentially with three modes:

- baseline AFL++ (no args)
- FS-10% (`AFL_FRAMESHIFT=1`)
- FS-full (`AFL_FRAMESHIFT=1` and `AFL_FRAMESHIFT_MAX_OVERHEAD=1`) -- this one should be functionally the same as no thresholding

vanhauser-thc:
cannot eval at the moment, also it would need to be done over all targets, not just 4

vanhauser-thc:
I can run fuzzbench experiment again :-)

So I have to set these at run time of afl-fuzz, right?
> * baseline AFL++ (no args)
> * FS-10% (`AFL_FRAMESHIFT=1`)
> * FS-full (`AFL_FRAMESHIFT=1` and `AFL_FRAMESHIFT_MAX_OVERHEAD=1`) -- this one should be functionally the same as no thresholding

can you please sync your repo to the current dev state?

vanhauser-thc:
and resolve the conflict :)

hgarrereyn:
@vanhauser-thc awesome! I just rebased/squashed the changes on top of the current `dev` branch, see https://github.com/AFLplusplus/AFLplusplus/commit/bf380a0d60f9937be7bd80bfe5e334350b5c9d33

FrameShift is an entirely runtime feature, there are no flags needed or changes to the way targets are compiled/instrumented.

At runtime, you need to set `AFL_FRAMESHIFT=1` to enable frameshift. By default, it will enforce a max analysis overhead of 10% (meaning that at maximum 10% of the total fuzz time can be allocated to running frameshift structure inference tests). This is configurable with the `AFL_FRAMESHIFT_MAX_OVERHEAD` runtime env argument which accepts floats in `(0, 1]`. i.e:
- `AFL_FRAMESHIFT_MAX_OVERHEAD=0.1` (default, 10% max overhead)
- `AFL_FRAMESHIFT_MAX_OVERHEAD=0.05` 5% max overhead
- `AFL_FRAMESHIFT_MAX_OVERHEAD=1` no limit on frameshift analysis

_Note that something has changed on the `dev` branch that broke the aflplusplus build when linking `afl-cmin` for me, I needed to add `CFLAGS=-pthread` to get this version to build._

vanhauser-thc:
it has started: https://storage.googleapis.com/fuzzbench-data/index.html?prefix=2026-01-21-aflpp/

vanhauser-thc:
sorry, can you please sync to the dev branch again? compiling worked fine for me and the CI, but not on fuzzbench ... :-(

hgarrereyn:
@vanhauser-thc merged! Compilation without extra cflags is working again for me. https://github.com/AFLplusplus/AFLplusplus/commit/bd90d7081483a0b610fd7f4637b37c8a7b0d2932

vanhauser-thc:
_fs = 25%
_fs2 = 10%

This is a normal fuzzbench run (many have a corpus):
```
aflplusplus2_fs2 1.000
aflplusplus2_fs 0.998
aflplusplus2 0.993
```
so with framshift active it is very slightly better

Then I did a fuzzbench run without a starting corpus:
```
aflplusplus2_fs 1.000
aflplusplus2_fs2 0.989
aflplusplus2 0.982
```
so apparently 10% is a good default value.
Maybe I will do another fuzzbench run with 20% and 30%

vanhauser-thc:
@hgarrereyn can you please once more sync to the current dev state?

20% was the best overall, but I have the feeling that an algorithm that goes from 50% down to 10% might be the best approach and I need to create better statistics. for this the new dev state helps me seeing if this it true.
thanks

hgarrereyn:
@vanhauser-thc synced!

vanhauser-thc:
So I am currently doing two fuzzbench runs:

one with an empty corpus and frameshift with 10, 20 and 30%
one with a saturated oss corpus and frameshift with 10 vs. std afl++

I hope this will give me the information for the optimal default setting for frameshift.
after that I will merge it and adapt the implementation to the optimal default config.

vanhauser-thc:
So I did the analysis twice to sure.

afl++ vs. frameshift 10% on a saturated corpus: frameshift_10 is basically the same (1.0 vs 0.999) - so enabled or not does not make a difference for a saturated corpus.
frameshift 10% vs 20% vs 30% on an empty corpus: 10% was best.

So analysis result: it is best if frameshift is enabled by default and set to 10%.
once the CI ran through I will merge. I will rename AFL_FRAMESHIFT to AFL_DISABLE_FRAMESHIFT and change the logic accordingly.

vanhauser-thc:
These are the modifications I did:

https://github.com/AFLplusplus/AFLplusplus/commit/3baa14aae4f9b84c2888623990ab01774951c8b7

and could you please do a small PR that adds documentation to `docs/env_variables.md` what frameshift does and the env to disable and the one to configure it?
thanks!

vanhauser-thc:
maybe even a small document excerpt, e.g. `docs/FRAMESHIFT.md`

vanhauser-thc:
ping @hgarrereyn  :)

hgarrereyn:
@vanhauser-thc done in https://github.com/AFLplusplus/AFLplusplus/pull/2701

Our paper was also recently accepted at ICSE 2026 (to be presented in Rio de Janeiro in April)! I'm planning to replace the arxiv link with the proper doi once the proceedings have been published.

PR #2573, opened 2025-11-05
GUIFuzz forkserver fix

Trowlett0:
Fixes related to PR #2570.

- Removed hardcoded `/usr/bin/python3` calls in the forkserver.
- Added a separate launcher script (`guifuzz_run.sh`) that resolves python3 with `command -v`.
- The launcher also checks for an X11 session, since PyAutoGUI is not compatible with Wayland.
- Cleaned up unnecessary and duplicated GUI handling code in the forkserver (fauxsrv).


domenukk:
Thank you! :)

vanhauser-thc:
thanks :)

PR #2596, opened 2025-11-19
Update build_unicorn_support.py to support 64-bit build

macenb:
Build broke for me on copy since it created a lib64 directory instead of lib, so I added a small exception handle for it to successfully build. Worked for my local build with this hot fix

### DIFF
--- a/unicorn_mode/build_unicorn_support.py
+++ b/unicorn_mode/build_unicorn_support.py
@@ -172,7 +172,10 @@ for ln in lns:
             if "linked_libs" in ln_json and any(["unicorn" in x for x in ln_json['linked_libs']]):
                 if "out_dir" in ln_json:
                     out_dir = Path(ln_json['out_dir'])
-                    shutil.copytree(out_dir / "lib", libs_path, dirs_exist_ok=True)
+                    try:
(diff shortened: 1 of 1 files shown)


unicorn_mode/build_unicorn_support.py
```diff
@@ -172,7 +172,10 @@ def detect_rustc_version():
             if "linked_libs" in ln_json and any(["unicorn" in x for x in ln_json['linked_libs']]):
                 if "out_dir" in ln_json:
                     out_dir = Path(ln_json['out_dir'])
-                    shutil.copytree(out_dir / "lib", libs_path, dirs_exist_ok=True)
+                    try:
+                        shutil.copytree(out_dir / "lib", libs_path, dirs_exist_ok=True)
+                    except FileNotFoundError:
+                        shutil.copytree(out_dir / "lib64", libs_path, dirs_exist_ok=True)
```
wtdcode, line 178: I would assume you are using rhel or some other derived distros where cmake ouputs at `lib64`.

The hot patch itself is fine and the root cause is that we need to set this via unicorn's `DCMAKE_INSTALL_LIBDIR`.

wtdcode (approved):

PR #2600, opened 2025-11-21
make testing easier

wtdcode:
Solve https://github.com/AFLplusplus/unicornafl/issues/51

From what I have tried locally, this finds the crash at 69th execution exactly and I aslo loose the condition by checking if there are extra edges.

### DIFF
--- a/test/test-unicorn-mode.sh
+++ b/test/test-unicorn-mode.sh
@@ -83,10 +83,10 @@ test -d "$DIR" && {
         $ECHO "$GREY[*] running afl-fuzz for unicorn_mode compcov, this will take approx 35 seconds"
         {
           export UNICORN_AFL_CMPCOV=1
-          ../afl-fuzz -m ${MEM_LIMIT} -V15 -U -i in -o out -d -- "$PY" ../unicorn_mode/samples/compcov_x64/compcov_test_harness.py @@ >>errors 2>&1
+          ../afl-fuzz -m ${MEM_LIMIT} -V35 -U -i in -o out -d -s 0 -- "$PY" ../unicorn_mode/samples/compcov_x64/compcov_test_harness.py @@ >>errors 2>&1
           unset UNICORN_AFL_CMPCOV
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
if it finds extra edges - is the only possiblity for this because of cmplog?

domenukk:
> Solve [AFLplusplus/unicornafl#51](https://github.com/AFLplusplus/unicornafl/issues/51)
> 
> From what I have tried locally, this finds the crash at 69th execution exactly and I aslo loose the condition by checking if there are extra edges.

FWIW there's a chance algorithms or underlying implementation changes at some point, would that be an issue for this test?I.e., deterministic mode changes, changes in weight for the scheduler, etc.

wtdcode:
> if it finds extra edges - is the only possiblity for this because of cmplog?

Yes, actually, this wasthe  original logic:

https://github.com/AFLplusplus/AFLplusplus/blob/4.00c/test/test-unicorn-mode.sh#L86

I made it harder in my previous PR.

wtdcode:
> > Solve [AFLplusplus/unicornafl#51](https://github.com/AFLplusplus/unicornafl/issues/51)
> > From what I have tried locally, this finds the crash at 69th execution exactly and I aslo loose the condition by checking if there are extra edges.
> 
> FWIW there's a chance algorithms or underlying implementation changes at some point, would that be an issue for this test?I.e., deterministic mode changes, changes in weight for the scheduler, etc.

Not too much I would say because it only need 1 extra edge to succeed, which should be only possible with compcov (or crash).

vanhauser-thc:
```
./test-all.sh: 115: ./test-unicorn-mode.sh: Syntax error: "}" unexpected (expecting "fi")
make: *** [GNUmakefile:373: tests] Error 2
Error: Process completed with exit code 2.
```

wtdcode:
not a bash expert obviously =)

vanhauser-thc:
thanks!

PR #2603, opened 2025-11-25
Add missing initialization for permission fields in afl_state and afl_forkserv

### LINKED ISSUE #2602: Input file permission wrong for `afl-showmap`
Hey!

The commit c8d1b66af3706ebd4da831e352d5720fc323312e introduces a new variable `mode_t perm` into the `struct afl_forkserver`. While this value is initialized when using `afl-fuzz`, it doesn't appear to be in the case of `afl-showmap`. This results in the input file being created with the permission `0o0`, which prevents the target from opening it. Other tools using the `struct afl_forkserver` may also be affected. 

Best,
Nils

graux-pierre:
Fixes #2602

### DIFF
--- a/src/afl-forkserver.c
+++ b/src/afl-forkserver.c
@@ -329,6 +329,8 @@ void afl_fsrv_init(afl_forkserver_t *fsrv) {
   fsrv->uid_set = 0;
   fsrv->gid_set = 0;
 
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
thanks!

PR #2605, opened 2025-11-26
Add afl-showmap tests

nbars:
This adds three tests in total:
  - Whether afl-showmap produces the same coverage for the same input in stdin mode.
  - whether afl-showmap produces the same coverage for the same input when using -i and @@;
  - whether afl-showmap produces different coverage for different inputs when using -i and @@ (see issue #2602).


vanhauser-thc:
thanks a lot!

Issue #2597, opened 2025-11-20
`build_gramatron_mutator` cannot compile with GCC14+


jubnzv:
**Describe the bug**
Executing `./custom_mutators/gramatron/build_gramatron_mutator.sh` fails when GCC14+ is used.

**To Reproduce**
```
cd ./custom_mutators/gramatron && ./build_gramatron_mutator.sh
```

**Expected behavior**
The mutator successfully built.

**Screen output/Screenshots**
It fails when compiling the `json-c` dependency:
```
arraylist.c: In function 'array_list_new':
arraylist.c:49:43: error: 'calloc' sizes specified with 'sizeof' in the earlier argument and not in the later argument [-Werror=calloc-transposed-args]
   49 |   if(!(arr->array = (void**)calloc(sizeof(void*), arr->size))) {
      |                                           ^~~~
arraylist.c:49:43: note: earlier argument should specify number of elements, later size of each element
``` 

**Additional context**
AFL++ version: `dev` branch; `f860aa2e132e3d110aa9b50a3f30fcb697a9fd86`
The `calloc-transposed-args` warning was introduced in GCC14. Previous versions of GCC compiled that code successfully.

vanhauser-thc:
json-c needs to be updated to a newer version (and build script switched to cmake)

vanhauser-thc:
fixed in dev branch, thanks @carvedcoder!

PR #2613, opened 2025-12-12
fix: Update gramatron build script to use CMake and fix GCC 14+ compatibility

CarvedCoder:
## Problem
The gramatron mutator failed to build with GCC 14+ due to:
1. `calloc-transposed-args` warnings in json-c
2. `SSIZE_T_MAX` preprocessor errors with newer compilers (after adding json-c 0.18)
3. Outdated autotools-based build system

## Solution
- **Switched from autotools to CMake** for building json-c (as requested by maintainer in #2597)
- **Automatic SSIZE_T_MAX fix** - Patches malformed preprocessor directives in old json-c versions
- **Smart compiler detection** - Auto-detects and uses the same compiler (GCC/Clang) that built AFL++
- **Robust submodule handling** - Works both as git submodule and standalone clone
- **Automatic cleanup** - Resets json-c to clean state before building to prevent conflicts (happened due to json-c 0.18)
- **Disabled unnecessary components** - Only builds the library (no tests/apps)

## Changes Made
- Updated `custom_mutators/gramatron/build_gramatron_mutator.sh`:
  - Replaced autotools build with CMake
  - Added Python/awk-based patch for SSIZE_T_MAX issue (in json-c 0.18)
  - Added compiler detection logic
  - Improved git submodule initialization with fallback to direct clone
  - Added proper include paths for json-c headers
  - Platform-specific linker flags
  - Changed the data of JSONC_VERSION to get json-c 0.18

## Testing
Tested successfully on:
- Linux with GCC 14
- Docker environment
- Fresh repository clone

## Build Output
```
root@ee7bd038e23f:/src/custom_mutators/gramatron# ./build_gramatron_mutator.sh
Gramatron Mutator build script (Linux)

[*] Performing basic sanity checks...
[*] Compiler: gcc-14
[+] All checks passed!
[*] Making sure json-c is checked out
[*] Detected git repository, attempting submodule initialization
[+] Submodule initialized successfully
[+] Got json-c.
[*] json-c already built

[+] Json-c successfully prepared!
[+] Building gramatron now.
/usr/bin/ld: ../../src/afl-performance.o: plugin needed to handle lto object
/usr/bin/ld: ../../src/afl-performance.o: plugin needed to handle lto object

[+] gramatron successfully built!
[*] Output: gramatron.so
```


domenukk:
Applying patches through python sounds quite horrible. Probably we should deprecated this and hook up the maintained [gramatron mutator](https://docs.rs/libafl/0.15.3/libafl/mutators/gramatron/index.html) inside LibAFL instead.

(this can be done with the LibAFL CustomMutator adapter:
https://github.com/AFLplusplus/AFLplusplus/blob/68b492b2c7725816068718ef9437b72b40e67519/custom_mutators/rust/custom_mutator/src/lib.rs#L438

vanhauser-thc:
Thanks for this!
Why does this need a SSIZE_T fix? Is json-c unmaintained or they don’t want to fix this?
the „use the same compiler“ is unnecessary imho but that is not important.

vanhauser-thc:
> Applying patches through python sounds quite horrible. Probably we should deprecated this and hook up the maintained [gramatron mutator](https://docs.rs/libafl/0.15.3/libafl/mutators/gramatron/index.html) inside LibAFL instead.
> 
> (this can be done with the LibAFL CustomMutator adapter:
> 
> https://github.com/AFLplusplus/AFLplusplus/blob/68b492b2c7725816068718ef9437b72b40e67519/custom_mutators/rust/custom_mutator/src/lib.rs#L438

why should rust be better for this than python?
Python is installed everywhere, rust is not and as gramatron is not rust based, that new huge dependency is not the right move.

CarvedCoder:
Json-c 0.18 version has some bugs because they define SSIZE_MAX as SSIZE_T_MAX which causes it to have unterminated #if and #ifndef

```
/src/custom_mutators/gramatron/json-c/json_object.c:45: error: unterminated #if
   45 | #if SIZEOF_SSIZE_T == SIZEOF_INT
/src/custom_mutators/gramatron/json-c/json_object.c:44: error: unterminated #ifndef
   44 | #ifndef SSIZE_T_MAX
```
This bug seemed to be in 0.17 version too

So i had to use python to remove that part, it falls back to awk if python is not found

CarvedCoder:
Json-c doesn't seem to be that well maintained unless you get the latest commit which could introduce more bugs , cjson would be a better alternative for json-c since it doesn't need any build system and seems to be well maintained

vanhauser-thc:
I am fine with switching to cjson

CarvedCoder:
Ok , I will change the previous APIs to use cjson
I'm a bit busy now so it will take ~12 hours 

Is there any other directories which relies on json-c parsing?

vanhauser-thc:
nope, nothing else uses it thanks! and no hurry.

CarvedCoder:
accidentally made the commit with a wrong branch

domenukk:
> > Applying patches through python sounds quite horrible. Probably we should deprecated this and hook up the maintained [gramatron mutator](https://docs.rs/libafl/0.15.3/libafl/mutators/gramatron/index.html) inside LibAFL instead.
> > (this can be done with the LibAFL CustomMutator adapter:
> > https://github.com/AFLplusplus/AFLplusplus/blob/68b492b2c7725816068718ef9437b72b40e67519/custom_mutators/rust/custom_mutator/src/lib.rs#L438
> 
> why should rust be better for this than python? Python is installed everywhere, rust is not and as gramatron is not rust based, that new huge dependency is not the right move.

I'm not talking python vs rust I'm talking maintained version of gramatron (in LibAFL) vs 4 year old unmaintained version that doesn't even build anymore..
Why not just keep maintaining one?

CarvedCoder:
The afl++ implementation is tied with the internals through afl_state_t and as much as I could understand libafl doesn't reproduce the same type type of corpus-weighted splice logic.

Won't these cause problems if it was replaced?

vanhauser-thc:
@domenukk I agree with CarvedCoder - I don't understand your argument. are they 100% compatible and can just be switched and it works in AFL++?

vanhauser-thc:
@CarvedCoder so Dominik proposed someone writes a new gramatron custom mutator based on the libafl implementation. Apparently it received some bug fixes (but unsure if that were fixes for bugs that were introduced when porting to rust or in the original implementation, have a look if you care). 

not sure who would want to do such an effort - but it isnt me :-)

merging this PR.
thanks!

domenukk:
> The afl++ implementation is tied with the internals through afl_state_t and as much as I could understand libafl doesn't reproduce the same type type of corpus-weighted splice logic.
> 
> Won't these cause problems if it was replaced?

I honestly don't know, so _maybe_? I do know some people fixed bugs in LibAFL's version, I fear you will have to do the same for this one (it's been unmaintained and maybe unused for 4 years now) - and come back with more patches next week.
Effectively, we, as an afl++ project, would end up maintaining two versions of this grammar fuzzer, that doesn't perform better than Nautilus, which we also maintain...

I'm not against merging this now, but not sure it's the right way going forward

vanhauser-thc:
@CarvedCoder 

It looks like you forgot to add required files:

```
In file included from gramfuzz.c:5:
gramfuzz.h:16:10: fatal error: json-parser.h: No such file or directory
   16 | #include "json-parser.h"
      |          ^~~~~~~~~~~~~~~
compilation terminated.
In file included from gramfuzz-helpers.c:6:
gramfuzz.h:16:10: fatal error: json-parser.h: No such file or directory
   16 | #include "json-parser.h"
      |          ^~~~~~~~~~~~~~~
compilation terminated.
In file included from gramfuzz-mutators.c:6:
gramfuzz.h:16:10: fatal error: json-parser.h: No such file or directory
   16 | #include "json-parser.h"
      |          ^~~~~~~~~~~~~~~
compilation terminated.
In file included from gramfuzz-util.c:6:
gramfuzz.h:16:10: fatal error: json-parser.h: No such file or directory
   16 | #include "json-parser.h"
      |          ^~~~~~~~~~~~~~~
compilation terminated.
cc1: fatal error: json_parser.c: No such file or directory
```

CarvedCoder:
Ahh I'm really sorry it was in the accidental pr that I made I'm adding it rn

CarvedCoder:
hi @vanhauser-thc  I added the json-parser.h in this pr #2615 , sorry for the trouble

CarvedCoder:
> > The afl++ implementation is tied with the internals through afl_state_t and as much as I could understand libafl doesn't reproduce the same type type of corpus-weighted splice logic.
> > Won't these cause problems if it was replaced?
> 
> I honestly don't know, so _maybe_? I do know some people fixed bugs in LibAFL's version, I fear you will have to do the same for this one (it's been unmaintained and maybe unused for 4 years now) - and come back with more patches next week. Effectively, we, as an afl++ project, would end up maintaining two versions of this grammar fuzzer, that doesn't perform better than Nautilus, which we also maintain...
> 
> I'm not against merging this now, but not sure it's the right way going forward

In long term deprecating this or migration to libafl  or nautilus would be a better choice while keeping this as legacy as some users may still use it , this pr was meant for maintenance to keep this version buildable and testable for current users. Writing a whole new custom_mutator will be very hard due to the internals it would be better to suggest nautilus for afl++ and libafl if they want rust

PR #2617, opened 2025-12-17
LibAFL Nautilus Custom Mutator

domenukk:
Adds a Nautilus custom Mutator.
It also makes it easy to use any other LibAFL mutator (or combination thereof)


custom_mutators/libafl_nautilus/README.md
```diff
@@ -0,0 +1,61 @@
...
+Or use the Makefile:
+```sh
+make
+```
+
+## Usage
+
+To run AFL++ with this mutator, you need to:
+1.  Set `AFL_CUSTOM_MUTATOR_LIBRARY` to the path of the compiled shared library.
+2.  Set `AFL_POST_PROCESS_KEEP_ORIGINAL=1` to ensure AFL++ stores the serialized tree in the queue, not the unparsed bytes.
```
vanhauser-thc: I understand but it makes the whole custom mutator moot. there will be no testcases other instances can import.

custom_mutators/libafl_nautilus/README.md
```diff
@@ -0,0 +1,61 @@
...
+3.  Set `NAUTILUS_GRAMMAR_FILE` to the path of your grammar JSON file.
+
+```sh
+export NAUTILUS_GRAMMAR_FILE=/path/to/grammar.json
+export AFL_CUSTOM_MUTATOR_LIBRARY=target/release/liblibafl_nautilus.so
+export AFL_POST_PROCESS_KEEP_ORIGINAL=1
+
+# Run AFL++ (add -n if you don't want deterministic fuzzing, usually good for grammar)
+afl-fuzz -i in -o out -n -- ./target @@
```
vanhauser-thc: with this flag, there is no coverage check and no inputs will ever be saved

vanhauser-thc:
if the queue does not contain generated output then this custom mutator is pointless.
for a similar grammar mutator it just creates it own data directory and reads/writes the binary form in there (same file name)

custom_mutators/libafl_nautilus/README.md
domenukk: True

custom_mutators/libafl_nautilus/README.md
domenukk: Well you can do grammar fuzzing with it, isn't that the point? :D

domenukk:
I can change it so it has two queues, effectively, but I don't think it's super clean. Would be more "correct" to use the exporter script in a loop on new testcases

custom_mutators/libafl_nautilus/README.md
vanhauser-thc: if you want to show how libafl can be integrated into AFL++ as a custom mutator and be useful then it is in your interest that it can actually help the fuzzing campaign :-)

vanhauser-thc:
you could also keep the binary form in memory - no great for restarts though.

the problem with the exporter script is that the other instances look in `./queue/id:xxxxxx,` so this would not help. needs to be in a second queue I think.

domenukk:
Ok I guess I'll rework that guy then

custom_mutators/libafl_nautilus/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+            }
+        };
+
+        // Mutate the input
+        self.mutator.mutate(&mut self.state, &mut input)?;
+
+        // Serialize the mutated input
+        self.serialized_buf.clear();
+        input.unparse(self.context, &mut self.serialized_buf);
+        self.cache.put(self.serialized_buf.clone(), input.clone());
```
domenukk: This is a bit ugly now, I can't just rely on storing the input anywhere for stacked mutations so I have to cache it

domenukk: (since we store target bytes in the corpus now, instead of the AST)

domenukk:
@vanhauser-thc this is ready to be played around with

domenukk:
Nevermind lol, need to fix more things

domenukk:
Now it should work

test/test-nautilus.sh
```diff
@@ -0,0 +1,223 @@
...
+
+$ECHO "$BLUE[*] Running libafl_nautilus unit tests"
+(cd "$CUSTOM_MUTATOR_PATH" && cargo test) || exit 1
+
+LIBAFL_BASE_PATH=$(cd $(pwd)/../custom_mutators/libafl_base;pwd)
+$ECHO "$BLUE[*] Running libafl_base unit tests"
+(cd "$LIBAFL_BASE_PATH" && cargo test --features=mutator) || exit 1
+
+  # Create the vulnerable program
+  cat > test-nautilus-target.c <<EOF
```
vanhauser-thc: you can remove this as you add the c file to the repo with this PR

test/test-nautilus.sh
```diff
@@ -0,0 +1,223 @@
...
+  AFL_CUSTOM_MUTATOR_LIBRARY="$LIB_PATH" timeout 60s ../afl-fuzz -V 20 -m ${MEM_LIMIT} -i in-nautilus -o out-nautilus -d -- ./test-nautilus-target >> errors 2>&1
+  echo "DEBUG: afl-fuzz finished with code $?"
+  ls -l errors
+  
+  unset NAUTILUS_GRAMMAR_FILE
+  unset AFL_NO_UI
+  unset AFL_CUSTOM_MUTATOR_ONLY
+
+  # Check if afl-fuzz ran successfully and loaded the mutator
+  if grep -F "Found 'afl_custom_mutator'" errors >/dev/null; then
```
vanhauser-thc: How does this work? the grammar does not contain the string that results in an abort in the harness

GNUmakefile
```diff
@@ -625,6 +625,9 @@ code-format:
 	./.custom-format.py -i qemu_mode/libcompcov/*.h
 	./.custom-format.py -i qemu_mode/libqasan/*.c
 	./.custom-format.py -i qemu_mode/libqasan/*.h
+	cargo +nightly fmt --manifest-path custom_mutators/rust/Cargo.toml --all
+	cargo +nightly fmt --manifest-path custom_mutators/libafl_base/Cargo.toml
+	cargo +nightly fmt --manifest-path custom_mutators/libafl_nautilus/Cargo.toml
```
vanhauser-thc: please add a `-` as otherwise the make target fails if no cargo nightly is installed, which is unnecessary :)

vanhauser-thc:
@domenukk 

The CI fails because:
```
[!] [A/B] Positive test failed: Nautilus did NOT find the HYBRID crash ('...token_0') in 30s.
[!] Check 'errors' file for details.
```
more antigravity to the rescue? ;-)

domenukk:
Aaah works on my machine, I guess the CI machine is too slow...
I'll think of something (maybe just increase timeout?)

PR #2618, opened 2025-12-17
cmplog buffer overread

nbars:
Just to trigger CI once, the fix will be added after CI

### DIFF
--- a/test/test-cmplog-overread.c
+++ b/test/test-cmplog-overread.c
@@ -1,0 +1,25 @@
+// Test: cmplog buffer over-read bug
+// __cmplog_rtn_hook_* functions copy 32 bytes regardless of string length
+// Compile: AFL_LLVM_CMPLOG=1 ./afl-clang-fast -o test-cmplog-overread test/test-cmplog-overread.c
+// Run: AFL_CMPLOG_DEBUG=1 ./test-cmplog-overread
+
+#include <stdio.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+int main(void) {
+    long ps = sysconf(_SC_PAGESIZE);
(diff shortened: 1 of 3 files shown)


nbars:
Hmm, interesting, the test works for me locally. I will recheck.

vanhauser-thc:
overread should not happen because **up-to** 32 bytes are copied. it is checked if the bytes leak over a page and checks the permission on the new page before copying.

nbars:
The validation happens, but even if it returns a value < 32, it still copies 32 bytes (constant) regardless. However, still not sure why this is not triggering on the CI. But should it be correct to use the validated length in the copy statements?

https://github.com/AFLplusplus/AFLplusplus/blob/68b492b2c7725816068718ef9437b72b40e67519/instrumentation/afl-compiler-rt.o.c#L2692

vanhauser-thc:
> The validation happens, but even if it returns a value < 32, it still copies 32 bytes (constant) regardless. However, still not sure why this is not triggering on the CI. But should it be correct to use the validated length in the copy statements?
> 
> https://github.com/AFLplusplus/AFLplusplus/blob/68b492b2c7725816068718ef9437b72b40e67519/instrumentation/afl-compiler-rt.o.c#L2692

ah damn, there was a PR I accepted that fixed something, but now looking at this, this introduced this wrong handling.

I pushed this fix:
https://github.com/AFLplusplus/AFLplusplus/commit/b3a5c5d568f5f22702f9cd2b2a9abae3022938fd

nbars:
Ah, great! I guess this can be closed then, since the test case is not working on CI for some reason.

vanhauser-thc:
please have a look at my change, just to be sure I did not introduce an error - 4 eyes ...

nbars:
Sure, looks good to me! Did not expect, even though it makes sense, that deleting the branch auto-closes the PR :D

PR #2619, opened 2025-12-17
cmplog back-edge detection bug

nbars:
To trigger CI once, the fix will be added after CI

### DIFF
--- a/instrumentation/cmplog-instructions-pass.cc
+++ b/instrumentation/cmplog-instructions-pass.cc
@@ -26,6 +26,7 @@
 #include <sys/time.h>
 
 #include "llvm/Config/llvm-config.h"
+#include "llvm/ADT/SmallPtrSet.h"
 #include "llvm/ADT/Statistic.h"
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
please provide an explanation for this PR

nbars:
The original code used `continue` inside a for-loop iterating over
successor blocks:
```
    for (BasicBlock *B : BR->successors())
      if (IsBackEdge(BR->getParent(), B, DT)) continue;
```
This only skips the inner for-loop iteration; consequently, it always
adds cmplog instrumentation, even for loop conditions. The fix uses a
boolean flag to properly track when a back-edge is found and skip the
comparison entirely.

Additionally, extend IsBackEdge() to follow chains of blocks that have
only one successor (containing just PHI nodes and an unconditional
branch). These empty blocks could eventually lead to a back-edge that would
otherwise be missed.

nbars:
The test checks if loops are instrumented, which caused it to fail without the patch.

vanhauser-thc:
ok, makes sense, thanks.

PR #2620, opened 2025-12-17
Fix splicing to copy from correct offset in target file

nbars:
The splicing code was copying from the beginning of new_buf (offset 0) instead of from split_at. This meant the spliced result contained:
- Bytes 0 to split_at-1 from in_buf (correct)
- Bytes 0 to len-split_at-1 from new_buf (incorrect)

The correct behavior for splicing is to take:
- Head (0 to split_at-1) from the current input
- Tail (split_at to len-1) from the target input

Fixed in both fuzz_one_original() and mopt_common_fuzzing().

### DIFF
--- a/src/afl-fuzz-one.c
+++ b/src/afl-fuzz-one.c
@@ -3424,7 +3424,7 @@ retry_splicing:
     len = target->len;
     afl->in_scratch_buf = afl_realloc(AFL_BUF_PARAM(in_scratch), len);
     memcpy(afl->in_scratch_buf, in_buf, split_at);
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thanks!

vanhauser-thc:
@nbars sorry for the ping - AFL++ has a regression between 4.34c and 4.35c and benchmarking the diffs turned up this PR being the main culprit ... any ideas why?

vanhauser-thc:
sorry ignore that, this is not the culprit ...

PR #2624, opened 2025-12-17
Fix K-CTX context restoration in return blocks

nbars:
When AFL_LLVM_CTX_K is enabled, the context restoration in return blocks was incorrectly using IRB (which inserts at block start) instead of Post_IRB (which inserts before the return instruction).


vanhauser-thc:
thanks

PR #2625, opened 2025-12-17
llvm-common: skip null entries in global_ctors instead of breaking

nbars:
Null entries can appear in the middle of llvm.global_ctors when optimization passes remove constructor functions. Using 'break' caused constructors after null entries to be missed.

Changed to 'continue' to match LLVM's CtorUtils.cpp behavior, cf. https://github.com/llvm/llvm-project/blob/c386d6d3bfa400e94ac9cadb85a0b1a72ed54b2b/llvm/lib/Transforms/Utils/CtorUtils.cpp#L97


vanhauser-thc:
thanks!

PR #2627, opened 2025-12-18
cmplog: use LoopInfo for loop condition detection

nbars:
Realized that this could be done better.

This replaces manual block-walking back-edge detection with LLVM's LoopInfo analysis for more robust loop-condition skipping.


vanhauser-thc:
thanks!

PR #2629, opened 2025-12-21
Added dictionary for ASCII STL file format

seanm:


vanhauser-thc:
1. only PRs to dev (changed this for you)
2. There seems to be already an stl dictionary, so fix the conflict please

seanm:
> 1. only PRs to dev (changed this for you)

Ah, didn't realize that. (Side note: I don't think I've ever seen a project where the default `git clone` command results a branch where development work is not expected to come from.)  I guess I should have read contributing.md :)

> 2. There seems to be already an stl dictionary, so fix the conflict please

Well, nothing to fix, they are basically identical.

Any reason the stable branch cannot have this dictionary as well?  Doesn't seem like optional dictionaries would make anything less stable...

vanhauser-thc:
it is in stable.
everything first lands in dev, and then, after some time, gets pushed to stable.

PR #2630, opened 2025-12-23
Fix afl-showmap returning exit code 2 (crash) instead of 1 for timeouts

nbars:
In non-forkserver mode, the timeout signal handler set a global `timed_out` flag, but the exit code calculation only checked `fsrv->last_run_timed_out` which was never set. This adds a `run_timed_out()` helper to check both flags and also uses `FSRV_RUN_CRASH` return value for proper crash detection in forkserver mode.


vanhauser-thc:
haven't looked if no_forkserver_timed_out is reset everywhere needed, but looks fine otherwise. thanks

PR #2631, opened 2025-12-23
afl-showmap: add streaming mode (-S) for efficient coverage collection

nbars:
Add -S flag for continuous coverage collection via stdin/stdout. This allows processing inputs without writing them to disk and is faster when inputs arrive one by one, leveraging forkserver speed without needing to handle implementation details. This would be useful for computing coverage on the go for evaluation, or even for building a fuzzer prototype without the hassle of implementing the forkserver protocol and so on.

Currently, problems arise when `afl-showmap` prints an error message, since it apparently writes to stdout, which is also used for communication. I guess these are the options:

1. Write errors to stderr (expected behavior anyway, but probably will break stuff).
2. Ignore this situation and assume that the pipe will break anyway when an error is printed (I think all error messages are related to fatal errors).
3. Create an fd and send it as the first message to `afl-showmap`.

```
Protocol:
- Input:  [u32 length][u8 data[length]]... [u32 0] (zero length = EOF)
- Output: [u16 status][u32 edge_count][{u32 edge_idx, u8 hit_count}...] Status: bits 0-1 = exit status (0=exited, 1=timeout, 2=crash)
```

### DIFF
--- a/include/debug.h
+++ b/include/debug.h
@@ -33,9 +33,9 @@
  * Terminal colors *
  *******************/
 
-#ifndef MESSAGES_TO_STDOUT
-  #define MESSAGES_TO_STDOUT
-#endif
+/* Define MESSAGES_TO_STDOUT to send diagnostic messages to stdout.
+   By default, messages go to stderr (correct Unix convention).
+   afl-fuzz defines this in afl-fuzz.h for its TUI. */
 
 #ifdef USE_COLOR
 
--- a/examples/afl_showmap_streaming_mode.py
+++ b/examples/afl_showmap_streaming_mode.py
@@ -1,0 +1,178 @@
+#!/usr/bin/env python3
+"""
+Simple wrapper class for afl-showmap streaming mode.
+
+Usage:
+    with AflShowmapStreaming("./afl-showmap", "./target", ["arg1"]) as showmap:
+        status, crash_signal, edges = showmap.get_coverage(b"test input")
+        print(f"Status: {status.name}, hit {len(edges)} edges")
+        if crash_signal:
+            print(f"Crashed with signal: {crash_signal.name}")
+
+        # Batch processing
+        results = showmap.get_coverage_batch([b"input1", b"input2"])
+        for status, crash_signal, edges in results:
+            print(f"{status.name}: {len(edges)} edges")
+"""
+
+import os
+import signal
+import struct
+import subprocess
+import sys
+from enum import IntEnum
+from pathlib import Path
+
+
+class ExitStatus(IntEnum):
+    """Exit status from streaming protocol (bits 0-1 of status field)."""
+
+    EXITED = 0   # Normal exit
+    TIMEOUT = 1  # Execution timed out
+    CRASH = 2    # Target crashed
+
+
+# Status field layout (u16):
+#   bits 0-1:  exit status (0=exited, 1=timeout, 2=crash)
+#   bits 2-7:  reserved (must be 0)
+#   bits 8-15: signal number (only valid when status=crash)
+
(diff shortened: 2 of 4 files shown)


nbars:
Also, option three would allow us to record the target's output. Any thoughts?

vanhauser-thc:
IMHO everything should be written to stderr. but we have to check afl-cmin.bash and afl-cmin.py that they still work and not rely on stdout texts (and fix them to look in stderr otherwise).

what is the speed increase of the streaming mode?

nbars:
I will think about the interface and also check what is currently being written to stdout.

The benchmark below used a relatively small target, i.e., differences would be less pronounced with slower targets. However, it's still nice not to have to deal with copying files (-i) or having hassle with duplicated file names (-I). Furthermore, when the files coming in one by one, e.g., if you are testing a generator (PER_INPUT) the modus is really handy. 

 - BATCH: All inputs are processed in a single afl-showmap invocation using batch mode -I (only afl-showmap runtime is measured, not copying files nor creating the file list)
 - STREAM_SYNC: Inputs are streamed to afl-showmap synchronously (one at a time, waiting for each result)
 - STREAM_ASYNC: Inputs are streamed asynchronously (multiple inputs are written to stdin)
 - PER_INPUT: Each input triggers a separate afl-showmap process invocation

image

vanhauser-thc:
numbers look good.
maybe think about if this can be made the default if stdin is used (of if this can break in some circumstances.

> I will think about the interface and also check what is currently being written to stdout.

please ping me when you think the PR is ready to be merged :)

vanhauser-thc:
ping :)

nbars:
I will continue to work on this by the end of this week

nbars:
Regarding the change to diagnostics written to stderr by default: As far as I can tell, all tools use `-o -` and do not rely on stdout containing diagnostics, so this should be fine. I would expect the same for downstream, but mentioning this in the release notes might still be good.

It adds LVs for stdout & stderr to the protocol, so this will not be a breaking change later. This, however, is not implemented yet (as noted in the help text). Thus, this PR is ready from my side. I will add the stdout & stderr features later.

include/debug.h
```diff
@@ -33,9 +33,9 @@
  * Terminal colors *
  *******************/
 
-#ifndef MESSAGES_TO_STDOUT
-  #define MESSAGES_TO_STDOUT
-#endif
+/* Define MESSAGES_TO_STDOUT to send diagnostic messages to stdout.
+   By default, messages go to stderr (correct Unix convention).
+   afl-fuzz defines this in afl-fuzz.h for its TUI. */
```
vanhauser-thc, line 38: let's see if this breaks anything :-)

vanhauser-thc:
the CI shows some issues:

```
 [*] compiling test-instr.plain...
[*] Test 1: streaming vs batch mode equivalence (input '0')
  File "<string>", line 5, in <module>
struct.error: unpack requires a buffer of 6 bytes
[+] streaming mode produces same coverage as batch mode for input '0'
[*] Test 2: streaming vs batch mode equivalence (input '1')
  File "<string>", line 5, in <module>
struct.error: unpack requires a buffer of 6 bytes
[+] streaming mode produces same coverage as batch mode for input '1'
[*] Test 3: streaming mode differentiates inputs
[+] streaming mode produces different coverage for different inputs
[*] Test 4: streaming mode determinism
  File "<string>", line 5, in <module>
struct.error: unpack requires a buffer of 6 bytes
  File "<string>", line 5, in <module>
struct.error: unpack requires a buffer of 6 bytes
```

vanhauser-thc:
thanks!

PR #2632, opened 2025-12-23
afl-showmap: set AFL_MAP_SIZE for targets with large coverage maps

nbars:
The instrumentation runtime checks AFL_MAP_SIZE to verify the parent tool supports large maps before starting the forkserver. Without this env var set, targets requiring >64KB maps fail with FS_ERROR_MAP_SIZE.

afl-fuzz already sets this, but afl-showmap did not.

### DIFF
--- a/src/afl-common.c
+++ b/src/afl-common.c
@@ -1367,6 +1367,20 @@ u8 *u_simplestring_time_diff(u8 *buf, u64 cur_ms, u64 event_ms) {
 
 }
 
+/* Validate map size, returns validated size or FATALs if invalid */
+u32 validate_map_size(u32 map_size) {
+
+  if (!map_size || map_size > (1U << 29)) {
+
+    FATAL("illegal AFL_MAP_SIZE %u, must be between %u and %u", map_size, 64U,
+          1U << 29);
+
+  }
+
+  return map_size;
+
+}
+
 /* Reads the map size from ENV */
 u32 get_map_size(void) {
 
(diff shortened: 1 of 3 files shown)


src/afl-showmap.c
```diff
@@ -1592,13 +1592,16 @@ int main(int argc, char **argv_orig, char **envp) {
...
-      fsrv->map_size = map_size;
+      fsrv->map_size = DEFAULT_SHMEM_SIZE;  // dummy temporary value
+      char vbuf[16];
+      snprintf(vbuf, sizeof(vbuf), "%u", DEFAULT_SHMEM_SIZE);
+      setenv("AFL_MAP_SIZE", vbuf, 1);
 
     } else {
 
-      fsrv->map_size = 4194304;  // dummy temporary value
+      fsrv->map_size = map_size;
```
vanhauser-thc, line 1603: if the map_siize is >= `1U << 29` then this has to FATAL (maximum possible map size that is supported).

nbars: The existing check uses `> 1U << 29`. Is this correct, or should it be `>=`? See the new push.

src/afl-common.c
```diff
@@ -1367,6 +1367,20 @@ u8 *u_simplestring_time_diff(u8 *buf, u64 cur_ms, u64 event_ms) {
 
 }
 
+/* Validate map size, returns validated size or FATALs if invalid */
+u32 validate_map_size(u32 map_size) {
+
+  if (!map_size || map_size > (1U << 29)) {
```
vanhauser-thc, line 1373: It needs to be `>=` for this `if`

src/afl-common.c
```diff
@@ -1367,6 +1367,20 @@ u8 *u_simplestring_time_diff(u8 *buf, u64 cur_ms, u64 event_ms) {
...
+
+    FATAL("illegal AFL_MAP_SIZE %u, must be between %u and %u", map_size, 64U,
```
vanhauser-thc, line 1375: if map size is 0 then a default should be set (eg MAP_SIZE = 64kb) and not fataled I think.

vanhauser-thc: also a map_size of 4 would be fine, because code below sets it to minimum 64 with `if (map_size % 64) { map_size = (((map_size >> 6) + 1) << 6); }`

vanhauser-thc:
I want to make a new release and have this PR in it. will merged and then make the necessary changes. thanks!

PR #2633, opened 2025-12-23
afl-cc: add check for using TSAN and ASAN/LSAN at the same time

Rosayxy:
According to the [gcc's specifications on instrumentation options](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html), the `-fsanitize=thread` option cannot be combined with `-fsanitize=address`,  `-fsanitize=leak`. Therefore, this commit adds check for using TSAN and ASAN/LSAN and exits the compiling process when the incompatibility issue is detected.


vanhauser-thc:
thanks, doesn't hurt.
for the future, please only send PRs to the dev branch, changed this for you

PR #2637, opened 2025-12-24
afl-cmin: reimplement afl-cmin using C language

kcwu:
New afl-cmin is roughly 10-30% faster than afl-cmin.py. If the fuzzing target is light weight and run cmin with high parallelism (>32), the speed up might go up to 50-100%.
 - reduce python overhead
 - reduce the overhead of calling external afl-showmap
 - optimizations: less syscals, less memory usage, less disk io

The command option -T is extened:
 -T N      N workers for everything (N workers to run the fuzzing
	   target and N workers for cmin calculation). This is aligned
	   with afl-cmin.py
 -T N:M    N workers to run the fuzzing taget, M workers for all other
	   tasks. For example, you may want to limit the number of
	   fuzzer execution because they consume too much resources.

The majority of this patch is written by gemini (antigravity) and reviewed by me.


vanhauser-thc:
Please undelete afl-cmin and rename it afl-cmin.py, and install all afl-cmin*

vanhauser-thc:
ping

kcwu:
I will work on it this weekend.

kcwu:
renamed afl-cmin to afl-cmin.awk.

@pwnalone FYI, it uses afl-fuzz's `set_sanitizer_defaults()` to initialize ASAN_OPTIONS. Does it work for you?

vanhauser-thc:
thanks for the changes. please give me some time to test it, I am under a high load currently :-)

vanhauser-thc:
I tried this on one example, it had the same number of files, same number of deduped files, same number of unique tuples, but:

```
2026-01-07 16:32:49,336 - INFO - narrowed down to 1419 files, saved in "cmin-py"
```
vs
```
    Written 1420 files, covered 9919/9919 tuples (109230.77/sec) [elapsed 0s]
[+] Wrote 1420 files.
```

kcwu:
the number is not guaranteed identical because it depends on the processing order. Parallel workers make the ordering non-deterministic.

vanhauser-thc:
yeah this makes sense.
thanks!

PR #2638, opened 2025-12-24
calibration: use max observed execution time per seed for q->exec_us

Om005:
During calibration some runs use a resumed persistent-loop (fast) while one run (the fork + initialization) can be significantly slower. The previous code used the average per-seed time to set q->exec_us. If the average is much smaller than the fork/init run, the first fork run may be killed by the timeout, causing a repeated fork->timeout loop and preventing persistent mode from ever being used.

This change tracks the maximum observed single-run time during calibration and sets q->exec_us to max(max_single_run, avg_run_time). That ensures timeouts are large enough to accommodate fork/init runs while remaining robust against transient small runs.

Files changed:
- afl-fuzz.c: calibrate_case()

### DIFF
--- a/src/afl-fuzz-run.c
+++ b/src/afl-fuzz-run.c
@@ -520,6 +520,7 @@ u8 calibrate_case(afl_state_t *afl, struct queue_entry *q, u8 *use_mem,
   u8 fault = 0, new_bits = 0, var_detected = 0, hnb = 0,
      first_run = (q->exec_cksum == 0);
   u64 start_us, stop_us, diff_us;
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
This will have an impact on seed selection and power schedule. This will need a rigorous benchmark before this can be merged.

rather than this wouldn’t it be better if the MIN() would be used and for persistent target where a fuzz exec will need a fork, increase the timeout? This will need quite some changes though

vanhauser-thc:
Why did you close the pr?

PR #2645, opened 2026-01-06
Fix use of ASAN_OPTIONS in afl-cmin.py

pwnalone:
When attempting to minimize a set of crashes using an ASAN-instrumented target binary, `afl-cmin` fails to detect any crashes because it unconditionally sets `ASAN_OPTIONS` to `detect_leaks=0`. This PR adds the `abort_on_error=1` and `symbolize=0` ASAN options and pulls in any other options the user may have set in the environment. This can be demonstrated by running a command similar to the following.

```sh
afl-cmin -C -i output/default/crashes -o unique-crashes -- /path/to/asan/target
```

### DIFF
--- a/afl-cmin.py
+++ b/afl-cmin.py
@@ -212,6 +212,14 @@ tuple_index_type_code = "I"
 file_index_type_code = None
 
 
+def get_asan_options():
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thanks!

PR #2647, opened 2026-01-09
clarify instrumentation modes for beginners

### LINKED ISSUE #2640: Clearer documentation?
**IMPORTANT**

Thank you for making AFL++ better!

**Describe the bug**
I'm not sure if this should be a feature request or not, but as a newbie to afl++ I found the documentation a little confusing and have suggestions for how some of it may be improved.

**To Reproduce**
DNA

**Expected behavior**
DNA

**Screen output/Screenshots**
DNA

**Additional context**
I saw that `AFL_INST_RATIO` was [outdated](https://aflplus.plus/docs/env_variables/) and tried to find more info.  I am using clang and so the page on the [llvm instrumentation](https://github.com/AFLplusplus/AFLplusplus/blob/stable/instrumentation/README.llvm.md) is pertinent.

As a newbie it's easy to get confused between building afl _itself_ and using afl to instrument code.  So that when you read `LLVM_CONFIG=llvm-config-21 make` it's not clear when this applies.  Even adding something as simple as "Build afl with ..." would be useful.  It might be useful also to have some way of knowing when you're already using this feature.  I believe I am because I built a [Docker container](https://github.com/AFLplusplus/AFLplusplus/blob/stable/Dockerfile).  So as I understand it `AFL_INST_RATIO` isn't having any effect.  There's a lot of output so I might have missed it but it might be nice to have a clear message indicating when this is not the case.

I think these two things alone could go a long way.  I hope this is useful.

Regards

bajpai07:
This pull request improves AFL++ documentation by clarifying the different
instrumentation modes (compiler-based, QEMU, Unicorn) in a more beginner-friendly
way.

The goal is to help new users understand:
- what instrumentation is,
- when to use each mode,
- and common pitfalls when choosing an instrumentation backend.

This change is based on confusion experienced as a first-time AFL++ user and
addresses feedback from issue #2640.
Fixes #2640


domenukk:
I don't think this is necessary for beginners, they should just use afl-cc

vanhauser-thc:
for me it does not make things more clear. I think @gitonthescene should say if that would have helped or not, and if not, what would have.

bajpai07:
Thanks for the feedback, that makes sense.

I agree that for most beginners, simply using afl-cc is usually sufficient. My intention was mainly to address confusion when users encounter multiple instrumentation modes mentioned across the documentation.
If this change does not improve clarity in practice, I’m completely fine with adjusting or dropping it.
I’ll wait for input from @gitonthescene as well to understand what would have been more helpful from a beginner’s perspective.

vanhauser-thc:
PR by gitonthescene is merged, thanks

bajpai07:
Thanks for the review and clarification. I understand the direction and will keep this in mind for future contributions.
Appreciate the feedback!

PR #2650, opened 2026-01-11
clarify doc

gitonthescene:
I was thinking something like the following.

[#2640](https://github.com/AFLplusplus/AFLplusplus/issues/2640)


vanhauser-thc:
LGTM, thanks

Issue #2648, opened 2026-01-09
afl-cmin.py is incompatible with Python 3.14


loqs:
**Describe the bug**
afl-cmin.py is incompatible with Python 3.14's change of start method from fork to forkserver for the multiprocessing module due to reliance on global variables.

**To Reproduce**
```
$ git clone https://github.com/AFLplusplus/AFLplusplus
$ cd AFLplusplus/
$ CC=/usr/bin/clang LD=/usr/bin/lld make
$ ./afl-clang-fast -o test-instr.plain -O0 test-instr.c
$ mkdir in in2
$ echo 0 > in/in
$ MEM_LIMIT=none
$ ./afl-cmin -m ${MEM_LIMIT} -i in -o in2 -- ./test-instr.plain
2026-01-09 22:54:52,101 - INFO - use 1 workers (-T)
search: 1 files [00:00, 20867.18 files/s]
2026-01-09 22:54:52,105 - INFO - Found 1 input files in 1 directories
usage: afl-cmin.py [-h] -i dir -o dir [-f file] [-m megs] [-t msec] [-O] [-Q]
                   [-U] [-X] [--crash-dir dir] [-A] [-C] [-e] [-T WORKERS]
                   [--as_queue] [--no-dedup] [--debug]
                   /path/to/target_app [args ...]
afl-cmin.py: error: the following arguments are required: -i, -o, /path/to/target_app
Traceback (most recent call last):
  File "/build/afl++/src/AFLplusplus-4.35c/afl-cmin.py", line 829, in <module>
    main()
    ~~~~^^
  File "/build/afl++/src/AFLplusplus-4.35c/afl-cmin.py", line 624, in main
    files, hash_list = dedup(files)
                       ~~~~~^^^^^^^
  File "/build/afl++/src/AFLplusplus-4.35c/afl-cmin.py", line 555, in dedup
    with multiprocessing.Pool(args.workers) as pool:
         ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/multiprocessing/context.py", line 119, in Pool
    return Pool(processes, initializer, initargs, maxtasksperchild,
                context=self.get_context())
  File "/usr/lib/python3.14/multiprocessing/pool.py", line 215, in __init__
    self._repopulate_pool()
    ~~~~~~~~~~~~~~~~~~~~~^^
[...]

PR #2653, opened 2026-01-13
Make afl-cmin.py compatible with Python 3.14

loqs:
Remove afl-cmin.py's use of global variables which is incompatible with Python 3.14's change of start method from fork to forkserver for the multiprocessing module on the Linux.

Fixes https://github.com/AFLplusplus/AFLplusplus/issues/2648.

### DIFF
--- a/afl-cmin.py
+++ b/afl-cmin.py
@@ -96,124 +96,126 @@ class HelpFormatter(argparse.HelpFormatter):
         self.add_text("%s" % prog)
 
 
-parser = argparse.ArgumentParser(formatter_class=HelpFormatter)
+def init_args():
+    parser = argparse.ArgumentParser(formatter_class=HelpFormatter)
 
-cpu_count = multiprocessing.cpu_count()
-group = parser.add_argument_group("Required parameters")
-group.add_argument(
-    "-i",
-    dest="input",
-    action="append",
-    metavar="dir",
-    required=True,
-    help="input directory with the starting corpus",
+    cpu_count = multiprocessing.cpu_count()
+    group = parser.add_argument_group("Required parameters")
+    group.add_argument(
+        "-i",
+        dest="input",
+        action="append",
+        metavar="dir",
+        required=True,
+        help="input directory with the starting corpus",
 )
-group.add_argument(
-    "-o",
-    dest="output",
-    metavar="dir",
-    required=True,
-    help="output directory for minimized files",
-)
+    group.add_argument(
+        "-o",
+        dest="output",
+        metavar="dir",
+        required=True,
+        help="output directory for minimized files",
+    )
 
-group = parser.add_argument_group("Execution control settings")
-group.add_argument(
-    "-f",
-    dest="stdin_file",
-    metavar="file",
-    help="location read by the fuzzed program (stdin)",
-)
-group.add_argument(
-    "-m",
-    dest="memory_limit",
-    default="none",
-    metavar="megs",
-    type=lambda x: x if x == "none" else int(x),
-    help="memory limit for child process (default: %(default)s)",
-)
-group.add_argument(
-    "-t",
-    dest="time_limit",
-    default=5000,
-    metavar="msec",
-    type=lambda x: x if x == "none" else int(x),
-    help="timeout for each run (default: %(default)s)",
-)
-group.add_argument(
-    "-O",
-    dest="frida_mode",
-    action="store_true",
-    default=False,
-    help="use binary-only instrumentation (FRIDA mode)",
-)
-group.add_argument(
-    "-Q",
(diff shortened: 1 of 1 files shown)


afl-cmin.py
```diff
@@ -543,11 +532,12 @@ def run(self):
 
 class CombineTraceWorker(multiprocessing.Process):
 
-    def __init__(self, pack_name, jobs, r_out):
+    def __init__(self, pack_name, jobs, r_out, tuple_index_type_code):
         super().__init__()
         self.pack_name = pack_name
         self.jobs = jobs
         self.r_out = r_out
+        self.tuple_index_type_code
```
kcwu, line 540: unused?

afl-cmin.py
```diff
@@ -96,128 +96,124 @@ def __init__(self, prog, *args, **kargs):
...
-    required=True,
-    help="input directory with the starting corpus",
-)
-group.add_argument(
-    "-o",
-    dest="output",
-    metavar="dir",
-    required=True,
-    help="output directory for minimized files",
+def init_args():
```
kcwu, line 99: Personally I prefer "create_parser" and let the caller invoke parser.parse_args(). If so, the function might be reused and/or extended in the future (say, unit test).

afl-cmin.py
```diff
@@ -357,7 +343,7 @@ def get_nyx_map_size(target_dir):
     return map_size
 
 
-def afl_showmap(input_path=None, batch=None, afl_map_size=None, first=False):
+def afl_showmap(input_path=None, batch=None, afl_map_size=None, first=False, args, afl_showmap_bin):
```
kcwu, line 346: does it work for you?
```
SyntaxError: parameter without a default follows parameter with a default
```

afl-cmin.py
```diff
@@ -628,21 +618,35 @@ def collect_files(input_paths):
...
+        level=log_level, format="%(asctime)s - %(levelname)s - %(message)s"
+    )
+    logger = logging.getLogger(__name__)
+
+    init(args, logger)
+
+    afl_showmap_bin = search_binary("afl-showmap")
 
-    files = collect_files(args.input)
+    files = collect_files(args, args.input)
```
domenukk: This seems pointless?

domenukk: As in, we probably want to remove the redundant parameter

loqs: Removed.

afl-cmin.py
```diff
@@ -820,7 +823,7 @@ def save_file(idx):
         if args.no_dedup:
             # Unless we deduped previously, we have to dedup the crash files
             # now.
-            crash_files, hash_list = dedup(crash_files)
+            crash_files, hash_list = dedup(crash_files, args)
```
domenukk, line 826: Just another small nit, can we use args as first parameter everywhere for consistency?

vanhauser-thc: the python variant is not the primary tool anymore, was switched out for the new C implementation, hence I would rather forego on the gold plating :-)

vanhauser-thc:
@loqs  thanks a lot!

afl-cmin.py
```diff
@@ -96,128 +96,124 @@ def __init__(self, prog, *args, **kargs):
...
+        help='output file name like "id:000000,hash:value"',
+    )
 group.add_argument(
     "--no-dedup", action="store_true", help="skip deduplication step for corpus files"
-)
-group.add_argument("--debug", action="store_true")
-
-parser.add_argument("exe", metavar="/path/to/target_app")
-parser.add_argument("args", nargs="*")
+    )
```
kcwu, line 211:
```
IndentationError: unexpected indent
```

afl-cmin.py
kcwu, line 346: this function need `tuple_index_type_code` and `logger` as well.
please make sure the program still works with your patch.

vanhauser-thc:
damn afl-cmin.py is broken now.
because CI was green I wasn't aware ... because of the switch to afl-cmin based on C the CI does not test the python script anymore ...

loqs:
https://github.com/AFLplusplus/AFLplusplus/pull/2658 contains what I believe are the fixes.

afl-cmin.py
loqs: No. I was running the C afl-cmin when testing.

afl-cmin.py
loqs: Do you want me to rename or get the fixes pushed first?

afl-cmin.py
loqs: Changed to self.tuple_index_type_code = tuple_index_type_code and run has been adjusted to use it.

Issue #2651, opened 2026-01-11
python setuptools not detected as it is not a binary


loqs:
**Describe the bug**
System install python setuptools because a binary is being searched for not a python module.

**To Reproduce**
```
$ python -c 'import setuptools'
$ echo $?
0
$ git clone https://github.com/AFLplusplus/AFLplusplus
$ cd AFLplusplus/
$ git switch dev
$ make
$ AFL_UCAFL_NO_VENV=1 make test
....
[*] Now building unicornafl python bindings
[!] You opt in installing the unicornafl to your current site packages, which probably won't work.
[!] We will add --user for you and install with intepreter from /usr/bin/python3
[*] We will install unicornafl to venv at /usr using /usr/bin/python3
[!] No setuptools, will install setuptools first
```
Full log below.
**Expected behavior**
The system installed setuptools should be located.

**Additional context**

[afl++-4.35c.r20.gea152aee-1-x86_64-check.log](https://github.com/user-attachments/files/24555116/afl%2B%2B-4.35c.r20.gea152aee-1-x86_64-check.log)

I do not expect the build to succeed. The issue is just the detection of setuptools being wrong and it is easier to demonstrate with the venv disabled. Two more very minor issues which may not warrant their own bug reports.
- When a venv is being used https://github.com/AFLplusplus/AFLplusplus/blob/stable/unicorn_mode/build_unicorn_support.py#L126C22-L126C29 will locate a system maturin but it will not be used by https://github.com/AFLplusplus/AFLplusplus/blob/stable/unicorn_mode/build_unicorn_support.py#L137 causing the build to fail.
- When a venv is used `$PATH` is changed so the venv's python will be used as python at https://github.com/AFLplusplus/AFLplusplus/blob/ea152aee0eb66b27e3700eec2a1a3643fa5c33a4/test/test-nyx-mode.sh#L30 so the system provided https://github.com/nyx-fuzz/packer/blob/7967b432c1994059a02013c9bc8fdd22230df48a/packer/nyx_config_gen.py#L3 is not found causing nyxtests to fail. I suspect you are not seeing this as building  nyxmode is disabled for CI. Which is also why I suspect you may not be aware nyxmode does not build without intervention due to https://github.com/nyx-fuzz/libxdc/issues/23 unless a system provided libcapstone4 is available.

XENON102:
Hi @loqs
Can i work on this issue...

vanhauser-thc:
[...]

PR #2654, opened 2026-01-13
Detect setuptools using import

loqs:
Fixes https://github.com/AFLplusplus/AFLplusplus/issues/2651


loqs:
Do you want to change `maturin` detection to the same approach as well?

domenukk:
Why do you run it as cmd and not just directly try to import?

loqs:
> Why do you run it as cmd and not just directly try to import?

@domenukk because if the venv is being used it will be a different python with different available modules.

PR #2655, opened 2026-01-14
Improve filtering when using AFL_PC_FILTER_FILE

pyoor:
This PR fixes several issues in the PC filtering mechanism:
- Decrementing the PC address caused the first guard to be disabled even when explicitly included by the PC filter.
- When dumping the pcmap and applying filtering, disabled guards could lead to writing the relative PC address at an incorrect index.
- PC filtering is currently only supported for a single module.  This caused the guards in all other modules to remain enabled if not explicitly defined by the PC filter.


vanhauser-thc:
you guys are the only known users of your feature here, so I blindly trust that you know what you are doing - as it is disabled by default for normal users :-)

Issue #2611, opened 2025-12-10
Coresight mode improvement


ram9ir:
Hello, I would like to make some adjustments to the coresight_mode directory, and its subrepositories in coresight-trace, coresight-trace/coresight-decoder. 

could you tell me how to do this, should i submit a pull request to /RICSecLab/coresight-trace ? From what I understand, these repositories are no longer available. What should I do?

vanhauser-thc:
are you using coresight?

ram9ir:
Yes

vanhauser-thc:
Then send a PR here

ram9ir:
But as I said some parts of this subsystem is a subrepo and are located here: https://github.com/Ricseclab/coresight-trace

Ricseclab/corrsight-trace seems unmaintained and I don't have any contacts with original developers.

Should I just remove the sub-repo and place code directly in aflplusplus? Or maybe make a fork?

vanhauser-thc:
I was thinking about removing coresight because I thought nobody is using it :-)

I forked off all repos to AFLplusplus, please push there. note that there are already other forks, maybe they have good changes too that you want to have

azanegin:
+1 for coresight!

vanhauser-thc:
is this the only PR for this issue? I doubt it, right @ram9ir ? So I reopened after you marked the PR as auto-closing the issue.

ram9ir:
No, its not the only PR, i will make changes to coresight-trace and coresight_mode folder from AFLplusplus.

PR #2664, opened 2026-01-19
Update coresight mode

ram9ir:
cs-trace/cs-proxy changes:

- libforksrv (replaces patched glibc)
- added the ability to trace shared libraries set via the CS_TRACE_LIB environment variable
- added a new board (hi1616)
- remove execv mode
- set a threshold value equal to the etr buffer size to avoid pausing the tracing

coresight-decoder changes: 

- added processing of Long Address 32-bit ISO Long packets for ETMv4
- improved handling of exception packets in the trace stream
- do not shut down the decoder in case of an error, instead return the error code

fixes https://github.com/AFLplusplus/AFLplusplus/issues/2611


vanhauser-thc:
is this PR ready?

ram9ir:
yes

PR #2669, opened 2026-01-21
[COMPLETED A TODO] :/* TODO use 2 * HASH_LENGTH *

Jay-1409:
- make code-format ✔️

### DIFF
--- a/src/afl-fuzz-init.c
+++ b/src/afl-fuzz-init.c
@@ -32,6 +32,9 @@
 
 #ifdef HAVE_AFFINITY
 
+  /*defines*/
+  #define HASH_LENGTH 20
+
 /* bind process to a specific cpu. Returns 0 on failure. */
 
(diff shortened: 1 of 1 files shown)


domenukk:
I don't see the benefit? What even is a hashlength in this context?

Jay-1409:
> I don't see the benefit? What even is a hashlength in this context?

You're right to question this 

After rethinking the change, I realise the 2 * 20 here checks whether a filename consists of exactly 40 hexadecimal characters (which happens to be the length of a SHA1 hash string representation). I had blindly followed the TODO without a second thought apologies for that and  it wont happen again.

The comment says "don't forget the SHA 1 files" - this is detecting SHA1 filenames like a1b2c3d4... (40 hex chars). The 20 refers to SHA1's 20-byte output, which becomes 40 hex characters when represented as a string.

The TODO comment itself might be questionable; I have removed the TODO comment.

Apologies for overseeing it.
Thank you so much for the review!!

PR #2672, opened 2026-01-21
push to stable

vanhauser-thc:


vanhauser-thc:
@kcwu I had to fix a bug and your default behaviour of the cmin output filenames.

afl-cmin.py
```diff
@@ -903,7 +916,10 @@ def save_file(idx):
             crash_files, hash_list = dedup(args, crash_files)
 
         for idx, crash_path in enumerate(crash_files):
-            fn = base64.b16encode(hash_list[idx]).decode("utf8").lower()
+            if use_sha1_filenames:
+                fn = base64.b16encode(hash_list[idx]).decode("utf8").lower()
+            else:
+                fn = os.path.basename(crash_path)
```
kcwu, line 922: path1/foo will overwrite path2/foo. Is it okay?

vanhauser-thc: yes I am aware. this will only happen if the same filenane is in different folders.
I am not happy about this, but backward tracability is important.
I think renaming duplicate names (but keeping the extension!) is better. But that "rename to hashes" broke some workflows so I had to do a quick fix.

PR #2674, opened 2026-01-22
Allow configurable input placeholder instead of hard-coded @@

### LINKED ISSUE #2608: Allow customizing the `@@` input placeholder
There's an uncommon convention for an at-prefixed argument like "`@argfile.txt`" to indicate a file from which more arguments should be read. (It's supported by [Java `javac`](https://docs.oracle.com/en/java/javase/21/docs/specs/man/javac.html#command-line-argument-files), [Python `argparse`](https://docs.python.org/3/library/argparse.html#fromfile-prefix-chars) (with customizable prefix), [Scala `sopt`](https://github.com/fulcrumgenomics/sopt?tab=readme-ov-file#argument-files), and [the program I want to fuzz](https://rgbds.gbdev.io/docs/v1.0.0/rgbfix.1#:~:text=%40at_file).) AFL++ replaces the first occurrence of "`@@`" with its input path, so I can't do "`@@@`" because that would become "`argfile.txt@`".

One solution would be to specify a different placeholder instead of "`@@`" for AFL++. This would be comparable with `xargs`, which replaces "`{}`" by default but can configure it with `-I <replace-str>`.

I could just build an "AFL-fuzzable" version of my program that changes its own argfile prefix to not involve the at sign. Or (with considerably more work) I could write a test harness that treats AFL's input file as specifying both CLI args and stdin content (with some separator that won't be confused with the CLI itself), and dispatch them to the actual program being fuzzed.

AbhayXplorer:
Fixes #2608

This change allows configuring the input file placeholder via -I <str>
instead of relying on the hard-coded @@ sequence.

Default behavior is unchanged.
This avoids conflicts with programs that use @argfile syntax.


src/afl-common.c
```diff
@@ -228,35 +228,47 @@ void detect_file_args(char **argv, u8 *prog_in, bool *use_stdin) {
...
   while (argv[i]) {
 
-    u8 *aa_loc = strstr(argv[i], "@@");
+    u8 *aa_loc = strstr(argv[i], input_placeholder);
 
     if (aa_loc) {
 
-      if (!prog_in) { FATAL("@@ syntax is not supported by this tool."); }
+      if (!prog_in) {
+        FATAL("@@ syntax is not supported by this tool.");
```
vanhauser-thc, line 246: shouldn't this be replaced with `input_placeholder` then too?

src/afl-fuzz.c
```diff
@@ -24,6 +24,9 @@
 
  */
 
+char *input_placeholder = "@@";
```
vanhauser-thc, line 27: no, put this into the afl struct in include/afl-fuzz.h

vanhauser-thc:
this is AI generated and is simply not working. please stop posting PRs.

PR #2680, opened 2026-01-25
instrumentation: detect and abort on exec* calls

### LINKED ISSUE #2492: Show a warning (Or crash the fuzzer) when the target does an `exec*()` on an instrumented binary
Hi everyone, thank you for making AFL++ better and better.

I experienced a problem with the coverage map in AFL++.
I have two binary, both instrumented by `afl-cc`. The first one does an `exec()` to the second one. It should not happen, but I did a mistake with the configuration.
The problem is, because the second binary is instrumented, it adds more edges to the coverage map, so the coverage map is bigger than the number output by launching the first binary with `AFL_DUMP_MAP_SIZE=1`.

So that, I propose to add a warning when an input hits an edge that have an ID > `__afl_map_size` of the fuzzed target.

Best regards,

gyanranjanpanda:
Hi! This PR addresses the issue where calling exec*() functions from an instrumented binary to another instrumented binary causes coverage map corruption.

Fixes #2492

## The Problem
When a target does exec() to another instrumented binary, both binaries try to use the same shared coverage map. If the second binary has more edges than the first, it can write beyond the allocated map size, causing memory corruption or crashes.

## The Fix
Following the discussion in #2492, I implemented a compile-time detection with runtime abort approach:

1. **Detection during instrumentation** (`SanitizerCoveragePCGUARD.so.cc`):
   - The existing `isExecCall()` helper already detects calls to exec family functions (execve, execl, execlp, execle, execv, execvp)
   - When found, we inject a call to `__afl_exec_error()` before the exec call

2. **Runtime error handler** (`afl-compiler-rt.o.c`):
   - Added `__afl_exec_error()` function that prints a clear error message explaining the issue
   - Points users to the GitHub issue for more context
   - Calls abort() to prevent silent corruption

## Testing
The instrumentation logic was already in place, I just added the missing runtime function. The error message guides users to either:
- Avoid exec() between instrumented binaries, or
- Compile the exec'd binary without AFL++ instrumentation

Let me know if you'd like me to add support for any other exec variants or adjust the error message!

### DIFF
--- a/instrumentation/afl-compiler-rt.o.c
+++ b/instrumentation/afl-compiler-rt.o.c
@@ -1089,6 +1089,13 @@ static void __afl_start_forkserver(void) {
 
   }
 
+  if (getenv("LD_BIND_LAZY") == NULL) {
+
+    // prevent further executed programs to fuck up the coverage
+    setenv("AFL_DISABLE_LLVM_INSTRUMENTATION", "1", 1);
+
+  }
+
   if (getenv("AFL_OLD_FORKSERVER")) {
 
     __afl_old_forkserver = 1;
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -9,6 +9,8 @@
     - added `AFL_FORCE_FASTRESUME` which will ignore the saved hash of the
       target - but note it will only work if the coverage map size did not
       change
+    - prevent further executed instrumented programs by the fuzz target to
+      manipulate the coverage
   - afl-cc:
     - LLVM 22 support (they are again switching around include files ...)
     - g_/curl_/xml_ string support for COMPCOV, thanks to @Prajwal-kp-18


vanhauser-thc:
Hi, thanks for the PR.
please note that you make a major change that I have not signed off on the solution yet.
Executing a program can be totally fine in a harness. it is a bad idea because it slows down the fuzzing, but is some circumstances this might be needed.
The problem is though if the further executed programs are also instrumented for fuzzing - that is the issue. not the execution of programs itself.

gyanranjanpanda:
Thanks for clarifying that makes sense. I understand now that exec* itself isn’t the problem; the real issue is when it ends up spawning another AFL instrumented binary and effectively nests fuzzing. Executing an uninstrumented child can be perfectly valid in a harness.

My current change was too broad since it treated all exec* as unsafe. I’ve updated it to allow an escape hatch via `AFL_ALLOW_EXEC`. This way, we safely abort by default, but users can explicitly allow uninstrumented children (like `cp`) when they know it's safe.

vanhauser-thc:
Please reverse it - `AFL_LLVM_DENY_EXEC` and only then block.
also for mac/linux differences, you can do `#if defined(__APPLE__) ` - if you need to be complete with the exec* variants for each platform, otherwise this is not helpful.
remember to add the env var to docs/envs*.md, include/env*.h and afl-fuzz -hh output

docs/env_variables.md
```diff
@@ -306,10 +306,19 @@ For more information, see
 
 #### Thread safe instrumentation counters (in all modes)
 
+Setting `AFL_LLVM_THREADSAFE_INST` will inject code that implements thread safe
+counters. The overhead is a little bit higher compared to the older non-thread
```
vanhauser-thc: you duplicated 2 lines here

instrumentation/SanitizerCoveragePCGUARD.so.cc
```diff
@@ -112,6 +112,20 @@ static const char *ijon_enabled;
...
+
+  llvm::Function *Callee = callInst->getCalledFunction();
+  if (!Callee) return false;
+
+  if (!Callee->hasName()) return false;
+
+  llvm::StringRef FuncName = Callee->getName();
+  return (FuncName == "execve" || FuncName == "execl" || FuncName == "execlp" ||
+          FuncName == "execle" || FuncName == "execv" || FuncName == "execvp");
+}
```
vanhauser-thc: system and popen are missing. likely a few more I currently don't think about. plus macos specific ones? (with #if defined() )

instrumentation/afl-compiler-rt.o.c
```diff
@@ -204,7 +202,6 @@ __thread u32 __afl_ijon_state_log = 0;
 typedef struct afl_module_info_t afl_module_info_t;
 
 struct afl_module_info_t {
-
```
vanhauser-thc, line 207: pls run "make code-format" ...

qemu_mode/libqasan/dlmalloc.c
```diff
@@ -6241,7 +6241,7 @@ mspace create_mspace(size_t capacity, int locked) {
   size_t msize;
   ensure_initialization();
   msize = pad_request(sizeof(struct malloc_state));
-  if (capacity < (size_t) - (msize + TOP_FOOT_SIZE + mparams.page_size)) {
+  if (capacity < (size_t)-(msize + TOP_FOOT_SIZE + mparams.page_size)) {
```
vanhauser-thc: code format ...

src/afl-fuzz.c
```diff
@@ -440,6 +440,7 @@ static void usage(u8 *argv0, int more_help) {
       "AFL_DEFER_FORKSRV: enforced deferred forkserver (__AFL_INIT is in a shared lib)\n"
       "AFL_FUZZER_STATS_UPDATE_INTERVAL: interval to update fuzzer_stats file in\n"
       "                                  seconds (default: 60, minimum: 1)\n"
+      "AFL_LLVM_DENY_EXEC: if set, aborts if the target calls exec*() functions\n"
```
vanhauser-thc: oh sorry, I was dumb. this belongs to afl-cc help output as it is compile time, not run time.

instrumentation/SanitizerCoveragePCGUARD.so.cc
```diff
@@ -1151,6 +1166,13 @@ bool ModuleSanitizerCoverageAFL::InjectCoverage(
 
       // printDebugInfo(IN);
 
+      // Check for exec* calls
+      if (isExecCall(&IN)) {
```
vanhauser-thc: you need to check for the AFL_LLVM_DENY_EXEC in the start of the plugin and then set a global variable, then check the variable here.

and implement the same functionality in SantizeCoverageLTO.so.cc :-)

gyanranjanpanda:
Hi @vanhauser-thc! I've addressed all the feedback:

## Changes Made

### ✅ Compile-Time Optimization
- Moved `AFL_LLVM_DENY_EXEC` check from runtime to compile-time in both LLVM plugins
- Added `deny_exec_calls` global variable checked during plugin initialization  
- Only inject `__afl_exec_error()` calls if variable is set at compile-time
- Removed runtime `getenv()` for better performance

### ✅ Full LTO Support
- Implemented `isExecCall()` function in `SanitizerCoverageLTO.so.cc`
- Added `AFLExecError` as class member for proper scope
- Complete parity with PCGUARD implementation

### ✅ Complete Function Coverage
- **Exec family**: `execve`, `execl`, `execle`, `execv`, `execvp`
- **macOS-specific**: `execvP`
- **Shell functions**: `system()`, `popen()`

### ✅ Documentation & Help
- Fixed duplicate lines in `docs/env_variables.md`
- Moved help text from `afl-fuzz.c` to `afl-cc.c` (compile-time, not runtime)
- Applied `make code-format`

instrumentation/SanitizerCoverageLTO.so.cc
```diff
@@ -129,6 +129,23 @@ void initializeModuleSanitizerCoverageLTOLegacyPassPass(PassRegistry &PB);
...
+
+  llvm::Function *Callee = callInst->getCalledFunction();
+  if (!Callee) return false;
+
+  if (!Callee->hasName()) return false;
+
+  llvm::StringRef FuncName = Callee->getName();
+  return (FuncName == "execve" || FuncName == "execl" || FuncName == "execlp" ||
+          FuncName == "execle" || FuncName == "execv" || FuncName == "execvp" ||
+          FuncName == "execvP" || FuncName == "system" || FuncName == "popen");
```
vanhauser-thc: there is also execvpe, execlpe, fexecve, execveat, posix_spawn, posix_spawnp. (also in pcguard pls)
I thought execvP was a typo but that really exists :-)

vanhauser-thc: maybe it makes sense to move this function to instrumentation/afl/llvm-common.cc ?

vanhauser-thc:
> Applied make code-format 

really? because so many lines are changed that should not be changed, files that you did not make any changes yourself

PR #2687, opened 2026-01-28
Fix/Improve Cmplog routines LLVM pass

kyakdan:
## Summary
- Fix signature checks for `strstr`-like functions that were listed but never matched due to checking for `int32` return type instead of pointer
- Add instrumentation for `memmem`, `strcasestr`, and `strnstr` functions
- Add `g_strstr_len` handling with correct 3-parameter signature (haystack, len, needle) 
- `apr_cstr_casecmp` (Apache APR)
- `sqlite3_stricmp` (SQLite)
- `sqlite3StrICmp` (SQLite internal)
- `apr_cstr_casecmpn` (Apache APR)
- `sqlite3_strnicmp` (SQLite)

## Added/Adjusted Tests
- [x] `test/test-cmplog-routines-pass.sh` verifies LLVM IR instrumentation
- [x] `test/test-cmplog.c` updated with `memmem` and `strstr` checks and `afl-fuzz` finds crash quickly with CmpLog enabled

### DIFF
--- a/instrumentation/cmplog-routines-pass.cc
+++ b/instrumentation/cmplog-routines-pass.cc
@@ -102,7 +102,7 @@ llvmGetPassPluginInfo() {
 bool CmpLogRoutines::hookRtns(Module &M) {
 
   std::vector<CallInst *> calls, llvmStdStd, llvmStdC, gccStdStd, gccStdC,
-      Memcmp, Strcmp, Strncmp, GStrstrLen;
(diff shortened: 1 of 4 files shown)


vanhauser-thc:
what a dumb oversight be me, thanks for fixing!

kyakdan:
> what a dumb oversight be me, thanks for fixing!

I also added a few comparison functions that were missing.

PR #2689, opened 2026-01-28
Fix GCC 13+ compilation error with returns_twice calls

### LINKED ISSUE #2541: Compile error with afl-gcc-fast: returns_twice call is not first in basic block
**IMPORTANT**

   - True
   - Error during compilation with `afl-gcc-fast`

**Describe the bug**
Since GCC 13 ([7a158a5](https://github.com/gcc-mirror/gcc/commit/7a158a5776f5ca95a318fcae0ca2dfefb0789538)), calls marked as `returns_twice` (e.g. `setjmp`) must be the *first* instruction in their basic block (BB). AFL’s inline instrumentation inserts statements *before* such calls, which violates GCC's invariant and triggers a hard error:

```
error: returns_twice call is not first in basic block …
```

This is a bug in the instrumentation pass (plugin); the GCC check is correct and required for valid `setjmp`/`longjmp` semantics.

**To Reproduce**

1. Build AFL++ with GCC >= 13:

```
$ gcc -v
gcc version 13.3.0 (Ubuntu 13.3.0-6ubuntu2~24.04) 
```

2. Compile minimal example with `-fchecking`:

```c
#include <setjmp.h>

jmp_buf env;
void bar ();

void foo () {
  bar ();
  setjmp (env);
}
```

```
$ afl-gcc-fast -fchecking ex.c
afl-cc++4.34a by Michal Zalewski, Laszlo Szekeres, Marc Heuse - mode: GCC_PLUGIN-DEFAULT
afl-gcc-pass ++4.34a by <oliva@adacore.com>
ex.c: In function ‘foo’:
ex.c:9:1: error: returns_twice call is not first in basic block 3
   9 | }
      | ^
_setjmp (&env);
ex.c:9: confused by earlier errors, bailing out
```

This example comes from Postgres: branch `REL_17_STABLE`, commit [591aeb5](https://github.com/postgres/postgres/commit/591aeb5068ae580fec423bd142641585d12d9537). The bug reproduces on other branches as well:

```
$ CC=afl-gcc-fast CXX=afl-g++-fast CFLAGS="-fchecking" ./configure
$ make
...
afl-gcc-fast -Wall -Wmissing-prototypes -Wpointer-arith -Wdeclaration-after-statement -Werror=vla -Wendif-labels -Wmissing-format-attribute -Wimplicit-fallthrough=3 -Wcast-function-type -Wshadow=compatible-local -Wformat-security -fno-strict-aliasing -fwrapv -fexcess-precision=standard -Wno-format-truncation -Wno-stringop-truncation -fchecking -I../../../src/include -D_GNU_SOURCE    -c -o autovacuum.o autovacuum.c
afl-cc++4.34a by Michal Zalewski, Laszlo Szekeres, Marc Heuse - mode: GCC_PLUGIN-DEFAULT
afl-gcc-pass ++4.34a by <oliva@adacore.com>
autovacuum.c: In function ‘AutoVacLauncherMain’:
autovacuum.c:3384:1: error: returns_twice call is not first in basic block 26
 3384 | }
      | ^
_138 = __sigsetjmp (&local_sigjmp_buf, 1);
autovacuum.c:3384: confused by earlier errors, bailing out
```

**Expected behavior**
Compilation succeeds; no `returns_twice` call is preceded by instrumentation within its basic block.

**Screen output/Screenshots**
See logs above.

**Additional context**

Without `-fchecking` compilation succeeds, but generated gimple is still invalid (`_setjmp` call is *not* the first instruction):

```
  <bb 3> [local count: 1073312329]:
  .afl_prev_loc.0_8 = __afl_prev_loc;
  ...    
  _setjmp (&env);
  goto <bb 5>; [99.96%]
```

---

**Attachments**

- [ex.c.344t.afl.txt](https://github.com/user-attachments/files/22281535/ex.c.344t.afl.txt) — gimple dump (attached).

Image

- CFG image — visualization:
  - red edges are *abnormal* — they are not executed in an actual program; this is an internal representation of setjmp/longjmp semantics.
  - black edges are *normal*.

**Observed CFG**

- `bb3`: instrumentation + `_setjmp`; invalid, which triggers GCC check.
- `bb5`: *normal* successor of `bb3`; executed after both normal and abnormal returns.
- `bb4`: `ABNORMAL_DISPATCHER`; reachable only via abnormal edges (not executed in the actual program). 

**What is wrong**

- `setjmp` call must be the first in `bb 3`. Any preceding statements are illegal. These statements run only in normal entry, while `setjmp` may also return via `longjmp`.
- Instrumenting `bb4` is unnecessary because it is not executable in the actual program, but this does not cause an error since GCC does not check it.

---

**Possible fixes**

**Common**

The core question is how to treat abnormal edges. Should fuzzing cover the two execution paths — normal return and abnormal return (via `longjmp`) — or not? The options below are ordered from simplest (but less precise) to more complex (and more semantically accurate).

**Notes**

- Abnormal edges exist only in CFG; they are not executed in the actual program.
- A `returns_twice` BB can be identified as the successor of the single `ABNORMAL_DISPATCHER` BB (at most one per function).
- CFGs may be more complex than the minimal example:

1. A predecessor of `returns_twice` BB (`bb2`) may have multiple *normal* successors.
2. A `returns_twice` BB (`bb3`) may have no outgoing abnormal edge (when `returns_twice` function is also `leaf`).

**A. Skip instrumentation in `returns_twice` BBs (quick hotfix)**

Avoid instrumenting any block that contains a `returns_twice` call.

*Result*: you lose coverage from the normal entry to `returns_twice` BB itself, and from its normal successors (`bb5`) when case (2) above applies.

**B. Split `returns_twice` BBs (preserve current behavior)**

For each `returns_twice` BB (`bb3`):

- Create a new BB containing only instrumentation.
- Redirect all *normal* incoming edges to this new BB.
- Let the new BB unconditionally jump into the `returns_twice` BB.
- Keep *abnormal* predecessor pointing directly to the `returns_twice` BB.

*Result*: you preserve all current coverage, but may introduce redundant instrumentation on normal entry to `returns_twice`. Such instrumentation is only *useful* when case (1) above applies.

**C. Path-sensitive instrumentation (complex but cleaner)**

Treat abnormal edges as non-executable in general. Still:

- If case (1) above applies, splitting as in **B** remains necessary.
- Keep instrumentation in *normal* successors of `returns_twice` BB to capture both normal and abnormal returns. Case (2) above does not affect this.

*Result*: you preserve all coverage without redundant instrumentation. This requires additional analysis to ensure correctness.

gyanranjanpanda:
Fixes #2541

## Problem

GCC 13+ requires that `returns_twice` calls like setjmp and vfork must be the first instruction in their basic block. The plugin was inserting instrumentation before these calls, which breaks this requirement and causes compilation to fail.

This was blocking compilation of projects like PostgreSQL on newer GCC versions.

## Solution

Skip instrumentation for any basic block that contains a `returns_twice` call. The check uses `gimple_call_flags()` to detect the `ECF_RETURNS_TWICE` flag.

## Trade-off

We lose coverage on these blocks. But:
- They're pretty rare in most code
- The alternative is a compilation error
- Can always revisit with a more sophisticated approach later if needed

## Changes

Just 15 lines added to afl-gcc-pass.so.cc, no other changes.

## Testing

Compiles cleanly and should fix the PostgreSQL issue mentioned in #2541.

### DIFF
--- a/instrumentation/afl-gcc-pass.so.cc
+++ b/instrumentation/afl-gcc-pass.so.cc
@@ -370,6 +370,21 @@ struct afl_pass : afl_base_pass {
 
     if (AFL_R(100) >= (long int)inst_ratio) return false;
 
+    /* Skip blocks with returns_twice calls (e.g., setjmp, vfork).
+       Since GCC 13, these calls must be first in their basic block. */
+    gimple_stmt_iterator gsi;
+    for (gsi = gsi_start_bb(bb); !gsi_end_p(gsi); gsi_next(&gsi)) {
+
+      gimple stmt = gsi_stmt(gsi);
+      if (is_gimple_call(stmt)) {
+
+        int flags = gimple_call_flags(stmt);
+        if (flags & ECF_RETURNS_TWICE) return false;
+
+      }
+
+    }
+
     edge          e;
     edge_iterator ei;
     FOR_EACH_EDGE(e, ei, bb->preds)


vanhauser-thc:
a) this is not against dev
b) I am not sure this is the right solution. I need someone who knows gimple well if this is how it should be done or not until I can merge this

gyanranjanpanda:
Hey @vanhauser-thc!

Rebased against `dev` now - should be all set on that front.

On the gimple thing - totally fair concern. I'm not super familiar with gimple internals, so I just went with what seemed like the safest approach: check if a block has a `returns_twice` call and skip instrumenting it. Basically just trying to avoid breaking the GCC 13+ requirement where those calls need to be first in the block.

Yeah, we lose coverage on those blocks, but figured that's better than the whole thing not compiling. These calls don't pop up too often anyway (mostly just setjmp/vfork stuff).

If there's a better way to do this that I'm missing, I'm definitely open to it. Happy to rework this if you or someone who knows gimple better has suggestions.

Thanks!

kyakdan:
Instead of skipping instrumentation, we can add a trampoline basic block where we add coverage instrumentation. This basic block then falls through to the "return_twice" block. This guarantees that the returns_twice call remains to be first in basic block.

At a high-level, this would convert the current instrumentation causing the error
```
[predecessor] ----> [instrumentation + setjmp]
                  ^        
                  |
[longjmp]   -------
```
to
```
[predecessor] ----> [trampoline: instrumentation] ---------------------> [setjmp]
                                                      fallthrough
                                                                       ^        
                                                                       |
[longjmp]  -------------------------------------------------------------
```
which respects the constraints by GCC 13+. Other edges targetting [setjmp] will be directed.

I guess this was already an option proposed in the original issue. I'm not a GIMPLE expert, but do have an initial implementation that works. @vanhauser-thc interested in a PR?

gyanranjanpanda:
Thanks for the detailed explanation, @kyakdan  the trampoline approach makes a lot of sense.

You're right that this option was mentioned in the original discussion. My current implementation (Solution A) focuses on a minimal and low-risk fix to unblock GCC 13+ users quickly, but it does come at the cost of skipping coverage for returns_twice blocks.

The trampoline design you outlined is clearly a more complete solution since it preserves coverage while respecting GCC 13+ constraints. I agree that it’s architecturally cleaner than simply skipping instrumentation.

Since this touches GIMPLE-level behavior and control flow shaping, I’d appreciate guidance from @vanhauser-thc on whether the project would prefer:

A minimal unblock fix (current approach), or

A more complete trampoline-based solution preserving coverage.

If the trampoline approach is preferred, I’m happy to:

Rework this PR to implement it, or

Open a follow-up PR with a clean implementation and tests.

Looking forward to your direction.

kyakdan:
With the trampoline basic block we ensure that we don't miss coverage instrumentation oppurtunities. For example this is how the initial implementation I mentioned instruments the example presented in the GitHub issue.

cfg

Given that the initial implementation is already there, my suggesstion would be that I push these changes in a PR and we continue the discussion there.

vanhauser-thc:
> I guess this was already an option proposed in the original issue. I'm not a GIMPLE expert, but do have an initial implementation that works. @vanhauser-thc interested in a PR?

yes please!

gyanranjanpanda:
the trampoline approach preserves more coverage but adds significant complexity (~60 lines of CFG manipulation vs. ~15 lines to skip). Given that 
returns_twice
 calls are rare, the simpler approach in my current PR might be more practical.

That said, if you prefer better coverage over simplicity, I'm happy to update my PR with the trampoline implementation,or if u want i can raise other pr with trampoline approch, Let me know!

PR #2691, opened 2026-02-01
ADD fix issue #2492

### LINKED ISSUE #2492: Show a warning (Or crash the fuzzer) when the target does an `exec*()` on an instrumented binary

taazcode:
#2492  pull request

Type of PR:
Bugfix (prevents out-of-bounds write and silent coverage corruption)

Related issue:
Fixes #2492

Purpose of this PR:
This PR fixes a silent coverage map corruption issue in
__sanitizer_cov_trace_pc_guard() by adding bounds checking and detecting
coverage map size mismatches.
Without this fix, interacting instrumented binaries can cause undefined
behavior and incorrect fuzzing results without any warning.

Testing done:
Yes. The changes were tested with instrumented binaries and verified that:

No out-of-bounds writes occur

Coverage mismatches are detected and reported

Normal fuzzing behavior is unchanged when no mismatch exists

### DIFF
--- a/COVERAGE_MAP_MISMATCH_FIX.md
+++ b/COVERAGE_MAP_MISMATCH_FIX.md
@@ -1,0 +1,581 @@
+# Coverage Map Mismatch Detection - User Guide
+
+## Problem Statement
+
+When fuzzing with AFL++, a common but difficult-to-debug issue occurs when your target binary calls another instrumented binary:
+
+```
+Binary 1 (instrumented) 
+    └─ exec() ──> Binary 2 (also instrumented) ⚠️ PROBLEM!
+```
+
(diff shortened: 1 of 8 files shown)


taazcode:
This fix is based on fundamental C safety principles (array bounds checking)
and adds minimal overhead (one comparison per edge) while significantly
improving reliability and debuggability.

PR #2693, opened 2026-02-02
nyx mode: fix exit code on errors when building

ksen-lin:
Hi!
This is a fix to return 1 on errors instead of 0 when building Nyx mode. This helps the caller know something went wrong so it can stop the build process (and is useful for automating the build of Nyx mode).
Not sure if those `exit 0` were intended though, but below i found `exit 1`'s as well.

### DIFF
--- a/nyx_mode/build_nyx_support.sh
+++ b/nyx_mode/build_nyx_support.sh
@@ -12,7 +12,7 @@ echo "[*] Performing basic sanity checks..."
 if [ "$CI" = "true" ]; then
 
   echo "[-] Error: nyx_mode cannot be tested in the Github CI, skipping ..."
-  exit 0
+  exit 1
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
Hmm this means if someone is building everything then it fails when on arm?
Did you run into a specific problem you needed to fix?
We could ignore errors from building nyx in the makefile but then compile errors would not error the build …
Hmm dunno

vanhauser-thc:
yeah I think it is safe to be merged. thanks.

ksen-lin:
thanks!

> Did you run into a specific problem you needed to fix?

Just in case, i was writing a script to build only the parts of AFL++ that i needed (x64 host), and my script was happy to continue despite a misinstalled rust

PR #2696, opened 2026-02-04
Fix GCC plugin instrumentation for returns_twice calls

kyakdan:
GCC 13+ requires returns_twice calls (like setjmp, sigsetjmp) to be the first instruction in their basic block. AFL++'s GCC plugin was inserting instrumentation before these calls, causing compilation errors with -fchecking enabled.

We fix this with a "trampoline" block using the following steps:
- detect blocks containing returns_twice calls
- split such a block to create an empty trampoline predecessor
- insert instrumentation in the trampoline
- redirect abnormal edges (e.g., from longjmp) to the trampoline successor (the original block, e.g. setjmp)

This ensures setjmp remains the first instruction in its block while still providing coverage that is symantically equivalent to the original instrumentation.

Fixes: https://github.com/AFLplusplus/AFLplusplus/issues/2541

### DIFF
--- a/instrumentation/afl-gcc-pass.so.cc
+++ b/instrumentation/afl-gcc-pass.so.cc
@@ -127,6 +127,7 @@
 #if defined(__has_include) && __has_include("memmodel.h")
   #include "memmodel.h"
 #endif
+#include <dominance.h>
 
 /* This plugin, being under the same license as GCC, satisfies the
    "GPL-compatible Software" definition in the GCC RUNTIME LIBRARY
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
thanks!

vanhauser-thc:
btw do you (or your company) would like to be the official maintainer of the gcc plugins?

kyakdan:
> btw do you (or your company) would like to be the official maintainer of the gcc plugins?

Yes, I can take care of them.

PR #2698, opened 2026-02-05
Fix the possibility of undefined behaviour in argument handling by validating argc before accessing 

Jay-1409:
**Type of PR**: BUG FIX
**Purpose of this PR**: Fix the possibility of undefined behaviour in argument handling by validating argc before accessing argv[2].


vanhauser-thc:
thanks

PR #2703, opened 2026-02-08
Fix undefined behavior by correcting the printf format specifier (%p used for an int instead of %d) and make the function safer to run and test.

Jay-1409:
# Your pull request

**Type of PR**: Feature or Bugfix or Enhancement
**Purpose of this PR**: Fix undefined behavior by correcting the printf format specifier (%p used for an int instead of %d) and make the function safer to run and test.
**I have tested the changes**: yes/no


vanhauser-thc:
thanks

PR #2705, opened 2026-02-10
Prevent orphaned forkserver

martinus:

**Type of PR**: Bugfix
**Purpose of this PR**: Earlier, when fetching map_size with `afl_fsrv_get_mapsize` is started. Here in fast_resume branch the forkserver is started again, so the old one should be killed to prevent having an orphaned forkserver.

I don't think leaving the orphaned forkserver hanging around is/was a big problem, but still it is more correct to kill it instead of leaving it orphaned.

## IMPORTANT

### DIFF
--- a/src/afl-fuzz.c
+++ b/src/afl-fuzz.c
@@ -3154,6 +3154,11 @@ int main(int argc, char **argv_orig, char **envp) {
 
     }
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thanks!

PR #2708, opened 2026-02-12
Fix cast<Instruction> crash when IRBuilder constant-folds (fixes #2707)

### LINKED ISSUE #2707: Error inside `llvm-19/include/llvm/Support/Casting.h` when instrumenting a library with UBSAN on Debian 13 i386
**Describe the bug**
Everything I'm about to describe takes place inside a Docker container running Debian 13, platform linux/i386. The host system is NixOS running Linux kernel 6.12.68 (x86_64).

I install LLVM 19 from official software repositories, and build AFL++ (dev branch, `6d7f0695b4e2d48`) from scratch, specifying `llvm-config-19`. Then, I fetch a third-party library I need for my project, called `freeradius-client` (https://github.com/FreeRADIUS/freeradius-client). I build this library using `afl-clang-fast`. The `./configure` stage completes, but during `make`, an error terminates the process:

```
clang: /usr/lib/llvm-19/include/llvm/Support/Casting.h:578: decltype(auto) llvm::cast(From *) [To = llvm::Instruction, From = llvm::Value]: Assertion `isa<To>(Val) && "cast<Ty>() argument of incompatible type!"' failed.
```

...followed by a long stack dump (full output along with diagnostic files attached at the end).

**To Reproduce**
I have prepared a trimmed down Dockerfile which reproduces the problem. Simply download it, rename to `Dockerfile` and, inside the same directory, run:

```sh
docker build --progress plain .
```

The error seems to be caused by the presence of `AFL_USE_UBSAN=1`, and it does not occur when this env is not set.

**Expected behavior**
The `freeradius-client` library should be compiled with instrumentation, even if `AFL_USE_UBSAN=1` is enabled.

**Additional context**
- `AFL_USE_ASAN=1` works fine
- I observed the same behavior on LLVM 18; I could not test LLVM 20+ due to missing LLVM packages for Debian 13 i386
- same problem present on `v4.35c` tag

**Attachments**
- [output.txt](https://github.com/user-attachments/files/25260877/output.txt)
- [env-b18173.c](https://github.com/user-attachments/files/25260875/env-b18173.c)
- [env-b18173.sh](https://github.com/user-attachments/files/25260876/env-b18173.sh)
- [Dockerfile.txt](https://github.com/user-attachments/files/25261061/Dockerfile.txt)

nbars:
`IRBuilder.crate*` returns `Value`, not `Instruction`, so unconditional casting may cause issues in specific circumstances, such as when a `Select` is folded into a constant.

Fixes #2707

### DIFF
--- a/instrumentation/SanitizerCoveragePCGUARD.so.cc
+++ b/instrumentation/SanitizerCoveragePCGUARD.so.cc
@@ -171,6 +171,7 @@ class ModuleSanitizerCoverageAFL
                                 ArrayRef<BasicBlock *> AllBlocks);
   void   updateCoverageForSelect(IRBuilder<> &IRB, Value *result,
                                  LoadInst *MapPtr, uint32_t &vector_cnt);
+  void   setNoInstrumentMetadata(Value *V);
 
   void SetNoSanitizeMetadata(Instruction *I) {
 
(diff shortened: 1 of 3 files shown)


nbars:
Just added release asserts to `split-compares-pass.so.cc` in case we have the same issue there. Probably worth opening a separate issue to track that.

vanhauser-thc:
Thanks! Instead of an assert why not fix it? :)

nbars:
Lazy and fix is not straightforward, also I would not expect this to actually happen. If the assert fires, we know what to do.

PR #2709, opened 2026-02-13
Add value profiling support

kyakdan:
# Summary
This PR adds value profiling (VP) to AFL++, a feature that extracts distance metrics from CmpLog comparison operands and uses them to provides an additional gradient signal by tracking how close those operands are to matching. It considers inputs that bring operands closer as "interesting". This is useful when compare operands are transformed (e.g., hashed or XOR'd) in ways that make CmpLog's direct solve attempts less effective. VP gives the fuzzer gradient information to incrementally converge on the correct values.

# Modes
This feature is controlled via `AFL_VALUE_PROFILE` environment variable and it requires CmpLog to collect comparison information:
- `AFL_VALUE_PROFILE=0`: disabled (default)
- `AFL_VALUE_PROFILE=1`: always active
- `AFL_VALUE_PROFILE=N` (N > 1): activate after N seconds of edge-coverage stagnation; deactivate when coverage recovers (after at least one full queue cycle)

# How it works
The following steps describe how value profiling works at a high level
1. VP extracts per-site distance features by running the CmpLog binary
  a. absolute distance and hamming distance for INS compares
  b. byte-level match counts and hamming distance for the first different byte for RTN compares
1. Features are hashed into a VP virgin bitmap; new bits represent progress
1. Inputs producing VP progress are saved to the queue (tagged `+vp`)
1. A per-site ranking approach favors inputs with the smallest distance to each comparison site
1. Queue culling includes VP-favored entries alongside coverage-favored ones

### Testing
The PR adds integration tests in `test-llvm.sh`. It also contains an example illustrating the benifits of value profiling. To run the example manually, you can run the following command from AFL++ source root:

```sh
./afl-clang-fast -o test/test-value-profile test/test-value-profile.c
AFL_LLVM_CMPLOG=1 ./afl-clang-fast test/test-value-profile.c -o test/test-value-profile.cmplog

AFL_LLVM_CMPLOG=1 ../afl-clang-fast -o test-value-profile.cmplog test-value-profile.c
mkdir -p /tmp/in && echo 00000000 > /tmp/in/seed

AFL_NO_UI=1 AFL_VALUE_PROFILE=1 AFL_I_DONT_CARE_ABOUT_MISSING_CRASHES=1 AFL_SKIP_CPUFREQ=1 AFL_BENCH_UNTIL_CRASH=1 ./afl-fuzz -i /tmp/in -o /tmp/out -c ./test/test-value-profile.cmplog -t 5000 -m none -- ./test/test-value-profile
```

Note: if you already ran the `afl-system-config` or `afl-persistent-config` scripts, you won't need all options/env vars mentioned in the `afl-fuzz` command above.


vanhauser-thc:
Thanks!

I have implemented it myself some time ago:
https://github.com/AFLplusplus/AFLplusplus/tree/vp

The problem with the libfuzzer style implementation is that it pollutes the corpus with outdated interesting entries. In my implementation they are disabled if the entry was only added because of a closer distance and a better is found. And I added a few other performance and efficiency hacks. On the other hand I likely have a bug because nested things could not be solved.

I like your approach of activating if no coverage gains are reached.

I think merging the best of both approaches is best, WDYT?

kyakdan:
Ah I didn't know about your implementation. I agree that merging the best of both approches makes sense. 

Indeed libfuzzer-style value profiling can blow up the corpus. On the other hand, only keeping queue entries that reach the best distance for a given location can also have negative effects in some cases. For example, a comparison at a given location can involve different operand pairs as in the below example. In this example, we have a loop that compares different elements from two arrays. If we only keep track of the input with the best distance for this location, we'd stop making progress when a single pair becomes close or solved.  

```c
int matches =  0;
for (int i = 0; i < MAX_ITERATIONS; i++) {
  if (needle[i] == haystack[i]) {
    matches += 1;
  }
}

// some checkes based on matches
```

That said, I guess it is always possible to come up with special cases where a certain approach wouldn't perform well. It would be straight forward to adjust my implementation to only add an entry to the queue if it reaches a better distance to what we saw so far.

Regarding nested checks, I slightly adjusted the test example and added nested checks and they were solved. 

How would you suggest we best merge the best of both approaches? Would it make senese to cherry-pick or reimplement the improvement from your branch on top of this one (given that your initial implementation is 1058 commits behind stable), or would it make more sense to start fresh?

vanhauser-thc:
I think it’s best to build on your pr and cherry pick from my implementation. Disabling corpus entries is important but we will need to take care of special cases like that.
Also the control[] set to not needing to walk the whole map for changes.
I am on my honeymoon and don’t have the time to take an in-depth look at it.
Go for it already if you want

kyakdan:
First of all Congratulations!

I'll take a close look at your implementation and start cherry picking from it. We can then finalize the implementation/discussion when you're back.

kyakdan:
I took a close look at your `vp` branch and thought about what would be the best way to merge both approaches. My proposal would be to enable two levels of value profile that configurable through a environment variable (analogous to the CmpLog levels through the `-l` flag):
- `AFL_VALUE_PROFILE_LEVEL=1`: light weight value profile (built on your `vp` ideas)
  - keep `control[]` to avoid full-map scans
  - Instead of only keeping the inputs producing the best distance for a given comparison site, we can configurably keep the best `K` inputs (count configurable via `AFL_VALUE_PROFILE_SLOTS` environment variable). This would enable the implementation handle loops/nested compares better while keeping the corpus under control
- `AFL_VALUE_PROFILE_LEVEL=2`: feature-rich path (current libFuzzer-style VP)
  - keep feature extraction/virgin feature tracking from compare operands
  - use this when we want stronger guidance even if it has more overhead

We can make `AFL_VALUE_PROFILE_LEVEL=1` the default option if value profile is active and keep the other one as opt-in.

The integration would keep your performance/efficiency optimizations and anti-corpus-bloat behavior, while preserving the richer guidance mode when needed. WDYT?

I have an intial implementation of this integration as described above and I can refine and clean it up to submit if we agreed on the approach.

vanhauser-thc:
I would prefer this to be a command line parameter, same as -l for cmplog.

two levels could make sense. Once everything is ready we can do some extensive testing for optimal values

vanhauser-thc:
tell me if you have it ready in a stage to perform testing :-)
(and note the merge conflicts)
thanks!

kyakdan:
I'm on the final stages refining the level 1 and we can perform testing/benchmarking shortly. I'll let you know when it's ready. I'll rebase on `dev` when finalizing the PR.

kyakdan:
@vanhauser-thc I seem to have closed this PR by mistake. Can you reopen it? I cannot do that.

kyakdan:
Nevermind, I'll open a new one

PR #2713, opened 2026-02-14
LTO perf regression

### LINKED ISSUE #2399: afl-clang-lto is 10x slower for some target
**Describe the bug**
`afl-clang-lto` might be much slower than normal `clang -flto=full` to link.

**To Reproduce**
```
git clone https://github.com/tstack/lnav.git
cd lnav
./autogen.sh
CC='time afl-clang-lto' CXX='time afl-clang-lto++' AR=llvm-ar-19 RANLIB=llvm-ranlib-19 ./configure
# use -j to reduce time before linking
make V=1 -j32
```
The last step, linking `src/lnav`, takes 40 minutes.

**Expected behavior**
Without AFL, 
```
CC='time clang-19' CXX='time clang++-19 AR=llvm-ar-19 RANLIB=llvm-ranlib-19 CFLAGS='-O3 -flto=full' CXXFLAGS='-O3 -flto=full' LDFLAGS='-O3 -flto=full' ./configure
make V=1 -j32
```
The last step takes only 4 minutes. In other words, AFL LTO mode is 10x slower.

I am not sure this is reasonable or not. Feel free to close this issue.

**Additional context**
I have tested clang 19.1.7 and 20.1.3

nbars:
This hoists the map ptr to the function level and uses an intrinsic for saturating addition instead of a manual implementation. This reduced the LTO compile time for `lnav` (see #2399) to 5 minutes for me. Without the patch, building took +20 minutes (canceled the build at that point).

The improvements have been applied to LTO and PCGUARD, as a drive-by, I refactored code shared between the two passes.

Fixes #2399

### DIFF
--- a/instrumentation/SanitizerCoverageLTO.so.cc
+++ b/instrumentation/SanitizerCoverageLTO.so.cc
@@ -205,17 +205,6 @@ class ModuleSanitizerCoverageLTO
   void InjectCoverageAtBlock(Function &F, BasicBlock &BB, size_t Idx,
                              bool IsLeafFunc = true);
 
-  void SetNoSanitizeMetadata(Instruction *I) {
-
-#if LLVM_VERSION_MAJOR >= 19
-    I->setNoSanitizeMetadata();
-#else
-    I->setMetadata(I->getModule()->getMDKindID("nosanitize"),
-                   MDNode::get(*C, None));
-#endif
-
-  }
-
   std::string    getSectionName(const std::string &Section) const;
   FunctionCallee SanCovTracePC /*, SanCovTracePCGuard*/;
   Type *IntptrTy, *IntptrPtrTy, *Int64Ty, *Int64PtrTy, *Int32Ty, *Int32PtrTy,
@@ -264,6 +253,7 @@ class ModuleSanitizerCoverageLTO
   GlobalVariable                  *AFLIJONState = NULL;
   const char                      *ijon_enabled = nullptr;
   Value                           *MapPtrFixed = NULL;
+  Value                           *FuncMapPtr = NULL;
   AllocaInst                      *CTX_add = NULL;
   std::ofstream                    dFile;
   size_t                           found = 0;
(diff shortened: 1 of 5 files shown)


nbars:
This also seems to have affected normal PCGUARD mode. Compile time went down from +8 minutes (canceled there) to 2min 17s. I actually encountered specific targets that build forever, so this probably fixed that as well.

instrumentation/SanitizerCoverageLTO.so.cc
```diff
@@ -2370,21 +2356,21 @@ void ModuleSanitizerCoverageLTO::InjectCoverageAtBlock(Function   &F,
...
+      setNoSanitizeMetadata(Counter);
 
+      Value *Incr;
       if (skip_nozero == NULL) {
 
-        auto cf = IRB.CreateICmpEQ(Incr, Zero);
-        markAflSkip(cf);
-        auto carry = IRB.CreateZExt(cf, Int8Tyi);
-        Incr = IRB.CreateAdd(Incr, carry);
+        Incr = IRB.CreateBinaryIntrinsic(Intrinsic::uadd_sat, Counter, One);
```
vanhauser-thc: this is wrong - you are switching from 
`if (++counter == 255) counter = 1` to `if (counter < 255) ++counter`
which is something completely different and results in worse coverage (this applies two all three passes).

vanhauser-thc:
cool optimization!
please note that you also introduced a major regression.

instrumentation/SanitizerCoverageLTO.so.cc
nbars: Well, expected this to be for avoiding the 0 case only. Unfortunately, this was the main cause of the perf regression. I will check if Icmp works better.

nbars:
Replaced the saturating add with a umax, yielding similar results. Emitted assembler also looks fine. Please check :)

instrumentation/SanitizerCoverageLTO.so.cc
```diff
@@ -1871,6 +1856,19 @@ void ModuleSanitizerCoverageLTO::instrumentFunction(
...
   };
 
+  /* Set up FuncMapPtr before the select/switch instrumentation loop,
+     because updateBitmapForResult uses it.  InjectCoverage (called later)
+     will reuse the same value. */
+  if (map_addr) {
+
+    FuncMapPtr = MapPtrFixed;
+
+  } else if (AFLMapPtr) {
```
vanhauser-thc: why does it need an `if`? and if it needs this, doesn't it then also need an `else`?

vanhauser-thc:
I think the umax is a better solution, yes.
please have a look at my code comment if it is relevant.
then I can merge it :-)
thanks!

instrumentation/SanitizerCoverageLTO.so.cc
nbars: replaced with else

vanhauser-thc:
thanks!

PR #2715, opened 2026-02-14
Rate-limit cycle-boundary syncing in main to reduce overhead for fast targets

martinus:
# Your pull request

**Type of PR**: Enhancement
**Purpose of this PR**

I somewhat arbitrarily choose 10 seconds as the minimum time between two syncs. In `benchmark.py` I see an improvement of about 2.34% on my AMD Epyc for multicore. No change beyond noise in singlecore.

|CPU                                                 | MHz   | threads | singlecore | multicore | afl-*-config |
|----------------------------------------------------|-------|---------|------------|-----------|--------------|
|AMD EPYC 9354P 32-Core Processor (before this PR) | 3802  | 64      | 42944      | 1384121   | system       |
|AMD EPYC 9354P 32-Core Processor (after this PR) | 3800  | 64      | 42982      | 1416570   | system       | 

I have also tried an implementation based on `sync_time_us` which works as well, but I find the hardcoded 10 second simpler:

```cpp
/* Rate-limit cycle-boundary syncing to avoid overhead with fast
   targets that complete queue cycles very quickly. Sync time shouldn't
   take more than 5% of the runtime. */
if (unlikely(afl->start_time * 1000 >
             afl->last_sync_time * 1000 + afl->sync_time_us * 20)) {
  sync_fuzzers(afl);
}
```

## IMPORTANT

### DIFF
--- a/src/afl-fuzz.c
+++ b/src/afl-fuzz.c
@@ -568,6 +568,30 @@ static void fasan_check_afl_preload(char *afl_preload) {
 
 }
 
+/* Throttle syncs by `sync_time` and `sync_interval_cnt`. Pass NULL for
+   sync_interval_cnt to only limit by sync_time. Main node sync time is half of
+   secondary nodes, and a third of SYNC_INTERVAL
+ */
+static void maybe_sync_fuzzers(afl_state_t *afl, u64 cur_time,
+                               u32 *sync_interval_cnt) {
+
+  u64 sync_time = afl->is_main_node ? afl->sync_time >> 1 : afl->sync_time;
+
+  if (unlikely(cur_time > sync_time + afl->last_sync_time)) {
+
+    u32 sync_interval = afl->is_main_node ? SYNC_INTERVAL / 3 : SYNC_INTERVAL;
+
+    if (NULL == sync_interval_cnt ||
+        !((*sync_interval_cnt)++ % sync_interval)) {
+
+      sync_fuzzers(afl);
+
+    }
+
+  }
+
+}
+
 /* Main entry point */
 
 int main(int argc, char **argv_orig, char **envp) {
@@ -3365,7 +3389,8 @@ int main(int argc, char **argv_orig, char **envp) {
 
         }
 
-        sync_fuzzers(afl);
+        /* sync only based on sync_time, not sync_interval_cnt */
+        maybe_sync_fuzzers(afl, get_cur_time(), NULL);
 
       }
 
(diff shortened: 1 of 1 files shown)


src/afl-fuzz.c
```diff
@@ -3365,7 +3365,14 @@ int main(int argc, char **argv_orig, char **envp) {
 
         }
 
-        sync_fuzzers(afl);
+        /* Rate-limit cycle-boundary syncing to avoid overhead with fast
+           targets that complete queue cycles very quickly. Sync at most
+           once every 10 seconds. */
+        if (unlikely(afl->start_time > afl->last_sync_time + 10000)) {
```
vanhauser-thc: don't you mean something like `curtime > afl->last_sync_time + 10000` 
And I think 5 minutes as a minimum makes more sense, or even 10 minutes - not seconds.

martinus: I used `afl->start_time` because that is set in line 3346, just a few lines above. With curtime do you mean to call `get_cur_time()` instead? That's fine by me, I just wanted to safe a call to that in a possibly tight loop

5 minutes sounds a bit much for me, at least how I use the fuzzer :slightly_smiling_face: I tend to start running all my fuzz tests with `AFL_EXIT_ON_TIME` of 1 minute, and once they all ran through I increase their runtime by a factor of 2 until I reach 60 minutes

martinus: I think the  most consistent would be actually to use `AFL_SYNC_TIME` / 2, that would then be 10 minutes by default, it's configurable, and that's also how its described in `env_variables.md`:

```md
  - `AFL_SYNC_TIME` allows you to specify a different minimal time (in minutes)
    between fuzzing instances synchronization. Default sync time is 20 minutes,
    note that time is halved for -M main nodes.
```

vanhauser-thc: Yes I agree

vanhauser-thc:
so this PR now transforms code that was inline into a specific function and otherwise there is not change?

martinus:
> so this PR now transforms code that was inline into a specific function and otherwise there is not change?

The functional change is that the sync on cycle boundaries at the beginning of the loop is now gated with the last sync time, for the other sync it behaves exactly as before.

I now saw what you've meant with `afl->start_time`, that was a bug on my side... I now directly call `get_cur_time()` instead.

vanhauser-thc:
thanks!

PR #2717, opened 2026-02-16
Fix two PCGUARD sub-block coverage bugs for atomic instructions

kyakdan:
# Summary
The current LLVM pass has two bugs in handling atomic instructions:
- `AtomicCmpXchgInst` was incorrectly gated by `isIntegerTy(1)`, so cmpxchg sub-block instrumentation was skipped.
- `AtomicRMWInst` min/max/umin/umax was incorrectly gated in counting, causing guard under-allocation while instrumentation still consumed guard slots.

The root cause for both issues came from type checks that seems to have been copied from scalar compare handling (`i1`), but:
- `cmpxchg` returns `{T, i1}` (struct), not `i1`
- `atomicrmw` returns the old operand type (e.g. `i32`), not `i1`

The fix was to 1) remove invalid type gating for `cmpxchg` in counting and instrumentation, and 2) remove invalid type gating for `atomicrmw` in counting so guard sizing matches emitted instrumentation.

### DIFF
--- a/instrumentation/SanitizerCoveragePCGUARD.so.cc
+++ b/instrumentation/SanitizerCoveragePCGUARD.so.cc
@@ -1049,17 +1049,10 @@ bool ModuleSanitizerCoverageAFL::InjectCoverage(
 
         } else if (cxchg) {
 
-          if (cxchg->getType()->isIntegerTy(1)) {
-
-            block_is_instrumented = true;
-            cnt_sel++;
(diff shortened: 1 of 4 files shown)


vanhauser-thc:
Thanks!
That was when I attempted use codex to write code. in hindsight - it is generally good, but makes quite a lot of mistakes so it does not really save time - better to code it yourself....

Issue #2723, opened 2026-02-17
Compiler error with cmplog and `annotate` attribute


ngg:
**Describe the bug**
If an `annotate` attribute is present on a struct field, and I compile with `AFL_CMPLOG=1`, it fails with an error `Undefined temporary symbol .L.str`.

Example C++ code:

```cpp
struct S {
    __attribute__((annotate("dummy"))) int x = 0;
};

void f() {
    S s;
}
```

**To Reproduce**
1. Create x.cc with the contents above
2. Try to compile with `AFL_CMPLOG=1 afl-clang-fast++ -c x.cc`

**Expected behavior**
It should compile fine, but it fails with an error like this:

```txt
$ AFL_CMPLOG=1 afl-clang-fast++ -c x.cc
afl-cc++4.21c by Michal Zalewski, Laszlo Szekeres, Marc Heuse - mode: LLVM-PCGUARD
Running split-switches-pass by laf.intel@gmail.com
SanitizerCoveragePCGUARD++4.21c
[+] Instrumented 2 locations with no collisions (non-hardened mode) of which are 0 handled and 0 unhandled selects.
Running cmplog-switches-pass by andreafioraldi@gmail.com
Running cmplog-instructions-pass by andreafioraldi@gmail.com
Running cmplog-routines-pass by andreafioraldi@gmail.com
error: Undefined temporary symbol .L.str
1 error generated.
```

**Additional context**
I could reproduce on Debian 13 with clang and AFL++ installed with apt, and also the current `dev` branch is affected.

kyakdan:
The root cause of the problem is that `__attribute__((annotate("dummy")))` makes Clang emit an LLVM IR like:
```
@.str = private unnamed_addr constant [6 x i8] c"dummy\00", section "llvm.metadata"
%12 = call ptr @llvm.ptr.annotation.p0.p0(ptr %11, ptr @.str, ...)
```
The call to ` @llvm.ptr.annotation.p0.p0` matches the `isPtrRtn` heuristic, which results in the pass instrumenting this call, creating `__cmplog_rtn_hook(@.str, ...)` . However, `@.str` lives in section `"llvm.metadata"`, a special section that the LLVM backend doesn't emit into the final object. The assembler sees a reference to `.L.str`  but the symbol was never emitted

kyakdan:
@ngg this should fix the issue https://github.com/AFLplusplus/AFLplusplus/pull/2725

[...]

PR #2725, opened 2026-02-17
fix CmpLog compilation failure with annotate attribute

kyakdan:
Skip LLVM intrinsics in cmplog-routines-pass to prevent instrumenting `llvm.ptr.annotation` calls whose string args live in llvm.metadata and cannot be referenced by emitted code.

Fixes https://github.com/AFLplusplus/AFLplusplus/issues/2723


vanhauser-thc:
Hey thanks for this!
There are more instances where `getCalledFunction` is used, I will check them and add the intrinsic check as well.

PR #2730, opened 2026-02-20
Fix afl-cmin qemu mode

b1gr3db:
# Fix afl-cmin qemu mode

The c implementation for `afl-cmin` fails to convert the target arguments for qemu mode, resulting in it never using `afl-qemu-trace` and failing to run.

**Type of PR**:  Bugfix
**Purpose of this PR**:  Fix qemu mode in `afl-cmin`


vanhauser-thc:
this is missing more than just qemu ... thanks
fyi @kcwu

Issue #1716, opened 2023-04-26
Clang + shared ASAN + AFL++ is incompatible


maxammann:
## Describe the bug

The way AFL++ defines the weak [__asan_region_is_poisoned](https://github.com/AFLplusplus/AFLplusplus/blob/7ca1b85c5e8229fa49620d0fb542c86965ef5abb/instrumentation/afl-compiler-rt.o.c#L1999-L2003) function is incompatible with shared ASAN.

Shared ASAN is default in GCC, but not in Clang, so this issue probably does not affect a lot of users. Though, it is also present in LibAFL, which also exports [__asan_region_is_poisoned](https://github.com/AFLplusplus/LibAFL/blame/main/libafl_targets/src/cmplog.c#L30-L36).

## To Reproduce

The following experiments are performed on Debian and NOT on Ubuntu. See below for more information.

Suppose the following harness, which copies from outside an allocated region to some other buffer (`y`).

```c
#include <string.h>
#include <stdio.h>
void main() {
  char *x = (char*)malloc(5 * sizeof(char));
  memset(x, 0, 5);
  char *y = (char*)malloc(10 * sizeof(char));
  memcpy(y, x, 10);
  free(x);
  free(y);
}
```

Compiling and executing this with ASAN through AFL++ should crash. And it is true in the default configuration.

| Command      | Corruption detected? |
| ----------- | ----------- |
| `AFL_USE_ASAN=1 afl-clang-fast harness.c -o harness -O0 && ./harness`      | ✅ |

Doing the same with LLVM's shared ASAN **on Debian** does not work. (Static ASAN is default in Clang)

| Command      | Corruption detected? |
| ----------- | ----------- |
| `AFL_USE_ASAN=1 afl-clang-fast harness.c -o harness -shared-libasan -O0 && LD_PRELOAD=$(clang -print-file-name=libclang_rt.asan-x86_64.so) ./harness`   | ❌ |

The reason for this is that `libclang_rt.asan-x86_64.so` exports `__asan_region_is_poisoned` and makes it relocatable on Debian. On Ubuntu the symbol is exported but not relocatable, and therefore the correct implementation from `libclang_rt.asan-x86_64.so` will be chosen.

With GCC shared ASAN is the default. And it works with static and shared ASAN.

| Command | Corruption detected? |
| ----------- | ----------- |
| `AFL_USE_ASAN=1 afl-gcc harness.c -o harness -O0 && ./harness`   |  ✅ |
[...]

PR #2735, opened 2026-02-22
Support shared ASAN by only using weak declarations without definitions

Scott-Guest:
**Type of PR**: Bugfix/Enhancement
**Purpose of this PR**: Provide an alternative fix to #1716 to support shared ASAN

The root of #1716 is that, with dynamic linking, AFL++'s `__attribute__((weak))` no-op ASAN stubs still end up interposing the real ASAN interceptors. The workaround added in #1966 was to just force `-static-libasan` for GCC where shared ASAN is the default.

This PR removes that workaround, and instead fixes the issue by replacing the weak definitions with weak declarations. This makes the symbols `STB_WEAK` + `SHN_UNDEF`, so the dynamic linker either resolves the real ones from `libasan.so` or leaves them `NULL` (which each callsite must check).

You can replicate the original issue and test the fix by running the following Dockerfile:

```
FROM debian:trixie
RUN apt-get update && apt-get install -y --no-install-recommends \
        build-essential clang llvm-dev libclang-rt-dev \
        gcc-14-plugin-dev \
        python3 git ca-certificates \
    && rm -rf /var/lib/apt/lists/*

RUN git clone --depth=1 --branch stable \
        https://github.com/AFLplusplus/AFLplusplus.git /aflpp-upstream && \
    cd /aflpp-upstream && make -j$(nproc) source-only NO_NYX=1 2>&1 | tail -1
RUN git clone --depth=1 --branch shared-asan \
        https://github.com/Scott-Guest/AFLplusplus.git /aflpp-fixed && \
    cd /aflpp-fixed && make -j$(nproc) source-only NO_NYX=1 2>&1 | tail -1

RUN cat > /harness.c <<'EOF'
#include <string.h>
#include <stdio.h>
#include <stdlib.h>
int main() {
  char *x = (char*)malloc(5 * sizeof(char));
  memset(x, 0, 5);
  char *y = (char*)malloc(10 * sizeof(char));
  memcpy(y, x, 10);
  free(x);
  free(y);
  return 0;
}
EOF

RUN cat > /test.sh <<'EOF'
#!/bin/bash
ASAN_LIB=$(clang -print-file-name=libclang_rt.asan-$(uname -m).so)

echo '=== upstream (afl-clang-fast) ==='
AFL_USE_ASAN=1 /aflpp-upstream/afl-clang-fast /harness.c -o /tmp/harness -shared-libasan -O0 2>/dev/null
LD_PRELOAD=$ASAN_LIB /tmp/harness

echo
echo '=== fixed (afl-clang-fast) ==='
AFL_USE_ASAN=1 /aflpp-fixed/afl-clang-fast /harness.c -o /tmp/harness -shared-libasan -O0 2>/dev/null
LD_PRELOAD=$ASAN_LIB /tmp/harness

echo
echo '=== fixed (afl-gcc-fast) ==='
AFL_USE_ASAN=1 /aflpp-fixed/afl-gcc-fast /harness.c -o /tmp/harness -O0 2>/dev/null
/tmp/harness
EOF
RUN chmod +x /test.sh

CMD ["/test.sh"]
```

### DIFF
--- a/instrumentation/afl-compiler-rt.o.c
+++ b/instrumentation/afl-compiler-rt.o.c
@@ -2694,18 +2694,19 @@ void __sanitizer_cov_trace_switch(uint64_t val, uint64_t *cases) {
 
 }
 
-__attribute__((weak)) void *__asan_region_is_poisoned(void *beg, size_t size) {
-
-  return NULL;
-
-}
+__attribute__((weak)) void *__asan_region_is_poisoned(void *beg, size_t size);
 
 // POSIX shenanigan to see if an area is mapped.
 // If it is mapped as X-only, we have a problem, so maybe we should add a check
 // to avoid to call it on .text addresses
 static int area_is_valid(void *ptr, size_t len) {
 
-  if (unlikely(!ptr || __asan_region_is_poisoned(ptr, len))) { return 0; }
+  if (unlikely(!ptr || (__asan_region_is_poisoned &&
+                        __asan_region_is_poisoned(ptr, len)))) {
+
+    return 0;
+
+  }
 
 #ifdef __HAIKU__
   long r = _kern_write(__afl_dummy_fd[1], -1, ptr, len);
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
thanks, looks good IMHO

PR #2737, opened 2026-02-23
Fix out-of-bounds read in to_base64() and rename misleading parameter

Jay-1409:
**Type of PR**: Bugfix
**Purpose of this PR**: Fix bounds checking issue in `to_base64` function which could lead to a out of bounds.

##Demonstration 
Lets keep the inputs `u8 input[] = "A"; ` and using the original `to_base64()` funciton, we end up in with the error 
image

Lets change the function to the updated version
we get this output 
image

you can test it yourself here: https://leetcode.com/playground/WRco59B8
toggle between them and try!
image


Jay-1409:
I have also updated the misleading variable names. 

**Meaning of the parameters**
src -> source buffer (input)
dst -> destination buffer (output)
 src_len -> source buffer length

vanhauser-thc:
Thanks for the fix. Please direct all future PRs to the dev branch, fixed that for you

PR #2740, opened 2026-02-27
Define _GNU_SOURCE in afl-cmin.c so that it compiles with older glibc

Scott-Guest:
**Type of PR**: Enhancement
**Purpose of this PR**: Define _GNU_SOURCE in afl-cmin.c so that it compiles with older glibc

`afl-cmin.c` uses `O_DIRECTORY`, which is only exposed when you `#define _GNU_SOURCE` for glibc 2.11 and earlier. 

(Obviously this is a very old version, but I am stuck using it at the moment :smile:, and this is the only AFL++ build issue I've run into with it.)

### DIFF
--- a/src/afl-cmin.c
+++ b/src/afl-cmin.c
@@ -27,6 +27,10 @@
 #define AFL_MAIN
 #define AFL_CMIN
 
+#ifndef _GNU_SOURCE
+  #define _GNU_SOURCE
+#endif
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
Thanks! Note that afl-cmin is currently replaced with the python script as it is not mature enough yet

PR #2741, opened 2026-02-27
Respect LDFLAGS when testing CC in GNUmakefile

Scott-Guest:
**Type of PR**: Bugfix
**Purpose of this PR**: Add missing LDFLAGS, fix a build failure when STATIC=1 without zlib

The GNUmakefile compiles a few test programs to test which features CC supports. These should respect LDFLAGS, e.g. so that any `-L` will be searched when checking for zlib with `-lz`.

Separately, also remove the hard-coded `-lz` when `STATIC=1`. It's already inserted by the zlib check elsewhere, and this makes the STATIC=1 build fail unnecessarily if zlib isn't present.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -141,7 +141,7 @@ ifdef STATIC
   PYTHON_INCLUDE = /
 
   override CFLAGS_OPT += -static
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
Thanks!

PR #2760, opened 2026-03-14
libqasan support arm thumb hotpatch

gursegal:
## libqasan support arm thumb hotpatch

**Type of PR**: Enhancement
**Purpose of this PR**: 
in libqasan hotpatching arm libc currently assumes that the patched functions are compiled in ARM mode. However, on some ARM systems the libc implementation compiles functions in Thumb mode. Because the current patching logic does not account for this, patching these functions fails and libqasan does not work correctly in those environments.

Adding support for handling such cases is fairly easy, by checking the patched function address to determine whether the target function is in ARM or Thumb mode. Based on this check, the patching logic adjusts accordingly so the instrumentation is applied correctly.

the patched jump was modified to `bx r12` to explicitly allow jumping between arm and thumb.


vanhauser-thc:
thanks!

PR #2766, opened 2026-04-03
Fix unicorn build script for some edge cases

rmalmain:
this is needed to fix `libafl-fuzz` build in the CI of libafl.

it fixes 2 issues:
- try to get cargo target path from `CARGO_TARGET_DIR` if it's defined. otherwise there is a mismatch with the hardcoded path in the script.
- copy build lib only if it exists. if unicorn is installed on the system, the build is skip so no static lib is built.


wtdcode (approved):
LGTM.

PR #2774, opened 2026-04-16
CmpLog: bound routine string hook reads

kyakdan:
Fixes the routine string CmpLog hooks so they validate the readable operand range before calling `strnlen()`.

Previously, `__cmplog_rtn_hook_str()` and `__cmplog_rtn_hook_strn()` could call `strnlen()` across a page boundary before checking whether the full range was readable. If an operand pointed near the end of a mapped page and the next page was inaccessible, the hook could crash the target.

The hooks now first use `area_is_valid()` to determine the readable prefix, cap string scanning to the readable range shared by both operands, and preserve the existing bounded copy behavior for routine CmpLog entries.

### DIFF
--- a/instrumentation/afl-compiler-rt.o.c
+++ b/instrumentation/afl-compiler-rt.o.c
@@ -2783,8 +2783,13 @@ static u8 get_prog_addr_attr(const void *addr) {
 
 #endif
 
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
sounds plausible, thanks

PR #2787, opened 2026-05-13
ijon: add terminal retirement support for IJON max entries

nj00001:
Introduce `UINT64_MAX` as a terminal/retired sentinel for IJON max guidance entries.

Add:

```c
IJON_MAX_UNTIL(x, limit)
IJON_MAX_UNTIL_AT(addr, x, limit)
```

These helpers guide the fuzzer to maximize a value until a target threshold is reached. Once the encoded value becomes `UINT64_MAX`, the corresponding IJON max entry is considered completed and is removed from the IJON scheduling pool.

This avoids repeatedly scheduling seeds for already-satisfied progress dimensions, reducing wasted fuzzing cycles.

Implementation details:

* reuse `max_map[i] == UINT64_MAX` as retired state
* stop storing/scheduling retired IJON max inputs
* delete persisted seed files for retired entries
* preserve existing shared-memory ABI
* no changes to `ijon_map`
* no changes to `MAP_SIZE_IJON_BYTES`
* no forkserver protocol changes

The encoding maps `x >= limit` directly to `UINT64_MAX`, making the terminal condition naturally representable without additional metadata.


nj00001:
In summary, two main features have been added:

1. When the indicator variable `IJON_MAX_UNTIL` grows to the limit, it means that the ijon annotation has guided `x` to reach the threshold, and subsequent fuzzing will no longer consider growth of `x` greater than the limit value.

2. The environment variable `AFL_IJON_RETIRE_MAX` will disable ijon entries in `IJON_MAX` that reach `u64_max` to improve fuzzing efficiency. If all `ijon_max` values ​​reach `u64_max`, no further ijon mutations will be performed.

vanhauser-thc:
Can you please provide a test that ensures this works? A test/test-ijon-retire.sh that exits 0 on success and 1 on error?

vanhauser-thc:
thanks!

PR #2788, opened 2026-05-14
make inplace initialization for SAND abstractions mods

koltiradw:
The previous code led to null pointer dereference, because the initialization of the map for SAND mods in the code is earlier than the definition of SAND mods.


vanhauser-thc:
thanks!

PR #2794, opened 2026-05-22
Add FreeBSD build support to AFL++.

jsaunders-rr:
**Type of PR**: Enhancement
**Purpose of this PR**: This adds basic support for building and running AFL on FreeBSD 15. The basic instrumentation (afl-cc) and fuzzing (afl-fuzz) tools work. Only the basic "make all" target works at the moment, though, but unsupported functionality is disabled by default. (Unicorn Engine does not run on FreeBSD 15 X86-64.) All code should be built as before on other targets with no changes.
**I have tested the changes**: Yes. Regression testing (make test) has been also performed on Linux (Ubuntu 24.04). Testing on FreeBSD has been more manual due to the lack of docker, but has worked in our test/use cases.

### DIFF
--- a/utils/argv_fuzzing/Makefile
+++ b/utils/argv_fuzzing/Makefile
@@ -27,6 +27,12 @@ _LDFLAGS_ADD=$(UNAME_SAYS_LINUX:1=)
 LDFLAGS_ADD=$(_LDFLAGS_ADD:0=-ldl)
 LDFLAGS  += $(LDFLAGS_ADD)
 
+# Needed for backtrace under FreeBSD
+ifeq "$(SYS)" "FreeBSD"
+    LDFLAGS += -lexecinfo
+	COMPILER_RT_LDFLAGS += -lexecinfo
+endif
+
 # on gcc for arm there is no -m32, but -mbe32
 M32FLAG = -m32
 M64FLAG = -m64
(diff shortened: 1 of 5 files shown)


utils/argv_fuzzing/Makefile
```diff
@@ -55,10 +61,18 @@ install: argvfuzz32.so argvfuzz64.so
 	if [ -f argvfuzz64.so ]; then set -e; install -m 755 argvfuzz64.so $(DESTDIR)$(HELPER_PATH)/; fi
 
 argv_fuzz_persistent_demo: argv_fuzz_persistent_demo.c
-	../../afl-cc -g -o $@ $^
+	ifeq "$(SYS)" "FreeBSD"
+		../../afl-cc -lexecinfo -g -o $@ $^
```
vanhauser-thc: as you add this flag in the afl-cc patch above, all these makefile changes should not be necessary anymore, right?

vanhauser-thc: (where afl-cc is used)

vanhauser-thc:
@jsaunders-rr ping ... :-)

utils/argv_fuzzing/Makefile
jsaunders-rr: Correct. That was an artifact of an earlier, hackier version of getting it to build on FreeBSD. Removed.

vanhauser-thc:
thanks!

PR #2796, opened 2026-05-24
afl-fuzz: keep slow calibration runs in timeout sizing

### LINKED ISSUE #1545: When execution time of fork run is much longer than one persistent iteration, all fork execuions will timeout.
**Describe the bug**
In persistent mode, there are 2 ways to execute a test case:

1. When child process is not created, fork server will call `fork` to create a process to execute the test case.
2. When child process is created but being stopped at `__afl_persistent_loop`, fork server will send a `SIGCONT` signal to resume the process to execute the test case.

In general, forking a process should be slower than resuming the loop, and this is the reason why persistent mode is designed. For example, in `aflpp_driver.c`, before entering the persistent loop, 2 functions, `LLVMFuzzerTestOneInput` and `__asan_poison_memory_region`, are called, plus the overhead of `fork`: 

```c
__afl_manual_init();

// Call LLVMFuzzerTestOneInput here so that coverage caused by initialization
// on the first execution of LLVMFuzzerTestOneInput is ignored.
LLVMFuzzerTestOneInput(dummy_input, 4);

__asan_poison_memory_region(__afl_fuzz_ptr, MAX_FILE);
size_t prev_length = 0;

// for speed only insert asan functions if the target is linked with asan
if (__asan_region_is_poisoned) {

  while (__afl_persistent_loop(N)) { /*...*/ }

} else {

  while (__afl_persistent_loop(N)) {

    LLVMFuzzerTestOneInput(__afl_fuzz_ptr, *__afl_fuzz_len);

  }

}
```

However, if fork execution is so slow such that its execution time is longer than timeout limit (which is calculated from initial seed corpus), all following execution will timeout. The reason is that as long as one timeout occurs, the process executing the test case will be killed so the next execution must fork a new process, which results in timeout again and forever. This is not desirable.

**To Reproduce**

The following code can trigger the problem. Although this case seems to be very handcrafted, this problem do occur in real world fuzzing, which is the reason why I found this issue. 

```c
#include <memory.h>
#include <stdint.h>
#include <unistd.h>

int LLVMFuzzerTestOneInput(const uint8_t *input, size_t r)
{
	if (r == 4 &&
		input[0] == '#' && input[1] == '#' &&
		input[2] == 'S' && input[3] == 'I')
	{ // make `LLVMFuzzerTestOneInput(dummy_input, 4);` stuck
		sleep(1);
	}
	return 0;
}
```

**My Solution**

The reason why timeout limit can be lower than time required to execute fork run is that when calculating `exec_us` of a seed in calibration, average time of all calibration executions is calculated. If only one of the execution uses `fork` and others are just resuming persistent loop, the average value will be much lower than time of `fork` execution, so when calculating `max_us`, it can be lower than time of `fork` execution. Therefore, instead of using average time, we get a max time for each seed and there must be one of them being the time of `fork` execution.

https://github.com/AFLplusplus/AFLplusplus/pull/1544

officialasishkumar:
# Your pull request

**Type of PR**: Bugfix
**Purpose of this PR**: Fixes #1545 by tracking the slowest successful calibration execution separately from the average execution time, then using that slowest run when AFL++ derives timeout limits. This keeps persistent-mode fork/init costs from being averaged away by faster loop iterations.

Validation:
- `make -j2 NO_NYX=1 NO_QEMU=1 NO_UNICORN=1 NO_FRIDA=1 all`
- `test/test-persistent-timeout.sh`

## IMPORTANT

### DIFF
--- a/src/afl-fuzz.c
+++ b/src/afl-fuzz.c
@@ -3316,10 +3316,17 @@ int main(int argc, char **argv_orig, char **envp) {
 
       u64 max_ms = 0;
 
-      for (entry = 0; entry < afl->queued_items; ++entry)
-        if (!afl->queue_buf[entry]->disabled)
-          if ((afl->queue_buf[entry]->exec_us / 1000) > max_ms)
-            max_ms = afl->queue_buf[entry]->exec_us / 1000;
+      for (entry = 0; entry < afl->queued_items; ++entry) {
+
+        struct queue_entry *q = afl->queue_buf[entry];
+        u64                 q_max_us;
+
+        if (q->disabled) { continue; }
+
+        q_max_us = MAX(q->exec_us, q->exec_us_max);
+        if ((q_max_us / 1000) > max_ms) { max_ms = q_max_us / 1000; }
+
+      }
 
       // Add 20% as a safety margin, capped to exec_tmout given in -t option
       max_ms *= 1.2;
--- a/include/afl-fuzz.h
+++ b/include/afl-fuzz.h
@@ -288,7 +288,8 @@ struct queue_entry {
       fuzz_level,                       /* Number of fuzzing iterations     */
       n_fuzz_entry;                     /* offset in n_fuzz                 */
 
-  u64 exec_us,                          /* Execution time (us)              */
+  u64 exec_us,                          /* Average execution time (us)      */
+      exec_us_max,                      /* Slowest calibration run (us)     */
       handicap,                         /* Number of queue cycles behind    */
       depth,                            /* Path depth                       */
       exec_cksum,                       /* Checksum of the execution trace  */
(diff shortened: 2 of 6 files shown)


src/afl-fuzz.c
```diff
@@ -3316,10 +3316,17 @@ int main(int argc, char **argv_orig, char **envp) {
...
-      for (entry = 0; entry < afl->queued_items; ++entry)
-        if (!afl->queue_buf[entry]->disabled)
-          if ((afl->queue_buf[entry]->exec_us / 1000) > max_ms)
-            max_ms = afl->queue_buf[entry]->exec_us / 1000;
+      for (entry = 0; entry < afl->queued_items; ++entry) {
+
+        struct queue_entry *q = afl->queue_buf[entry];
+        u64                 q_max_us;
+
+        if (q->disabled) { continue; }
```
vanhauser-thc, line 3324: this should be unlikely() and moved just after the struct queue_entry for a bit of extra speed

vanhauser-thc:
I don't think this is the right patch for the problem.
It just coincidences that sometimes such a problem will not happen with this PR.
the problem has nothing to do with a per seed calibration.

the problem is that when a new process is forked (targt crashed, timed out, or the loop count ran out), then a new instance is forked and that will run much longer than a normal input. this is the case that needs to be handled.

vanhauser-thc:
@officialasishkumar  ping

PR #2812, opened 2026-06-13
bugfix(forkserver): memory leak when used with persistence mode

Jay-1409:
**Type of PR**: Bugfix
**Purpose of this PR**: There seemed to be a memory leak in the fork server.
**I have tested the changes**: yes i also ran `make code-fromat`


Jay-1409:
I am trying to resolve the merge conflict, but GitHub is lagging for some reason and gets stuck at this point where its not taking in my commit. (I will try again in a while)

 i will also remove that log file.

src/afl-forkserver.c
```diff
@@ -2246,91 +1889,84 @@ void afl_fsrv_start(afl_forkserver_t *fsrv, char **argv,
...
   }
 
   fsrv->fsrv_pid = -1;
   fsrv->child_pid = -1;
+#ifdef AFL_PERSISTENT_RECORD
+  if (fsrv->persistent_record_data) {
+    for (u32 i = 0; i < fsrv->persistent_record; ++i) {
+      afl_free(fsrv->persistent_record_data[i]);
+    }
+    ck_free(fsrv->persistent_record_data);
```
Jay-1409, line 2294: This is a part of my changes

src/afl-forkserver.c
```diff
@@ -2973,40 +2512,33 @@ store_persistent_record: {
...
-
+  LIST_FOREACH(&fsrv_list, afl_forkserver_t, { afl_fsrv_kill(el); });
 }
 
 void afl_fsrv_deinit(afl_forkserver_t *fsrv) {
 
   afl_fsrv_kill(fsrv);
   list_remove(&fsrv_list, fsrv);
 
+#ifdef AFL_PERSISTENT_RECORD
```
Jay-1409, line 3029: this is also a part of my changes

vanhauser-thc: isnt this missing the _len one you have in the other change?

vanhauser-thc:
a) it is not a memleak, because the memory is not lost, and it is automatically freed at exit.
but it avoids a leaksan false positive. so I do not mind. please see my comment if you maybe have missed something.

b) are you on macos? because the clang formatter there is different than on linux, dunno why. please revert the formatting, as you can see it makes a lot of noise

Jay-1409:
> a) it is not a memleak, because the memory is not lost, and it is automatically freed at exit. but it avoids a leaksan false positive. so I do not mind. please see my comment if you maybe have missed something.
> 
> b) are you on macos? because the clang formatter there is different than on linux, dunno why. please revert the formatting, as you can see it makes a lot of noise

To address a
--

When we call `afl_fsrv_deinit`, `afl_fsrv_kill` is called inside it, so that piece of code runs once. Unless there is a gap in my understanding of the code, i think it is already taken care of. 
```C++
void afl_fsrv_deinit(afl_forkserver_t *fsrv) {

  afl_fsrv_kill(fsrv);        <------------------
  list_remove(&fsrv_list, fsrv);

#ifdef AFL_PERSISTENT_RECORD
  if (fsrv->persistent_record_dir) {

    ck_free(fsrv->persistent_record_dir);
    fsrv->persistent_record_dir = NULL;

  }

#endif

}
```

To address b
-- 

yup I am on macOS, I reformatted it, I just spun a container with Ubuntu and reformatted the code, I hope it's better now. [apon digging, i found out that the issue was with my clang-format version, i was running on some other version or probably some other format, and upon switching to the correct version `clang-format-18` it formats just like required.

vanhauser-thc:
Thanks

PR #2818, opened 2026-06-16
use correct precision for bfloat16 in split-compares pass

ngg:
**Type of PR**: Bugfix
**Purpose of this PR**: use correct precision for bfloat16 in split-compares pass

`splitFPCompares()` hard-codes 16-bit floats as IEEE half (`precision = 11`). `bfloat16` is also 16 bits but has precision 8 (8-bit exponent, 7-bit mantissa), so its masks, `NaN_lowend`, and field widths come out wrong.

Fixing calculation of the `precision` variable fixes all derived constants. The LLVM version guard is needed because `bfloat16` support only exists from LLVM 11.

### DIFF
--- a/instrumentation/split-compares-pass.so.cc
+++ b/instrumentation/split-compares-pass.so.cc
@@ -987,9 +987,10 @@ size_t SplitComparesTransform::splitFPCompares(Module &M) {
     const unsigned int precision = sizeInBits == 32    ? 24
                                    : sizeInBits == 64  ? 53
                                    : sizeInBits == 128 ? 113
-                                   : sizeInBits == 16  ? 11
-                                   : sizeInBits == 80  ? 65
-                                                       : sizeInBits - 8;
+                                   : sizeInBits == 16
+                                       ? (op0->getType()->isBFloatTy() ? 8 : 11)
+                                   : sizeInBits == 80 ? 65
+                                                      : sizeInBits - 8;
 
     const unsigned           shiftR_exponent = precision - 1;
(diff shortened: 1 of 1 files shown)


instrumentation/split-compares-pass.so.cc
```diff
@@ -987,9 +987,14 @@ size_t SplitComparesTransform::splitFPCompares(Module &M) {
     const unsigned int precision = sizeInBits == 32    ? 24
                                    : sizeInBits == 64  ? 53
                                    : sizeInBits == 128 ? 113
-                                   : sizeInBits == 16  ? 11
                                    : sizeInBits == 80  ? 65
-                                                       : sizeInBits - 8;
+                                   : sizeInBits == 16
+#if LLVM_MAJOR >= 11
```
vanhauser-thc: we only support LLVM 14+ :-)

vanhauser-thc:
good that someone looks at our float stuff :-)
left one comment

instrumentation/split-compares-pass.so.cc
ngg: I didn't know that, `afl-cc.c` contains `#if LLVM_MAJOR >= 3` too :) Removed the guard now.

vanhauser-thc: yes, originally we support 3.7, then only llvm 6.0, then 11, now the minimum is 14. there are still old relicts, will remove them at some point I guess, that is an easy AI task :-)

PR #2824, opened 2026-06-18
push to stable

vanhauser-thc:


src/afl-cc.c
```diff
@@ -178,7 +178,7 @@ typedef struct aflcc_state {
 
   u8 instrument_mode, instrument_opt_mode, ngram_size, ctx_k;
```
ittybittybit, line 179: Hi can i please find the info to do my taxes and returns 2026 i need some detail input to update my investment cost and loss and total Canada and US send me your information to my email addr; jolydaniel715@gmail.com or jolyd764@icloud.com please thank you

Issue #2832, opened 2026-06-26
Building libqasan for musl


celi0n:
I am fuzzing a binary linked to musl with Qasan. Therefore I need to build libqasan with gcc-musl, which results in the error:

`invalid application of 'sizeof' to incomplete type 'FILE'` at this line:
https://github.com/AFLplusplus/AFLplusplus/blob/edaf3ad04d64185cae856c75ca4676cbb14a35b5/qemu_mode/libqasan/hooks.c#L213

Apparently in musl, FILE is opaque per the C standard, only glibc exposes its full definition. 
The best patch I found for this issue would be to do:
```
#ifdef __GLIBC__
#define QASAN_FILE_CHECK_SIZE sizeof(FILE)
#else
#define QASAN_FILE_CHECK_SIZE sizeof(void *)
#endif
```
Not sure if the wrong size cause issue. AI says if it is lower, it wont cause false positive, but I don't know we can trust it.
What do you think ?

vanhauser-thc:
The patch would work. But instead if would rather like something

```
#ifdef __MUSL__
#define QASAN_FILE_CHECK_SIZE sizeof(void *)
#else
#define QASAN_FILE_CHECK_SIZE sizeof(FILE)
#endif
```
whatever the define for muslibc is ...
please send a PR

celi0n:
Unfortunately it looks like it doesn't exist:
https://catfox.life/2022/04/16/the-musl-preprocessor-debate/

Ok i will open a PR

vanhauser-thc:
The proposed fix was wrong and impacted all other non-glibc

leinn32:
Checking at it!

Chaitanya-970:
Hi, seeing no activity here recently, 

I'd like to work on this issue. Is it still open to take?

@vanhauser-thc

vanhauser-thc:
@Chaitanya-970  I do not think this is easily solvable.

PR #2836, opened 2026-06-29
Fix musl build for libqasan

celi0n:
# Your pull request

**Type of PR**: Bugfix
**Purpose of this PR**: Fix musl build for libqasan (#2832)


qemu_mode/libqasan/hooks.c
```diff
@@ -203,14 +203,20 @@ void free(void *ptr) {
 
 }
 
+#ifdef __GLIBC__
+#define QASAN_FILE_CHECK_SIZE sizeof(FILE)
+#else
+#define QASAN_FILE_CHECK_SIZE sizeof(void *)
```
vanhauser-thc, line 209: This is wrong.
sizeof(FILE) is the size of the structure, NOT the pointer, so the value on my linux is for example 209

vanhauser-thc:
please provide a real fix, and one that does not break non-musl-glibc setups (dietlibs, ulibc, etc.)

vanhauser-thc:
ping @celi0n

qemu_mode/libqasan/hooks.c
celi0n: Yes, I know, but since musl does'nt allow us to have access to the structure size I have to put the smallest size this structure could have. If I understand correctly it should only be less effective at catching bugs in fgets arguments (which is better than not compiling at all).
Looks like they have the same problem here and used an similar solution  
https://github.com/wadkhlemshem/cbmc/blob/48c5b108c8fde6e08ede0da35d46fa3f978989c5/src/ansi-c/library/stdio.c#L71
I can change the define to work with more libc but I wont be able to do better

vanhauser-thc: I will only merge this if this supports more libc alternatives. For musl you could check the internal size and subtract a bit from it to make it safe

PR #2838, opened 2026-07-03
afl-cc: set have_c for -x c++-header / -x c-header (PCH builds)

spomatasmd:
**Type of PR**: Enhancement
**Purpose of this PR**: Fixing afl-clang-lto compilation flow when PCH rules are present (e.g. Verilator 5)

Verilator >= 5.019 (2023-10-01) added automatic precompiled-header support in verilated.mk.  Its PCH rules invoke the compiler as:

    $(CXX) ... -x c++-header $< -o $@

In AFL-LTO mode, add_runtime() unconditionally appends afl-compiler-rt.o and afl-llvm-rt-lto.o to every invocation unless have_c is set.  Those artefacts are link-time only; injecting them into a PCH build gives clang multiple inputs against a single -o:

    clang: error: cannot specify -o when generating multiple output files

Fix: in parse_misc_params(), when x_set is already true and the current argument is "c++-header" or "c-header", set have_c via SCAN_KEEP — the same mechanism the "-c" handler uses.
This suppresses the runtime injection (and the add_lto_linker/add_lto_passes calls) while leaving all instrumentation defines and optimisation flags intact.
The resulting .gch file is a valid CPCH binary.

### DIFF
--- a/src/afl-cc.c
+++ b/src/afl-cc.c
@@ -2840,6 +2840,17 @@ param_st parse_misc_params(aflcc_state_t *aflcc, u8 *cur_argv, u8 scan) {
 
     SCAN_KEEP(aflcc->x_set, 1);
 
+  } else if (aflcc->x_set &&
+             (!strcmp(cur_argv, "c++-header") ||
+              !strcmp(cur_argv, "c-header"))) {
+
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thanks!

PR #2855, opened 2026-07-20
push to stable

vanhauser-thc:


src/afl-fuzz-state.c
```diff
@@ -54,6 +54,7 @@ void afl_state_init(afl_state_t *afl, uint32_t map_size) {
   afl->schedule = EXPLORE;              /* Power schedule (default: EXPLORE)*/
   afl->havoc_max_mult = HAVOC_MAX_MULT;
   afl->clear_screen = 1;                /* Window resized?                  */
+  //afl->primary_trace = 1;              /* trace_bits holds the coverage map */
```
domenukk: maybe remove stuff instead of commenting out?

PR #2865, opened 2026-08-03
Accept the documented AFL_GCC_ONLY_FSRV spelling

karpovantonme:

**Type of PR**: Bugfix
**Purpose of this PR**: make the documented `AFL_GCC_ONLY_FSRV` spelling actually work
**I have tested the changes**: yes — see below

## The problem

`AFL_GCC_ONLY_FSRV` and `AFL_GCC_ONLY_FRSV` are both live in the tree, and they are not the same variable to the code.

Spelled `FSRV` (the documented one):

- `docs/env_variables.md` documents `AFL_LLVM_ONLY_FSRV`/`AFL_GCC_ONLY_FSRV`
- `src/afl-cc.c` — `add_gcc_plugin()` returns early on `getenv("AFL_GCC_ONLY_FSRV")`
- `src/afl-common.c` — when the deprecated `AFL_SAN_NO_INST` is used it prints "AFL_LLVM_ONLY_FSRV/AFL_GCC_ONLY_FSRV is induced and set instead" and calls `setenv("AFL_GCC_ONLY_FSRV", "1", 0)`

Spelled `FRSV`:

- `include/envs.h` — the registry of known variables
- `instrumentation/afl-gcc-pass.so.cc` — the plugin help text, the `fsrv_only` check, the message
- `instrumentation/afl-compiler-rt.o.c` — the check that swaps `__afl_area_ptr` to the dummy map

So the name in the documentation is unknown to the registry:

```
$ AFL_GCC_ONLY_FSRV=1 ./afl-fuzz
Did you mean AFL_GCC_ONLY_FRSV?
[!] WARNING: Mistyped AFL environment variable: AFL_GCC_ONLY_FSRV=1
```

Following the documentation gets a warning, a two second sleep, and a suggestion to use the typo. The transposed spelling passes silently. And since `afl-common.c` sets `AFL_GCC_ONLY_FSRV` itself when it deprecates `AFL_SAN_NO_INST`, AFL++ ends up warning about a variable it set on the user's behalf. On top of that, a user who exports the documented name does not get the dummy coverage map in the runtime, because that check only looks for `FRSV`.

## The change

Accept `FSRV` in all three places, and keep `FRSV` working — the plugin help text has been advertising it, so someone may well have copied it. Help text and messages now name the documented spelling.

I went with accepting both rather than a straight rename because the rename would silently stop working for anyone who took the name from the plugin's own `--help`. If you would rather have a clean rename, or `AFL_GCC_ONLY_FRSV` moved into `afl_environment_deprecated[]` so it warns, say so and I will redo it.

## Testing

Built `afl-fuzz` on macOS/arm64 with `AFL_NO_X86=1`. Before the change, `AFL_GCC_ONLY_FSRV=1 ./afl-fuzz` produced the warning above; after it, both spellings are accepted with no warning.

I could not build the GCC plugin or the LLVM passes here (macOS), so the `afl-gcc-pass.so.cc` and `afl-compiler-rt.o.c` hunks have only been read, not compiled — they are a `||` on an existing `getenv` and two string literals, but I would rather say it than let CI find out.

`clang-format` only on the touched line ranges, since running it over the whole files reflows a lot of unrelated code with the version I have here.

## While I was there

The same sweep — comparing every `AFL_*` name in `docs/env_variables.md` against `afl_environment_variables[]` — turns up six more documented variables that are live in the code but missing from the registry, so they produce the same "Mistyped" warning: `AFL_DUMP_MAP_SIZE`, `AFL_FORKSERVER_PARENT`, `AFL_FRIDA_INST_NO_BACKPATCH`, `AFL_FRIDA_SECCOMP_FILE`, `AFL_LLVM_BUG_ALLOCSIZE_STACK`, `AFL_USE_RTSAN`. And one, `AFL_LLVM_DISABLE_INSTRUMENTATION`, is documented but appears nowhere in the code at all.

I deliberately left them out of this PR to keep it to one thing. **If you want them fixed, just say so in a comment here and I will send a follow-up PR adding the six to the registry and sorting out the seventh** — I already have the sweep, so it is quick. Equally happy to leave them alone if any of them are unregistered on purpose.


vanhauser-thc:
Can you please add those to the env include file please: AFL_FRIDA_INST_NO_BACKPATCH, AFL_FRIDA_SECCOMP_FILE, AFL_LLVM_BUG_ALLOCSIZE_STACK, AFL_USE_RTSAN

thanks!

karpovantonme:
Done, added in the second commit. Checked each one after rebuilding, none of them warn now:

```
AFL_FRIDA_INST_NO_BACKPATCH=1 ./afl-fuzz   -> no warning
AFL_FRIDA_SECCOMP_FILE=1 ./afl-fuzz        -> no warning
AFL_LLVM_BUG_ALLOCSIZE_STACK=1 ./afl-fuzz  -> no warning
AFL_USE_RTSAN=1 ./afl-fuzz                 -> no warning
```

Two more from the same sweep that I left out, since you did not list them and I am not sure either is meant to be user facing:

- `AFL_DUMP_MAP_SIZE` - read in `afl-compiler-rt.o.c` in two places and documented in `env_variables.md`, so this one looks like a plain omission.
- `AFL_FORKSERVER_PARENT` - `afl-forkserver.c` sets it for the child and the runtime unsets it, so it is an internal marker rather than something a user sets. It would only warn in a nested run. Probably not worth registering, but say the word and I will add it.

Also documented in `env_variables.md` but absent from the code entirely: `AFL_LLVM_DISABLE_INSTRUMENTATION`. Looks like a leftover from a rename. Happy to send a docs-only PR for that separately.

vanhauser-thc:
Thanks!

Issue #1699, opened 2023-04-11
afl-fuzz seg faults when using a python custom mutator


r00tus3r:
Compiled AFLplusplus with `ASAN_BUILD=1` + `AFL_DEBUG=1`. Segmentation fault seems to occur during calibration stage 2.

Btacktrace 
```
==3737653==ERROR: AddressSanitizer: heap-use-after-free on address 0x7f9a66109800 at pc 0x7f9a69b9c397 bp 0x7ffcd41b17c0 sp 0x7ffcd41b0f68
READ of size 2014 at 0x7f9a66109800 thread T0
    #0 0x7f9a69b9c396 in __interceptor_memcpy ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:827
    #1 0x7f9a6956e825 in PyByteArray_FromStringAndSize (/lib/x86_64-linux-gnu/libpython3.10.so.1.0+0xcd825)
    #2 0x562920c00ccd in post_process_py src/afl-fuzz-python.c:537
    #3 0x562920c1efe3 in write_to_testcase src/afl-fuzz-run.c:87
    #4 0x562920c21c7b in calibrate_case src/afl-fuzz-run.c:481
    #5 0x562920bc240f in perform_dry_run src/afl-fuzz-init.c:894
    #6 0x562920bb241c in main src/afl-fuzz.c:2228
    #7 0x7f9a69268d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f)
    #8 0x7f9a69268e3f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x29e3f)
    #9 0x562920b9d804 in _start (/usr/local/bin/afl-fuzz+0x24804)
```

On a high level, the python custom mutator had the following code:
```
def init(seed):
    pass

def post_process(buf):
    # Modify buf into the format required by the program being fuzzed
    # If buf is not good, return b"" else return the updated buf in out_buf
    return out_buf

def deinit():
    pass
```

I tried to follow the example code to do this, maybe I'm missing something. I'll be happy to provide any additional information. Thanks!

vanhauser-thc:
but you have not defined out_buf in your python script? if you do not want to change anything then "return buf".

r00tus3r:
I have defined it, `len(out_buf) > len(buf)`, data in `buf` is copied to `out_buf`which is in the format expected by the program.

vanhauser-thc:
[...]

Issue #1424, opened 2022-05-24
A FATAL BUG when using custom mutator 


BlackBird-BB:
I tried to use [afl-protobuf-mutaor](https://github.com/thebabush/afl-libprotobuf-mutator) to fuzz a ELF FILE. Everything is prpared well. Here is the mutator.cc I used in afl-protobuf-mutaor.
``` c
size_t afl_custom_post_process(void* data, uint8_t *buf, size_t buf_size, uint8_t **out_buf) {
    #pragma unused (data)
    // new_data is never free'd by pre_save_handler
    // I prefer a slow but clearer implementation for now
    
    static uint8_t *saved_buf = NULL;

    menuctf::ChoiceList msg;
    std::stringstream stream;
    // Tried to load input_encode
    if (protobuf_mutator::libfuzzer::LoadProtoInput(true, buf, buf_size, &msg))
      ProtoToDataHelper(stream, msg);
    else
      abort();
    const std::string str = stream.str();

    uint8_t *new_buf = (uint8_t *) realloc((void *)saved_buf, str.size());
    if (!new_buf) {
      *out_buf = buf;
      return buf_size;
    }
    *out_buf = saved_buf = new_buf;

    memcpy((void *)new_buf, str.c_str(), str.size());

    return str.size();
  }

  int32_t  afl_custom_init_trim(void *data, uint8_t *buf, size_t buf_size) {
    /// NOTE: disable trim
    return 0;
  }
  
  size_t afl_custom_trim(void *data, uint8_t **out_buf) {
    /// NOTE: unreachable
    return 0;
  }
```
When AFLplusplus show
> calibration stage 2/8

The process abort.
After a long time debuging. I find that It abort at my mutator.cc function afl_custom_post_process:
``` c
// Tried to load input_encode
    if (protobuf_mutator::libfuzzer::LoadProtoInput(true, buf, buf_size, &msg))
      ProtoToDataHelper(stream, msg);
    else
      abort();
```
Then I use gdb to find out what happened. 

The Main process is :

> perform_dry_run-> calibrate_case -> write_to_testcase -> afl_custom_post_process

[...]

PR #2869, opened 2026-08-05
Fix custom post-process buffer ownership

### LINKED ISSUE #2868: Out-of-place custom post_process causes buffer use-after-free
**IMPORTANT**

1. Verified present on the current `dev` branch at `1ce1ac1b` (2026-08-05).
2. Command line and environment used (reproducer details below):

```
AFL_CUSTOM_MUTATOR_LIBRARY=./libpostprocessmutator.so AFL_NO_UI=1 \
AFL_SKIP_CPUFREQ=1 AFL_I_DONT_CARE_ABOUT_MISSING_CRASHES=1 \
ASAN_OPTIONS=detect_leaks=0:abort_on_error=1 \
./afl-fuzz -s 9 -V15 -m none -i in -o out -- ./target
```

**Describe the bug**

On current `dev`, an out-of-place custom `afl_custom_post_process()` can make
`write_to_testcase()` return a pointer whose allocation is subsequently
reallocated through `afl->out_scratch_buf`. A caller that retains the returned
pointer then passes freed memory to a later post-processing call.

ASan reports a heap-use-after-free during `calibrate_case()`. The allocation
is freed by `afl_realloc()` in `write_to_testcase()` and subsequently read by
the custom post-processor. Python is not required; a C post-processor
returning a separate buffer reproduces it.

The same buffer aliasing can manifest as `memcpy-param-overlap` during havoc.
`AFL_POST_PROCESS_KEEP_ORIGINAL=1` avoids some affected paths but is not a
general fix for the default behaviour.

**To Reproduce**

1. Check out `dev` at `1ce1ac1b`.
2. Build with `make ASAN_BUILD=1 afl-fuzz`.
3. Load a C custom mutator whose `afl_custom_post_process()` copies its input
   to a separately allocated buffer and returns that buffer (doubling the
   input works well; see #2869 for a ready-made fixture).
4. Run `afl-fuzz -s 9 -V8` with two small seeds.
5. Observe the ASan heap-use-after-free during calibration, normally within a
   few seconds.

**Expected behavior**

Post-processing should affect the testcase sent to the target without
invalidating buffers retained by AFL++ mutation stages.

**Screen output/Screenshots**

ASan trace on stock `dev` at `1ce1ac1b` (fixed seed `-s 9`, deterministic):

```
==2676853==ERROR: AddressSanitizer: heap-use-after-free
READ of size 160 at 0x7cb321be0908 thread T0
    #0 memcpy (/usr/lib/libasan.so.8)
    #1 afl_custom_post_process test/test-custom-post-process.c:57
    #2 write_to_testcase   src/afl-fuzz-run.c:156
    #3 calibrate_case      src/afl-fuzz-run.c:622
    #4 save_if_interesting src/afl-fuzz-bitmap.c:1320
    #5 common_fuzz_stuff   src/afl-fuzz-run.c:1644
    #6 fuzz_one            src/afl-fuzz-one.c:3776
```

**Additional context**

This appears to be a remaining instance of the ownership problems previously
discussed in #1699 and #1424. A C-only regression test and fix are in #2869.

matthiasgoergens:
Fixes #2868

`write_to_testcase()` stages out-of-place custom `post_process` results in
`afl->out_scratch_buf` and swaps the global `out` and `out_scratch` buffers.
Callers retain their own by-value buffer pointers, so a later call can
reallocate the same scratch allocation while a caller still references it.
On current `dev` (verified at `1ce1ac1b`), a C-only out-of-place
post-processor reproduces a heap-use-after-free during calibration within
seconds under ASan; the same ownership error can also produce overlapping
copies during havoc.

Keep post-processed testcases in dedicated AFL-owned buffers associated with
the custom mutator instead. Two buffers are used because calibration can call
`write_to_testcase()` again with the result of an earlier call. The mutation
engine's `out` and `out_scratch` buffers are no longer swapped by the custom
post-processing path.

The failed-post-processing fallback now restores both the original length and
the original pointer. Previously, multiple post-processors could leave the
pointer from an earlier processor paired with the original length.

Includes a C-only regression test wired into `test/test-custom-mutators.sh`;
it goes red on unpatched `dev` under ASan within seconds. Python is not
required to trigger the bug.

Fixes: 8f6d9d66 ("fix post_process")

This is related to the post-processing ownership failures discussed in #1699
and #1424, but remains reproducible on current `dev` without
`AFL_POST_PROCESS_KEEP_ORIGINAL`.

**Type of PR**: Bugfix
**Purpose of this PR**: Prevent custom post-processing from invalidating AFL++ mutation-buffer pointers

### DIFF
--- a/include/afl-fuzz.h
+++ b/include/afl-fuzz.h
@@ -1115,6 +1115,7 @@ struct custom_mutator {
   char       *name_short;
   void       *dh;
   u8         *post_process_buf;
+  u8         *post_process_buf_scratch;
   u8          stacked_custom_prob, stacked_custom;
 
   void *data;                                    /* custom mutator data ptr */
--- a/src/afl-fuzz-mutators.c
+++ b/src/afl-fuzz-mutators.c
@@ -156,6 +156,13 @@ void destroy_custom_mutators(afl_state_t *afl) {
 
       }
 
+      if (el->post_process_buf_scratch) {
+
+        afl_free(el->post_process_buf_scratch);
+        el->post_process_buf_scratch = NULL;
+
+      }
+
       ck_free(el);
 
     });
--- a/src/afl-fuzz-run.c
+++ b/src/afl-fuzz-run.c
@@ -152,8 +152,11 @@ u32 __attribute__((hot)) write_to_testcase(afl_state_t *afl, void **mem,
     ssize_t new_size = len;
     u8     *new_mem = *mem;
     u8     *new_buf = NULL;
+    struct custom_mutator *staging_mutator = NULL;
 
     LIST_FOREACH(&afl->custom_mutator_list, struct custom_mutator, {
+
+      staging_mutator = el;
 
       if (el->afl_custom_post_process) {
 
@@ -183,6 +186,7 @@ u32 __attribute__((hot)) write_to_testcase(afl_state_t *afl, void **mem,
       if (fix) {
 
         new_size = len;
+        new_mem = *mem;
 
       } else {
 
@@ -193,6 +197,7 @@ u32 __attribute__((hot)) write_to_testcase(afl_state_t *afl, void **mem,
     }
 
     ssize_t valid_size = new_size;
+    u8     *original_mem = *mem;
 
     if (unlikely(new_size < afl->min_length && !fix)) {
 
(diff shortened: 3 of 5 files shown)


vanhauser-thc:
thanks!

PR #2878, opened 2026-08-23
Link afl-cc with librt to support glibc < 2.17

Scott-Guest:
**Type of PR**: Bugfix
**Purpose of this PR**: Support older glibc by linking librt

`gettimeofday()` was [recently replaced](https://github.com/AFLplusplus/AFLplusplus/commit/f3c739fc8c29e725832e60e23401684530b0e02d#diff-4508068dbd706015292112facf49308042c1e24e57f20fa12324afcb5aaa3e01R1108-R1130) with `clock_gettime()`. Prior to glibc 2.17, `clock_gettime()` lived in `librt`, so we need a `-lrt`. Only `GNUmakefile.llvm` needs this fix; other makefiles already include `-lrt`.

This also exposed some missing `override` keywords, which are fixed as well.

### DIFF
--- a/GNUmakefile.llvm
+++ b/GNUmakefile.llvm
@@ -33,6 +33,10 @@ VERSION     = $(shell grep '^ *$(HASH)define VERSION ' ./config.h | cut -d '"' -
 
 SYS = $(shell uname -s)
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thanks

PR #2879, opened 2026-08-23
Fix afl-merge man page generation

Scott-Guest:
**Type of PR**: Bugfix
**Purpose of this PR**: Fix man page generation
**I have tested the changes**: yes
  
`afl-merge.8` was [recently added](https://github.com/AFLplusplus/AFLplusplus/commit/c12d296b4e45917643e7743648cff42ae522cb67) to `MANPAGES`, but Make had no rule for its `afl-merge` prerequisite. This broke `make man` and builds that install man pages. This PR adds an explicit `afl-merge` target that creates the required symlink.

PR #2883, opened 2026-08-27
Improve CI

chris-srlabs:
**Type of PR**: Bugfix
**Purpose of this PR**: Detect failing `make install` next time

This is related to fix #2879, that added a missing target to the Makefile.
In `ci.yml` now a `make install` is done instead of `make all` (since: `install: all $(MANPAGES)` also includes the manpages). In addition checkout actions were updated.


vanhauser-thc:
danke dir!

PR #2884, opened 2026-08-27
frameshift: fix narrow length field overflow

felix314159:
## Overview

**Type of PR**: Bugfix
**Purpose of this PR**: 

* reject insertions that exceed a tracked length field's representable range
* disable invalid `FrameShift` relations through the existing tracking path
* add unit tests for narrow-field wrapping

## Problem

`rel_on_insert()` updates tracked length fields by adding the inserted size, masking the result to the field width, and then comparing it with the previous value to detect overflow.

Masking before the comparison misses some overflows. For example, adding `256` to an 8-bit length field containing `5` wraps back to `5` and is accepted as a valid update:

```text
before: enabled=1 value=5 byte=99
after: enabled=1 value=5 byte=5
```

Standard havoc mutations can generate insertions larger than an 8-bit field can represent. When such an update is accepted, `fs_sanitize()` keeps applying the wrapped relation and writes an incorrect length into the mutated testcase. This can turn otherwise useful mutations into malformed inputs and reduce the coverage benefit provided by `FrameShift`.

The fix checks the insertion size against the field's remaining representable range before performing the addition. Existing tracking behavior then disables the relation instead of applying a wrapped value.

## Validation

* added a test case where adding `256` to an 8-bit relation must fail and a test confirming `fs_track_insert()` disables the relation and `fs_sanitize()` leaves the testcase unchanged
* ran existing unit tests and `make code-format`
* built `afl-fuzz`
* formatted the modified sources with the project formatter

### DIFF
--- a/src/afl-fuzz-frameshift.c
+++ b/src/afl-fuzz-frameshift.c
@@ -48,13 +48,12 @@ int rel_on_insert(fs_relation_t *rel, u64 idx, u64 size) {
   // Check if we should update the value of the field.
   if (idx >= rel->anchor && idx <= rel->insert) {
 
-    u64 pre = rel->val;
+    u64 max = rel->size < 8 ? (1ULL << (rel->size * 8)) - 1 : UINT64_MAX;
+
+    // Check before adding: masking first misses increments that wrap by a
+    // complete field width, such as adding 256 to an 8-bit length.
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
thanks! the same issue is in the discovery loop, will fix that one too

PR #2896, opened 2026-09-21
afl-cc: do not treat -fcf-protection as a request for CFISAN

EduardOrehov:
## Summary

`afl-cc` records a plain `-fcf-protection` on the command line as `have_cfisan`. That is the same flag `AFL_USE_CFISAN` sets, so the CFI-sanitizer code path is taken: for clang it injects `-flto` and `-fvisibility=hidden`. The PCGUARD pass is a pre-link pipeline callback and never runs under `-flto`, so the target ends up with **no instrumentation at all** and nothing warns about it.

This PR records the flag as `have_fcf` instead, which is exactly what the CFISAN path already checks before adding its own `-fcf-protection=full`. `AFL_USE_CFISAN=1` behaves as before.

## Why it matters

`-fcf-protection` is part of the default hardened `CFLAGS` on Fedora and RHEL (`redhat-rpm-config`), so anyone who builds a target with the distro flags gets a silently uninstrumented binary. It also makes `make test_build` fail during a distro package build, because the RPM `CFLAGS` are passed to the self-test.

## Reproduce (5.03c, Fedora 42, LLVM 20)

```
$ AFL_PATH=. ./afl-cc -fcf-protection ./test-instr.c -o test-instr
$ ./afl-showmap -m none -q -o .t0 ./test-instr </dev/null
$ echo 1 | ./afl-showmap -m none -q -o .t1 ./test-instr
$ cmp .t0 .t1 && echo IDENTICAL
IDENTICAL          # 0 tuples in both maps
```

With `AFL_DEBUG=1` the final clang command line shows the injected flags:

```
clang -fpass-plugin=./SanitizerCoveragePCGUARD.so -fcf-protection ./test-instr.c -o test-instr -flto -fvisibility=hidden ...
```

Compiling the same file directly with `clang -fpass-plugin=./SanitizerCoveragePCGUARD.so -fcf-protection` works fine, so the pass itself is not the problem.

## Testing

With the patch, in the same environment:

| Flags passed to `afl-cc`                         | Result                          |
|--------------------------------------------------|---------------------------------|
| none                                             | 11/13 tuples (baseline)         |
| `-fcf-protection`                                | 11/13 tuples                    |
| full Fedora hardened CFLAGS/LDFLAGS (without LTO)| 12/14 tuples                    |
| `AFL_USE_CFISAN=1`                               | still injects `-flto -fsanitize=cfi -fvisibility=hidden` |

`make test_build` now passes with `CFLAGS` set to the Fedora hardened flags (minus `-flto=auto`, which is a separate, known limitation of the pass-plugin approach).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vanhauser-thc:
thanks!

